PR #3, opened 2023-07-06
RPC reflector

tomasciccola:
This branch is a good starting point for a mapeo-mobile with rpc-reflector integrated.
Thinks to take into account:
* Initially there were two different wrappers (frontend and backend) for the channel thats used for bridging. I basically merged them into one file trying to accommodate both cases. This now lives on `src/shared/lib/message-port-like.js`
* For shared files between the frontend and backend I added a `src/shared` folder. I don't if this is a good approach or if there's a best practice.
* The api (now called MapeoClient) lives on `src/shared`. Its a plain object but probably I'll turn it into a class when adding the mocked api stuff
* I modified the `scripts/build_backend.sh` file so to copy the shared files in a way that preserves the relative file path and not mess imports. I don't know if this is a good approach, but it works
*


src/frontend/App.tsx
```diff
@@ -21,29 +23,35 @@ const App = () => {
...
-      // @ts-expect-error
-      subscription.remove();
+      channel.close();
     };
   }, []);
-
-  return {
-    send: (msg: string) => nodejs.channel.send(msg),
-  };
+  return clientApi;
```
gmaclennan, line 55: `nodejs.start()` shouldn't be in here - that should just happen once when the JS file first loads, not tied to the react lifecycle.

I'm not sure if the channel and clientApi need to be in here either, nor tied to the useEffect. This isn't a side-effect, and there is no async here, so this code is rendering the app once with no clientApi, then calling `setClientApi()` and rendering again. If we want to do a `useClientApi()` it should probably just be a wrapper for `useMemo()`, since we only want it once, or we could just put the client api in the global scope.

gmaclennan (commented):
I few bits of feedback. I think it's ok to leave the `useNodejsMobile()` hook here for this PR, because it will work ok, but we should address that in a later PR.

Also to address in a later PR: https://github.com/digidem/react-native-with-mapeo-core-next/issues/4

src/frontend/App.tsx
achou11: ah yeah this stuff I worked on - not sure why I did this but fully agree that it should be set up outside of react

package.json
```diff
@@ -10,9 +10,14 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
+    "assert": "^2.0.0",
+    "buffer": "^6.0.3",
     "nodejs-mobile-react-native": "^0.8.2",
     "react": "18.1.0",
-    "react-native": "0.70.6"
+    "react-native": "0.70.6",
+    "readable-stream": "^4.4.1",
```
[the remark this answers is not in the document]
tomasciccola: Same with this, I had to leave the readable-stream dep for the app to work. But the metro translation to `stream` wasn't necessary. react-native mysteries I guess...

package.json
```diff
@@ -10,9 +10,14 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
+    "assert": "^2.0.0",
```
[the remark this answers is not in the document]
tomasciccola: It seems that removing this breaks the app, don't know the real reason though...

src/frontend/App.tsx
achou11: ah remember why i had it set up this way - mapeo mobile does it similarly (where `api.startServer` calls `nodejs.start()`): 

https://github.com/digidem/mapeo-mobile/blob/ca167c257b90ba74b0e63e623f405aed98d86145/src/frontend/AppLoading.tsx#L77-L81

either way, agree it probably could just be moved outside for the sake of this project

achou11: addressed in https://github.com/digidem/react-native-with-mapeo-core-next/pull/7

achou11 (commented):
Noticed some issues as I explored these changes locally:

1. `rpc-reflector` was not being specified as a dep for the _backend_ project. It was only being specified for the top-level. Guessing that this causes some of the issues related to certain deps seeming to be needed but not really

2. Needed to polyfill `process`, `util`, and `events` in order for rpc-reflector to not complain on the React Native side. Using something like https://github.com/ionic-team/rollup-plugin-node-polyfills to achieve and updating the metro config worked for now

3. The MessagePortLike class needs to add an `off` method, which is what rpc-reflector uses to determine if it's a viable message port interface: https://github.com/digidem/rpc-reflector/blob/9bcc507edd6e498145593bf491a927e3ab5ba652/lib/is-message-port-like.js#L19-L21

4. The MessagePortLike doesn't fully account for the differences between the nodejs mobile `channel` on React Native vs NodeJS Mobile. On NodeJS Mobile, it extends the `EventEmitter`, so that's fine. On React Native, it extends a [custom implementation](https://github.com/facebook/react-native/blob/v0.70.6/Libraries/vendor/emitter/EventEmitter.js) of an EventEmitter, which behaves differently from Node's EventEmitter

5. Not as important, but currently this PR introduces deps changes via npm 6 (previously was 8). Ideally we'd keep it at 8 but Tomas mentioned potential issues getting 8 to work in his environment, so may need some looking into

ErikSin:
@tomasciccola I scheduled a meeting with you for tuesday to talk about these things:

1. I am unable to build this app on my phone/emulator
2. Typescript typing is not working for me.
3. Review general architecture
4. Can we talk about strategies of creating fake data (ts object, json objects....etc)

ErikSin:
As discussed in out meeting today, this is the list of todos for the mock API for observations

### To do (Backend)
- [ ] Generate observations (as JSON) using 'Mapeo-Mock-Data'. 
- [ ] Convert JSON into objects, and creating them as files in `src/utils'
- [ ] On creation of Mapeo-Core, write them objects to database

### To do (Front-end)
- [ ] Ask Gregor/Andrew about typings
- [ ] Move `createClient()`outside of react lifecycle/hook
- [ ] Create `useObservation` and `useObservations` hook
- [ ] Bring Screens from `mapeo-mobile` which utilize observations

tomasciccola:
So, the type of `mapeoClient.observation` is now just a plain object, this was so that the generated docs from mapeo-mock-data would be valid observations. In the, not so distant future we are going to get the type from mapeo-core-next so I didn't took the time to get more strict typing. 
Additionally I created `src/utils/mockData.js` as javascript (not typescript) despite what initially discussed with Erik, this is 'cause the backend, which is plain js, is loading this file and transpiling didn't make much sense.
I know @ErikSin we talked about using ts for this file, but in the future the JSDoc on the mockData file won't be much more than
```js
/* @type {import('mapeo-core-next').Observation[]} */
const mockData = [...]
```

tomasciccola:
So, with @ErikSin we went back and forth with setting `clientApi` as a function initialized on `App.tsx` or as a hook. For now its a hook, but I like the approach that @gmaclennan proposed, copying and pasting from the chat:

```tsx
export function WithClientApi({ children, client }) {
  return (
    <ClientApiContext.Provider value={client}>
      {children}
    </ClientApiContext.Provider>
  )
}

// this will probably be smth like initClient() that sets up the MessagePort like and coordinates server state
const client = createClient(port) 

<WithClientApi client={client}>
  <App />
</WithClientApi>

// That way if you want to unit-test components then you can do
const client = {
  observation: {
    getMany(): [...mockedData]
  }
}

<WithClientApi client={client}>
  <ComponentToUnitTest />
</WithClientApi>
```
So basically use a react context. We can set up that in this PR, or merge it in the state that's in, and do it later

src/backend/index.js
```diff
@@ -1,21 +1,36 @@
...
+import MessagePortLike from './message-port-like.js';
+import {mockData} from './mockData.js';
 
 // TODO: Account for args passed from node.startWithArgs
 debug.enable('*');
 
-const log = debug('mapeo-mobile-node-next');
+// @ts-expect-error
+const channel = new MessagePortLike(rn_bridge.channel);
+channel.start();
```
gmaclennan, line 15: Can you move this (`.start()`) to after the RPC server is created? Strictly this will be ok as-is because everything happens in the same tick, but if someone changes that, e.g. with an await before `createServer()`, then there would be a chance that messages would be lost.

src/backend/message-port-like.js
```diff
@@ -0,0 +1,76 @@
...
+  #messageHandler;
+  /** @type {String} */
+  #API_EVENT_NAME = '@@API_MESSAGE';
+  /** @type {Channel} */
+  #channel;
+
+  /**
+   * @param {Channel} channel
+   */
+  constructor(channel) {
```
gmaclennan, line 30: Since this is strictly tied to the static import `nodejs-mobile-react-native`.Channel, I think this should just be an import in this file, and not a constructor parameter, so this file encapsulates the Channel dependency.

src/frontend/lib/MessagePortLike.ts
```diff
@@ -0,0 +1,119 @@
...
+  #channel: Channel;
+  #eventsSubscriptions = new Map<
+    string,
+    WeakMap<Listener, EventSubscription>
+  >();
+  #channelSubscription: EventSubscription;
+  #state: ServerState = 'idle';
+  #queuedMessages: any[] = [];
+
+  constructor(channel: Channel) {
```
gmaclennan, line 20: Since this is strictly tied to the static import `nodejs-mobile-react-native`.Channel, I think this should just be an import in this file, and not a constructor parameter, so this file encapsulates the Channel dependency.

src/frontend/lib/MessagePortLike.ts
```diff
@@ -0,0 +1,119 @@
...
+      // TODO: Call this.#subscriptions.delete(event) if no more listeners?
+    }
+
+    return this;
+  }
+
+  removeAllListeners() {
+    this.#eventsSubscriptions.clear();
+    this.removeAllListeners();
+  }
```
gmaclennan, line 106: This seems to be re-implementing an event emitter, which is a hard thing to do and is full of edge cases. Why not use an existing library and extend from that rather than the react-native eventemitter?

gmaclennan (changes requested):

src/frontend/lib/MessagePortLike.ts
achou11: think the context for this (as well as the other review comments) mostly lives with stuff that I updated via https://github.com/digidem/react-native-with-mapeo-core-next/pull/8 (see PR description)

but yeah if there's something we can use to simplify this, probably a better approach

achou11: think the main trickiness was that since this is tied to nodejs mobile's Channel implementation, it doesn't have the expected interface as the traditional Node EventEmitter, which RPC reflector relies on. 

React Native's EventEmitter uses a subscription based API i.e. `addListener` returns a `Subscription` and there are no methods like `on` and `off`, hence why this class is reimplementing some of those while also tying it with the Nodejs mobile Channel

gmaclennan: Yes, but why extend MessagePortLike from the react-native EventEmitter? Why not extend EventEmitter from [events](https://www.npmjs.com/package/events) or [EventEmitter3](https://www.npmjs.com/package/eventemitter3) - those have the methods that rpc-reflector checks. I would have thought that would work fine? There is no relationship between the type of EventEmitter that Channel extends and what MessagePortLike needs to be, except how you internally handle attaching and removing listeners from the channel.

achou11: ah right. didn't think about that when initially implementing but yeah that makes sense 👍

gmaclennan: yeah, because what you're doing here is essentially reimplementing an EventEmitter, and also creating a compatibility layer with the react native EventEmitter, which is a maintenance overhead, and probably needs a lot more work to catch all the edge cases that the npm libs already check for.

achou11: right, so the actionable changes here would be:

1. use one of the aforementioned libraries to extend for this class, instead of React Native's EventEmitter

2. hook into the `on` and `off` methods in order to you manage the nodejs channel listeners as well, which probably still requires some tracking similar to what we have in the current implementation

achou11: Don't think so because it relies on Node's EventEmitter, which isn't available in React Native's runtime

achou11:
Having trouble getting this to work when running locally. Just to check, is it working for other people?

src/frontend/lib/MessagePortLike.ts
gmaclennan: `tiny-typed-emitter` is just [require('events').EventEmitter](https://github.com/binier/tiny-typed-emitter/blob/master/lib/index.js#L3). It works in node, but you need to have a polyfill for react native. You can just use [`events`](https://www.npmjs.com/package/events) which is handy because it has the same name - you just need to install it from npm in the frontend - or you can use EventEmitter3 with tiny-typed-emitter if you configure Metro to package `events` -> `EventEmitter3`

tomasciccola:
> Having trouble getting this to work when running locally. Just to check, is it working for other people?

Yep, up until now, when I implemented the changes proposed by @gmaclennan, now the app crashes upon opening it, and logcat doesn't give any useful output. Gonna rollback the chances and introduce them one by one to see where it fails...

PR #22, opened 2023-08-15
more proposed changes to the rpc-reflector integration

achou11:
The current state of https://github.com/digidem/react-native-with-mapeo-core-next/pull/3 wasn't working for me locally, so I ended up diving in to figure out what the issue was, and ended up just cleaning up a bunch of things to see if it felt more approachable. Note that I'm not adamant about getting this merged as is, but wanted to at least make it easier to consider changes that might be helpful to incorporate in the final result.

Notable changes:

- updates the MessagePortLike implementations for frontend and backend based on Gregor's suggestions from the other PR
- moving away from the using a hook for creating the frontend api client. wasn't keen on having the client be part of react state and felt that it would be much easier to work with the api as an "always available" variable that can be called from anywhere (no more null/undefined checks needed). so the proposed approach (for now) is more like a singleton pattern, where there's a semi-mutable variable exported from a module that gets updated in the appropriate React component, which is the `Loading` component in this case. only caveat to this approach is making sure that the variable is set before the actual usage of it, which is relatively straightforward in this case but of course is prone to errors if misused.
- updating the backend and frontend communication to be more similar to how we do it in mapeo mobile. basically, in `Loading`, we do 2 things:
  - set up listeners for `status` messages that the backend sends
  - initialize the nodejs mobile process and create the frontend api client
  
  Didn't want to create hooks for these things because these should only really happen in one place (hooks are best for defining reusable behavior)


tomasciccola (approved):
LGTM :rocket:  !

PR #25, opened 2023-08-21
Ej/port navigation

lightlii:
This is still WIP because I'm having some trouble with Typescript (perhaps some leftover flow code as Gregor mentioned [here](https://digidem.slack.com/archives/C01GJ3V8D62/p1692623035077299).


ErikSin:
You can just stub out the custom header for now. Just create a very basic custom header with just a back button and title. 

Most of your problems are being introduced by draft observations, which we shouldnt be dealing with right now.

Also, the map screen can be stubbed out, so you dont need the any of the gps pill/gps stuff.

Also take out dev experiments. 

Anything related to UI should either not be included or stubbed out. We are just trying to introduce react-navigation with typings

PR #26, opened 2023-08-21
chore: added pre commit step

ErikSin:
Added husky and prettier staged as pre-commit steps.


achou11 (changes requested):
A few items:

1. prettier should be explicitly installed
2. I would avoid using `pretty-quick` since it currently is broken with prettier v3 and the maintenance on it has kind of staggered (which is usually okay but for something that works with a heavily updated dep like prettier, is prone to maintenance issues). Instead would use https://github.com/okonet/lint-staged. See https://github.com/digidem/mapeo-mock-data/blob/main/package.json for an idea of how to set these up.

PR #29, opened 2023-08-22
React Native Upgrade and Expo Modules

ErikSin:
Upgraded React-Native from `70.xx` to `71.13`. 

Added expo as a dependency (and configured the android files)

bumped mmkv to `@latest` as the version being used is not compatible with React-Native < v71.xx


achou11 (approved):
not much for me to review but I don't see any glaring issues. would say if the app runs locally then you're good 🤷‍♂️ 

will eventually have to integrate native modules but think it'd be easier to worry about that later. initial explorations to get that working live in https://github.com/digidem/react-native-with-mapeo-core-next/blob/build-native-deps/

PR #30, opened 2023-08-23
Ej/port navigation

lightlii:
Installs and configures react-navigation and creates basic setup with navigator + custom header

closes https://github.com/digidem/react-native-with-mapeo-core-next/issues/12


lightlii:
I've attempted porting over the flow typing in [src/frontend/types/index.ts](https://github.com/digidem/react-native-with-mapeo-core-next/pull/30/files#diff-0dc7ba143a2d6c2b0d35f6c810c111a56f2fd78a422a45f52f6ce1f6af8b7443) as well as the [declarations for react-navigation](https://github.com/digidem/react-native-with-mapeo-core-next/pull/30/files#diff-5ef0552e05e1f80cbec5be4233c83ccbd9650ad6b06c0c33188a2874a0f91c77) but I still get the error `Cannot find module 'react-navigation' or its corresponding type declarations.` ...Not sure I've well understood how this is working

ErikSin:
> I've attempted porting over the flow typing in [src/frontend/types/index.ts](https://github.com/digidem/react-native-with-mapeo-core-next/pull/30/files#diff-0dc7ba143a2d6c2b0d35f6c810c111a56f2fd78a422a45f52f6ce1f6af8b7443) as well as the [declarations for react-navigation](https://github.com/digidem/react-native-with-mapeo-core-next/pull/30/files#diff-5ef0552e05e1f80cbec5be4233c83ccbd9650ad6b06c0c33188a2874a0f91c77) but I still get the error `Cannot find module 'react-navigation' or its corresponding type declarations.` ...Not sure I've well understood how this is working

Perhaps i don't fully understand what is going on here, but these look like manual typings of react-nav.  I assumed they were used by flow? We don't need these flow typings anymore, as the TS typing are part of the react-nav library, except for the custom types created in [sharedTypes.ts](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/sharedTypes.ts).

android/app/src/main/java/com/mapeonext/MainActivity.java
ErikSin, line 1: what is this for?

src/frontend/hooks/useNavigationWithTypes.ts
```diff
@@ -0,0 +1,16 @@
+import {BottomTabNavigationProp} from '@react-navigation/bottom-tabs';
+import {CompositeNavigationProp, useNavigation} from '@react-navigation/native';
+import {NativeStackNavigationProp} from '@react-navigation/native-stack';
+import {AppStackList, HomeTabsList} from '../Navigation/AppStack';
```
ErikSin, line 4: `HomeTabsList` is importing from the wrong place

src/frontend/types/index.ts
ErikSin, line 1: can be deleted

src/frontend/types/react-navigation.d.ts
ErikSin, line 1: can be deleted

ErikSin (changes requested):
We should bring over the basic <CustomHeaderLeft/>. Take out any code that isnt related to the title and the back button.

Also we should follow the patter of pages (expect for the bottom tab navigator screen) having a `title` property  so that it can be injected into the header eventually

And we should bring over all the [typings](https://github.com/digidem/mapeo-mobile/blob/ca167c257b90ba74b0e63e623f405aed98d86145/src/frontend/sharedTypes.ts#L41).

android/app/src/main/java/com/mapeonext/MainActivity.java
lightlii: https://reactnavigation.org/docs/getting-started#installation

ErikSin (approved):

ErikSin:
navigation is blocking alot of tickets! So i just added the header and cleaned it up @lightlii

PR #31, opened 2023-08-23
chore: update persisted state to follow best practices

ErikSin:
Updated persisted store to not be statically created. Previous implementation (which followed Mapeo Mobile), meant that the persisted store was being dynamically created. This dynamic creation meant that a context was still need to propagate the state to the rest of the app, going against zustand best practices.

While this will mean more refactoring, I think this will come with an obvious benefit of a more simplified code and persisted store implementation!

### DIFF
--- a/src/frontend/hooks/persistedState/createPersistedState.ts
+++ b/src/frontend/hooks/persistedState/createPersistedState.ts
@@ -9,7 +9,7 @@ import {MMKV} from 'react-native-mmkv';
 
 const storage = new MMKV();
 
-type PersistedStoreKey = 'test' | 'MapeoLocale';
+type PersistedStoreKey = 'MapeoLocale';
 
 const MMKVZustandStorage: StateStorage = {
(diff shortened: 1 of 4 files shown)


src/frontend/hooks/persistedState/createPersistedState.ts
```diff
@@ -0,0 +1,50 @@
...
+    const value = storage.getString(name);
+    return value ?? null;
+  },
+  removeItem: name => {
+    return storage.delete(name);
+  },
+};
+
+/**
+ * alskdfasdf
```
lightlii, line 28: whats this? 👀

ErikSin:
> I'm unable to run this currently, even with downgrading react native for some reason..
> 

The mmkv version is tied to the react native version. You would also need to downgrade mmkv if you use an older react-native version

lightlii:
ok this runs now I've fixed my rn issue! Tested and working

PR #32, opened 2023-08-23
Es/internationalization

ErikSin:
Adds React-Intl as a dependency.

Due to hermes not supporting [RelativeTimeFormat](https://hermesengine.dev/docs/intl/#not-yet-supported), the [Intl.RelativeTimeFormat Polyfill](https://formatjs.io/docs/polyfills/intl-relativetimeformat) was added. This polyfill also required the [Intl.Locale Polyfill](https://formatjs.io/docs/polyfills/intl-locale) and the [Intl.PluralRules Polyfill](https://formatjs.io/docs/polyfills/intl-pluralrules/)

## To Do:
- [ ] crowdin integration


src/frontend/languages.json
ErikSin, line 1: I added this file for testing purposed

src/frontend/translations/messages.json
ErikSin, line 1: added for testing purposes

achou11 (approved):
would be helpful to have a small example of changing the language, but may become too out-of-scope I guess

PR #60, opened 2023-09-20
Static assets

lightlii:
closes https://github.com/digidem/react-native-with-mapeo-core-next/issues/36

Just misses `sharedComponents/icons/CategoryIcon.tsx` as this component depends on the api...


ErikSin (approved):
Switch to non-default assets

PR #61, opened 2023-09-20
Shared components

lightlii:
closes https://github.com/digidem/react-native-with-mapeo-core-next/issues/37

Currently lacks `sharedComponents/GpsPill.tsx` because it relies on code in `lib/utils` which is integrated in [this PR](https://github.com/digidem/react-native-with-mapeo-core-next/pull/60) and [contexts that are not yet ported over](https://github.com/digidem/react-native-with-mapeo-core-next/issues/38)


lightlii:
fixed those default exports now @ErikSin !

PR #63, opened 2023-09-22
Es/create observation screen

ErikSin:


src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
ErikSin, line 1: @achou11 Can you look over this file and just tell me if all of this makes sense and is aligned with `@mapeo/schema`

PR #71, opened 2023-10-09
Es/design changes

ErikSin:
- Updates button to have border radius of 30
- Adds support for svg images
- Adds CoMapeo Logo and Topographical Lines as Svgs

Closes [#1125](https://github.com/digidem/mapeo-mobile/issues/1125)


ErikSin:
These svgs will be used for the splash screen, so there is no current screen using them. But I have tested them out and are rendering as expected

achou11:
the svgs are _only_ used for the splash screen right? if that's the case, not sure the integration with react-native-svg is strictly needed (although we'll probably need it at some point anyways).

would ideally make sure that the assets we use align with splash screen libraries use these assets. for example, would take a look at the docs for https://github.com/zoontek/react-native-bootsplash and make sure that there's a path for the assets we're using to work with the library, because splash screens are typically they're own activity (or whatever android calls it), so they're not just basic react native navigation screens in a strict sense.

ErikSin:
They are also used in one [non-splash screen.](https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1103-15015&mode=design&t=hLJOCl49e8PkX0zI-4) Also there has been talk about converting some icons to svgs eventually (but nothing concrete on that)

achou11 (approved):
overall the changes make sense to me 👍 can worry about the assets in the context of the splash screen later

ErikSin:
> would ideally make sure that the assets we use align with splash screen libraries use these assets. for example, would take a look at the docs for https://github.com/zoontek/react-native-bootsplash or https://docs.expo.dev/versions/latest/sdk/splash-screen/ and make sure that there's a path for the assets we're using to work with the library, because splash screens are typically they're own activity (or whatever android calls it), so they're not just basic react native navigation screens in a strict sense.

Good point!

Looking at these 2 libraries, expo doesn't support svgs. So we should probable use the other one. It is a little unclear whether or not the file needs to be a certain folder and you are going to be creating the splash screen this sprint. So my other works relies on these svgs. So why don't we merge it as is, and if you do need to move these files, we can adjust when we know!

PR #72, opened 2023-10-11
feat: internationalization

ErikSin:
This pr allows for internationalization through crowdin, formatjs, and react-intl

- Crowdin.yml has been added, which configures which files are pulled into the translation website
- Translation build script has been created. This build scripts compiles all the translations into one file, and provides the react app with the list of available locales
- Format js has been added, which extracts all the messages from the components, and adds them as a JSON file. This JSON file is used by crowdin.


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 npx lint-staged
+npm run extract-messages
```
achou11, line 5: maybe put this in the lint-staged configuration and only run when relevant files are staged? e.g.

```json
  "lint-staged": {
    "*.{js,jsx,ts,tsx}": "npm run extract-messages"
  },
```

package.json
```diff
@@ -90,6 +93,7 @@
     "@typescript-eslint/parser": "^5.37.0",
     "babel-jest": "^29.2.1",
     "eslint": "^8.19.0",
+    "glob": "^9.3.0",
```
achou11, line 96: is this a peer dep of the formatjs cli?

achou11 (approved):
overall looks good! couple of non-blocking comments

package.json
ErikSin: It used by the `build-translations` script!

ErikSin:
@achou11 I just fixed a couple things
1. The two translations scripts are fairly cheap scripts so I added them to npm run start command
2. I changed the location of where the translation is created, to be outside of the src folder. This matches the pattern in Mapeo-Mobile
3. And I added the extract-message as a concurrent step with prettier in lint staged

PR #73, opened 2023-10-11
Es/device naming

ErikSin:
Adds Screens which allow user to name their device.

Conditionally renders the device naming screens if device name does not exist (held in persisted storage), and renders home screens if the name does exists

## To Do:
- [ ] Send name to back end on initialization

## Screen Shots

image

image

image

Closes [#1117](https://github.com/digidem/mapeo-mobile/issues/1117)


src/frontend/screens/DeviceNaming/GetStarted.tsx
```diff
@@ -0,0 +1,70 @@
...
+    defaultMessage: 'Collaborate on mapping and monitoring projects',
+  },
+  getStarted: {
+    id: 'screens.DeviceNaming.GetStarted.getStarted',
+    defaultMessage: 'Get Started',
+  },
+});
+
+const TopoLinesWidth = 360;
+const TopoLineHeight = 640;
```
achou11, line 23: Mostly interested in adding `original` somewhere, to make it clearer about they're representing. the screaming snakecase is just personal preference

```suggestion
const TOPO_LINES_SVG_ORIGINAL_WIDTH = 360;
const TOPO_LINES_SVG_ORIGINAL_HEIGHT = 640;
```

src/frontend/screens/DeviceNaming/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  function handleAddNamePress() {
+    if (invalidName) {
+      setErrorTimeout();
+      return;
+    }
+
+    navigation.navigate('Success', {deviceName: name});
+  }
```
achou11, line 57: mentioned in other comment, but i think the persistence of the device name should happen here and you should only navigate to the success screen after that happens without error

achou11 (changes requested):
looking good! some blocking comments that may require changes

src/frontend/screens/DeviceNaming/index.tsx
ErikSin: There is a semi annoying complication with conditional rendering of screens. If we set the name here the name is set, and these screen will not be rendered anymore (including the next screen). The conditional rendering is nice because it is declarative, aka the screens are based of the state of the device Name. 

We could make it imperative, and set the name here, and just navigate to the success screen, and then navigate to the map screen. It would just require the app to imperatively check the state on every open of the app. Aka, on the open of the app we would have to do the following at the top level:

```ts
// imperative
const deviceId = getDeviceInfo()
if (!deviceId) {
navigate("deviceNaming)
} else {
navigate("map)
}
```

ErikSin: Also ff we do it imperatively, after the hit `go to map`, they are navigated to the map. But if they hit the back button, those screen will be on the nav stack, so they would be able to go back to those screen. We could do a check on those screens:
``` ts
// in all device naming screens
useLayoutEffect(()=>{
   if(!!deviceId) navigate("map")
})
```

that feels equally as clunky as the AppState Event listener

PR #74, opened 2023-10-11
Es/settings

ErikSin:
Introduces Settings Screen. Only introduces the navigation for the setttings and nested settings. Subsequent PR will handle the navigation for the items in the settings screen.

## Screen Shots
image

image

image


src/frontend/sharedComponents/List/ListItemIcon.tsx
```diff
@@ -1,14 +1,29 @@
 import React from 'react';
 import {StyleSheet, View} from 'react-native';
+import Icon from 'react-native-vector-icons/MaterialIcons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
 
+type ListItemIconProps =
+  | {
+      iconName: string;
+    }
+  | {icon: React.ReactNode};
```
achou11, line 10: wondering if it would just be easier to pass the icon as a child? (or just using the `icon` prop here and always passing a react node)

guessing the reasoning for this was to account for using icons from a different icon set (e.g. material icons vs material community icons). subsequent question would be is it possible to just rely on the community icons for all of our icons, or do we need to be pulling from both? that way this change wouldn't be necessary to begin with, but i'm definitely lacking context so maybe it's not feasible for now

achou11 (approved):
looks good! one question but non-blocking

src/frontend/sharedComponents/List/ListItemIcon.tsx
ErikSin: I think in the future we should just pass an icon (not icon name) as a child, and the parent component can handle it. It does seem like we will be using a wider array of icons and svgs in the future, and that would allow for that to happen.

PR #75, opened 2023-10-12
Es/rename device

ErikSin:
Rename Device 

image

image

image

Closes [#1132](https://github.com/digidem/mapeo-mobile/issues/1132)


src/frontend/hooks/store/useDeviceNameStore.ts
ErikSin, line 1: the save button is located in the header, which is a parent component of the screen. So i am lifing the state of the form using zustand so the header can access it

ErikSin:
closing as this is being replaced by #137

PR #76, opened 2023-10-13
approximation of api

ErikSin:


src/frontend/AppLoading.tsx
ErikSin, line 1: @achou11 lmk if this makes sense in terms of what you think app loading should look like

PR #78, opened 2023-10-17
Es/observation details

ErikSin:
Adds the observation fields (previously known as observation details) to the observation screen. Field can either be text answer, single select answer, or a multiple select answer.

## Text field:
image

## Single Select
image

## Multi Select
image


ErikSin:
Closing as this needs to be updated

PR #79, opened 2023-10-18
Es/invite flow

ErikSin:
Allows user to list devices on the same network, and invite them to the project!

### to note:
There seems to be a lot of files that were edited, but a large number of them are just SVGs (aka, this isn't a huge PR)

image

image

image

image

image

image


src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+  const membersQuery = useProjectMembers();
+  const deviceInfo = useDeviceInfo();
+  const coordinators = React.useMemo(
+    () =>
+      !membersQuery.data
+        ? undefined
+        : membersQuery.data.filter(
+            member =>
+              member.capabilities.name === 'Coordinator' ||
+              member.capabilities.name === 'Project Creator',
```
ErikSin, line 68: is there a better way of determining the role? right now it is not strongly typed : `member.capabilities.name` is just type `string`

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -0,0 +1,83 @@
...
+      {/* Divider */}
+      <View style={{marginTop: 20}}></View>
+
+      {/* List available devices here */}
+      {devices &&
+        devices.map(device => {
+          const name = device.name;
+          const deviceId = device.deviceId;
+          // this is not exposed yet
+          const deviceType = 'mobile';
```
ErikSin, line 55: Do we expose the `deviceType` (whether it is `mobile` or `desktop`)?

src/frontend/screens/Settings/ProjectSettings/YourTeam/WaitingForInviteAccept.tsx
```diff
@@ -0,0 +1,100 @@
...
+    project.$member
+      .invite(route.params.deviceId, {roleId: route.params.role})
+      .then(() => {
+        queryClient.invalidateQueries({queryKey: ['projectMembers']}),
+          navigation.navigate('InviteAccepted', route.params);
+      })
+      .catch(err => {
+        openSheet();
+      });
+  }, []);
```
ErikSin, line 51: The architecture is a little odd here. But essentially, when the coordinator clicks to invite someone, they are navigated to this page to wait for the invite. So im deferring the actual invite till after the navigation has happened so we have access to the promise. Otherwise, the user clicks invite, the invite promise is created, and then the user is navigated to this page. Since react-nav doesnt allow for non-serializable things to be passed between pages, it would not have access to the promise.

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+      {coordinators &&
+        coordinators.map(coordinator => (
+          <DeviceCard
+            key={coordinator.deviceId}
+            style={{marginTop: 10}}
+            name={coordinator.name || ''}
+            deviceId={coordinator.deviceId}
+            deviceType="mobile"
+            // This is a weak check. We should be using deviceIds, but those are not exposed
```
gmaclennan, line 136: Good point, fix coming in https://github.com/digidem/mapeo-core-next/pull/454

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -0,0 +1,83 @@
...
+  },
+});
+
+export const SelectDevice: NativeNavigationComponent<'SelectDevice'> = ({
+  navigation,
+}) => {
+  const ssid = useLocalDiscoveryState(state => state.ssid);
+  const {formatMessage: t} = useIntl();
+
+  const devices = useLocalPeers();
```
gmaclennan, line 35: Note, this will include devices that are already members of the project until #167 is fixed

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewInvitation.tsx
```diff
@@ -0,0 +1,85 @@
...
+});
+
+// this should come from mapeo core ideally
+
+export const ReviewInvitation: NativeNavigationComponent<
+  'ReviewInvitation'
+> = ({route, navigation}) => {
+  const {formatMessage: t} = useIntl();
+  const {role, ...rest} = route.params;
+  const project = useProject();
```
gmaclennan, line 45: Unused

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectInviteeRole.tsx
```diff
@@ -0,0 +1,114 @@
...
+          })
+        }
+      />
+      <RoleCard
+        style={{marginTop: 10}}
+        role="coordinator"
+        onPress={() =>
+          navigation.navigate('ReviewInvitation', {
+            ...route.params,
+            role: 'coordinator',
```
gmaclennan, line 62: This is used in the `WaitingForInviteAccept` component as the roleId, which is a unique ID. This should be renamed maybe, and use the `roleId`, however I'm not sure if we correctly expose this in the API. I think the API is confusing here, and maybe we can quickly jump on a call to discuss how to improve it.

gmaclennan (changes requested):
Good work on this, it's looking good, however I think the approach of sending an invite as a side-effect of rendering the `WaitingForInviteAccept` screen is prone to bugs (especially as we don't have a meaningful way to cancel invites yet).

There are a couple of issues this throws up around the limitations of the API, and I've created some issues for those. I think the role names and IDs also needs cleaned up - it's created to be forwards compatible with custom role definitions, but I think it's not clear how to use it right now and maybe we are not exposing enough from the backend. It would help to discuss this together and figure out how to adapt the API, and the best workaround for the meantime.

gmaclennan (changes requested):
Good work on this, it's looking good, however I think the approach of sending an invite as a side-effect of rendering the `WaitingForInviteAccept` screen is prone to bugs (especially as we don't have a meaningful way to cancel invites yet).

There are a couple of issues this throws up around the limitations of the API, and I've created some issues for those. I think the role names and IDs also needs cleaned up - it's created to be forwards compatible with custom role definitions, but I think it's not clear how to use it right now and maybe we are not exposing enough from the backend. It would help to discuss this together and figure out how to adapt the API, and the best workaround for the meantime.

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
gmaclennan: Yes good point, we need to think this through, I'd like to loop @EvanHahn in and maybe have a quick chat about it with you.

src/frontend/hooks/server/projects.ts
```diff
@@ -9,3 +10,13 @@ export function useProject() {
...
   return projectContext.project;
 }
+
+export function useProjectMembers() {
+  const project = useProject();
+  return useQuery({
+    queryFn: async () => {
+      return await project.$member.getMany();
+    },
+    queryKey: ['projectMembers'],
```
gmaclennan, line 20: I think we should add an optional [`refetchInterval`](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery) to this hook, so that we can poll the backend to changes in this list. Project membership could change due to another device inviting a different device - that information would sync in the background with this device. I think updating this list of project members every 2-3 seconds would be ok. This is most relevant in training situations with many devices potentially inviting each other.

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
EvanHahn: Happy to chat about this any time.

src/frontend/screens/Settings/ProjectSettings/YourTeam/WaitingForInviteAccept.tsx
ErikSin: I updated this.

I dont think we need to worry about all 4 states in one screen. One screen just needs to handle sending the invite, and tracking the promise that comes along with it. As soon as the promise is resolved or rejected, we can just navigate to a new page. So iI[updated the page](https://github.com/digidem/CoMapeo-mobile/blob/4ceee0526952470d6cab3be8a92ec878c7fd1afd/src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewInvitation.tsx#L50) to do that.

Unfortunately we don't get the transition animation between sending and invite and waiting for an invite that react navigation gave us, but for now I think that is ok.

ErikSin:
I tested this with 2 devices today. Unfortunately I do not think peer discover is running. I had both devices running the dev environment (I learned today that if you plug in 2 devices, metro will simultaneouly run on both devices). `api.on('local-peers', ...)` is not being called when there is another device around.

@gmaclennan

src/backend/patches/@mapeo+core+9.0.0-alpha.4+001+fix-quickbit-dynamic-require.patch
```diff
@@ -21,3 +21,68 @@ index d4cfdfd..41cc2db 100644
...
+-    if (!address) {
+-      this.#log(`service up (${service.name}) with no ipv4 addresses`)
++    // const address = service.addresses?.filter(isIpv4)[0]
++    const { name, port, referer } = service
++    if (!referer) {
++      this.#log(`service up (${service.name}) had no referer`)
+       return
+     }
++
++    const { address } = referer
```
ErikSin, line 51: Evan and I discovered that `address` was showing up in `referer`, but not in `service`. @gmaclennan

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewInvitation.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const {openSheet, ...restBottomSheet} = useBottomSheetModal({
+    openOnMount: false,
+  });
+  const project = useProject();
+
+  function sendInvite() {
+    setInviteSent(true);
+    project.$member
+      .invite(rest.deviceId, {roleId: role})
+      .then(val => navigation.navigate('InviteAccepted', {...route.params}))
```
ErikSin, line 54: This is timing out after 5 seconds:

`Error: Server timed out after 5000ms.
            The server could be closed or the transport is down.`
            
Otherwise, the other device is recieving the invite and is able to accept it and join the project @gmaclennan

src/backend/patches/@mapeo+core+9.0.0-alpha.4+001+fix-quickbit-dynamic-require.patch
gmaclennan: Hmm I have a feeling that the changes in this patch will break things in other network environments, but if it’s enough to get it working for initial testing then that’s ok for now.

ErikSin: This doesn't change how the front end is being implemented, so if we need to revert it and fix it another way, it shouldn't make a difference

gmaclennan (approved):
Thanks Erik, I mainly focussed on reviewing the logic - I have not tested this in the emulator or on a device. There are a few small things, but nothing blocking this I don't think - we can fix things up in a follow-up as long as you open the issues. Two questions here about text size accessibility, and possible error when cancelling an invite then the invite errors. I think cancellable invites is an important missing feature from the backend that we need to address.

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewInvitation.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const queryClient = useQueryClient();
+
+  function sendInvite() {
+    setInviteSent(true);
+    project.$member
+      .invite(rest.deviceId, {roleId: role})
+      .then(val => {
+        if (val == 'ACCEPT')
+          queryClient.invalidateQueries({queryKey: ['projectMembers']});
+        navigation.navigate('InviteAccepted', {...route.params});
```
gmaclennan, line 59: Just realized there is no code path for when an invite is rejected, nor do we have a screen for that. Maybe a follow-up?

gmaclennan (changes requested):
Sorry I missed something in my first review. An invite response can be `ACCEPT`, `REJECT`, `ALREADY` (device being invited is already part of the project) or `UNRECOGNIZED` (this is an unexpected/unknown response, I think actually the method should throw instead of resolving with this response)

src/frontend/sharedComponents/Text.tsx
```diff
@@ -8,7 +8,7 @@ export const Text = ({
   ...otherTextProps
 }: React.PropsWithChildren<TextProps>) => (
   <RNText
-    style={[{fontFamily: 'Roboto', color: BLACK}, style]}
+    style={[{fontFamily: 'Roboto', color: BLACK, fontSize: 16}, style]}
```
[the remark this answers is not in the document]
achou11: as long as [`allowFontScaling`](https://reactnative.dev/docs/text#allowfontscaling) is enabled (by default it is) and you don't make any changes to [`maxFontSizeMultiplier`](https://reactnative.dev/docs/text#maxfontsizemultiplier), font scaling should respect the device settings, using the set `fontSize` as the base size when the font scale is the OS default.

Example using this branch:

- Before (default OS font scale)

    image

- After (OS font scale set to 2x):

    image

src/frontend/screens/Settings/ProjectSettings/YourTeam/WaitingForInviteAccept.tsx
```diff
@@ -0,0 +1,86 @@
...
+            navigation.navigate('YourTeam');
+          }}
+        />
+      </View>
+      <ErrorModal
+        sheetRef={sheetRef}
+        closeSheet={closeSheet}
+        isOpen={isOpen}
+        clearError={() => navigation.navigate('YourTeam')}
+      />
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/6456542191246db03cc9ee3ac2fca01121940278

package.json
```diff
@@ -25,7 +25,7 @@
     "@formatjs/intl-pluralrules": "^5.2.4",
     "@formatjs/intl-relativetimeformat": "^11.2.4",
     "@gorhom/bottom-sheet": "^4.5.1",
-    "@mapeo/ipc": "^0.1.3",
+    "@mapeo/ipc": "^0.1.5",
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/1422058b06ff139d9c986d9a48f9d825e8bdf511

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+  },
+});
+
+export const YourTeam: NativeNavigationComponent<'YourTeam'> = ({
+  navigation,
+}) => {
+  const {formatMessage: t} = useIntl();
+  const membersQuery = useProjectMembers();
+  const deviceInfo = useDeviceInfo();
+  const coordinators = React.useMemo(
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/6270848df21da033e3cc15f56f2495d2ddc3b8a5

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Text>{ssid}</Text>
+      </View>
+      <Text style={{marginTop: 10}}>{t(m.notSeeingDevice)}</Text>
+      <Text style={{marginLeft: 10}}>{`\u2022 ${t(m.sameWifi)}`}</Text>
+      <Text style={{marginLeft: 10}}>{`\u2022 ${t(m.sameVersion)}`}</Text>
+      {/* Divider */}
+      <View style={{marginTop: 20}}></View>
+
+      {/* List available devices here */}
+      {devices &&
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/982d425bba1cc26564698a248b477df56ef4a932

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const DeviceNameWithIcon = ({
+  deviceType,
+  name,
+  deviceId,
+  thisDevice,
+  iconSize,
+  style,
+}: DeviceNameWithIconProps) => {
+  const {formatMessage} = useIntl();
+  console.log({name});
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/853a09940c16ecb004793421af78e35e845937b9

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectInviteeRole.tsx
```diff
@@ -0,0 +1,119 @@
...
+    defaultMessage:
+      'As a Coordinator this device can invite and remove users, and manage project details.',
+  },
+});
+
+export const SelectInviteeRole: NativeNavigationComponent<
+  'SelectInviteeRole'
+> = ({route, navigation}) => {
+  const {formatMessage: t} = useIntl();
+  console.log({route});
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/79/commits/853a09940c16ecb004793421af78e35e845937b9

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewInvitation.tsx
ErikSin: Yeah, I discussed this with Sabella. We are just going to show the error screen with a message. Ill create a ticket and we can do it as follow up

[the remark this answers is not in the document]
ErikSin: We could also add an extra layer, where when the invitee accepts, the device inviting then also has to confirm that they still want to do invite them. I don't know the extent of how invites are broadcasted, but perhaps a simpler alternative than broadcasting that an invite is cancelled

ErikSin:
> Sorry I missed something in my first review. An invite response can be `ACCEPT`, `REJECT`, `ALREADY` (device being invited is already part of the project) or `UNRECOGNIZED` (this is an unexpected/unknown response, I think actually the method should throw instead of resolving with this response). Currently all responses result in navigation to the InviteAccepted screen. I think maybe we need to show the different responses on that screen.

I updated so it only navigates on accept. Ill create a ticket for handling the other states (Im just going to throw them and pass a message to the error screen)

Issue #54, opened 2023-09-18
React Lifecycle, Server state, and Splash Screen


ErikSin:
Listen to the nodejs-mobile process and only renders children when the server is up and running and listening.
On initial app load it waits for the server before dismissing the native splash screen.
If it doesn't hear a heartbeat message for timeout, it displays a screen to the user so they know something is wrong.

## To do
- [x] Frontend initiates the backend using the nodejs channel messaging (done in Splash Screen)
- [x] Wait for the start messaage before showing the app (done in Splash Screen)
- [x] Show error message if it does not correctly start
- [x] When the app goes into the background, [pause the server, and when it comes back to the foreground, resume the server
](https://code.janeasystems.com/nodejs-mobile/react-native/bridge#rn_bridgeapponevent-callback)

PR #87, opened 2023-10-24
initial frontend integration of Mapeo Core Next

achou11:
Towards #54 

Follow-up for #81 to actually get things linked with the app. 

Most of the changes are related to introducing queries + hooks to get ipc communication functional. Some areas that could potentially use improvements:

- ipc communication. Erik mentioned that there are some subtle issues with the client api usage. my guess is that it would be related to the ipc setup
- app background + foregrounding. i did some basic work but probably not enough to cover more specific scenarios
- metro bundle refreshes apparently cause the app to crash...


src/backend/src/app.js
```diff
@@ -1,61 +1,83 @@
...
+  const state = {
+    activeProjectId: null,
+  }
 
+  // 1. Initialize Mapeo
   const manager = new MapeoManager({
     rootKey: KeyManager.generateRootKey(),
+    // TODO: Use actual file storage instead of memory
     dbFolder: ':memory:',
     coreStorage: () => new RAM(),
```
achou11, line 36: not exactly sure what we want to specify here so keeping this to memory at the moment. any input @gmaclennan ?

src/backend/src/server.js
```diff
@@ -0,0 +1,47 @@
...
+  /** @type {{ close: () => void } | null} */
+  #apiServer = null
+
+  /**
+   * @param {MapeoManager} mapeoManager
+   */
+  constructor(mapeoManager) {
+    this.#channel = new MessagePortLike()
+    this.#mapeoManager = mapeoManager
+    this.#status = new ServerStatus()
```
achou11, line 23: figured it makes sense to couple the status to this server instance, but not entirely sure. in mapeo mobile, they're kept separate but the coordination of their calls is similar to what's done in this class, hence my decision to colocate. open to input though

src/backend/src/app.js
```diff
@@ -1,61 +1,83 @@
...
   })
 
-  const { close } = createMapeoServer(manager, channel)
+  // Automatically create initial project if no projects exist yet
+  if ((await manager.listProjects()).length === 0) {
+    const projectId = await manager.createProject()
+    log(`Created initial project with id ${projectId}`)
+    state.activeProjectId = projectId
```
ErikSin, line 43: i know that I said storing the projectId in memory was fine for now...but Im realizing that it would be good to test having things persist. So can you persist the project id before we merge.

src/backend/src/app.js
```diff
@@ -1,61 +1,83 @@
...
+  // 2. Set up event listeners
+  rnBridge.channel.on('update-active-project-id', (id) => {
+    // TODO: Write to persisted file
+    state.activeProjectId = id
+  })
+
+  rnBridge.channel.on('get-active-project-id', async () => {
+    // TODO: Read from persisted file
+    rnBridge.channel.post('app:active-project-id', state.activeProjectId)
+  })
```
ErikSin, line 57: We should also create a listener that passes the `Manager.deviceInfo()` to the front end here (I don't think we need to use react-query in the front end as it should only be passed once and never changes).

src/frontend/sharedComponents/ApiLoading/index.tsx
ErikSin, line 1: Just a note so I don't forgot. We should include the logic to hide the splash screen here as well

ErikSin (changes requested):
just 2 small changes

src/backend/src/app.js
ErikSin: oh I realize now that this is accessible to the front end already with the `mapeoClient`...

gmaclennan:
Just jumping in here for a small comment I noticed while checking in: I don't think we should persist the active project in the backend. This is front end state, so should be persisted there, I think trying to persist this in the backend might create problems in the future

gmaclennan:
I'll be able to do a more thorough review on Monday, and I can hopefully help with some of the bugs as I suspect we have already dealt with a lot of them with the current Mapeo mobile, eg when metro bundled refreshes you need to remember that the front end code is going to reload, but the backend will not, so you need to handle that.

gmaclennan (changes requested):
Good work getting it this far, this integration work is complicated.

A few issues with this though:

- Starting and stopping the entire backend integration (e.g. the creation of the IPC client and the message channel) is unnecessary and will lead to a lot of tricky race conditions. The only thing we need to do (and not entirely convinced we actually need to, but best to be safe) is to stop listening on a network port when the app is in the background.
- It doesn't make sense (to me) to store the active project id in the backend - it's only really using the backend as a complicated persisted storage mechanism.
- I don't think we need (at this stage) all the server status setup with the heartbeat - this was all set up to workaround how the previous Mapeo worked. The only thing we need to wait for is the server listening, but the `getUrl` methods handle that, so there is no need for the frontend to worry about the backend state other than the initial startup.
- The front end was not actually sending a 'status' message to the backend, so if the backend started before the front-end attached the event listener (e.g. when you reload the react native side) it was freezing.

Since we're across tricky timezones I went ahead and wrote the code for fixing these things, rather than going back-and-forth which could take a day or two. I've pushed this as a new branch https://github.com/digidem/CoMapeo-mobile/tree/gm/integration-fixes which is on top of this branch, so if you want to incorporate those changes into this PR, you can do on this branch `git reset --hard gm/integration-fixes` (I think that's the command, but don't count on it - I use a git UI for this kind of stuff)

There are two main changes in my PR

1. simplifying server startup and status bfb2e2dc2e8b8332a4d6fe0eea9572e6779accfe
2. moving active project state to the front end 712fc47c93c1b4a46367032fb55016413d2994b9

I also tweaked some of the gradle settings to remove some build warnings

Let me know if me changing code as part of a review is helpful, or if you would prefer to just receive feedback and comments and I leave the code alone.

@ErikSin pinging you because my changes will affect the work you have stacked on this.

Some other notes on the way I've approached this:

- I want us to be able to unit test or render with Storybook or similar the components we use, so I've followed practice of passing in an api instance to the `<ApiProvider>` that can be replaced with a mock if needed
- I've removed all dependency on rnBridge from the `ApiProvider` and have a separate component (no context) for waiting for the backend to start. I don't attempt to render the app until the nodejs process has started (we can call mapeoApi methods before it starts because they are queued in the MessagePort, but on slow devices they could timeout waiting for nodejs to start up - we have seen it take several seconds some times)
- My `ProjectContext` is a bit messy - it's been a while since I wrote React code and I really don't like how much boiler plate is needed messing with useEffect! There might be a cleaner way to do this with react-query or with zustand, but what I have should work ok, it's just aesthetics I think.

achou11:
thanks for helping out with this @gmaclennan! the points you outlined make sense and I don't have any strong objections to them. Erik and I were going back-and-forth about the active project id persistence and I thought keeping it in backend made sense because of some edge cases, but 

> Let me know if me changing code as part of a review is helpful, or if you would prefer to just receive feedback and comments and I leave the code alone.

in this case I think it's easier to have just implemented the changes you were thinking of yourself instead of doing a back-and-forth. created a PR from your branch that points to this to make it easier to review for myself and eventually integrate here: https://github.com/digidem/CoMapeo-mobile/pull/90

---

Do you have any input on the database locations when initializing the MapeoManager on the backend? Basically what to do for the `dbFolder` and `coreStorage` opts? Guessing `coreStorage` would be using random-access-file but not sure what the configuration for that is, since I haven't used it.

src/frontend/contexts/ProjectContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+// If we don't have an activeProjectId set, then we use either the first project
+// returned by mapeo.listProjects() or we create a new default project
+async function getExistingProjectIdOrCreate(mapeoApi: MapeoClientApi) {
+  const projects = await mapeoApi.listProjects();
+  // We shouldn't normally get here, except when somehow we end up with an
+  // invalid active project id
+  if (projects[0]) return projects[0].projectId;
+  return mapeoApi.createProject();
+}
```
ErikSin, line 89: @gmaclennan This might cause a problem if the user does not have a way of switching a project (which is not possible for the mvp). 

If the user joins a new project, they will have 2 projects. The front end will save the id of the new project to persisted state. But if the persisted state cache gets cleared, this functions behaviour will put the user to the first project, which is presumably the default project. Since there is not a way of switching projects in the front end, the user will not be able to access the other project. 

Down the line, that will be ok because they will be able switch between projects. But for the MVP, we could possibly kick people out of their projects by accident.

gmaclennan: Is there a reason that persisted state would be cleared? Is it less likely to happen on the backend than the front end?

ErikSin: Good point, Im actually just assuming that is easier to clear the cache in the front end. A user can clear data cache and data from an app in android settings. That will delete persisted storage in the front end. I don't actually know the architecture of how information is persisted with the backend, so I don't know how those actions would effect it.

I'm happy to keep your architecture as well for simplicity. This is an edge case that won't last for long once we have projects.

gmaclennan: By default mmkv [stores data in files storage](https://github.com/Tencent/MMKV/wiki/android_advance#customize-location) (not cache storage). It would be deleted if the user deleted the app data, but not if they deleted the app cache. However the backend works the same way - there is no data that can persist across the user deleting app data. Therefore I think it's fine to persist this in the frontend - there is no extra security against deletion that comes with storing in the backend. We just need to be careful that the app does not delete this piece of persisted data

achou11: wasn't aware that mmkv doesn't use cache storage, which was my main concern around keeping the active project id persisted in the frontend. given that, think gregor's assertion about it living there make sense 👍

PR #89, opened 2023-11-01
Added root key and root key parsing

ErikSin:
This Pr introduces a secure root key being created in the front end, and sending the root key to the back end to initialize Mapeo Manager.

Stores the root key in [expo-secure-storage](https://docs.expo.dev/versions/latest/sdk/securestore/)

Creates the root key using [expo-crypto](https://docs.expo.dev/versions/latest/sdk/crypto/)

### Process of root-ket
- On initialization of the app, searches for root-key in secures storage
   - If root key does not exit, creates a root key using expo crypto. This key is converted to a hex and stored in secure storage
- Sends root key to backend using `nodejs.startWithArgs`
- Root key is parsed and used to initialize `MapeoManager`


src/frontend/hooks/useSecureStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+    (val: string) => {
+      setStatus('loading');
+      console.log(val);
+      setItemAsync(key, val)
+        .then(() => {
+          setValue(val);
+        })
+        .catch(err => {
+          throw new Error(err);
+        })
```
achou11, line 30: thoughts on introducing an `'error'` status and updating this to set that instead of throwing here? then that error status can be used to show some default error screen or something similar when this hook is used. note that the `finally` would have to be removed/updated if you do change this.

leaving this as is would require having the error boundary set up (which i don't think it is yet) or making sure you wrap any call to `setToSecureStore()` to catch and handle this thrown error

src/frontend/sharedComponents/ApiLoading/index.tsx
```diff
@@ -24,10 +27,17 @@ export const ApiLoading = ({children}: React.PropsWithChildren<{}>) => {
...
+    setRootKey(uint8ArrayToHex(getRandomBytes(16)));
+  }
+
   React.useEffect(() => {
-    // TODO: Use startWithArgs and pass in other args that the backend expects
-    nodejs.start('loader.js');
-  }, []);
+    if (!rootKey) return;
+
+    nodejs.startWithArgs(`loader.js --rootKey ${rootKey}`);
```
achou11, line 39: ideally the start would only be called once. in theory, this could be at risk calling start multiple times. 

not sure what the proper solution should be but for now, perhaps adding a ref to indicate that it's been called already and checking it here would make sense in addition to this?

src/frontend/hooks/useSecureStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+      })
+      .finally(() => {
+        setStatus('idle');
+      });
+  }, [key, getItemAsync]);
+
+  const setToSecureStore = useCallback(
+    (val: string) => {
+      setStatus('loading');
+      console.log(val);
```
achou11, line 23: Can remove

```suggestion
      console.log(val);
```

src/frontend/hooks/useSecureStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    getItemAsync(key)
+      .then(storageVal => {
+        setValue(storageVal);
+      })
+      .finally(() => {
+        setStatus('idle');
+      });
+  }, [key, getItemAsync]);
```
achou11, line 18: probably doesn't affect anything but kind of confusing to have `getItemAsync` here since it's a purely stable reference as an imported function (i.e. not defined within the component)

```suggestion
  }, [key]);
```

src/frontend/sharedComponents/ApiLoading/index.tsx
ErikSin: Is there a state from the backend that we could check that indicates whether nodejsmobile has been started? That seems like a better solution. 

If that's not possible, maybe it would be better to make sure that the root key cannot be changed. Like do an explicit check any time `setRootKey` is being run, and don't set it unless the `rootKey` Value is null

achou11: > Is there a state from the backend that we could check that indicates whether nodejsmobile has been started? That seems like a better solution.

ah yeah now that you mention it, we use the server status to initiate the nodejs mobile script:

https://github.com/digidem/mapeo-mobile/blob/56528dcccf291d4248beb6348767ca50a6ce92be/src/frontend/AppLoading.tsx#L77-L81

so perhaps that should be incorporated here

ErikSin: So does [this](https://github.com/digidem/CoMapeo-mobile/pull/89/commits/47fbe4cd6d19b0dd9628cd1c0ed9271442a32ea5) make sense?

src/frontend/hooks/useSecureStorage.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/89/commits/e8deba6c5f0d8705f606d5e1e2a158e2cdc706d2

src/frontend/hooks/useSecureStorage.ts
ErikSin: addressed https://github.com/digidem/CoMapeo-mobile/pull/89/commits/e8deba6c5f0d8705f606d5e1e2a158e2cdc706d2

src/frontend/hooks/useSecureStorage.ts
ErikSin: I was just using it for trying to debug something, I didn't mean for it to end up in the code. Lets leave out the error for now? Maybe lets talk next week about adding some sort of error boundry. Now would be the time to do it given the state of the app.

ErikSin:
@achou11 this is has been rebased and updated

src/frontend/contexts/ProjectContext.tsx
```diff
@@ -5,15 +5,23 @@ import {usePersistedProjectId} from '../hooks/persistedState/usePersistedProject
 
 type MapeoProject = Awaited<ReturnType<MapeoClientApi['getProject']>>;
 
-type ActiveProjectContext = {
-  project: MapeoProject | undefined;
+type ActiveProjectContextType = {
+  project?: MapeoProject;
```
achou11, line 9: Not sure it'll make a difference in our case, but there is a notable distinction between the before and after:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAUFKYBOcBWEAxsAPwBcUAzsDgJYB2A5gNzoC+66okUAQgEMcyNJmx5CJSjXrMoAHygBXBgBMIAM0YRVbTuiJwGNKMErxEKVPsPHgpgEyVBwq+yA

basically the original definition ensures that the property exists on the value that's annotated with the type (even it it's undefined), which is helpful in certain cases. would lean towards keeping the original definition if it's not affecting anything

```suggestion
  project: MapeoProject | undefined;
```

src/frontend/hooks/persistedState/usePersistedProjectId.ts
```diff
@@ -2,8 +2,8 @@ import {type StateCreator} from 'zustand';
 import {createPersistedState} from './createPersistedState';
 
 type ProjectIdSlice = {
-  projectId: string | undefined;
-  setProjectId: (id: string | undefined) => void;
+  projectId?: string;
```
achou11, line 5: same comment about the TS optional field distinction. would lean towards the original definition because we know we instantiate the state with a `projectId: undefined` (see line 10 in this file)

```suggestion
  projectId: string | undefined;
```

src/frontend/hooks/useSecureStorage.ts
achou11, line 1: do we still need this?

src/frontend/App.tsx
```diff
@@ -22,12 +22,14 @@ import {QueryClient, QueryClientProvider} from '@tanstack/react-query';
...
 import {createMapeoClient} from '@mapeo/ipc';
-import {ApiLoading} from './sharedComponents/ApiLoading';
+import {ServerLoading} from './ServerLoading';
 import {ActiveProjectProvider} from './contexts/ProjectContext';
+import {initializeNodejs} from './initializeNodejs';
 
 const queryClient = new QueryClient();
 const messagePort = new MessagePortLike();
 const mapeoApi = createMapeoClient(messagePort);
+initializeNodejs();
```
achou11, line 32: what happens in the app if this throws? we may need to catch the error and depending on how the app responds, do something to improve the ux in the case of failure

src/backend/index.js
```diff
@@ -5,16 +5,22 @@ import { init } from './src/app.js'
...
 }
 
 try {
   const args = parseArgs({ options })
 
   const { values } = args
 
+  if (typeof values.rootKey !== 'string') {
+    throw new Error('backend did not recieve root key from front end')
+  }
```
achou11, line 18: don't think this check is needed because of the `parseArgs()` call, which strictly checks the flags passed and errors if it doesn't match the expectations. 

```suggestion
  if (typeof values.rootKey !== 'string') {
    throw new Error('backend did not recieve root key from front end')
  }
```

src/frontend/initializeNodejs.ts
```diff
@@ -0,0 +1,23 @@
...
+      await setItemAsync(ROOT_KEY, newRootKey);
+      rootKey = newRootKey;
+    } catch (err) {
+      throw new Error(
+        `Error initializing root key: ${typeof err === 'string' ? err : ''}`,
+      );
+    }
+  }
+
+  nodejs.startWithArgs(`loader.js --rootKey ${rootKey}`);
```
achou11, line 22: I think the `=` is needed in order to appease node's `parseArgs()` without additional configuration (tried it out in the repl and ran into issues without it)

```suggestion
  nodejs.startWithArgs(`loader.js --rootKey=${rootKey}`);
```

achou11 (changes requested):
looking good! Only blocking comment is regarding the `startWithArgs()` call. All the others are generally non-blocking

src/backend/index.js
ErikSin: this also checks for type error. `rootkey` can technically be a boolean as well

achou11: the options configuration is set to make sure that we pass a string I think

https://nodejs.org/docs/latest-v16.x/api/util.html#utilparseargsconfig

src/frontend/initializeNodejs.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/89/commits/1a9a0ef294e7d4226dc3159d35cf89c0a973022e

src/frontend/hooks/persistedState/usePersistedProjectId.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/89/commits/1a9a0ef294e7d4226dc3159d35cf89c0a973022e

src/frontend/contexts/ProjectContext.tsx
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/89/commits/1a9a0ef294e7d4226dc3159d35cf89c0a973022e

src/backend/index.js
ErikSin: yeah, its typescript that's complaining though. The other option would be to use `ts-expect-error`, but this throw, while unnecessary, does feel a little clearer

achou11: ah got it! thanks for clarifying

src/frontend/App.tsx
ErikSin: Yeah, I think we should plan out where we want to put error boundaries (although this technically would not be caught by an error boundaries since it is outside the react component, so I don't actually know the best way to handle this).

gmaclennan: To do this:

```suggestion
function noop () {}
const nodejsInitPromise = initializeNodejs();
// make sure we catch so we don't have an uncaught error - but we want to store the original promise (above) so we can check if it rejected later
nodejsInitPromise.catch(noop)
```

Then in a component:
```js
React.useEffect(() => {
  nodejsInitPromise.then(() => {
    setState('initialized')
  }).catch(e => {
    setState('error')
  })
}, []) // <-- no need for dep, since function def is not inside component
```

Can show an error based on state, or render rest of the app once initialized.

PR #90, opened 2023-11-06
integration fixes

achou11:
Stacked on #87 

See https://github.com/digidem/CoMapeo-mobile/pull/87#pullrequestreview-1714729570


src/frontend/contexts/ProjectContext.tsx
achou11, line 1: as Gregor noted in the original comment, there's probably a more intricate way of doing this with zustand/react-query, but think the implementation is fine for now

achou11 (commented):
Overall looks good! haven't tested on device yet but guessing it will run as expected.

@ErikSin mind taking a look at the changes + testing on device. Will defer to you in terms of how + when this is merged into the other branch, since I know you wanted your rootkey work to be merged too. Gregor mentioned that this may affect your PR in terms of git conflicts, so whatever's easier for you is fine

PR #94, opened 2023-11-13
add backend database and index persistence

achou11:
Before we were just storing things in memory. This _should_ make it so that we save things to disk so that data survives app restarts

~The implementation is a bit more verbose than ideally necessary only because of an issue in the version of @mapeo/core that we're using. The next release will solve this and the implementation here will be very minimal as a result~ UPDATED


ErikSin:
I can test this on my device, but I perhaps @gmaclennan you should look at the backend code?

gmaclennan (approved):
Sorry I had reviewed this but not submitted.

This looks fine, but I have a non-blocking comment on semantics: The naming suggests that `coreStorage` is an index, which it isn't, this is where the actual user data is stored (e.g. observations etc) and is the most important data.

- `dbFolder`: this contains sqlite databases, but they have different functions:
  - Client/Manager sqlite DB persists important project keys and is necessary to access projects, also contains a materialized view of project settings.
  - Project sqlite DBs (one for each project) stores materialized views of core storage data. They can be rebuilt, but we don't have a way of exposing that yet.
- `coreStorage` There is a separate folder of storage for each project. Within a project there are two folders:
  - `corestore`: this is all the hypercore data, e.g. all the data in the project
  - `indexer`: this persists the indexing state for each core in `corestore`, it is used for indexing into the sqlite materialized views

I suggest naming these folders `sqlite-databases` and `hypercore-storage`, since that is the most accurate description of what is in the folders (for anyone debugging a phone and manually browsing files)

src/backend/src/app.js
```diff
@@ -45,12 +49,22 @@ export async function init({ version, rootKey }) {
   log('Starting app...')
   log(`Device version is ${version}`)
 
-  // 1. Initialize Mapeo
+  const privateStorageDir = rnBridge.app.datadir()
+  const dbDir = join(privateStorageDir, DB_DIR_NAME)
+  const indexDir = join(privateStorageDir, INDEX_DIR_NAME)
+
+  if (!existsSync(dbDir)) {
```
gmaclennan, line 56: It's best practice to avoid checking existence before an action, see https://nodejs.org/api/fs.html#fs_fs_exists_path_callback, although in this case I can't imagine a case where it would cause an issue. However calling mkdirSync with `recursive: true` will work regardless of whether the folder already exists

gmaclennan:
Since @achou11 is not around this week, I went ahead and added some changes. This is ready to merge imho @ErikSin if you need it this week.

PR #95, opened 2023-11-14
chore: add minification option for bundling backend

achou11:
Adds the option for minifying the backend bundle using https://github.com/egoist/rollup-plugin-esbuild#standalone-minify-plugin. Not enabled by default with this PR since it's helpful to not minify for debugging purposes, but we'll want to use this when creating actual distributed builds.

Basic testing saw the bundle size go from ~3 MB to ~1.5 MB

### DIFF
--- a/scripts/build-backend.mjs
+++ b/scripts/build-backend.mjs
@@ -1,9 +1,18 @@
 #!/usr/bin/env node
 
+import {parseArgs} from 'util';
 import {$} from 'execa';
 import path from 'path';
 import fs from 'fs';
 import {downloadPrebuilds} from './download-prebuilds.mjs';
+
+const {
+  values: {prod},
+} = parseArgs({
+  options: {
+    prod: {type: 'boolean'},
+  },
+});
 
 const $$ = $({stdio: 'inherit'});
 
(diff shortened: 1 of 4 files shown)


scripts/build-backend.mjs
```diff
@@ -1,10 +1,19 @@
 #!/usr/bin/env node
 
+import {parseArgs} from 'util';
 import {$} from 'execa';
 import path from 'path';
 import fs from 'fs';
 import {downloadPrebuilds} from './download-prebuilds.mjs';
 
+const {
+  values: {prod},
```
ErikSin, line 10: Where is this `prod` flag getting passed in?

achou11: it's not getting used anywhere at the moment but when we want to enable prod-specific things in this script (e.g. backend minifying), we'd run this using `--prod` (open to renames)

achou11: we may want to enable minifying by default (even in dev), which is what we do in mapeo mobile. mostly because it helps with app startup time

ErikSin (approved):
LGTM

PR #96, opened 2023-11-14
chore: small adjustments to backend initialization

achou11:


src/backend/index.js
```diff
@@ -2,25 +2,22 @@ import { parseArgs } from 'util'
...
-
 try {
-  const args = parseArgs({ options })
-
-  const { values } = args
+  const { values } = parseArgs({
+    options: {
+      version: { type: 'string' },
+      rootKey: { type: 'string' },
+    },
```
achou11, line 10: turns out you need to inline the options into parseArgs in order for proper type inference (as opposed to storing as a separate variable and passing it in)

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/16295fb55f1cf70f5fe6791ba3d87d0c7ef9066a/types/node/util.d.ts#L1492-L1493

ErikSin (approved):
tyepscript is confusing at times..

PR #97, opened 2023-11-14
feat: add splash screen

### LINKED ISSUE #54: React Lifecycle, Server state, and Splash Screen

achou11:
Closes #54
Towards #100 

Uses [`react-native-bootsplash`](https://github.com/zoontek/react-native-bootsplash) after experimenting with both [`expo-splash-screen`](https://github.com/expo/expo/tree/main/packages/expo-splash-screen/) and [`react-native-splash-screen`](https://github.com/crazycodeboy/react-native-splash-screen).

TODO:

- ~replace the image assets with what we actually want on the screen~ done
- ~investigate if it would be possible to add the version pill via a very hacky `windowSplashScreenBrandingImage`~ not possible since it's only supported for Android 12+

---

Preview:

https://github.com/digidem/CoMapeo-mobile/assets/18542095/245a859d-3f67-4952-9a87-353d0824ed72


ErikSin (commented):
So this doesn't match Sabella's original sizing. I am not sure if you discussed this with Sabella.

It's not really possible to get it much bigger with bootsplash. But I think we can achieve a larger image with [react native spash screen](https://github.com/crazycodeboy/react-native-splash-screen). 

If Sabella is ok with this sizing, then lets keep it, otherwise we should export react-native splash screen.

achou11:
> So this doesn't match Sabella's original sizing. I am not sure if you discussed this with Sabella.
> 
> It's not really possible to get it much bigger with bootsplash. But I think we can achieve a larger image with [react native spash screen](https://github.com/crazycodeboy/react-native-splash-screen).
> 
> If Sabella is ok with this sizing, then lets keep it, otherwise we should try react-native splash screen.

Yeah I discussed with Sabella and she's fine with it. I have hesitations around react-native-splash-screen due to changes in React Native + Android. Also, I _think_ scaling would be tricky given its approach

ErikSin (commented):
Do you think it would make sense to let [React Navigation handle the splash screen](https://github.com/zoontek/react-native-bootsplash#how-should-i-use-it-with-react-navigation)?

achou11:
> Do you think it would make sense to let [React Navigation handle the splash screen](https://github.com/zoontek/react-native-bootsplash#how-should-i-use-it-with-react-navigation)?

Potentially! guess only concern is if it would be a significant amount of time added before the splash screen disappears, but worth trying

ErikSin:
> Potentially! guess only concern is if it would be a significant amount of time added before the splash screen disappears, but worth trying

I tried both methods on my device, and timed it, and the timing was identical

ErikSin (approved):

achou11 (commented):
re: c63b03441a66f94861cd26ca89ab8ce09549d330

turns out we actually can't do this in our case because of the way we conditionally render based on the server status. this change prevents the bootsplash from disappearing if an error server status is returned when initializing the app

PR #98, opened 2023-11-15
chore: add prestart npm script

achou11:
Adds a `prestart` npm script, which will run before `start`. Moves the translations and intl build steps into this step.

Not a very necessary change but felt a bit cleaner, especially if you wanted to do the 2 steps above but not actually start react native (for whatever reason)


ErikSin (approved):
Thanks! That does feel much cleaner.

PR #104, opened 2023-11-22
call manager start() and stop() in app init

achou11:
Haven't tested the changes on a device yet.

the latest version of core introduced `start()` and `stop()` methods on MapeoManager, which are currently used to handle the media server. due to this, i believe we should reintroduce using the rnBridge app lifecycle hooks, which presents a couple of questions:

- Do we need more variants of a server status now? e.g. `stopping`, `stopped`
- Does the message port need to also be started/stopped in the rnBridge lifecycle hooks?
  - may need to introduce a `pause()` method on the MessagePortLike if so


src/backend/src/app.js
```diff
@@ -53,6 +52,19 @@ export async function init({ version, rootKey }) {
     coreStorage: () => new RAM(),
   })
 
+  await manager.start()
+
+  rnBridge.app.on('pause', async (pauseLock) => {
+    log('App went into background')
+    await manager.stop()
+    pauseLock.release()
+  })
```
achou11, line 61: does `messagePort` need to be stopped here too?

src/backend/src/app.js
```diff
@@ -53,6 +52,19 @@ export async function init({ version, rootKey }) {
...
+
+  rnBridge.app.on('resume', () => {
+    log('App went into foreground')
+    manager.start()
+  })
```
achou11, line 66: does messagePort need to be started here too?

gmaclennan:
> * Do we need more variants of a server status now? e.g. `stopping`, `stopped`

No, because the "server" in this "server status" refers to the IPC server, not the HTTP server, and that is not stopping. The `getUrl` methods internally wait for the server to start before resolving, so whether the HTTP server is stopped should not be relevant to using the API. The edge case is a "stale" URL - e.g. the frontend has a URL from before app pause, and when the app resumes it tries to load that URL. If the port has changed then that will not work. We can discuss how to resolve that in a separate issue.

> * Does the message port need to also be started/stopped in the rnBridge lifecycle hooks?

No, the message port doesn't open any ports or use any resources, so there is no need to start or stop it.

>   * may need to introduce a `pause()` method on the MessagePortLike if so

No need I don't think.

src/backend/src/app.js
```diff
@@ -53,6 +52,19 @@ export async function init({ version, rootKey }) {
     coreStorage: () => new RAM(),
   })
 
+  await manager.start()
```
gmaclennan, line 55: No need to await this I don't think, getUrl methods will wait internally. We might want to catch errors on these start and stops, however the unhandled rejection handler will do that and probably suffice for now unless we have a different message we want to show to the user in this case.

src/backend/src/app.js
gmaclennan: no

src/backend/src/app.js
gmaclennan: no

gmaclennan (approved):
Looks good. One non-blocking perf issue - no need to await http server start before initializing the IPC server and starting the message port.

achou11:
> No, because the "server" in this "server status" refers to the IPC server, not the HTTP server, and that is not stopping

ah that makes sense, thanks for the clarification!

src/backend/src/app.js
gmaclennan, line 55:
```suggestion
// Don't await, methods that use the server will await this internally
manager.start()
```

gmaclennan:
@ErikSin this is also good to merge when you are ready.

PR #105, opened 2023-11-23
Coordinate format

ErikSin:
Created screen for user to choose coordinate format

Created persisted setting state to save format and propagate throughout the app.

image


achou11 (approved):
lgtm!

PR #106, opened 2023-11-27
fix: Hide splashscreen after permissions and project name determined

achou11:
Changes:

- fixes a general issue with the splashscreen hiding in certain cases (see https://github.com/digidem/CoMapeo-mobile/pull/97#pullrequestreview-1750882320)
- deletes permission context. Replaces it with asking user permissions at the app.tsx level, and using the expo hooks to determine if the permissions are granted.
- ~~Created a zustand store to set and determine if all the appropriate properties are loaded, which is used to determine when to hide the splash screen. Due to the device naming flow, the navigation needs to know if a device name has been created. Whether it has or has not been created conditionally renders different screens. Therefore the api to determine if a device name has to be resolved before the `appNavigator` determines which screen needs to be rendered, and therefore needs controls the splash screen. This zustand store allows the `appNavigator` to access all the other properties related to closing the splash screen~~ Realized that i could just pass this as a prop as `appNavigator` is exposed in `app.tsx`
- Observation context was removed. The observation context was waiting for observations api to resolve before rendering its children, and rendering a loading state while waiting for this observation api to resolve. Since the `appNavigator` is a child of the `ObservationContext` This loading and rendering of children was happening while the splash screen was up, since the `appNavigator` controls when the splash screen is closed. This could potentially mean very long wait times as it waits for all the observations to be loaded. In order to get rid of this context, the following things needed to happen
     - The observation and presets queries were converted to [useSuspenseQuery](https://tanstack.com/query/latest/docs/react/guides/suspense) to handle the loading state that the observationContext was previously doing.
     - `useSuspenseQuery` does not have an `enabled` property. The `enabled` property was being used to wait for the `project`. In order to get around this, the `ActiveProjectContext` now only renders its children if `project!==undefined`.
     - `React.Suspense` was added around the `appNavigator` to handle the loading states


ErikSin:
Sorry @achou11, Im just seeing this now!

Thanks for getting rid of the provider, but I actually think we can also get rid of zustand and use the hooks that come out of the box with expo:

1. [useCameraPermission](https://docs.expo.dev/versions/latest/sdk/camera/#usecamerapermissionsoptions)
2. for location: [useForegroundPermission](https://docs.expo.dev/versions/latest/sdk/location/#useforegroundpermissionsoptions)

We can request the permission at the app.tsx level.

This will get rid of all the boiler plate and there is no plan to use any tools outside of expo that require permissions (aka, the next thing will be the microphone, which will most likely use expo for)

src/frontend/contexts/ProjectContext.tsx
```diff
@@ -2,11 +2,12 @@ import * as React from 'react';
 import {type MapeoClientApi} from '@mapeo/ipc';
 import {useApi} from './ApiContext';
 import {usePersistedProjectId} from '../hooks/persistedState/usePersistedProjectId';
+import {Loading} from '../sharedComponents/Loading';
 
 type MapeoProject = Awaited<ReturnType<MapeoClientApi['getProject']>>;
 
 type ActiveProjectContextType = {
-  project: MapeoProject | undefined;
+  project: MapeoProject;
```
ErikSin, line 10: I made it so that the context returns a loading state when undefined, and only renders its children when the mapeo project has been initialized. this allows us to use `useSuspenseQuery` which doesn't have an `enabled` property

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -28,13 +33,20 @@ import {useNavigationFromRoot} from '../hooks/useNavigationWithTypes';
 
 export const AppNavigator = () => {
   const {formatMessage} = useIntl();
-  const navigation = useNavigationFromRoot();
+  const isLoaded = useEverythingLoaded();
+  const deviceName = useDeviceName();
+
+  if (isLoaded && !deviceName.isPending) {
+    BootSplash.hide();
+  }
```
ErikSin, line 41: this will be used in a subsequent pr to determine what should be loaded: either the map, or the device naming screens

achou11 (commented):
@ErikSin I wasn't aware that these changes would be coming into this PR but probably would've been better to separate into a stacked one (no worries though). ~do you mind summarizing the react-query changes here for posterity?~ EDIT: ah i see you updated the PR description, thanks!

some notes/questions i have:

- seems like we should have a proper error boundary set up if we're planning to use suspense and the suspense queries. fine to do as a follow-up i guess
- i'm a little unfamiliar with suspense as a whole, but seems like this + react query's suspense hooks are basically a (maybe?) cleaner way of writing data-fetching components without all of the status checking boilerplate when using plain react query. does that sound about right? is there anything notably different about doing all of this with suspense vs the "normal" way? just wondering if using suspense somehow locks us into a specific way of writing react code (for better or worse). 

is there a better name + description for this PR at this point? feels like it's moved way beyond just fixing the permissions issue 😅

ErikSin:
> @ErikSin I wasn't aware that these changes would be coming into this PR but probably would've been better to separate into a stacked one (no worries though).

I thought about doing that but everything in the PR is just related to the splash screen. There is alot of work done with react-query, but all that was to make sure the splash screen was timed properly.

>     * seems like we should have a proper error boundary set up if we're planning to use suspense and the suspense queries. fine to do as a follow-up i guess

I agree. I think we need to determine where to put the error boundaries though so we have appropriate error handing that doesn't just blow up the app on errors. Ill create an issue so we can start a discussion.

>     * i'm a little unfamiliar with suspense as a whole, but seems like this + react query's suspense hooks are basically a (maybe?) cleaner way of writing data-fetching components without all of the status checking boilerplate when using plain react query. does that sound about right? is there anything notably different about doing all of this with suspense vs the "normal" way? just wondering if using suspense somehow locks us into a specific way of writing react code (for better or worse).

Suspense leverages [React-Suspense](https://react.dev/reference/react/Suspense). Essentially the parent suspense component deals with the loading state. Same architecture as Error-Boundary, where the errors are caught by the parent Error-Boundary Component, and that component determines what to render.  We don't actually have to use suspense for all our queries. It is really nice for observations because so many children use it, and now we don't have to code a loading state for all those components. But for all the other queries, i think using the plain queries are fine, and having the components using those queries determine what do with the loading state is what we should generally stick with. So I don't think suspense locks us into a specific way of code because we don't (and probably shouldn't) have to use it.

ErikSin:
> Suspense leverages [React-Suspense](https://react.dev/reference/react/Suspense). Essentially the parent suspense component deals with the loading state. Same architecture as Error-Boundary, where the errors are caught by the parent Error-Boundary Component, and that component determines what to render. We don't actually have to use suspense for all our queries. It is really nice for observations because so many children use it, and now we don't have to code a loading state for all those components. But for all the other queries, i think using the plain queries are fine, and having the components using those queries determine what do with the loading state is what we should generally stick with. So I don't think suspense locks us into a specific way of code because we don't (and probably shouldn't) have to use it.

I should add that suspense makes it very nice to use with typescript. The queries always return `query.data` because the loading state is handled by the parent. And therefore typescript know that the component is only rendered when the api call has been called and the promise has been resolved.

achou11:
> > Suspense leverages [React-Suspense](https://react.dev/reference/react/Suspense). Essentially the parent suspense component deals with the loading state. Same architecture as Error-Boundary, where the errors are caught by the parent Error-Boundary Component, and that component determines what to render. We don't actually have to use suspense for all our queries. It is really nice for observations because so many children use it, and now we don't have to code a loading state for all those components. But for all the other queries, i think using the plain queries are fine, and having the components using those queries determine what do with the loading state is what we should generally stick with. So I don't think suspense locks us into a specific way of code because we don't (and probably shouldn't) have to use it.
> 
> I should add that suspense makes it very nice to use with typescript. The queries always return `query.data` because the loading state is handled by the parent. And therefore typescript know that the component is only rendered when the api call has been called and the promise has been resolved.

cool, thanks for the explanations! can definitely see the value of leaning into it

PR #107, opened 2023-11-27
useLocationHook

ErikSin:
Replaced Location Context with a useLocationHook.

### Noteable changes
- [ ] Got rid location context
- [ ] Replaced asking for location permissions, which use to live in the location context, to App.tsx (The location context was initialized in the app.tsx level, so the permission were requested when App.tsx was initialized).
- [ ] Uses expo hook [useForegroundPermission](https://docs.expo.dev/versions/latest/sdk/location/#useforegroundpermissionsoptions) to ask for permissions instead of the react-native methods.
- [ ] Added `useLocation()` hook which subscribes to user location
- [ ] Added a debounce function, which only update react state if 
    - distance moved is greater than 15m OR 
    - the time is greater than `MaxTimeInterval` AND distance moved is greater than `minDistanceInterval`


src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+      const rulerLatDisplacement = Math.abs(rulerLat - coords[1]);
+      if (!ruler || rulerLatDisplacement > MAX_RULER_LAT_DISPLACEMENT) {
+        rulerLat = coords[1];
+        ruler = new CheapRuler(rulerLat, 'meters');
+      }
+      const distance = ruler.distance(coords, lastCoords);
+      const time = location.timestamp - lastLocation.timestamp;
+      if (
+        distance > maxDistanceInterval ||
+        (time > maxTimeInterval && distance > minDistanceInterval)
```
achou11, line 101: just to be a bit clearer about what this represents
```suggestion
      const timeElapsed = location.timestamp - lastLocation.timestamp;
      if (
        distance > maxDistanceInterval ||
        (timeElapsed > maxTimeInterval && distance > minDistanceInterval)
```

src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const [permissions] = useForegroundPermissions();
+
+  useFocusEffect(
+    React.useCallback(() => {
+      if (!permissions || !permissions.granted) return;
+
+      let ignore = false;
+      const LocationSubscriptionProm = watchPositionAsync(
+        {accuracy: Accuracy.BestForNavigation},
```
achou11, line 41: not sure if this is helpful or any clearer, but could set `distanceInterval: debounceOptions.minDistanceInterval`, which then saves part of a check in the debounce function (can remove the `distance > minDistanceInterval` part)

src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+        (time > maxTimeInterval && distance > minDistanceInterval)
+      ) {
+        lastLocation = location;
+        callback(location);
+      }
+    };
+  };
+}
+
+function getCoords(location: LocationObject) {
```
achou11, line 110: more useful to work with a tuple in this case (TS infers `number[]` without this)

```suggestion
function getCoords(location: LocationObject): [number, number] {
```

src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+        return;
+      }
+      const coords = getCoords(location);
+      const lastCoords = getCoords(lastLocation);
+      const rulerLatDisplacement = Math.abs(rulerLat - coords[1]);
+      if (!ruler || rulerLatDisplacement > MAX_RULER_LAT_DISPLACEMENT) {
+        rulerLat = coords[1];
+        ruler = new CheapRuler(rulerLat, 'meters');
+      }
+      const distance = ruler.distance(coords, lastCoords);
```
achou11, line 97: a little bit tricky to wrap my head around the ruler memoization. can understand why it might be useful to memoize, but curious if it justifies the additional code (the initialization for CheapRuler seems very inexpensive)

Basically wondering if this would be okay, in addition to removing `ruler` and `rulerLat`, and `MAX_RULER_LAT_DISPLACEMENT` +  adding the return type to `getCoords()`:

```suggestion
      const ruler = new CheapRuler(lastCoords[1], 'meters')
      const distance = ruler.distance(lastCoords, coords);
```

achou11 (approved):
Left some mostly non-blocking comments but overall seems like a good change.

src/frontend/hooks/useLocation.ts
ErikSin: no, unfortunately we cannot do that as that would mean the `debounce` function would only get called when the user has moved 15m. We are also updating the location if the user has moved more than 1m after 2 seconds. That wouldn't be called if we add `distanceInterval`

achou11: sorry I meant the 1 meter distance with my original comment, if that wasn't clear. but yeah given your reasoning, can disregard

ErikSin: wait, but them I think your right, we could do that...

achou11: ah ok, un-resolved this in case you want to consider it then 😄 my head keeps getting mixed up about the requirements. it may make sense, but also not the biggest deal left alone

src/frontend/hooks/useLocation.ts
ErikSin: It is called "cheap ruler". So I took out the memoization: https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a

src/frontend/hooks/useLocation.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a

src/frontend/hooks/useLocation.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a

src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+          if (ignore) return;
+          setLocation(location);
+        }),
+      );
+
+      return () => {
+        ignore = true;
+        LocationSubscriptionProm.then(sub => sub.remove());
+      };
+    }, [permissions]),
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a

src/frontend/hooks/useLocation.ts
```diff
@@ -0,0 +1,113 @@
...
+  useFocusEffect(
+    React.useCallback(() => {
+      if (!permissions || !permissions.granted) return;
+
+      let ignore = false;
+      const LocationSubscriptionProm = watchPositionAsync(
+        {accuracy: Accuracy.BestForNavigation},
+        debounceLocation({
+          ...debounceOptions,
+        })(location => {
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a

src/frontend/hooks/useLocation.ts
ErikSin: addressed with [this PR](https://github.com/digidem/CoMapeo-mobile/pull/107/commits/fa942647c4ca4cb1be9dfbf9cc3a53aed71d531a), But also adjusted the code, see [this comment](https://github.com/digidem/CoMapeo-mobile/pull/107/files#r1414621394) for reasoning

ErikSin:
Feel free to merge after reviewing if this is blocking your PR

PR #125, opened 2023-12-07
fix: return error state in useLocation

gmaclennan:
- keeps previous location in the case of an error
- doesn't return derived state (coord tuple) from the hook, just the
  single source of truth (LocationObject)
- small style adjustment: camelCase variable name.


ErikSin (approved):

ErikSin:
I'm happy to do it this way. But just wanted to mention that the `coords` object was a derived state from `location`. There was no state syncing happening, so it still is only one source of truth, the `location` variable.

PR #127, opened 2023-12-07
fix: fix freeze after hot reload / fast refresh

### LINKED ISSUE #113: Server Status getting stuck on `STARTING`
On metro refresh (eg, developer presses `r` in the terminal running metro, or an edit is made in react code) the screen goes blank. This is because [ServerStatus](https://github.com/digidem/CoMapeo-mobile/blob/503a1cdb1b976671394765a4579397c8c34602a8/src/frontend/ServerLoading.tsx#L12) is being set to null. this was determined by doing a `console.log(ServerStatus)`

The current workaround, is to go to the ServerLoading file, make a change (the change doesn't actually need to effect the code. Aka, I just add a space), and save the file. This causes the metro server to refresh again, solving the issue

Image

gmaclennan:
Fixes #113

After reload, the frontend is sending a `get-server-status` message,
which the nodejs process receives, and tries to send back the current
status. However this message back never reaches the frontend, likely due
to the way the RN bridge is re-created during a reload.

One fix that worked was to add a small delay when responding to
`get-server-status`, but analysing the logs showed that, from the nodejs
process perspective, the app is going into the background and back into
the foreground after the reload, and _after_ the `get-server-status`
message is received. The fix that this implements is to re-send the
server status whenever the app comes back into the foreground, this
seems to fix things because it seems like nodejs bridge messages sent
while the app is in the background do not get through.

### DIFF
--- a/src/backend/src/status.js
+++ b/src/backend/src/status.js
@@ -18,13 +18,16 @@ export class ServerStatus {
   #state = 'STARTING'
 
   constructor() {
-    rnBridge.channel.on('get-server-status', () => {
+    rnBridge.channel.on('get-server-status', async () => {
       log('status request -> ' + this.state)
-      rnBridge.channel.post(
-        'server:status',
-        /** @type {StatusMessage} */
-        { value: this.state },
-      )
+      this.#postState()
+    })
+    // Sometimes rnBridge.channel.post doesn't work if the app is in the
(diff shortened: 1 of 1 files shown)


achou11 (approved):
ah interesting! wouldn't have expected the fast refresh to technically cause the app to go into the background but good to know. thanks for looking into this and finding a small solution!

PR #128, opened 2023-12-08
created useLocationProviderStatus

ErikSin:
Creates a Seperate hook to determine if location status is enabled


gmaclennan (changes requested):
I'm not sure it makes sense to combine the location provider status with the useLocation hook. My understanding of the UX is that we would want to show this information in different places on the screen, and we would not want to set the location to `undefined` when the location provider status is false - that would mean the users location would disappear from the map. I think we can address the details of this later: perhaps using the timestamp of the last location update, and after a certain period of time has passed since the last location update then we can display the user location in a different colour to indicate it's "staleness", but I think we also show that location is turned off in the GPS pill.

PR #129, opened 2023-12-08
set debounce in react state

ErikSin:
Here is an alternative solution to the memory leak caused by the interval in useLocation()

### DIFF
--- a/src/frontend/hooks/useLocation.ts
+++ b/src/frontend/hooks/useLocation.ts
@@ -110,7 +110,10 @@ export function useLocation({
           accuracy: Accuracy.BestForNavigation,
           distanceInterval,
         },
-        debounceLocation(debounceOptions),
+        location => {
+          if (ignore) return;
+          debounceLocation(debounceOptions)(location);
+        },
       );
 
       locationSubscriptionProm.catch(error => {
--- a/src/frontend/hooks/useLocation.ts
+++ b/src/frontend/hooks/useLocation.ts
@@ -61,7 +61,7 @@ export function useLocation({
           if (!(await hasServicesEnabledAsync())) {
             lastLocation.current = undefined;
             setLocation({
-              location,
+              location: undefined,
               error: new Error('Location service is not enabled'),
             });
             clearInterval(interval.current);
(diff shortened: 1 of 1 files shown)


src/frontend/hooks/useLocation.ts
```diff
@@ -38,8 +38,68 @@ export function useLocation({
...
+          if (!(await hasServicesEnabledAsync())) {
+            lastLocation.current = undefined;
+            setLocation({
+              location,
+              error: new Error('Location service is not enabled'),
+            });
+            clearInterval(interval.current);
+            return;
+          }
+        }, LOCATION_TIMEOUT);
```
ErikSin, line 70: this needs to be an interval because `distanceInterval` is being used.

If its was a timer instead of an interval the following could happen:
- Interval set to 2 m
- User does not move more than 2 meter in 10 seconds. Because of this, the location callback is not fired for 10 seconds. 
- setTimeout() would be called. location service would still be enabled, so nothing would happen...
- user turns off location service
- Since the setTimeout() has already been called, we don't check again
- this hook would then return a stale location

This would be avoided with an interval as it would continue checking until the the location Service is disables OR the component has unmounted

gmaclennan (commented):
I think moving all this into React state complicates things and opens us up to hard-to-catch bugs, e.g. in the code here `debounceLocation` needs to be a dependency for the useEffect. There is a hard-to-reason race condition to deal with after `await hasServicesEnabledAsync()` and I fined it really hard to figure out what can happen during that await - I think at the least we need to check `ignore` before setting state, and probably we shouldn't clear the refs. However I think this approach makes us need to do all this work - I'll write a quick PR with an alternative approach that I hope will simplify things and reduce edge-case bugs.

src/frontend/hooks/useLocation.ts
gmaclennan: We're slightly muddling two approaches here:

1. We can continually poll the location provider state with a setInterval.
2. We can listen to position updates, and if we have not had a position update in X milliseconds, then check the location provider state.

This PR kind of mixes both approaches, which to me makes it hard to reason with and potentially introduces subtle bugs.

PR #133, opened 2023-12-11
chore:created useLastSavedLocation()

ErikSin:
Creates a react-query hook that retrieves the last known location (previously found in the location context).

### DIFF
--- a/src/frontend/Navigation/AppNavigator.tsx
+++ b/src/frontend/Navigation/AppNavigator.tsx
@@ -11,6 +11,7 @@ import {useIntl} from 'react-intl';
 import BootSplash from 'react-native-bootsplash';
 import {useDeviceName} from '../hooks/server/deviceInfo';
 import {Loading} from '../sharedComponents/Loading';
+import {usePrefetchLastKnownLocation} from '../hooks/useLastSavedLocation';
 
 // import {devExperiments} from '../lib/DevExperiments';
 
@@ -33,6 +34,7 @@ import {Loading} from '../sharedComponents/Loading';
 export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
   const {formatMessage} = useIntl();
   const deviceName = useDeviceName();
+  usePrefetchLastKnownLocation();
 
   if (permissionAsked && !deviceName.isPending) {
     BootSplash.hide();
--- a/src/frontend/hooks/useLastSavedLocation.ts
+++ b/src/frontend/hooks/useLastSavedLocation.ts
@@ -1,0 +1,10 @@
+import {useQuery} from '@tanstack/react-query';
+import {getLastKnownPositionAsync} from 'expo-location';
+
+export const useLastSavedLocation = () => {
+  return useQuery({
+    queryKey: ['lastPostition'],
+    queryFn: async () =>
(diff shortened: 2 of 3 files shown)


src/frontend/hooks/useLastSavedLocation.ts
```diff
@@ -0,0 +1,10 @@
+import {useQuery} from '@tanstack/react-query';
+import {getLastKnownPositionAsync} from 'expo-location';
+
+export const useLastSavedLocation = () => {
+  return useQuery({
+    queryKey: ['lastLocation'],
+    queryFn: async () =>
+      getLastKnownPositionAsync().catch(err => console.log(err)),
```
achou11, line 8: If error-handling or an error object isn't needed for us, maybe return `null` for return type simplification

> Gets the last known position of the device or `null` if it's not available or doesn't match given requirements such as maximum age or required accuracy.

```suggestion
      getLastKnownPositionAsync().catch(err => {
        console.log(err)
        return null
      }),
```

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -74,6 +59,10 @@ export const MapScreen = () => {
     setFollowing(true);
   }
 
+  if (savedLocation.isLoading) {
+    return <Loading />;
+  }
```
achou11, line 64: does the whole screen need to be blocked by this? what happens if this is omitted?

achou11 (approved):
couple of notes but seems reasonable overall

src/frontend/hooks/useLastSavedLocation.ts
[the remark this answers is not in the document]
ErikSin: I don't think to. There are 2 use cases: have a coordinate to show when setting the coordinate format and for showing a location on the map, while the actual location loads.

In both cases, the info doesn't actually matter (aka. were not saving the information). We use it because it is better than having undefined data. So setting a maxAge or requiredAccuracy, means there is a greater chance the information is null, which defeats the purpose of using this hook.

ErikSin: no! not needed, was just using it for debugging purposes. Got rid of it here: https://github.com/digidem/CoMapeo-mobile/pull/133/commits/87d01473ddb5ec3bba6f7b1f0e8e278d40d741f3

src/frontend/screens/MapScreen/index.tsx
ErikSin: its seems to be fine if we omit it. Addressed here: https://github.com/digidem/CoMapeo-mobile/pull/133/commits/e7a70b87055e9662767ad04ad3e00327e11d1575

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -33,6 +34,7 @@ import {Loading} from '../sharedComponents/Loading';
 export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
   const {formatMessage} = useIntl();
   const deviceName = useDeviceName();
+  usePrefetchLastKnownLocation();
```
ErikSin, line 37: this is a relatively inexpensive call AND prefetching avoids and jumpy ui in the map!

Issue #135, opened 2023-12-13
Add rules-of-hooks and exhaustive-deps eslint rules


gmaclennan:
We should add the eslint plugin to ensure we are following the [rules of hooks](https://legacy.reactjs.org/docs/hooks-rules.html). It's too easy to make an error and miss a dependency, which results in a hard to track bug.

achou11:
Already done. closing

PR #137, opened 2023-12-18
With react hook forms

ErikSin:
Screens to allow user to rename their device.

## To Note
- Introduces [React Hook form](https://www.react-hook-form.com/get-started/#ReactNative), which handles validation and submission of the form.
- Uses [optimistic updates](https://tanstack.com/query/latest/docs/react/guides/optimistic-updates#via-the-ui) to update the name before it is saved to the DB.

image

image

image

Closes [#1132](https://github.com/digidem/mapeo-mobile/issues/1132)


src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -17,12 +22,12 @@ export const useEditDeviceInfo = () => {
...
   return useMutation({
-    mutationKey: ['device'],
+    mutationKey: ['deviceInfo'],
     mutationFn: async (name: string) => {
       return mapeoApi.setDeviceInfo({name});
     },
-    onSuccess: () => {
-      queryClient.invalidateQueries({queryKey: ['deviceInfo']});
+    onSuccess: async () => {
+      return await queryClient.invalidateQueries({queryKey: ['deviceInfo']});
```
ErikSin, line 30: This allows for optimistic updates. Essentially it forces the cache to to become valid only AFTER it has resolved the promise of the API call. Otherwise it is considered a valid cache as soon as the call is made, even before the promise has resolved.

gmaclennan: Do we want optimistic updates? Previously we would show a saving state (it was a disabled state of the save button for observation edit - maybe not the best UI), and then navigate on success. We had discussed showing a "save success" message, but never got round to implementing it.

It's arguable that otpimisitc updates are more trouble and complexity than they are worth for a local-only app, where all mutation actions are going to be very quick, and they could result in a confusing user experience: save -> appears to have updated -> then see error message.

I think we need more UX work on this overall, perhaps for all forms which are async having states for "saving", "error saving" and "saved". However in the meantime maybe it's not worth the work on optimistic updates given it doesn't give us a clear way to handle errors.

src/frontend/screens/Settings/ProjectSettings/DeviceName.tsx
```diff
@@ -0,0 +1,144 @@
...
+          handleSubmit={handleSubmit(val => {
+            mutate(val.name);
+            setIsEditting(false);
+          })}
+          setIsEditting={setIsEditting}
+          isEditting={isEditting}
+        />
+      ),
+    });
+  }, [navigation, isEditting, handleSubmit, setIsEditting]);
```
gmaclennan, line 71:
```suggestion
  }, [navigation, isEditting, handleSubmit, setIsEditting, mutate]);
```

Can you fix #135 - bugs from something like this are really hard to track down.

src/frontend/screens/Settings/ProjectSettings/DeviceName.tsx
```diff
@@ -0,0 +1,144 @@
...
+        ) : (
+          // Optimistically updates the device name
+          <Text>{isPending ? newDeviceName : deviceInfo.data?.name}</Text>
+        )
+      ) : (
+        <React.Fragment>
+          <HookFormTextInput
+            name="name"
+            control={control}
+            rules={{required: true, maxLength: 60}}
```
gmaclennan, line 119: How are validation errors displayed to the user? E.g. what would happen if they entered a name over 60 characters? And what would happen if they tried to delete the device name and save?

gmaclennan (commented):
This looks good, react-hook-form is fairly easy to understand (reviewing this having never looked at it before).

An unrelated question though: what is your reasoning for using the same screen with two states (`isEditing: true | false`) rather than two separate screens (one `DeviceNameView` and one `DeviceNameEdit`). Previously in the app the "back" action during an edit cancels the edit and returns to the view. In this current implementation as I understand it the back action during an edit will return to the previous screen. It's possible to also interrupt the back action and change its behaviour based on the `isEditing` state, but it seems like you end up with a lot of code which is harder to read all to avoid having an extra screen.

Consider also simplifying the HeaderRight button into something more generic / re-usable, e.g. a save button with an `onPress` and an edit button with an `onPress`, and handling which is displayed in the main component. It seems like the current component is doing too much, which is fine for this screen, but when you start adding lots of similar screens with similar behaviour, you might end up needing to write lots of similar but slightly different header right components.

ErikSin:
> It's arguable that otpimisitc updates are more trouble and complexity than they are worth for a local-only app, where all mutation actions are going to be very quick, and they could result in a confusing user experience: save -> appears to have updated -> then see error message.

Generally I agree, and I don't think we should use optimistic updates that often. And in most places when we update something we navigate to a new page after the mutation happens. And actually the suggestion is to call a function that prefetches the cache right before the navigation. Errors will always happen before the prefetch, so we dont have to worry about the user seeing data that isn't saved and then seeing an error.

The reason I opted to a optimistic update in this case is because of the simplicity of the this data type (` {name:string}`) and because there is no navigation. Due to the simplicity of the data type it feels like there is a small likely hood of error. And with no navigation, there is instant feedback; In otherwords, as soon as the user clicks save, they see their device name instantaneously. Without the optimistc update the user sees a brief flash of their old name, then their new name. 

Since we are not error handling in this case, if there was an error and we optimistically updated, what would happen is the user would see the "new" name. Then the cache would be updated, but since there was an error and it didn't save, the cache would just be populated with the original name. It would cause a flicker of new name, then old name. I think there is a smaller chance of this happening, so i think it makes sense in this case to keep the optimistic update.

Also, the [optimistic update code](https://github.com/digidem/CoMapeo-mobile/blob/a7bfb675c0585cc81eaca3872dc334e6519242fc/src/frontend/screens/Settings/ProjectSettings/DeviceName.tsx#L111) is quite simple. It does get a little more complex if mutation is not collocated with component, but as I said before, I don't think we should use optimistic updates that often.

src/frontend/hooks/server/deviceInfo.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/137#issuecomment-1861325797

src/frontend/screens/Settings/ProjectSettings/DeviceName.tsx
ErikSin: [We highlight the input box as red](https://github.com/digidem/CoMapeo-mobile/blob/a7bfb675c0585cc81eaca3872dc334e6519242fc/src/frontend/screens/Settings/ProjectSettings/DeviceName.tsx#L123) if there is an error

ErikSin:
> An unrelated question though: what is your reasoning for using the same screen with two states (`isEditing: true | false`) rather than two separate screens (one `DeviceNameView` and one `DeviceNameEdit`). Previously in the app the "back" action during an edit cancels the edit and returns to the view. In this current implementation as I understand it the back action during an edit will return to the previous screen. It's possible to also interrupt the back action and change its behaviour based on the `isEditting` state, but it seems like you end up with a lot of code which is harder to read all to avoid having an extra screen.

I separated the screens. It surprisingly added some complications for how we navigate back. the react navigation life cycle is not that straighforward due to the limitation with native navigation. So before when the screens were one. I only had to worry about going back when the back button was pressed. So I could always intercept the back action. 

With this architecture of separated screens, when the user press saves, they have to navigate back to the previous screen (previously they would press save, and the state of the screen would change, but they didn't navigate back). This means i can't just intercept the back action, because the back action is different based on whether the user hits back or they hit save. If they hit save, we don't want to interrupt and ask if they want to "discard" or "continue".  And since there is no state associated with saving, there is no way of determining if the back action is from the save button or from the back button. I also can't synthetically create a `isEditting` state here. Due to lifecycle of react, if user a presses save, I should 1.mutate, 2. setIsEditting(false), 3. goBack(). But the goBack() is called before the next render, so the react is not able to update its state and the back action is not able to know that `isEditting` should be false

So to get around we need to program seperate back events based on which button is pressed. If the user pressed the back button or the hardware back button, we interuppt and ask to confirm. If the press the save button, we just go back.

gmaclennan:
> I separated the screens. It surprisingly added some complications for how we navigate back. the react navigation life cycle is not that straighforward due to the limitation with native navigation. So before when the screens were one. I only had to worry about going back when the back button was pressed. So I could always intercept the back action.

Sorry to add this complexity to your life! If you use `navigation.pop()` in the save button, then you could use the same code as you have before (`beforeRemove` handler) but use the check `event.type === 'pop'` to distinguish between a navigation action from save vs. hardware back or header back button.

ErikSin:
Closing to separate concerns and refactor

PR #139, opened 2023-12-19
Es/sync

ErikSin:


src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+export const Sync: NativeNavigationComponentNoHeader<'Sync'> = ({
+  navigation,
+}) => {
+  const ssid = useWifiName();
+  const sync = useProject().$sync;
+  const projectSettings = useProjectSettings();
+  const deviceInfo = useDeviceInfo();
+  const [connectedPeers, setConnectedPeers] = React.useState<number>(0); //this should initialize with `sync.getState().connectPeers`, but it is undefined
+  const [isSyncing, setIsSyncing] = React.useState(false);
+  const [hasDataToSync, setHasDataToSync] = React.useState<boolean>(); //this should initialize with `sync.getState().data.dataToSync`, but it is undefined
```
ErikSin, line 73: [getState()](https://github.com/digidem/mapeo-core-docs/tree/main/client-docs#syncgetstate) is not returning the right object. @gmaclennan

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const {formatMessage} = useIntl();
+
+  console.log(sync.getState()); // sync.getState() === {"_h": 0, "_i": 0, "_j": null, "_k": null, "clear": [Function anonymous]}
+
+  React.useEffect(() => {
+    const syncListener = (val: State) => {
+      setConnectedPeers(val.connectedPeers);
+      setIsSyncing(val.data.syncing);
+      setHasDataToSync(val.data.dataToSync);
+    };
```
ErikSin, line 83: is there a good way for me to catch any errors?

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+  if (!hasDataToSync) {
+    return (
+      <Button fullWidth onPress={() => {}}>
+        {formatMessage(m.caughtUp)}
+      </Button>
+    );
+  }
+
+  return (
+    <Button fullWidth onPress={startSync}>
```
ErikSin, line 187: Should the user be able to press [sync.start()](https://github.com/digidem/mapeo-core-docs/tree/main/client-docs#syncstart) even if there are no connected peers

src/frontend/screens/Sync/index.tsx
ErikSin, line 1: Should i [wait for the `initial` sync](https://github.com/digidem/mapeo-core-docs/tree/main/client-docs#syncwaitforsynctype) to finish before I load this page?

ErikSin:
@gmaclennan Can you take a look at my questions for the initial implementations of sync?

src/frontend/screens/Sync/index.tsx
gmaclennan: sync.getState() is an async function, it returns a promise.

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const deviceInfo = useDeviceInfo();
+  const [connectedPeers, setConnectedPeers] = React.useState<number>(0); //this should initialize with `sync.getState().connectPeers`, but it is undefined
+  const [isSyncing, setIsSyncing] = React.useState(false);
+  const [hasDataToSync, setHasDataToSync] = React.useState<boolean>(); //this should initialize with `sync.getState().data.dataToSync`, but it is undefined
+  const {formatMessage} = useIntl();
+
+  console.log(sync.getState()); // sync.getState() === {"_h": 0, "_i": 0, "_j": null, "_k": null, "clear": [Function anonymous]}
+
+  React.useEffect(() => {
+    const syncListener = (val: State) => {
```
gmaclennan, line 79: I don't see the advantage of splitting the state into separate pieces. It's just going to mean as we add features here we keep needing to add additional states, and it introduces the possibility of these different states getting out of sync, which they shouldn't - there is one sync state.

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+        ),
+        headerRight: () => <SettingsButton />,
+      });
+  }, [ssid]);
+
+  React.useLayoutEffect(() => {
+    if (!ssid) {
+      navigation.navigate('NoWifi');
+    }
+  }, [ssid, navigation]);
```
gmaclennan, line 109: navigating based on an effect seems wrong. If we were to do it (which I don't think we should) then we need to figure out how to replace screens in the stack, because currently joining and leaving a wifi network will just keep adding screens to the stack I think.

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+      navigation.setOptions({
+        headerTitle: () => (
+          <View style={{flexDirection: 'row', alignItems: 'center'}}>
+            <MaterialIcons name="wifi" size={20} color={BLACK} />
+            <Text style={{marginLeft: 5}}>{ssid}</Text>
+          </View>
+        ),
+        headerRight: () => <SettingsButton />,
+      });
+  }, [ssid]);
```
gmaclennan, line 103: The header icon can just read this from the hook, it doesn't need any of the screen state.

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+      setConnectedPeers(val.connectedPeers);
+      setIsSyncing(val.data.syncing);
+      setHasDataToSync(val.data.dataToSync);
+    };
+
+    sync.addListener('sync-state', syncListener);
+
+    return () => {
+      sync.removeListener('sync-state', syncListener);
+    };
```
gmaclennan, line 89:
```suggestion
    let ignore = false
    sync.getState().then(syncListener)
    return () => {
      ignore = true; // add a check for ignore to syncListener
      sync.removeListener('sync-state', syncListener);
    };
```

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+  React.useLayoutEffect(() => {
+    if (!ssid) {
+      navigation.navigate('NoWifi');
+    }
+  }, [ssid, navigation]);
+
+  return (
+    <View style={styles.container}>
+      <View>
+        {projectSettings.data && projectSettings.data.name && (
```
gmaclennan, line 114:
```suggestion
        {projectSettings.data?.name && (
```

src/frontend/screens/Sync/index.tsx
gmaclennan: A question for the co-design team I think. From a technical perspective this makes sense: it puts the peer in "syncing mode" where any peers that join will be able to start syncing.

src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -2,21 +2,27 @@ import React from 'react';
...
       <LinearGradient
         style={styles.linearGradient}
         colors={['#0006', '#0000']}
       />
-      {/* Stand in for styling,  */}
-      <Text>""</Text>
+      <IconButton
+        style={styles.leftButton}
+        onPress={() => {
+          navigation.navigate(ssid ? 'Sync' : 'NoWifi');
```
gmaclennan, line 22: yeah really not sure about using screens for wifi state - it's going to confuse the navigation stack a lot, and it means that lots of screens need to know about wifi state so it breaks encapsulation (adding a sync button to anywhere requires adding ssid state to that component where you add it)

ErikSin: I was just trying to leverage react navigation animations for transition between screens. Without it, it is pretty harsh, but I think having no animation is ok for the MVP, and then we can create a custom animation later

src/frontend/screens/Sync/index.tsx
ErikSin: Actually React-Navigation is really good at catching these kinds of things, and will not just add them endlessly to the stack. 

Regardless, I think we shouldn't have theses as 2 stack screens, and we can just keep them as one

src/frontend/screens/Sync/index.tsx
gmaclennan: I see the problem here, the types returned by @mapeo/ipc were incorrect. I have made a fix in https://github.com/digidem/mapeo-ipc/pull/2

ErikSin:
Closing as this is unfinished, and will reopen when ready for an actual PR

PR #142, opened 2023-12-20
added loading state to button

ErikSin:
Adds a loading state to button.

I discussed these loading states with Sabella. Eventually we want to animate the button to contract into the loader.

https://github.com/digidem/CoMapeo-mobile/assets/67773827/9d7b8341-0e70-4311-b390-30acbec2db3f

### DIFF
--- a/src/frontend/screens/Success.tsx
+++ b/src/frontend/screens/Success.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import {StyleSheet, View} from 'react-native';
-import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
 
 import SuccessIcon from '../images/Success.svg';
 import NewDeviceLogo from '../images/NewDeviceLogo.svg';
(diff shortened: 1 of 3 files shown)


src/frontend/screens/Success.tsx
```diff
@@ -48,16 +46,17 @@ export const Success = ({
       </View>
       <Button
         fullWidth
+        // due to conditional rendering, there is a brief second, where the state is no longer pending, but the user has not been navigated. So if we don't do catch all these state, the button flickers briefly before navigating
```
ErikSin, line 49: just to note, this flickering only happens during conditinal rendering. If we were to do:

```ts
const {mutation} = useSomeReactQueryMutation()

mutate(val, {
    onSuccess:() => navigate("nextPage")
})

```
this onsucess callback is called immediately and we wont see any flickering

src/frontend/sharedComponents/Button.tsx
achou11, line 1: Feeling somewhat conflicted about integrating loading state into this component as it currently stands, as opposed to conditionally rendering one level up. I think if it looked more coherent (e.g. with the transition animation and all) i'd be less concerned but right now, it seems safer to conditionally render the button vs the loading indicator in the components that hold the loading state. My guess is that most buttons we use won't need to worry about a loading state so i'm generally hesitant to initially make that part of this component and complicating its implementation 

Another issue with integrating the loading state here is making sure it's visually consistent based on the button's styling configurations. For example, I'd make sure that both the button and the loading indicator are rendered in the same or similar "container" view to ensure visual consistency in terms of space taken up.

Thoughts?

ErikSin: As I think about this, I think you are right. These are 2 seperate component, completly different ui and different function.

When we animate the button into a loader it would make sense to have it as one component. but for now we should probably just handle it in the parent component

ErikSin:
I'm going to close this in favor of just creating a separate loader component.

PR #144, opened 2023-12-20
feat: added loading spinner

ErikSin:
Added a loading spinner: 

 https://github.com/digidem/CoMapeo-mobile/assets/67773827/9d7b8341-0e70-4311-b390-30acbec2db3f

### DIFF
--- a/src/frontend/screens/Success.tsx
+++ b/src/frontend/screens/Success.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import {StyleSheet, View} from 'react-native';
-import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
 
 import SuccessIcon from '../images/Success.svg';
 import NewDeviceLogo from '../images/NewDeviceLogo.svg';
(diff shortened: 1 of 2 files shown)


achou11 (approved):
probably would make the commit title a little specific but otherwise changes lgtm!

ErikSin:
Im realizing that this pattern doesnt actually make sense. if there is an error, we should be catching it BEFORE this screen. the loader should happen the screen before...its just complicates thing because of the conditional rendering @achou11

achou11:
> Im realizing that this pattern doesnt actually make sense. if there is an error, we should be catching it BEFORE this screen. the loader should happen the screen before...its just complicates thing because of the conditional rendering @achou11

yeah pretty sure I had brought this up before: https://github.com/digidem/CoMapeo-mobile/pull/73#discussion_r1356900866

think the suggestion i made there is still potentially useful: https://github.com/digidem/CoMapeo-mobile/pull/73#discussion_r1357246210

ErikSin:
> think the suggestion i made there is still potentially useful: [#73 (comment)](https://github.com/digidem/CoMapeo-mobile/pull/73#discussion_r1357246210). the gist is that i think you need some separate client state to represent whether or not to enter the device naming flow which is updated based on the mutation later on, and conditionally render based on that

the issue with this is that retrieving deviceName is async. So i would need to create another wrapper, and only render the nav container when deviceName is established...I think it might just be easier to not conditionally render in this case

ErikSin:
Closing as this should be broken up into 2 separate Prs

PR #149, opened 2024-01-10
Es/fatal error

ErikSin:
Shows an error page when server has incorrectly started.

image

### DIFF
--- a/src/frontend/ServerLoading.tsx
+++ b/src/frontend/ServerLoading.tsx
@@ -13,7 +13,6 @@ export const ServerLoading = ({
   const [serverStatus, setServerStatus] = React.useState<StatusMessage>({
     value: 'STARTING',
   });
-  console.log(serverStatus);
 
   React.useEffect(() => {
     const subscription = nodejs.channel.addListener('server:status', msg => {
--- a/src/frontend/ServerLoading.tsx
+++ b/src/frontend/ServerLoading.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import {SafeAreaView, Text} from 'react-native';
 import nodejs from 'nodejs-mobile-react-native';
+import BootSplash from 'react-native-bootsplash';
 
 import type {StatusMessage} from '../backend/src/status';
 import {MessagePortLike} from './lib/MessagePortLike.js';
+import {FatalError} from './screens/FatalError';
 
 export const ServerLoading = ({
   messagePort,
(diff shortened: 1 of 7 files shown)


src/frontend/ServerLoading.tsx
```diff
@@ -38,12 +40,8 @@ export const ServerLoading = ({
...
 
   if (serverStatus.value === 'ERROR') {
-    return (
-      <SafeAreaView
-        style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
-        <Text>Error!</Text>
-      </SafeAreaView>
-    );
+    BootSplash.hide();
+    return <FatalError />;
```
ErikSin, line 44: We could possibly put this in an error boundary, but I am unsure of the best practices with fatal errors, so im curious what you think @achou11 .

All that being said, even if this was encapsulated in an error boundary, I think we still would need to put this `FatalError` Component here as well. `BootSplash.hide()` is async, so throwing an error after `hide()` would not get caught by the ErrorBoundary, as they are only captured within the react render cycle

src/frontend/ServerLoading.tsx
```diff
@@ -12,6 +13,7 @@ export const ServerLoading = ({
   const [serverStatus, setServerStatus] = React.useState<StatusMessage>({
     value: 'STARTING',
   });
+  console.log(serverStatus);
```
achou11, line 16:
```suggestion
```

src/frontend/ServerLoading.tsx
achou11: i think it makes sense here. the error boundary is more for react rendering errors that come up, whereas this is capturing an error happening outside of react (the server). mapeo mobile essentially does what you're doing here, if that's any confirmation 😄

achou11 (approved):
aside from a stray console log, lgtm!

src/frontend/screens/FatalError.tsx
```diff
@@ -0,0 +1,78 @@
...
+            height={50}
+            style={{position: 'absolute', bottom: 0, right: 0}}
+          />
+        </View>
+        <Text style={styles.text}>{formatMessage(m.somethingWrong)}</Text>
+      </View>
+      <Button
+        fullWidth
+        onPress={() => {
+          RNRestart.restart();
```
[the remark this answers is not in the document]
achou11: i think it should suffice, as long as that means that `nodejs.start()` is being called again. hard to know without testing though...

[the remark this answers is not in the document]
achou11: the safe alternative is to just provide copy about closing and re-opening the app

achou11:
hm looking into the restart dep and starting to have questions about whether the app bundle restart is enough in the case of a server crash 🤔 ideally can confirm that it works by throwing an error in the backend startup script and testing the button, but haven't tried that myself yet.

Can confidently say that the UI implemented here lgtm, but yeah just some questions around how that restart button actually works 🤷

PR #154, opened 2024-01-11
Es/retry error

### LINKED ISSUE #147: Error State: Retry Action
Create a [bottom sheet modal that handles errors](https://www.notion.so/digidem/Alerts-and-Error-States-1f7889785bfb44bda2d3e6b3c47dddc4?pvs=4#aaa06952767349949f75076e14530bfd). This modal should take a function that re-tries the api call and close the modal.

## To Do
- Create a full size Bottom Sheet Modal, using [RN Bottom Sheet Modal](https://ui.gorhom.dev/components/bottom-sheet/modal)
- `Go Back` button should close modal
- `Try Again` button should call api again, and close modal

## Screen shots
image

## Technical notes
- if a mutation is being tried again, [mutation needs to be reset](https://tanstack.com/query/latest/docs/react/guides/mutations#resetting-mutation-state)

ErikSin:
Created an error modal that notifies user of error and asks them to try it again.

https://github.com/digidem/CoMapeo-mobile/assets/67773827/b8187ff2-f7ad-41a5-9f10-7f79cee248d0

## Technical notes
- Ideally this modal would be declarative, and show up when there is an error, but gorham bottom sheet is designed to be imperatively opened and closed.
- This should be paired with a loading state. Ideally the user activates the api call, the button is turned into a loader, and then when promise is resolved/rejected, the appropriate ui is shown.
- If possible, reset the error state, so that the user can try again and the react can pick up the new error state by this new attempt. tanstack/query has a [Reset Mutation State](https://tanstack.com/query/v4/docs/react/guides/mutations#resetting-mutation-state)
- When being used with tanstack/query, call `openSheet` in `onError` Callback:
``` ts
const someMutation = useMutation(....)

function handleMutation(){
    someMutation.mutate({}, {onError:()=>{openSheet()})
}

return(
    <>
        <Button onPress={handleMutation}>
            Mutate
        </Button>
        <ErrorModal {...props/>
    </>


```

closes #147

### DIFF
--- a/src/frontend/sharedComponents/BottomSheetModal/index.tsx
+++ b/src/frontend/sharedComponents/BottomSheetModal/index.tsx
@@ -105,15 +105,7 @@ export const BottomSheetModal = React.forwardRef<RNBottomSheetModal, Props>(
         snapPoints={!fullHeight ? undefined : ['100%']}
         enableDynamicSizing={!fullHeight}
         handleComponent={() => null}>
-        <BottomSheetView
-          style={{
-            padding: 20,
-            paddingTop: 30,
-            // need to add paddingbottom due to bug: https://github.com/gorhom/react-native-bottom-sheet/issues/791
-            paddingBottom: 20,
-          }}>
-          {children}
-        </BottomSheetView>
+        <BottomSheetView>{children}</BottomSheetView>
       </RNBottomSheetModal>
     );
   },
--- a/src/frontend/sharedComponents/BottomSheetModal/index.tsx
+++ b/src/frontend/sharedComponents/BottomSheetModal/index.tsx
@@ -8,7 +8,7 @@ import {
 } from '@gorhom/bottom-sheet';
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 
-import {LIGHT_GREY} from '../../lib/styles';
+import {DARK_GREY, LIGHT_GREY} from '../../lib/styles';
 
 export const MODAL_NAVIGATION_OPTIONS: NativeStackNavigationOptions = {
   presentation: 'transparentModal',
@@ -22,15 +22,15 @@ export const useBottomSheetModal = ({openOnMount}: {openOnMount: boolean}) => {
 
   const closeSheet = React.useCallback(() => {
     if (sheetRef.current) {
-      sheetRef.current.close();
       setIsOpen(false);
+      sheetRef.current.close();
     }
   }, []);
 
   const openSheet = React.useCallback(() => {
     if (sheetRef.current) {
-      sheetRef.current.present();
       setIsOpen(true);
+      sheetRef.current.present();
     }
   }, []);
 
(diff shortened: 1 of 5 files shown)


src/frontend/sharedComponents/ErrorModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  'openSheet'
+> & {clearError?: () => void};
+
+/**
+ *
+ * should be used with 'useBottomSheetModal()' Hook
+ */
+export const ErrorModal = ({
+  clearError,
+  closeSheet,
```
achou11, line 35: seems like it would make sense to specify a single prop instead of having two separate props in this case and defining the `handleGoBack()` function (which is the only thing that uses these props)

 e.g.

```tsx
const SomeComponent = () => {
  return <ErrorModal 
    onDismiss={() => {
      // Clear the error if necessary
      // Close the sheet
    }}
  />
}
```

Then this ErrorModal component only needs to pass that `onDimiss` prop directly to the `onPress` prop for the button.

src/frontend/sharedComponents/BottomSheetModal/index.tsx
achou11, line 1: do the changes here affect any other places that use this component? Feel like this is a very finicky and easy-to-break component (that I mostly implemented 🥲) - probably deserves a complete refactoring/simplification...

achou11 (changes requested):

src/frontend/sharedComponents/BottomSheetModal/index.tsx
ErikSin: no, it shouldn't. Also, this component is only being used by the passcode stuff (which is not accessible to the user anyways). The other components that were using this in Mapeo have not been ported over yet. 

With the latest bottom sheet dep, snap points are no longer needed, it can be replaced by [enableDynamicSizing](https://ui.gorhom.dev/components/bottom-sheet/props#enabledynamicsizing). So now we dont need to calculate the snap points anymore. That is why that prop has been taken out.

The other notable change is styling. And when we actually port over the components, I think we can catch any styling changes then, and fix accordingly. There was a bug in the padding, which i realize the better way to deal with is to have the children just deal with its own styling (which i think is just good practice anyway).

achou11: cool, thanks for the explanation! good to know that this isn't being too widely used just yet in this app 😄

ErikSin: Also, I think the finicky thing with your implementation was the snap point calculator, which made sense to do for what we needed at the time. But now since we don't need that, the implementation is quite straightforward and seems to follow the docs.

Maybe we should refactor the [<BottomSheetContent/>](https://github.com/digidem/CoMapeo-mobile/blob/main/src/frontend/sharedComponents/BottomSheet/Content.tsx) that feels slightly over-engineered. But even that, I still think it is easy to use, so not urgent.

src/frontend/sharedComponents/ErrorModal.tsx
ErikSin: its need to take the `closeSheet` from the hook, and then the `clearError` needs to be defined by the api call. So we could pass an `onDismiss`, but we would have to make sure that the onDismiss includes the `closeSheet` function call. So it feels more intuitive to just have both props.

achou11: fair enough!

achou11 (approved):

PR #155, opened 2024-01-15
Provide Android external storage path to backend on initialization

achou11:
Introduces the following:

- Fastify plugin for serving "static" maps. The code for it is essentially copied over from https://github.com/digidem/mapeo-map-server/pull/101 because it'll be easier to get this integrated in the app if we don't have to worry about integrating the whole map server dep (which introduces an older version of fastify, better-sqlite3, etc). Its functionality is pretty isolated from the map server so feels okay to vendor that implementation here while I gradually work on updating the map server dep in sorts of ways.

- native module for getting necessary filesystem paths. basically vendors a very specific part of https://github.com/itinance/react-native-fs, which felt like overkill to introduce for this PR.

Still haven't gotten any tiles to render on the map so leaving as a draft until i figure that out 🤔 Seems like the style is read correctly (although not entirely sure)

### DIFF
--- a/android/app/src/main/java/com/mapeonext/FileSystemModule.java
+++ b/android/app/src/main/java/com/mapeonext/FileSystemModule.java
@@ -1,0 +1,38 @@
+package com.mapeonext;
+
+import com.facebook.react.bridge.NativeModule;
+import com.facebook.react.bridge.ReactApplicationContext;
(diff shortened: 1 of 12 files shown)


achou11:
retrofitting this PR to only include the native module creation and passing the path argument to the backend. will open a separate PR for integrating the map server when ready

PR #156, opened 2024-01-17
React hook form component

ErikSin:
Creates a text input that uses [react-hook-form](https://www.react-hook-form.com/get-started/)

Shows errors according to [Design System](https://www.notion.so/digidem/UX-08163e7bebc94d9390d7e360fd9e9cbe?p=f2bf435bd2be41c18e6c43176e59e17f&pm=c)

https://github.com/digidem/CoMapeo-mobile/assets/67773827/bd5c880b-1710-451f-99df-ffc05277272f

### Technical Notes
- for validation, if `required=true`, then  `minLength` must also be greater than 0. Otherwise, an empty value is still a value, so the `required` validation will be satisfied.


src/frontend/lib/styles.ts
```diff
@@ -13,7 +13,7 @@ export const MAGENTA = '#e2003f';
 export const DARK_MANGO = '#ed6109';
 export const MANGO = '#ff9933';
 export const MEDIUM_GREY = '#666666';
-export const LIGHT_GREY = '#EAEAEA';
+export const LIGHT_GREY = '#ceced6';
```
ErikSin, line 16: Updated this to match design specs (and it was barely showing up)

src/frontend/sharedComponents/HookFormTextInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+type TextInputProps<InputFields extends FieldValues> = {
+  containerStyle?: ViewStyleProp;
+  showCharacterCount?: boolean;
+  control: Control<InputFields>;
+} & Pick<
+  React.ComponentProps<typeof RNTextInput>,
+  'placeholder' | 'placeholderTextColor'
+> &
+  Omit<UseControllerProps<InputFields>, 'control'>;
```
ErikSin, line 26: `control` is optional (there is a `formProvider` that allows for control to be passed to useController), so I made it required here

src/frontend/sharedComponents/HookFormTextInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+import {ViewStyleProp} from '../sharedTypes';
+import {Text} from './Text';
+
+type TextInputProps<InputFields extends FieldValues> = {
+  containerStyle?: ViewStyleProp;
+  showCharacterCount?: boolean;
+  control: Control<InputFields>;
+} & Omit<
+  React.ComponentProps<typeof RNTextInput>,
+  'value' | 'onBlur' | 'onChange'
```
ErikSin, line 24: I know we talked about only using the placeholder prop, but we actually use other props in other components, so I think it would be good to have access to them. Perhaps we should also omit the styling prop because it could be a little confusing since there is technically a `view` around it that controls certain aspects of the style.

src/frontend/sharedComponents/HookFormTextInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+        )}
+        {maxLength && showCharacterCount && (
+          <Counter
+            isMaxLengthError={error?.type === 'maxLength'}
+            control={control}
+            name={name}
+            maxLength={maxLength}
+          />
+        )}
+      </View>
```
achou11, line 89: I would consider updating this so that you work with static nodes to reduce some logic:

```tsx
      <View style={[styles.underContainer, {flexDirection: 'row', justifyContent: 'space-between'}]}>
        <View>
          {errorMessage && (
            <Text style={{color: RED}}>{errorMessage.toString()}</Text>
          )}
        </View>

        <View>
          {maxLength && showCharacterCount && (
            <Counter
              isMaxLengthError={error?.type === 'maxLength'}
              control={control}
              name={name}
              maxLength={maxLength}
            />
          )}
        </View>
      </View>
 ```

summary of suggestion:

1. Be explicit about the flex direction (in react native the default is column, which differs from the row default for the web). in this case you want to ensure it's row and not rely on implicit flexbox-spec rules.

2. Set justify content to space-between instead of dynamically determining it. reduces the logic overhead and while not an issue in this case, dynamically calculating layout changes like that is pretty expensive.

3. wrap each node of interest with a view. this ensures that the justify-content from 2 is always being applied on two direct child nodes, regardless of the rendering outcome based on the logic there

File: src/frontend/sharedComponents/HookFormTextInput.tsx [hunk 5]
```diff
@@ -0,0 +1,132 @@
...
+import {BLACK, LIGHT_GREY, RED} from '../lib/styles';
+import {ErrorIcon} from './icons';
+import {ViewStyleProp} from '../sharedTypes';
+import {Text} from './Text';
+
+type TextInputProps<InputFields extends FieldValues> = {
+  containerStyle?: ViewStyleProp;
+  showCharacterCount?: boolean;
+  control: Control<InputFields>;
+} & Omit<
```
achou11, line 22: just a note that `Omit` is rather expensive in Typescript (especially on object types) so ideally we don't need this unless you're actually providing specific native text input props when using this component

achou11 (approved):
overall changes seem okay but have some suggestions that I would prefer to see before merging

src/frontend/sharedComponents/HookFormTextInput.tsx
achou11: if that's the case, i would suggest using `Pick` e.g. `Pick<React.ComponentProps<typeof RNTextInput>, 'placeholder'>`

src/frontend/sharedComponents/HookFormTextInput.tsx
achou11: saw your other comment about why this is included. would consider the suggestion in https://github.com/digidem/CoMapeo-mobile/pull/156#discussion_r1458028985

src/frontend/sharedComponents/HookFormTextInput.tsx
ErikSin: sorry, i wasnt the clearest with my comment. Im actually saying that "placeholder" is not the only one we use. For example in the manual GPS form, we use `keyboardType`, `accessibilityLabel`, and `underlineColorAndroid`. So I think we should expose them to the user. Im using omit because `  'value' | 'onBlur' | 'onChange' ` are handled by the Controller, so those are the ones we do not want to expose

achou11: ah yeah if you think there will be a variety that are needed, then keeping the Omit makes sense. You can use Pick to get multiple values, but if there's an indefinite list then it's just too verbose 😄

src/frontend/sharedComponents/HookFormTextInput.tsx
achou11: on additional thought, not as adamant about this suggestion being implemented - just personally find that having to deal with layouts that are conditionally affected in a couple of places can be a little tricky

ErikSin: I think your suggestion is a little cleaner. It might cause a slight problem if we were using an inspector like debugger because we would have 2 empty `views`, but I think that would be easy to figure out and none of us use that kind debugger.

src/frontend/sharedComponents/HookFormTextInput.tsx
ErikSin: so were only using `name` and `control`. So do you think we should just use `Pick< ... ,'name'>` and not expose the other ones?

ErikSin: Ohhh nevermind this comment is referring to textInput props, not Controller Props

ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/156/commits/172927279e7822ccb2b3f4496ceafa430ed5b5e6

src/frontend/sharedComponents/HookFormTextInput.tsx
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/156/commits/172927279e7822ccb2b3f4496ceafa430ed5b5e6

achou11: think you still need to add `justifyContent` to the `underContainer` style?

src/frontend/sharedComponents/HookFormTextInput.tsx
achou11: referring to wrong commit? not sure what i should be looking at 😅

src/frontend/sharedComponents/HookFormTextInput.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const errorMessage =
+    errorType && rulesWithMessage
+      ? rulesWithMessage[errorType]?.message
+      : undefined;
+
+  const maxLength = rulesWithMessage && rulesWithMessage['maxLength']?.value;
+
+  const rules = transformRules(rulesWithMessage);
+  return (
+    <React.Fragment>
```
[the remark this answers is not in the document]
achou11: resolved by https://github.com/digidem/CoMapeo-mobile/commit/172927279e7822ccb2b3f4496ceafa430ed5b5e6

src/frontend/sharedComponents/HookFormTextInput.tsx
ErikSin: yup, sorry you can ignore

src/frontend/sharedComponents/HookFormTextInput.tsx
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/156/commits/b4cd6d73ac039f43d2c960a2bbdfadcf3fa420d3

PR #157, opened 2024-01-18
Es/observation edit

ErikSin:
User can create ~~and edit~~ an observation.

## Introduces:
- useMostAccurateLocation hook
- loading and error state for creating an observations

### To Note
- Manual GPS screen has not been implemented (will be a seperate PR)


src/frontend/hooks/server/observations.ts
```diff
@@ -36,13 +37,14 @@ export function useCreateObservation() {
...
-    }) => {
+    mutationFn: async ({value}: {value: ClientGeneratedObservation}) => {
       if (!project) throw new Error('Project instance does not exist');
-      return project.observation.create(value);
+
+      return project.observation.create({
+        ...value,
+        schemaName: 'observation',
+        attachments: [],
+      });
```
ErikSin, line 47: According to the typings, this should be a `promise` but it is being returned as `{"_h": 0, "_i": 0, "_j": null, "_k": null, "clear": [Function anonymous]}`

This is causing a problem with our loading and success state, as nothing is being resolved or rejected.

It is still creating an observation though.

gmaclennan: I looked into this. I checked the tests on rpc-reflector to make sure this is being tested, and I tried adding this code to CoMapeo (in the project context):

```js
project.observation.create({ ... }).then(console.log)
```

This worked successfully, resolving and logging the result. What you are seeing when you log the value being returned is because of the way that rpc-reflector uses a `Proxy`, and how Hermes implements it. This is what I see on mobile:

```js
const p = project.observation.create({ ... })
console.log(p) // {"_h": 0, "_i": 0, "_j": null, "_k": null, "clear": [Function anonymous]}
console.log('then' in p) // true
console.log(typeof p.then) // function
```

TL;DR As far as I can test it, `project.observation.create()` is returning a promise, and it does resolve, and it also rejects if an invalid value is passed. I think the error must be elsewhere.

gmaclennan: Update: I also tried running this PR as-is on an Android emulator. I created a new observation and pressed save, and both `onSuccess` handlers correctly fired (I checked by adding console.log), and it navigated back to the map screen. Not sure what behaviour is not working, let me know and I can investigate further.

achou11 (commented):
Was trying to test the changes locally and ran into a couple of errors. Not sure if it's because of the local changes I made in order to access the edit screen (basically just updating the navigation that happens when pressing an observation row in the observations list).

- entering the edit screen (and waiting for the coordinates to resolve) 

    image

- As i was editing the notes section
 
    image


The fixes may be unrelated to the changes made in this PR, but wanted to surface in case you have a better idea.

ErikSin:
> Was trying to test the changes locally and ran into a couple of errors. Not sure if it's because of the local changes I made in order to access the edit screen (basically just updating the navigation that happens when pressing an observation row in the observations list).
> 
>     * entering the edit screen (and waiting for the coordinates to resolve)
>         image
> 
>     * As i was editing the notes section
>         image
> 
> 
> The errors may be unrelated to the changes made in this PR, but wanted to surface in case you have a better idea.

Sorry, Im realizing this is a naming issue. So Edit Observation and Create Observation are the same screen! This PR is about creating an observation. For the MVP, we are not going to be editting any observations, only creating. So the errors you are running into, are because I haven't dealt with the editting part, just the creating part (there are some subtle differences even though they are the same screen).

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -55,25 +56,41 @@ const m = defineMessages({
...
+    );
   }
 
   function editObservation() {
     if (!value) throw new Error('no observation saved in persisted state ');
+    if (!observationId) throw new Error('Need an observation Id to edit');
     if (!('versionId' in value))
       throw new Error('Cannot update a unsaved observation (must create one)');
+    // @ts-expect-error
     editObservationMutation.mutate({id: observationId, value});
```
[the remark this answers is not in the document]
ErikSin: Again, sorry, I put this `// @ts-expect-error` here because I was going to deal with it later and it was out of scope of this PR.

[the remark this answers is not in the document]
achou11: Yeah no worries! just thought it was worth keeping in the back of the mind for the mapeo core work in case we needed to make adjustments there based on this

achou11:
@ErikSin ah i see. makes more sense then! was able to test the simple creation flow and it looks like things work as expected. was noticing some unexpected behavior around draft persistence but not sure if that's within the scope of this PR

achou11 (approved):
for posterity, the primary functionality that this PR addresses are:

- creating an observation
- listing the observation (on the map and in the observations list)

seems like both are achieved based on local testing and the changes involved. anything else can be addressed in a follow-up

Issue #151, opened 2024-01-11
Peer Discovery


ErikSin:
Requirements:

- Provide a list of peers that could be invited to the project (connected peers, but filtering peers already part of the project)
- Stop discovery when the app closes, or wifi is turned off.
- Start discovery when the app opens, or wifi is turned on.
- Restart discovery if the wifi network changes, or the device IP changes.

Implementation plan:

- [ ] `useDiscoveryController(): 'started' | 'stopped' | 'error'`: hook that starts and stops the discovery service according to the requirements above. Should be a global singleton - if used in multiple places, should all reference the same instance / state. See #158 
- [ ] `useLocalPeers(projectId: string): LocalPeers[]`: hook to return a list of locally discovered peers, with an array of ids of the projects they are members of. Can include disconnected peers (`status: 'connecting' | 'connected' | 'disconnected'`)  if a peer has previously connected but then disconnected (will include a `disconnectedAt` property, which can be used to filter "old" peers - connection status is persisted for the lifecycle of the app). See #159

Testing plan:

- Because these hooks have complicated logic, use [`renderHook` from React Testing Library](https://testing-library.com/docs/react-testing-library/api/#renderhook) to unit test logic, mocking out the Mapeo API and React Native APIs for network connectivity.

Unknowns / possible problems:

- ~~I am unclear how `@react-native-community/netinfo` handles multiple connections. E.g. if a phone is connected to wifi, but does not have an internet connection via wifi, and is also connected to cellular, then is the wifi state still returned? If not then this could be a problem for Mapeo, so we will need a different library to get the wifi state.~~ Edit: `@react-native-community/netinfo` returns info about the "active data connection", and prioritises wifi over cellular if both are connected to the internet. However, if wifi is connected but has no internet connection, and there is a cellular data connection, then I _think_ `NetInfo` returns information about the cellular connection instead. The `useLocalDiscoveryState()` hook attempts to get around this by explicitly checking the wifi state when the active network state changes, but there can still be some circumstances when the wifi state might be "stale", so local discovery might not correctly start and stop. This will only be an issue if a user is (A) using a Wifi connection with no internet _and_ (B) has an active cellular data connection with internet connectivity.

PR #158, opened 2024-01-22
feat: useLocalDiscoveryState() hook

gmaclennan:
Partial fix for #151

Implements:

1. `LocalDiscoveryController` which is created and started when the app loads - this starts and stops peer discovery based on wifi network connectivity and whether the app is in the foreground/background.
2. `useLocalDiscoveryState()` which returns the status of peer discovery (is it started, or stopped), as well as info about the wifi network name and whether wifi is turned on or off. The former (peer discovery status) is more useful for debugging, since we don't have a plce to expose this in the app, but the wifi state I think we need in the UI?

TODO:

- ~~Add unit tests for hook logic~~ - tested on a device for now, unit tests are going to take a day or two to write because of the complexity of mocking everything. Unfortunately e2e tests are not possible using android emulator, and a testing setup with multiple real devices for e2e testing is a lot more work too.

Known Issues:

- Sometimes it seems like devices do not discover each other, so one of the devices needs to restart discovery by either switching away from teh app and returning, or turning wifi on and off.
- Currently when the app is put into the background, as well as stopping discovery of any new peers (and stopping advertising), it also disconnects any connected peers. This means that sync will stop immediately without completion. We might need to figure out a way of allowing sync to continue in the background for a fixed period of time when the app goes into the background.


src/frontend/hooks/useLocalDiscoveryState.ts
```diff
@@ -0,0 +1,210 @@
...
+      (ipAddressChanged || ssidChanged) &&
+      (state.status === 'starting' || state.status === 'started');
+
+    const shouldDiscoveryBeStopped =
+      !isLocalWifiConnected &&
+      (state.status === 'starting' || state.status === 'started');
+    if (shouldDiscoveryRestart) {
+      // StartStopStateMachine will ensure these happen consecutively
+      sm.stop().catch(noop);
+      sm.start().catch(noop);
```
ErikSin, line 159: what happens if you call `sm.start()` when `state === 'stopping'`? Aka, does it need to be completely stopped before we can start it again?

gmaclennan: When you call `sm.start()` when `state === 'stopping'`, it will wait for `stopped` before starting. The [state machine docs](https://github.com/digidem/start-stop-state-machine#api) attempt to explain the logic. Dealing this this stuff has so many edge cases it's a pain to deal with, but hopefully the start-stop-state-machine allows the code here to be a little simpler.

gmaclennan:
After testing the `NetInfo` interface on my device (Moto X4) I realized turning wifi on and off when wifi does not have an internet connection does not result in an event. I think the only way to reliably get the wifi state (including avoiding bugs/problems described in #151) is to poll the wifi state. Because we don't re-render unless this actually changes, I think this is ok and should not have a significant performance overhead, but we should maybe keep an eye on it.

I also realised from testing that the netinfo events were sometimes returning before the `fetch`, which throws things out of order, so I changed the code to cancel an active fetch when an netinfo event fires.

package.json
```diff
@@ -86,6 +87,7 @@
     "@mapeo/schema": "3.0.0-next.13",
     "@react-native-community/cli": "^9.3.2",
     "@react-native-community/eslint-config": "^3.2.0",
+    "@testing-library/react-native": "^12.4.3",
```
ErikSin, line 90: let remove this.

src/backend/patches/bonjour-service+1.1.1.patch
ErikSin, line 1: Can you add a description to the `backend/patches/README.md explaining this please

ErikSin (approved):
Im going to merge this so I can use it. 2 small comments, but those can be addressed with a separate PR

PR #165, opened 2024-01-24
add sharedStoragePath flag to backend initialization

achou11:
Precursor to the work to integrate the maps serving functionality for the backend. The backend needs to know where on the device the directory containing the map assets is located. 

~This implements a native module that can be used for getting Android-specific file paths. Basically vendors a very specific part of https://github.com/itinance/react-native-fs, which felt like overkill to introduce for this PR.~ EDIT: decided to use `react-native-fs` proper (with an intention to migrate to a more maintained fork later on)

The flag is not used by the backend at the moment. That will come in a follow-up PR


achou11:
Decided to go with the original `react-native-fs` dep for now, to avoid needing to do a React Native upgrade (see https://github.com/digidem/CoMapeo-mobile/pull/165/commits/88457ffd2307232bfb4082e61257590ff7e47e18)

src/frontend/lib/file-system.ts
```diff
@@ -0,0 +1,3 @@
+import {ExternalDirectoryPath} from 'react-native-fs';
+
+export {ExternalDirectoryPath as EXTERNAL_FILES_DIR};
```
achou11, line 3: Same variable used by Mapeo Mobile, just to confirm: https://github.com/digidem/mapeo-mobile/blob/56528dcccf291d4248beb6348767ca50a6ce92be/src/frontend/api.ts#L420

src/frontend/lib/file-system.ts
achou11, line 1: creating a separate file so that it's easier to migrate to a different library in the future

src/frontend/initializeNodejs.ts
```diff
@@ -19,5 +21,11 @@ export async function initializeNodejs() {
...
   }
 
-  nodejs.startWithArgs(`loader.js --rootKey=${rootKey}`);
+  const flags = [`--rootKey=${rootKey}`];
+
+  if (EXTERNAL_FILES_DIR) {
+    flags.push(`--sharedStoragePath=${EXTERNAL_FILES_DIR}`);
+  }
+
+  nodejs.startWithArgs(`loader.js ${flags.join(' ')}`);
```
ErikSin, line 30: @achou11 I remember that we had this conversation a while ago but don't remember the details. `startWithArgs` is not in the [docs](https://nodejs-mobile.github.io/docs/api/react-native-bridge). Can you remind me what this Api is?

package.json
```diff
@@ -56,6 +56,7 @@
     "react-native-bootsplash": "^5.1.3",
     "react-native-confirmation-code-field": "^7.3.1",
     "react-native-device-info": "^8.7.1",
+    "react-native-fs": "^2.20.0",
```
ErikSin, line 59: Does this need [configuration in the native code](https://github.com/itinance/react-native-fs?tab=readme-ov-file#usage-android)?

src/frontend/lib/file-system.ts
ErikSin, line 3: Do we have a plan for iOS? This is only supported by Android

package.json
achou11: from my understanding it doesn't. react native's automatic autolinking works with it. just from running the app, that seems to be the case

src/frontend/initializeNodejs.ts
achou11: it's a way of starting a node js process that will pass provided CLI flags to the process.

src/frontend/lib/file-system.ts
achou11: no plan for now, as i'm not super familiar with the filesystem apis for iOS. my guess is that it would be what react-native-fs calls the `DocumentDirectoryPath` (https://developer.apple.com/documentation/foundation/nssearchpathdirectory/nsdocumentdirectory). happy to add this in if you think it's worth doing now vs later

ErikSin (approved):

PR #172, opened 2024-02-05
Es/listen for invites hook

ErikSin:
Introduces the ability to invite and add users to a project. This is stacked on #79. 

https://github.com/digidem/CoMapeo-mobile/assets/67773827/64f9e3e6-ea94-4303-af0f-595fb10eca91

### Current Issues:
- ~~local peers list is not updating when a peer is leaves the network~~
- app name is not available as soon as someone joins a project (see end of video above)


src/frontend/hooks/useLocalPeers.ts
```diff
@@ -48,14 +48,19 @@ function createLocalPeerState(api: MapeoClientApi) {
...
+      const statusOfPeerChanged = existing && existing.status !== peer.status;
+      if (listOfPeersChanged || statusOfPeerChanged) {
         stateUpdated = true;
       }
     }
     if (stateUpdated) {
-      state = Array.from(peersById.values());
+      state = Array.from(peersById.values()).filter(
+        peer => peer.status !== 'disconnected',
+      );
```
ErikSin, line 63: @gmaclennan  I updated this logic to also update state when a user disconnects

gmaclennan (changes requested):
This looks good as in the logic appears correct and it should work correctly, however I think the architecture makes it quite hard to understand what is happening and is going to make maintenance difficult and bugs hard to catch.

The logic around invites is spread across several places in the AppNavigator, ProjectInviteBottomSheet, and the useProjectInviteListener. Consider moving everything into a single hook, and encapsulating that logic into the bottom sheet. e.g. something like

```js
// type would be better as discriminate union based on inviteState
function useProjectInvites(): {
    accept: () => void,
    reject: () => void
    invite?: Invite
    inviteState: 'noInvite' | 'pendingInvite' | 'accepting' | 'accepted' | 'rejected' | 'cancelled'
} {
  // handle logic for selecting which invite to show in the hook
  // deal with clearing invites inside the hook
}

const InviteBottomSheet = () => {
  const { accept, reject, invite, inviteState } = useProjectInvites()
  const navState = useNavigationState()
  const isOnEditingScreen = isEditingScreen(navState) 
  const {isOpen, sheetRef, openSheet, closeSheet} = useBottomSheetModal({
    openOnMount: false,
  });
  
  if (!isOpen && !isOnEditingScreen && invite) {
    openSheet()
  }

  // render bottom sheet content
```

I think an approach similar to that is going to make it much easier to maintain this in the future, and add functionality like being able to cancel invites.

src/frontend/hooks/useLocalPeers.ts
gmaclennan: as discussed today, rather than filter by disconnectedAt since 10 seconds, filter based on disconnected since the user opened this screen / first saw the list.

gmaclennan:
Following up on our discussion yesterday with some options for how to implement this, I think useSyncExternalStore might be complicating things, because we are already using tanstack query, which abstracts lots of this. @EvanHahn the invite API that you are developing affects the implementation, and I think it would be helpful to make some tweaks to work better with the front end implementation.

Currently we don't have an API for getting "pending invites" from the backend, e.g. invites that have been received and are pending a response. We only have an event emitter for each invite received, which means that the front end needs to subscribe as early as possible in the application lifecycle, but still risks "missing" an invite that is received before a listener for the `invite-received` event is attached. We also don't have a way of knowing if any of the pending invites have changed state, e.g. if one of them has been cancelled.

Things would be better if there is an API method to return all pending invites from the backend e.g. `client.invite.getPending()`. This should probably return invites in a deterministic order (e.g. order received) or expose an `receivedAt` prop so the front-end can sort them (possible better). It might also be helpful to expose invite ids, for tracking cancellation, and making sure the backend and frontend are "in agreement" about which invite is being responded to when the client accepts or rejects (it _should_ always be the first received for a projectId, but I guess there could always be a bug introduced). 

To keep the client-side list of invites updated, we either need (1) multiple events for when the list of pending invites changes, or (2) an `invites-updated` event that either returns a new list of pending invites, or requires the client to re-request.

1. Multiple events: we need an event for any time the backend changes the pending invites without client intervention, which is an invite being received, or an invite being cancelled (due to disconnection or the inviter explicitly cancelling it). The front-end can then update its "cache" of pending invites based on these events (e.g. remove these invites from the cache). For changes to pending invites by the front-end, the front-end can update its cache itself (e.g. when responding with "reject" or "accept"). Note we don't currently need an event for when a duplicate invite is responded to with "already" when the invitee responds (which changes the list of pending invites), because [currently we do not expose duplicate pending invites in the API](https://github.com/digidem/mapeo-core-next/blob/0d2a17a87e15dec574dc2f6a1c72a4b2138a42e6/src/invite-api.js#L89-L93).
2. Single `invites-updated` event - would be emitted whenever the backend receives and invite or an invite is cancelled. Could either call listeners with the new list of pending invites, or require the front-end to re-fetch pending invites from `invite.getPending()`.

Note if we use (2) we probably still need to keep an `invite-cancelled` event so that we can show the user a cancelled screen if they are currently viewing the invite that has been cancelled (rather than that invite just disappearing).

For implementation in the front-end, I suggested using useSyncExternalStore, however since we are using react-query throughout, which already handles this, then I think it makes sense to lean into react query (which uses `useSyncExternalStore` behind the scenes) and use that. "Dumb" (simple) way:

```js
function usePendingInvites() {
  const api = useMapeoApi()
  const queryClient = useQueryClient()
  const fetchInvites = () => api.invite.getPending()

  useEffect(() => {
    function onChange() {
      queryClient.invalidateQueries({queryKey: ['invites']});
    }
    // If we go with option (1)
    api.invite.on('invite-received', onChange)
    api.invite.on('invite-cancelled', onChange)
    // If we go with option (2)
    api.invite.on('invites-updated', onChange)

    return () => {
      api.invite.off('invite-received', onChange)
      api.invite.off('invite-cancelled', onChange)
    }
  }, [api, queryClient])

  return useSuspenseQuery({ queryKey: ['invites'], queryFn: fetchInvites })
}
```

However that's not the most efficient, because if this hook is used in multiple places, we invalidate (and re-fetch) the query in multiple places, so that `useEffect` could be moved into the parent component of the app and only use it once, but it's probably fine as-is.

The other optimization we could make is to just update the query cache, rather than invalidate it and re-request, eg.

```js
  useEffect(() => {
    api.invite.on('invite-received', (invite) => {
      queryClient.setQueryData(['invites'], data => [...data, invite])
    })
    api.invite.on('invite-cancelled', (invite) => {
      // Assuming we add an `id` to invites
      queryClient.setQueryData(['invites'], data => data.filter(({ id }) => id !== invite.id)
    })
    // If we go with option (2)
    api.invite.on('invites-updated', (invites) => {
      queryClient.setQueryData(['invites'], data)
    })

    return () => {
      // unsubscribe
    }
  }
```

And we probably would want mutation hooks, following the pattern we already have in the app:

```ts
function useAnswerInvite(response: 'accept' | 'reject' | 'cancel') {
  const api = useMapeoApi()
  const queryClient = useQueryClient()
  return useMutation({
    mutationFn: async (projectId) => {
      await api.invite[response](projectId)
      return projectId
    }
    onSuccess: (projectId) => {
      // simple option
      queryClient.invalidateQueries({ queryKey: ['invites'] })
      // "clever" option
      queryClient.setQueryData(['invites'], data => data.filter(invite => invite.projectId !== projectId)
    }
}
```

For handling cancellations, I'm not sure what the best pattern is. Maybe if we wrap all this into a single hook:

```js
function useReceivedInvite({ onCancel }) {
  const api = useMapeoApi()
  // requires deterministic order
  const currentInvite = usePendingInvites()[0]
  
  useEffect(() => {
    function onCancelled(inviteId) {
      if (inviteId === currentInvite.id) onCancel()
    }
    api.invite.on('invite-cancelled', onCancel)
    return () => api.invite.off('inviteCancelled', onCancel)
  }

  return {
    invite: currentInvite
    accept: useAnswerInvite('accept')
    reject: useAnswerInvite('reject')
  }
}
```

Then you can do something like:

```js
const InviteBottomSheet = () => {
  const { isOpen, openSheet, closeSheet } = useBottomSheet()
  const { invite, accept, reject } = useReceivedInvite({
    onCancel() {
      closeSheet()
      showInviteCancelledSheet()
      // or...
      setIsCancelled(true) // and render bottom sheet with cancelled message
    }
  }
  if (!isOpen && invite && notOnEditPage) openSheet()
  // ...
}
```

We could handle all this state and async transactions ourselves in `useSyncExternalStore` and set up a global store for this, or we could handle it all in a context provider, but we will end up re-implementing a lot of what tanstack-query already does to handle all the edge cases, and since we already use that I think maybe use the same pattern?

In terms of the options for how this is implemented in terms of changes to the backend API, I'll let you @ErikSin and @EvanHahn discuss and decide which pattern you want to use.

Seeing all this, and all the other hooks, makes me think it would be worthwhile at some point abstracting a lot of this logic into a separate `react-mapeo` module, that exposes all these hooks we use, so that the front-end development team does not have to worry about them too much, since the implementation is closely tied to the backend.

ErikSin:
So I did an exploration of this yesterday and I think tanstack query isn't the best idea. Tanstack is good for promises, but doesnt handle subscriptions that well. As you demonstrated, everytime the event emitter gets fired, we have to manually invalidate or populate the cache. 

Also another thing to consider is that we always need to be running this subscription at the app level, because it needs to be running independent of what component is mounted. AND it always needs to be running in the background, because we are always listening for invites. Because of that, I actually think it would make sense to add it to global state using a context. The intended use for useSyncExtrnalStore is to have each component subscribe to it when that component is mounted, and then unsubscribe when unmounted. It creates a nice optimization that the react state is always in line with the backend state, but only updated when needed. In our case, we don't need this optimization because we know the subscription is always running in the background.

I still think that we can use `useSyncExternalStore` at the top level, but then just propogate it using a context (if we want to use it as global state -which is not necessary at the moment). In order to do this, we would still need an API that exposes the state of all invites.

## Example if we want to keep the api similar to it is now
```ts
function createInviteStore(){
     let listener :()=>void

     function subscribe(fireUpdateSubscription){
          listener=fireUpdateSubscription
          api.invite.on('invite-received', (invite)=>{listener()})
     }

     function rejectInvite(inviteId){
          return api.invite.reject(inviteId).then(val=>{listener()})
     }

     function acceptInvite(inviteId){
          return api.invite.accept(inviteId).then(val=>{listener()})
     }

     function getSnapshot(){
          return api.invite.getPending()
     }

    return {subscribe, acceptInvite, getSnapshot, rejectInvite}
}

//the props should come from the function above
const InvitesProvider = ({children, subscribe, acceptInvite, getSnapshot, rejectInvite}) => {
     
     const invites = useSyncExternalStore(subscribe, getSnapShot)
     
     return (
          <InvitesContext value={{invites, acceptInvite, rejectInvite}}>
               {children}
          </InvitesContext />
     )

}

```

Example if we update the invite listener
```ts
function createInviteStore(){
     let listener :()=>void

     function subscribe(fireUpdateSubscription){
          listener=fireUpdateSubscription
           //This would update anytime the pending invites invites, eg, invite is cancelled, accepted, rejected, or cleared
          api.invite.on('invite-updated', (invite)=>{listener()})
     }

     function getSnapshot(){
          return api.invite.getPending()
     }

    return {subscribe, getSnapshot}
}

//the props should come from the function above
const InvitesProvider = ({children, subscribe, getSnapshot,) => {
     
     const invites = useSyncExternalStore(subscribe, getSnapShot)
    
     //invite is accepted of rejected in the modal, and uses the invite directly, not through a invitedId as used above   
     return (
          <InvitesContext value={{invites}}>
               {children}
          </InvitesContext />
     )

}


```

gmaclennan:
Yeah I think it's fine to use this or react query. But getSnapshot needs to be synchronous, it cants return a promise - you'll need to get pending invites and store them in state, and modify that state when you get the events, or re-request the pending invites - the same as react query cache invalidation (the query client implementation is quite similar). Either works though.

ErikSin:
>But getSnapshot needs to be synchronous

I did not know that about `useSyncExternalStore`.

Based on that, then I think you are right, lets use react-query, but with the knowledge that we are always running the invite listener at the app level, we can just update the cache at the top level. 

> However that's not the most efficient, because if this hook is used in multiple places, we invalidate (and re-fetch) the query in multiple places, so that useEffect could be moved into the parent component of the app and only use it once, but it's probably fine as-is.

Knowing that it is always being run at the top level we don't have this problem anymore. We can simply run the use effect once. And the we can just provide hooks that deal with the query and mutations. And we know that anytime a mutation happens, the `'invite-added'` or `'invite-removed'` event will be fired at the top level, and the cache will be updated.

## Example:
```ts
//should be run at the app level
function initializeInviteListeners(api){
  const queryClient = useQueryClient

  useEffect(() => {
      function onChange() {
        queryClient.invalidateQueries({queryKey: ['invites']});
      }

      api.invite.on('invite-received', onChange)
      api.invite.on('invite-cancelled', onChange)
  
      return () => {
        api.invite.off('invite-received', onChange)
        api.invite.off('invite-cancelled', onChange)
      }
    }, [api, queryClient])
  }
  
const useListInvites =()=>{
  const fetchInvites = () => api.invite.getPending()
  return useQuery({ queryKey: ['invites'], queryFn: fetchInvites })
}

```

ErikSin:
Also Evan and I had a long conversation and we decided that having an `invite-added` and `invite-removed` event would be the best. The front end needs to know about the cancellations and have the user acknowledge them. So having the removed as a seperate event, makes it sllightly easier to catch when a cancellation happens and allow us to do a seperate action (rather than just update the cache)

gmaclennan:
Sounds good. Example code looks good, should work.

EvanHahn:
We also decided a few other small things:

- The `ALREADY` response will be removed and folded into `REJECT` for simplicity.
- The frontend, not the backend, will be responsible for handling multiple invites to the same project. This simplifies the code on the backend a lot (no need to keep track of all the invites for a particular project), and on the frontend a little (invites can't "disappear" out from under the frontend unexpectedly).
- My first PR will not include cancellations and will be added in a followup.

ErikSin:
> ```js
> // type would be better as discriminate union based on inviteState
> function useProjectInvites(): {
>     accept: () => void,
>     reject: () => void
>     invite?: Invite
>     inviteState: 'noInvite' | 'pendingInvite' | 'accepting' | 'accepted' | 'rejected' | 'cancelled'
> } {
>   // handle logic for selecting which invite to show in the hook
>   // deal with clearing invites inside the hook
> }
> 
> const InviteBottomSheet = () => {
>   const { accept, reject, invite, inviteState } = useProjectInvites()
>   const navState = useNavigationState()
>   const isOnEditingScreen = isEditingScreen(navState) 
>   const {isOpen, sheetRef, openSheet, closeSheet} = useBottomSheetModal({
>     openOnMount: false,
>   });
>   
>   if (!isOpen && !isOnEditingScreen && invite) {
>     openSheet()
>   }
> 
>   // render bottom sheet content
> ```

I deviated from this approach a little in a new branch with the updated API. I started with your approach but there are a couple issues

ErikSin:
> ```js
> // type would be better as discriminate union based on inviteState
> function useProjectInvites(): {
>     accept: () => void,
>     reject: () => void
>     invite?: Invite
>     inviteState: 'noInvite' | 'pendingInvite' | 'accepting' | 'accepted' | 'rejected' | 'cancelled'
> } {
>   // handle logic for selecting which invite to show in the hook
>   // deal with clearing invites inside the hook
> }
> 
> const InviteBottomSheet = () => {
>   const { accept, reject, invite, inviteState } = useProjectInvites()
>   const navState = useNavigationState()
>   const isOnEditingScreen = isEditingScreen(navState) 
>   const {isOpen, sheetRef, openSheet, closeSheet} = useBottomSheetModal({
>     openOnMount: false,
>   });
>   
>   if (!isOpen && !isOnEditingScreen && invite) {
>     openSheet()
>   }
> 
>   // render bottom sheet content
> ```

I updated [here](https://github.com/digidem/CoMapeo-mobile/pull/206/files#diff-86e6ce7ea0b0198fb98c5e498b762ae89a93ba045782f198a7e419d4021cb996). I deviated from your suggestion. Ill leave a comment explaining why on that PR.

PR #173, opened 2024-02-07
Es/create or join project

ErikSin:
Create form for user to create project.

### Features
- Detects if user is already on a project (right now we are just checking if more than one project exists).
- Form that validates name
- tanstack query hook that creates a new project

### Screenshots
image

image

image

image

image


src/frontend/hooks/server/projects.ts
```diff
@@ -9,3 +11,30 @@ export function useProject() {
...
   return projectContext.project;
 }
+
+export function useAllProjects() {
+  const api = useApi();
+
+  return useQuery({
+    queryFn: async () => await api.listProjects(),
+    queryKey: ['projects'],
+    throwOnError: true,
```
achou11, line 21: mostly because i'm not super familiar, but can you provide some context for this? guessing it has something to do with relying on suspense or an error boundary instead of manually handling error states, but wondering what the criteria is for when you'd want this set. also, does it make sense for it to be the default for this hook, or an option to allow more flexibility case-by-case?

src/frontend/hooks/server/projects.ts
```diff
@@ -9,3 +11,30 @@ export function useProject() {
...
+  const queryClient = useQueryClient();
+  const updateProject = useUpdateActiveProjectId();
+
+  return useMutation({
+    mutationKey: ['createProject'],
+    mutationFn: async (name: string) => {
+      return await api.createProject({name});
+    },
+    onSuccess: async data => {
+      updateProject(data);
```
achou11, line 36: wondering if this shouldn't be done here and should be handled by the caller of the hook instead? having this here assumes that a created project should always be immediately switched to, which is accurate for now but seems overly strict/opinionated

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
achou11, line 1: is error-handling for this screen going to be a follow-up?

src/frontend/screens/Settings/CreateOrJoinProject/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const projects = useAllProjects();
+
+  if (projects.data) {
+    return (
+      <ScrollView style={styles.container}>
+        <View style={styles.greyBox}>
+          <Text style={{fontWeight: 'bold'}}>{t(m.whatIsAProject)}</Text>
+          <Text>{t(m.projectDescription)}</Text>
+        </View>
+        {projects.data.length > 1 ? (
```
achou11, line 63: what's the more proper check for this? checking if there exists a project is not a "default" project? If needed, maybe add a todo comment to note that this needs to be revisited

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
```diff
@@ -0,0 +1,86 @@
...
+  useFocusEffect(
+    React.useCallback(() => {
+      const subscription = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => true,
+      );
+
+      return () => subscription.remove();
+    }, []),
+  );
```
achou11, line 41: probably should have a PR that adds a shared hook for this 😄

achou11 (commented):
Overall flow looks solid. Had a couple of questions that would be good to get some clarification on, but otherwise nothing super blocking at the moment

src/frontend/hooks/server/projects.ts
achou11: to more succinctly put it, seems safer to let the caller of the mutation handle side effects and only let the mutation handle mutation-specific things (e.g. interacting with the api)

achou11 (approved):
Would categorize the comments I left earlier as easy to address later if needed and would consider them non-blocking

src/frontend/hooks/server/projects.ts
ErikSin: Agreed, this should be handled by the user. Especially in the future when there are multiple project that can be toggled by the user! But for this first QA build, we have decided that if you join a project, you are automatically just part of that project. Ill add a note here that this is not how it should be handled in the future

src/frontend/screens/Settings/CreateOrJoinProject/index.tsx
ErikSin: Yeah, maybe ill open an issue. This was the workaround for now given that we only want the user to (seemingly) be only on one project. Since there is no default project key, this was the best way to determine that.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
ErikSin: yeah....

src/frontend/hooks/server/projects.ts
ErikSin: Im going to get rid of this. I was playing around with Error Boundaries, but decided not to use this, and just forgot to take it out. As you said, this doesn't allow for flexibility with errors (it just makes all error one catch all), which is not the desired ui.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
ErikSin: yes, good point. Ill add a error modal.

ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/173/commits/e7fbc819a81146b93a3bbf210de36cec3b5e7db7

src/frontend/hooks/server/projects.ts
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/173/commits/1c9761ff81ed2cb6cfba0acfd6dddf374b5af0da

src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -30,6 +30,9 @@ export const useBottomSheetModal = ({openOnMount}: {openOnMount: boolean}) => {
   const openSheet = React.useCallback(() => {
     if (sheetRef.current) {
       setIsOpen(true);
+      if (Keyboard.isVisible()) {
+        Keyboard.dismiss();
+      }
```
ErikSin, line 35: @achou11 the keyboard was persisting when the modal opened. I added this so that it consistently gets dismissed before opening the modal

PR #180, opened 2024-02-12
chore: update app name to CoMapeo

### LINKED ISSUE #178: Change App Name

achou11:
Closes #178 

- PR updates app name from `MapeoNext` to `CoMapeo` (the former was used prior to deciding on a name I think).
- also makes some updates to iOS-related files since that's where most of the naming changes needed to happen, but it doesn't really affect us at the moment since we're focusing on Android
- some of the iOS changes are related to making sure that basic initial steps (e.g. pods setup) work. 
  - ran `npx install-expo-modules` in order to get expo-related changes for iOS
  - in theory it does, but we don't have the secret key setup needed in order to install the Mapbox iOS sdk (similar to what we have for Android), so technically doing `pod install` doesn't work as of right now. Will create a separate issue to address later since it's not urgent right now (see #181)

---

Preview:

image


achou11:
> Should we also change the package-json?

yep forgot that! updated to `comapeo-mobile` via https://github.com/digidem/CoMapeo-mobile/pull/180/commits/58435e5326d171b24661e1f5b63d4c989d874fbc

ErikSin (approved):
Tested on android, and nothing seems to be broken!

PR #182, opened 2024-02-12
Es/observation fields

ErikSin:
Draft pr for observation fields

### DIFF
--- a/docs/ObservationPresets.md
+++ b/docs/ObservationPresets.md
@@ -1,0 +1,41 @@
+### Flow for saving a field to an observation
+
+Each observation can have a several fields associated with it. According to Mapeo Schema:
+
+> A field defines a form field that will be shown to the user when creating or editing a map entity. Presets define which fields are shown to the user for a particular map entity. The field definition defines whether the field should show as a text box, multiple choice, single-select, etc. It defines what tag-value is set when the field is entered.
+
+- The available `Fields` come from the user chosen `Preset`
+  - `Preset` has a `fieldsIds` property. This is an array of `Field's` `docIds`
+  - based on this array we can retrieve all the fields with matching `docIds` using [useFieldQueries()](../src/frontend/hooks/server/fields.ts)
+- A `Field` is saved to the `tags` property of an `Observation`
+  - it is saved in the format of `[key]:value`, where `key === Field['tagId']`
+  - `key !== Field['docId']`
+
+```ts
+type Field = {
+    ...rest
+    tagKey:string
+    docId:String
+}
+
+type Preset = {
+    ...rest,
+    fieldIds:Array<Field['docId']>
+    // What are these tags, and are they related
+    tags:Tags
+}
+
+type Observation = {
+    ...rest,
+    tags:{
+        ...otherTags,
+        // Saves Fields here
+        [key:Field['tagKey']]:any
+    }
+}
+
+```
+
+### Open Question
+
+- [ ] is the `Preset['tags']` related to saving a `Field`?
--- a/messages/en.json
+++ b/messages/en.json
@@ -257,6 +257,18 @@
     "description": "Title of observation screen showing (non-editable) view of observation with map and answered questions",
     "message": "Observation"
   },
+  "screens.ObservationDetails.done": {
+    "description": "Button text when all questions are complete",
+    "message": "Done"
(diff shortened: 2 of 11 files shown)


docs/ObservationPresets.md
ErikSin, line 1: @gmaclennan Can you confirm that this is the correct way of saving a `Field` to an `observation`. I tried referencing Mapeo-Mobile but alot of the names have changed, so it is quite confusing. I left a question at the bottom of this doc as well

gmaclennan:
Thanks for bearing with this API without the full documentation. The logic is largely correct. I'll write out how I would explain it for docs:

> A preset defines how an observation should be displayed to the user, e.g. what icon should be displayed, what it should be called, and how the tags should be displayed.
>
> The preset tags (`preset.tags`) are used to match a preset to an existing observation. When choosing a preset for an observation, the tags defined in `preset.tags` should be applied to the observation.
>
> The fields associated with a preset define the UI for editing tags on an observation. Currently we only allow users to edit observation tags that have a field definition, and we only display tags that have a field definition. A field defines the UX for editing a tag value. The `tagKey` property of a field defines which tag is edited, and the rest of the field definition defines how the form interface should be displayed to the user, e.g. as a text field or a list of options. E.g. a field with `field.tagKey = 'foo'` would set the value of `observation.tags.foo`. The field also defines the labels and helper text for these form elements.

In terms of the implementation, I think the abstraction [`Field`](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/screens/ObservationEdit/Field.tsx) used in mapeo-mobile is helpful, and I would keep that rather than refactoring it out, since it reduces repeated logic. You could write this as a hook if you prefer.

For selecting a preset or changing a preset, the logic is here: https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/screens/CategoryChooser.tsx#L84-L106. In the [current schema](https://github.com/digidem/mapeo-schema/blob/main/schema/preset/v2.json#L71-L78) we also have `addTags` and `removeTags`, but I think it's ok to ignore those for now and I think maybe we remove them from the initial schema definition. This logic would be in your [`updatePreset` function](https://github.com/digidem/CoMapeo-mobile/blob/0da59bc751e215d255c3ddf204564324e4439b0e/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts#L113). Note that in mapeo-mobile we currently additionally add a `categoryId` tag. It would be nice to remove this for the MVP and instead write a new `matchPreset` function.

The logic for matching a preset to an observation (either draft or during editing) in mapeo-mobile is in the [`matchPreset` function](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/lib/utils.js#L80-L87). It is currently very simple, but ideally this function would be similar to [iD Editor](https://github.com/openstreetmap/iD/blob/70a117a38cfaf5518d0099adf70816bc5d196138/modules/presets/index.js#L206-L267). I'm happy to help with this logic. I would maybe consider not persisting the preset as a property of the draft observation, but rather matching it like [mapeo-mobile does it](https://github.com/digidem/mapeo-mobile/blob/56528dcccf291d4248beb6348767ca50a6ce92be/src/frontend/hooks/useDraftObservation.ts#L311-L317), otherwise you are keeping two sources of truth for an associated preset (it would be defined by both the  matching tags and the `preset` property) which could easily get out of sync and cause bugs.

Also consider a `setTags(tags => newTags)` function for the draft observation, which can be used by `updatePreset`, `updateObservationNotes` and `updateObservationField` to reduce repeated logic, and keep the logic of those functions clearer. Also note that a note is just a text field with `tagKey='note'` and can use the same logic. A function like:

```js
function updateTag(tagKey, value) {
  setTags(tags => ({ ...tags, [tagKey]: value }))
}
```

Could then be used for updating a note as `updateTag('note', noteValue)` and a field as `updateTag(field.tagKey, fieldValue)`, and I think reduce a lot of code.

The preset function is a bit more complicated, but would be something like:

```js
function updatePreset(preset) {
  setTags(tags => {
    const prevPreset = matchPreset(tags)
    return { ...omit(tags, Object.keys(prevPreset)), ...preset.tags }
}
```

Where `omit` is something like [lodash omit](https://docs-lodash.com/v4/omit/) (I think we might already have an implementation?)

PR #183, opened 2024-02-14
ensure light theme is used for native styles

achou11:
Fixes an issue where having a dark theme set for the OS causes native styes to be incompatible with the app styles (e.g. scrollbar color for scrollviews).

Once our app styles account for light vs dark themes, we can revert this change.

---

Preview (see right-hand side of images):

- Before:

    image

- After:

    image


achou11:
@ErikSin there's a good chance that this might affect other parts of the app that took respected the old value - not sure if you have an idea of what to check just to make sure it the change isn't wildly unintended.

ErikSin (approved):

PR #184, opened 2024-02-14
implement project settings device name screen

### LINKED ISSUE #163: Device Name/Rename

achou11:
Closes #163 

Remaining items/questions:

- [x] Do we need some sort of loading UI once you submit? If we were using a standard button, this question would be easier to answer since I believe there's an agreed-on pattern (I think?), but given the usage of a nav bar button, it's a little less clear to me about what to do.

- [x] What happens if the submission fails? How do we surface the error? Is it going to follow the (not yet standardized) pattern of bringing up a modal with the error?

- [X] ~Would like some thoughts on the cancel behavior when you're in editing mode. I've implemented it so that when you either (1) press the nav back button, (2) use the hardware back button, or (3) use the back gesture, it will go back to the "display" mode as opposed to going to the previous screen. This seemed like reasonable behavior but i'm working on personal assumptions since it's not really specified.~ EDIT: no longer relevant due to implementation of edit state as separate screen

- There are some changes/fixes to existing components that I would like to make but didn't want to introduce those in this PR for the sake of scope. Happy to create an issue or do a follow-up PR to address those separately. Mostly related to the HookFormInput

---

Preview:

https://github.com/digidem/CoMapeo-mobile/assets/18542095/6f62e106-0d8b-4a89-831c-b246a91ed3c4


src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+export const DeviceName: NativeNavigationComponent<'DeviceName'> = ({
+  route,
+  navigation,
+}) => {
+  const {mode} = route.params;
+  const {data} = useDeviceInfo();
+
+  return (
+    <ScrollView contentContainerStyle={styles.container}>
+      {data?.name &&
```
achou11, line 60: not really sure what the best approach to this is i.e. when data is potentially loading or fails to load. In theory, i think we don't even render the app if that's the case, but just thinking more generally...

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+      navigation.setOptions({
+        headerRight: () => (
+          <IconButton onPress={() => navigation.setParams({mode: 'edit'})}>
+            <EditIcon />
+          </IconButton>
+        ),
+      });
+    },
+    [navigation],
+  );
```
achou11, line 92: not a fan of this pattern from React Navigation but it's necessary because we update the nav header bar based on a navigation param value, so it needs to re-render the headers each time we set those params. to my surprise, the `options` field for the Screen component doesn't do that, as it seems to only handle the case of navigating to the screen initiallly

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+              text: t(m.discardChangesAlertConfirmText),
+              onPress: () => {
+                // TODO: Is going to display mode desired behavior here? Or should it go back to previous screen
+                navigation.setParams({mode: 'display'});
+              },
+            },
+          ]);
+        } else {
+          // TODO: Is going to display mode desired behavior here? Or should it go back to previous screen
+          navigation.setParams({mode: 'display'});
```
achou11, line 142: highlighting these as it's related to the navigation behavior when exiting the "editing" mode either through (1) cancelling your changes or (2) saving your changes. This implementation currently assumes that we want to switch to the "display" mode in both cases, as opposed to navigating to some other screen (or the previous screen)

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+      navigation.setOptions({
+        headerRight: () => {
+          return (
+            <IconButton
+              onPress={handleSubmit(value => {
+                updateDeviceName(value.deviceName, {
+                  onSuccess: () => navigation.setParams({mode: 'display'}),
+                  onError: _err => {
+                    // TODO: Handle errors
+                  },
```
achou11, line 164: highlighting just in case i need to do something here. my guess is that this could be a follow-up that implements an error modal pattern

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+  return (
+    <FieldRow label={t(m.deviceNameEditHeader)}>
+      <HookFormTextInput
+        control={control}
+        name="deviceName"
+        rules={{maxLength: 100, required: true, minLength: 1}}
+        // TODO: Update HookFormTextInput implementation so that either:
+        // - the implementation fully determines the text input's base style and this component doesn't allow custom styling
+        // - this style prop is properly merged with the text input's base style in the implementation
+        style={{flex: 1, color: BLACK, fontSize: 16}}
```
achou11, line 185: highlighting as it's something i noticed as I was working on this. didn't feel necessary to include such changes in this PR so would defer to a follow-up if desired

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+        showCharacterCount
+        autoFocus
+      />
+    </FieldRow>
+  );
+}
+
+export function createNavigationOptions({
```
achou11, line 193: I found it kind of tedious to define the header options for a screen in a different file from the screen's implementation (i.e. in `AppScreens`). Felt much easier to define in the same file. Wondering if there are any thoughts against this?

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+      headerRight: () =>
+        route.params.mode === 'display' ? (
+          <IconButton onPress={() => navigation.setParams({mode: 'edit'})}>
+            <EditIcon />
+          </IconButton>
+        ) : (
+          <IconButton onPress={() => {}}>
+            <SaveIcon />
+          </IconButton>
+        ),
```
achou11, line 213: although the screen calls `navigation.setOptions()` internally to override this, it's useful+recommended to provide a value here as stated by the [React Navigation docs](https://reactnavigation.org/docs/navigation-prop/#setoptions):

> When using navigation.setOptions, we recommend specifying a placeholder in the screen's options prop and update it using navigation.setOptions. This makes sure that the delay for updating the options isn't noticeable to the user. It also makes it work with lazy-loaded screens.


Basically it serves as the placeholder/default for when you initially navigate to this screen (either from within the app or a deeplink)

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
achou11, line 1: starting to wonder if I should implement this as two separate react navigation screens vs a screen with 2 states (represented by route params). currently using the latter but feels like i'm fighting against react navigation by using `setOptions()` and `setParams()`  in effects, even though there's some official docs that align with that usage 🤔 

not really sure what the idiomatic usage of react navigation is in this case and when to decide. if i were to implement 2 screens, would something like this make sense or be worth doing?

```tsx
<RootStack.Group key="projectSettingsDeviceName" screenOptions={{ title: "Device Name" }}>
  <RootStack.Screen name="DeviceNameDisplay" ... />
  <RootStack.Screen name="DeviceNameEdit" ... />
</RootStack.Group>
```

seems like it would need some additional screen configuration e.g. disable nav transitions

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
ErikSin: In this case, would it not make sense to show a blank name? Device Name is optional in schema (to allow for the instance where you freshly install a CoMapeo and and you are instantly added to a project, but have not created your name yet) 

In otherwords it is not necessarily a loading error. It could be the case that someone just does not have a name.

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
ErikSin: Yeah, I like this. I think for most screens, which just pass a header title, this is overkill. But for cases where there are quite a lot of changes, I think it is nice to co-locate it with the component

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
ErikSin: Yeah, i think it would make more sense to have these as 2 separate components. Even though the UI is really similar, I think all the logic involved in maintaining the different states would be simplified if they were their own page.

ErikSin (changes requested):
I think lets separate into 2 nav pages. It will be easier to follow and we can reduce the use effects/seperate the useEffects

src/frontend/screens/Settings/ProjectSettings/DeviceName/index.tsx
achou11: will attempt an update to this - does the note about using a stack group make sense at all, or is it completely unnecessary?

achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/184/commits/963048767ae82d8ce8ff67afc0a861c9c7ec82b7

implementation feels much cleaner!

ErikSin (changes requested):
We can add the loading state to the checkmark button in the header, while it is loading. On a fast phone it is fine, but on a slow phone, there is a akward wait.

Also, I think we can add the error state of opening a modal

src/frontend/screens/Settings/ProjectSettings/DeviceName/DisplayScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <ScrollView contentContainerStyle={styles.container}>
+      <FieldRow label={t(m.deviceNameLabel)}>
+        <Text>{deviceName}</Text>
+      </FieldRow>
+    </ScrollView>
+  );
+};
+
+DisplayScreen.navTitle = m.title;
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/184/commits/1c4678d9e4356ee767280b5ed4e61e0e7f6a8fb7

src/frontend/screens/Settings/ProjectSettings/DeviceName/EditScreen.tsx
```diff
@@ -0,0 +1,154 @@
...
+          style={{flex: 1, color: BLACK, fontSize: 16}}
+          showCharacterCount
+          autoFocus
+        />
+      </FieldRow>
+    </ScrollView>
+  );
+};
+
+EditScreen.navTitle = m.title;
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/184/commits/1c4678d9e4356ee767280b5ed4e61e0e7f6a8fb7

PR #187, opened 2024-02-20
implement query hooks for saving attachments and retrieving attachment urls

### LINKED ISSUE #119: Create Images Hook
**Unfinished** this should be updated when 
Create a react hook that serves and saves images associated with an observations

### To do:
- [ ] create a method that saves an image.
- [ ] create a method that serves an image. It should take the 3 variants as an option, and return the approriate variant

achou11:
Closes #119 

This PR implements the following query hooks:

- `useCreateBlobMutation()`: mutation that accepts a photo object (presumably coming from a observation draft) and saves it as a blob

- `useAttachmentUrlQueries()`: query that accepts a list of observation attachments and performs a query to get the http url for each of them

Perhaps out of scope, but this PR also uses `useCreateBlobMutation()` in the `SaveButton` implementation. `useAttachmentUrlQueries()` is currently not used anywhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "react-native-scale-bar": "^1.0.6",
     "react-native-screens": "^3.24.0",
     "react-native-svg": "^13.14.0",
+    "react-native-url-polyfill": "^2.0.0",
     "react-native-vector-icons": "^10.0.0",
     "readable-stream": "^4.4.1",
     "rpc-reflector": "^1.3.10",
--- a/src/frontend/hooks/server/media.ts
+++ b/src/frontend/hooks/server/media.ts
@@ -1,23 +1,39 @@
 import {Observation} from '@mapeo/schema';
 import {BlobVariant} from '@mapeo/core/dist/types';
-import {useMutation, useQuery} from '@tanstack/react-query';
+import {useMutation, useQueries} from '@tanstack/react-query';
+import {SetRequired} from 'type-fest';
+import {URL} from 'react-native-url-polyfill';
 
-import {Photo} from '../../contexts/PhotoPromiseContext/types';
+import {DraftPhoto} from '../../contexts/PhotoPromiseContext/types';
 import {useProject} from './projects';
+
+type SavablePhoto = SetRequired<
+  Pick<DraftPhoto, 'originalUri' | 'previewUri' | 'thumbnailUri'>,
+  'originalUri'
+>;
 
 export function useCreateAttachmentsMutation() {
   const project = useProject();
 
   return useMutation({
-    mutationFn: async (photos: Photo[]) => {
+    mutationFn: async (photos: SavablePhoto[]) => {
       if (!project) throw new Error('Project instance does not exist');
 
       return Promise.all(
         photos.map(p => {
-          // TODO: Fix
+          const {originalUri, previewUri, thumbnailUri} = p;
+
           return project.$blobs.create(
-            {original: 'abc'},
-            {mimeType: 'img/png'},
+            {
+              original: new URL(originalUri).pathname,
+              preview: previewUri ? new URL(previewUri).pathname : undefined,
+              thumbnail: thumbnailUri
+                ? new URL(thumbnailUri).pathname
+                : undefined,
+            },
+            // TODO: DraftPhoto type should probably carry MIME type info that feeds this
+            // although backend currently only uses first part of path
+            {mimeType: 'image/jpeg'},
           );
         }),
       );
@@ -25,28 +41,59 @@ export function useCreateAttachmentsMutation() {
   });
 }
 
-// TODO: Fix `type` issues
-export function useObservationAttachmentUrl(
-  attachment: Observation['attachments'][number],
-  // @ts-expect-error
-  variant: BlobVariant<(typeof attachment)['type']>,
+export function useAttachmentUrlQueries(
+  attachments: Observation['attachments'],
+  variant: BlobVariant<
+    Exclude<Observation['attachments'][number]['type'], 'UNRECOGNIZED'>
+  >,
 ) {
   const project = useProject();
 
-  return useQuery({
-    queryKey: ['attachmentUrl'],
-    queryFn: async () => {
-      if (!project) throw new Error('Project instance does not exist');
-      return project.$blobs.getUrl({
-        driveId: attachment.driveDiscoveryId,
-        // @ts-expect-error
-        type: attachment.type,
-        variant,
-      });
-    },
-    enabled: !!project,
-    // setting staleTime and CacheTime to infinity means it will only call the api once (aka photos do not get editted so we do not need to keep getting from the server)
-    staleTime: Infinity,
-    gcTime: Infinity,
+  return useQueries({
+    queries: attachments.map(attachment => {
+      return {
+        staleTime: Infinity,
+        gcTime: Infinity,
+        queryKey: [
+          'attachmentUrl',
+          attachment.driveDiscoveryId,
+          attachment.type,
+          variant,
+          attachment.name,
+        ],
+        queryFn: async () => {
+          if (!project) throw new Error('Project instance does not exist');
+
+          switch (attachment.type) {
+            case 'UNRECOGNIZED': {
+              throw new Error(
+                'Cannot get URL for unrecognized attachment type',
+              );
+            }
+            case 'video':
+            case 'audio': {
+              if (variant !== 'original') {
+                throw new Error('Cannot get URL of attachment for variant');
+              }
+
+              return project.$blobs.getUrl({
+                driveId: attachment.driveDiscoveryId,
+                name: attachment.name,
+                type: attachment.type,
+                variant,
+              });
+            }
+            case 'photo': {
+              return project.$blobs.getUrl({
+                driveId: attachment.driveDiscoveryId,
+                name: attachment.name,
+                type: attachment.type,
+                variant,
+              });
+            }
+          }
+        },
+      };
+    }),
   });
 }
--- a/src/frontend/hooks/server/media.ts
+++ b/src/frontend/hooks/server/media.ts
@@ -17,8 +17,6 @@ export function useCreateAttachmentsMutation() {
 
   return useMutation({
     mutationFn: async (photos: SavablePhoto[]) => {
-      if (!project) throw new Error('Project instance does not exist');
-
       return Promise.all(
         photos.map(p => {
           const {originalUri, previewUri, thumbnailUri} = p;
@@ -60,8 +58,6 @@ export function useAttachmentUrlQueries(
           attachment.name,
         ],
         queryFn: async () => {
-          if (!project) throw new Error('Project instance does not exist');
-
           switch (attachment.type) {
             case 'UNRECOGNIZED': {
               throw new Error(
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -69,6 +69,7 @@
     "react-native-scale-bar": "^1.0.6",
     "react-native-screens": "^3.24.0",
     "react-native-svg": "^13.14.0",
+    "react-native-url-polyfill": "^2.0.0",
```
achou11, line 72: had to introduce this because Hermes' `URL` implenetation does not have `pathname`, which is needed to parse the file URIs that are held in a draft photo object.

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
 import {Observation} from '@mapeo/schema';
 import {BlobVariant} from '@mapeo/core/dist/types';
-import {useMutation, useQuery} from '@tanstack/react-query';
+import {useMutation, useQueries} from '@tanstack/react-query';
+import {SetRequired} from 'type-fest';
+import {URL} from 'react-native-url-polyfill';
```
achou11, line 5: opted for the "convenient" usage of the polyfill here (https://github.com/charpeni/react-native-url-polyfill?tab=readme-ov-file#option-3-convenient--ponyfill). we could consider https://github.com/charpeni/react-native-url-polyfill?tab=readme-ov-file#option-1-simple but didn't think it was necessary to for now, given the very limited use case.

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
...
+            {
+              original: new URL(originalUri).pathname,
+              preview: previewUri ? new URL(previewUri).pathname : undefined,
+              thumbnail: thumbnailUri
+                ? new URL(thumbnailUri).pathname
+                : undefined,
+            },
+            // TODO: DraftPhoto type should probably carry MIME type info that feeds this
+            // although backend currently only uses first part of path
+            {mimeType: 'image/jpeg'},
```
achou11, line 36: not sure if there's ever a case where we take photos of different mime types but probably useful to include the mime type info as part of the photo type in general moving forward. better to do as a follow-up though

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
...
+      return {
+        staleTime: Infinity,
+        gcTime: Infinity,
+        queryKey: [
+          'attachmentUrl',
+          attachment.driveDiscoveryId,
+          attachment.type,
+          variant,
+          attachment.name,
+        ],
```
achou11, line 63: does this query key make sense? it's a query describing each attachment, so having the various fragments seems necessary, but not entirely sure

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
...
-    },
-    enabled: !!project,
-    // setting staleTime and CacheTime to infinity means it will only call the api once (aka photos do not get editted so we do not need to keep getting from the server)
-    staleTime: Infinity,
-    gcTime: Infinity,
+  return useQueries({
+    queries: attachments.map(attachment => {
+      return {
+        staleTime: Infinity,
+        gcTime: Infinity,
```
achou11, line 56: is this correct? i kept this from the unused pre-existing implementation of this hook, but wasn't entirely sure if it's still relevant or correct to keep

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -69,20 +71,43 @@ export const SaveButton = ({
   const navigation = useNavigationFromRoot();
   const createObservationMutation = useCreateObservation();
   const editObservationMutation = useEditObservation();
+  const createAttachmentsMutation = useCreateAttachmentsMutation();
```
ErikSin, line 74: on line 171 we need add 'isPending` state to the button.

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -69,20 +71,43 @@ export const SaveButton = ({
...
-        onError: () => {
-          if (openErrorModal) openErrorModal();
-        },
-        onSuccess: () => {
-          navigation.navigate('Home', {screen: 'Map'});
-        },
+
+    const savablePhotos = photos.filter(isSavablePhoto);
+
+    createAttachmentsMutation.mutate(savablePhotos, {
```
ErikSin, line 81: Do we need to do a check for instances where there are no photos? And then if there are no photos, skip creating the attachment, or will the empty array handle that?

ErikSin (approved):
I know you haven't tested this IRL yet as we have to add a bunch of UI to do that. So I am going to approve this knowing that we can test things in a follow up pr

src/frontend/hooks/server/media.ts
achou11: > Unless there is a possibility that 2 pieces of media has the same driveDiscoveryId

i believe this is expected to happen (e.g. multiple photos that you created + saved will live in the same hyperdrive).

after thinking about it, think what I have is closer to necessary than unnecessary. we want to have a unique query identifier _for each attachment request_ and including each of these pieces of information seems necessary to guarantee that

src/frontend/hooks/server/media.ts
achou11: think i wrote it in the past with your guidance because i was unfamiliar with both options 😄 good to know that we probably don't need them

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
...
+
 export function useCreateAttachmentsMutation() {
   const project = useProject();
 
   return useMutation({
-    mutationFn: async (photos: Photo[]) => {
+    mutationFn: async (photos: SavablePhoto[]) => {
       if (!project) throw new Error('Project instance does not exist');
 
       return Promise.all(
```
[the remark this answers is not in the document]
achou11: yeah that makes sense. how does the retry stuff translate code-wise? is there something built into react query to handle the case of "we failed to save 1 of 3 photos, retry for that 1", or do we have to do something manually e.g. re-attempt a mutation in the `onSuccess` callback?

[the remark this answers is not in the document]
achou11: part of me is wondering if this should be updated to only handle a single photo and then the consumer can decide to make concurrent calls if needing to save several photos.

not sure if there's an idiomatic approach that react-query suggests when it comes to saving "batches" of data, but my intuition is that mutations generally hold up better when they map 1-to-1 with the api that you're interacting with. for example, `project.$blobs.create()` accepts a single photo as opposed to many photos, so it feels more natural for the mutation to look similar. all just pure feeling though...

would updating this mutation so that it accepts a single photo and then calling several mutations in parallel make sense, especially given what you described? for example, after pressing the button to save the observation, saving the attachments could look like this:

```ts
const photosToSave = []

// Can also do Promise.all() but wrapped with a try/catch to manually handle the errors since react query won't in that case
const results = await Promise.allSettled(photosToSave.map((photo) => {
  return mutation.mutateAsync(photo))
})

// Depending on if we configure the `retry` option when creating the mutation, we could do checks on `results`

```

src/frontend/screens/ObservationEdit/SaveButton.tsx
achou11: the empty array should handle that but if we're looking to optimize a bit, we can do the check and completely avoid calling the mutation in the case of no savablePhotos 👍

src/frontend/hooks/server/media.ts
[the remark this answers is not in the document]
achou11: more of a UX question, but what should happen in the case that saving some (or all) of the attachments doesn't work, even after the retries? should you still be allowed to save the observation, but only with the attachments that successfully saved? (e.g. observation may be missing some that you took)

doesn't seem like we do much safeguarding in Mapeo Mobile (https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/hooks/useDraftObservation.ts#L272), so probably not a good reference for this question 😄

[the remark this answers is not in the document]
achou11: summarizing the problem that you've highlighted:

> Currently, we abort the process of saving an observation if saving any number of photos fails to save,
but this approach is prone to creating "orphaned" blobs.
> The alternative is to save the observation but excluding photos that failed to save, which is prone to an odd UX of an observation "missing" some attachments.
> This could potentially be alleviated by a more granular and informative UI about the photo-saving state, but currently there is nothing in place.
> Basically, which is worse: orphaned attachments or saving observations that seem to be missing attachments?

from just basic testing between the two paths, i actually found the case of "missing" attachments to be much more jarring, although that's mostly because of a lack of UI to surface that prior to saving the observation. 

For the sake of consistency and lack of a simple solution, I'll have the implementation maintain the same behavior as Mapeo Mobile and have the "orphaned" attachment problem, at least temporarily. Wondering if there's maybe some backend work that we can do to make this API usage less prone to such issues.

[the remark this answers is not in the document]
achou11: re: batching mutations, just realized that if i go with the approach of a mutation per photo, i'll have to do more work to get the pending state of "saving all photos". the `isPending` from `useMutation()` will update for each concurrent mutate call I make. seems like i'd have to use [`useIsMutating()`](https://tanstack.com/query/latest/docs/framework/react/reference/useIsMutating) or [`useMutationState()`](https://tanstack.com/query/latest/docs/framework/react/reference/useMutationState) in order to achieve that 🤔

src/frontend/hooks/server/media.ts
```diff
@@ -1,52 +1,99 @@
...
+  'originalUri'
+>;
+
 export function useCreateAttachmentsMutation() {
   const project = useProject();
 
   return useMutation({
-    mutationFn: async (photos: Photo[]) => {
+    mutationFn: async (photos: SavablePhoto[]) => {
       if (!project) throw new Error('Project instance does not exist');
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/187/commits/cd376ad053091c145c7428c3259ae25972f75cdf

src/frontend/screens/ObservationEdit/SaveButton.tsx
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/187/commits/98f0255207d762169a288d57d093d91509d705b1

src/frontend/screens/ObservationEdit/SaveButton.tsx
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/187/commits/a499581e2fa5d709603999dae356f1410a0a456e

src/frontend/hooks/server/media.ts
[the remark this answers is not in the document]
achou11: (somewhat) addressed via https://github.com/digidem/CoMapeo-mobile/pull/187/commits/db95f650e6c96024644aa57811fdf0d726b0b55e

src/frontend/hooks/server/media.ts
ErikSin: ohh, interesting, `driveDiscoveryId` is more like a foreign key, not a primary key... Ok, yeah, than what you have makes sense

Issue #171, opened 2024-02-05
Fix Slow Start Up Time


ErikSin:
With slower phones, the splash screen is taking over 10 seconds to be hidden. 

Splash Screen is being hidden in the [AppNavigator](https://github.com/digidem/CoMapeo-mobile/blob/9b59f2e3a4dcc1cf4c16838cb9da2968f044c7e3/src/frontend/Navigation/AppNavigator.tsx#L42). 

## To do
- [ ] Determine what step, before app navigator, that is causing the delay.

achou11:
experimenting a little with this and I have some suspicions that a lot of the boot time has something to do with Expo initialization (on the native side). lots of the hanging time is before the react part of the app is even started it seems

PR #194, opened 2024-02-26
chore: update nodejs mobile to v18

achou11:
Towards #171 

- 1/N steps towards potentially addressing the app startup issues. This is not the culprit of our issues but allows us to make upgrades to the development environment and other important deps that are more likely to blame (looking at you expo...)
- No app code changes
- Improves build scripts
- Updates the development node version to 20 (both frontend and backend)
- Updates minimum android version to 6 (sdk 24) - this is required by nodejs mobile react native.

---

Confirmation log via adb:

> 02-26 18:23:47.848 14728 14828 E NODEJS-MOBILE: RUNNING NODE v18.17.1


android/build.gradle
```diff
@@ -35,13 +35,13 @@ loadDotEnv()
...
-        minSdkVersion = 21
+        minSdkVersion = 24
         compileSdkVersion = 33
         targetSdkVersion = 33
         kotlinVersion = findProperty("android.kotlinVersion") ?: "1.8.10"
 
-        // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
-        ndkVersion = "23.1.7779620"
+        // We use NDK >= 24 due to NodeJS Mobile React Native
+        ndkVersion = "24.0.8215888"
```
achou11, line 44: I don't think this was strictly necessary since we don't really rely on the ndk for nodejs-mobile, but maybe better to keep here (i didn't try without this change)

scripts/download-prebuilds.mjs
```diff
@@ -85,19 +89,38 @@ export async function downloadPrebuilds({verbose} = {verbose: false}) {
...
 
-/**
- * @param {{name: string, version: string, target: string}} opts
- * @returns {string}
- */
-function getArtifactTarballName({name, version, target}) {
-  return `${name}-${version}-node-${NODE_ABI}-${target}.tar.gz`;
+function getNodeJsMobileNodeVersions() {
+  const nodeVersionFilePath = new URL(
+    'android/libnode/include/node/node_version.h',
```
achou11, line 94: this corresponds to the header file that the NodeJS project places several version-related constants in

ErikSin (approved):
Everything (that was working before) seems to be working!

PR #195, opened 2024-02-28
chore: upgrade React Native and Expo

### LINKED ISSUE #141: Laggy navigation animation when moving from map screen to other screens
Noticed this while testing on a relatively newer phone (Pixel 6, Android 14).

https://github.com/digidem/CoMapeo-mobile/assets/18542095/20d38eaf-95b8-46b3-a3b7-a7bb08cd2be3

This issue isn't present on my older device (Pixel 2, Android 11), but that's probably because it's not using the same transition to begin with (i think due to being an older OS version with less available nav APIs under the hood)

### LINKED ISSUE #171: Fix Slow Start Up Time

achou11:
Fixes #171 
Fixes #141 

Notes:

- Updates React Native to 0.73.5
  - Updates target to Android 14 (SDK 34)
  - Requires Android build and SDK tools 34
  - Requires Android NDK 25.1.8937393
  - Now requires Java 17 to build the Android app locally.
  - Updates Gradle to 8
  - Converts existing Java files to Kotlin (Kotlin is much nicer to work with, not that it matters for us right now...)
  - Updated a bunch of deps that seemed likely to be affected by the upgrade in React Native (even if their changelogs didn't indicate that). Most notably, worth keeping an eye on `@rnmapbox/maps`
  
- Updates Expo SDK to `50` (and uses the `npx expo install --fix` command to update relevant modules)
  - many changes are derived from [Expo's new upgrade helper](https://docs.expo.dev/bare/upgrade/?fromSdk=48&toSdk=50#androidappsrcmainjavacomhelloworldmainapplicationjava) which is 1) very helpful, but 2) also prone to me copying over changes that may not be necessary or wanted. I think some of the changes there are assuming that you're using Expo's CLI tooling to run/build the app instead of React Native's CLI. Tried my best to be judicious but be mindful about anything that seems broken that could be because of this.

- Replaces `react-native-fs` with `@dr.pogodin/react-native-fs`, which was an intended follow-up from https://github.com/digidem/CoMapeo-mobile/pull/165#issuecomment-1910921347 that I took to liberty of including here

- Updates TypeScript to 5


patches/react-native-svg-transformer+1.3.0.patch
```diff
@@ -0,0 +1,30 @@
...
+--- a/node_modules/react-native-svg-transformer/index.js
++++ b/node_modules/react-native-svg-transformer/index.js
+@@ -1,6 +1,11 @@
+ const { resolveConfig, transform } = require("@svgr/core");
+ const resolveConfigDir = require("path-dirname");
+ 
++// Implementation uses a fragile approach to figure out which babel transformer to use
++// Makes it uses the one for React Native CLI, which is what's used to bundle our app
++// https://github.com/kristerkari/react-native-svg-transformer/issues/333
++const babelTransformer = require("@react-native/metro-babel-transformer");
```
ErikSin, line 12: Can you add this description to the readme.md as well

ErikSin (approved):
I tested the app on a device, everything seems to be working as expected.

I don't love that a bunch of (seemingly) unrelated dependencies were upgraded...but I think where we are at this stage of the app it is ok as testing is pretty manageable.

I am approving, but just note that I didn't do an extensive code review as most of your changes come from the upgrade helper...let me know if you think I should review it another way

ErikSin:
Another note. It has shortened the app start up time on my very slow device, but it is still uncomfortably long (more than 5 seconds). Probably be better in a non dev build, but just wanted to mention that here

achou11:
> Another note. It has shortened the app start up time on my very slow device, but it is still uncomfortably long (more than 5 seconds). Probably be better in a non dev build, but just wanted to mention that here

Good to know! would be curious to know how long the startup for you is with a bare react native app, just to compare. My guess is that the overhead of Expo is still somewhat present, although improved compared to before this PR.

patches/react-native-svg-transformer+1.3.0.patch
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/195/commits/daf2f56083f551eb6654520048b465e821a5701e

PR #200, opened 2024-03-12
Es/saved observations

ErikSin:
Loads initial screen based on following factors
- if the user has no name set, user is navigated to intro screen where they will be prompted to add a name
- if the user has a name set, and no saved observation, they are navigated to the home (map) screen
- if the user has a saved observation, but no matching preset, implies user has started and observation and has not chosen a preset yet. User will be navigated to `PresetChooser` screen
- If the user has a saved observation with a matching preset, the user is navigated to `ObservationEdit` Screen

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -19,6 +19,7 @@ import {
   LocalDiscoveryProvider,
   createLocalDiscoveryController,
 } from './contexts/LocalDiscoveryContext';
+import {Loading} from './sharedComponents/Loading';
 
 const queryClient = new QueryClient();
 const messagePort = new MessagePortLike();
@@ -56,7 +57,9 @@ const App = () => {
               <ActiveProjectProvider>
                 <PhotoPromiseProvider>
                   <SecurityProvider>
-                    <AppNavigator permissionAsked={permissionsAsked} />
+                    <React.Suspense fallback={<Loading />}>
+                      <AppNavigator permissionAsked={permissionsAsked} />
+                    </React.Suspense>
                   </SecurityProvider>
                 </PhotoPromiseProvider>
               </ActiveProjectProvider>
--- a/src/frontend/Navigation/AppNavigator.tsx
+++ b/src/frontend/Navigation/AppNavigator.tsx
@@ -13,6 +13,7 @@ import {useDeviceInfo} from '../hooks/server/deviceInfo';
 import {Loading} from '../sharedComponents/Loading';
 import {createDeviceNamingScreens} from './ScreenGroups/DeviceNamingScreens';
 import {usePrefetchLastKnownLocation} from '../hooks/useLastSavedLocation';
+import {usePersistedDraftObservation} from '../hooks/persistedState/usePersistedDraftObservation';
 
 // import {devExperiments} from '../lib/DevExperiments';
 
@@ -34,6 +35,9 @@ import {usePrefetchLastKnownLocation} from '../hooks/useLastSavedLocation';
 
 export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
   const {formatMessage} = useIntl();
+  const existingObservation = usePersistedDraftObservation(
+    store => store.value,
+  );
 
   const deviceInfo = useDeviceInfo();
   usePrefetchLastKnownLocation();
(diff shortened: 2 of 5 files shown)


src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -48,14 +60,44 @@ export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
...
-        initialRouteName="IntroToCoMapeo"
-        screenOptions={NavigatorScreenOptions}>
-        {deviceInfo.data && deviceInfo.data.name
-          ? createDefaultScreenGroup(formatMessage)
-          : createDeviceNamingScreens(formatMessage)}
-      </RootStack.Navigator>
-    </React.Suspense>
+    <RootStack.Navigator
+      initialRouteName={setInitialRouteName({
+        hasName: !!deviceInfo.data?.name,
```
achou11, line 65: nit:

```suggestion
        hasDeviceName: !!deviceInfo.data?.name,
```

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -48,14 +60,44 @@ export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
...
+        : createDeviceNamingScreens(formatMessage)}
+    </RootStack.Navigator>
   );
 };
+
+function setInitialRouteName(
+  initialInfo:
+    | {hasName: false}
+    | {
+        hasName: true;
```
achou11, line 81: nit: 

```suggestion
    | {hasDeviceName: false}
    | {
        hasDeviceName: true;
```

src/frontend/App.tsx
```diff
@@ -56,7 +57,9 @@ const App = () => {
               <ActiveProjectProvider>
                 <PhotoPromiseProvider>
                   <SecurityProvider>
-                    <AppNavigator permissionAsked={permissionsAsked} />
+                    <React.Suspense fallback={<Loading />}>
+                      <AppNavigator permissionAsked={permissionsAsked} />
+                    </React.Suspense>
```
achou11, line 62: not familiar with suspense enough to really understand why this is needed - but it definitely sticks out as something i'd have additional questions about. guess my main question would be: what is this suspending on, and is this the preferred way of doing it vs some conditional rendering in one of the wrapping providers?

achou11 (changes requested):
Getting a (development-only) warning in the case of being directed immediately to a screen with a header left back button and then trying to go back (since there's no history at this point). in production, i guess it would just no-op, which doesn't seem ideal

Think the behavior i'd expect in that case would be to go back to the map screen. thoughts?

ErikSin:
> Getting a (development-only) warning in the case of being directed immediately to a screen with a header left back button and then trying to go back (since there's no history at this point). in production, i guess it would just no-op, which doesn't seem ideal
> 
> Think the behavior i'd expect in that case would be to act as though we're trying to to go back to the map screen. thoughts?


Hmm this actually is an error with the presets screen. It should be asking the user whether they want to `Discard Observation` or `Cancel`, which obviously is not happening

ErikSin:
> > Getting a (development-only) warning in the case of being directed immediately to a screen with a header left back button and then trying to go back (since there's no history at this point). in production, i guess it would just no-op, which doesn't seem ideal
> > Think the behavior i'd expect in that case would be to act as though we're trying to to go back to the map screen. thoughts?
> 
> Hmm this actually is an error with the presets screen. It should be asking the user whether they want to `Discard Observation` or `Cancel`, which obviously is not happening

Adressed: https://github.com/digidem/CoMapeo-mobile/pull/200/commits/df51b4153a8d22cad7f90ac35db15c76b00162ff

src/frontend/App.tsx
ErikSin: Previously this `<React.Suspense />` was inside the App Navigator. Any useSuspenseQuery(), triggers a loading state, and then the fallback component in the nearest suspense component up the tree is shown to the user until loading is done.

So the reason I had to move it out of the App Navigator, is that component needs access to all the presets (in order to check if the saved observations has an exisiting preset, which uses the `presetMatcher` function). The presets are retrieved through a useSuspenseQuery. The suspense has to be a parent component to any suspenseQuery.

The suspense API is now a stable API, and the [caveats of use](https://react.dev/reference/react/Suspense#caveats) don't apply to anything we do. Overall it is also a much cleaner code as opposed to conditional rendering. 

I think i made a comment the other day that suspense is not the preferred API, but i was confusing it with `ErrorBoundary`...

src/frontend/Navigation/AppNavigator.tsx
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/200/commits/9f371e91b95e13606bbb1fce43259fd60106faa7

src/frontend/Navigation/AppNavigator.tsx
ErikSin: https://github.com/digidem/CoMapeo-mobile/pull/200/commits/9f371e91b95e13606bbb1fce43259fd60106faa7

src/frontend/App.tsx
achou11: thanks for the explanation!

achou11 (approved):
I suggest pulling in the latest updates from the main branch just to make sure that these changes still work as expected

achou11:
Just for posterity: i prematurely created an issue that I thought existed prior to this PR, but was actually specific to this PR

https://github.com/digidem/CoMapeo-mobile/issues/209

PR #201, opened 2024-03-13
fix: fix creation of intl polyfills file

### LINKED ISSUE #175: Fix Date Created in Observation List
Currently the date/time that an observation is created, in the observation list, is showing up as a negative number. Should be a date, and time

Image

achou11:
Fixes #175 

Fixes a few issues:

- ~Adds a polyfill that's depended on by some of the other polyfills (https://formatjs.io/docs/polyfills/intl-getcanonicallocales)~ EDIT: technically not necessary, but keeping it unless strongly opposed (https://github.com/digidem/CoMapeo-mobile/pull/201#issuecomment-1997795429)
- Fixes build script to add import lines for the locale data for each polyfill

---

Generated polyfill file before:

```ts
// This file is automatically generated through scripts/build-intl-polyfills.js. Do not edit this directly!
// If you need to rebuild it, you can run `npm run build:intl-polyfills`
import "@formatjs/intl-locale/polyfill";
import "@formatjs/intl-pluralrules/polyfill";
import "@formatjs/intl-relativetimeformat/polyfill";
```

Generated polyfill file after:

```ts
// This file is automatically generated through scripts/build-intl-polyfills.mjs. Do not edit this directly!
// If you need to rebuild it, you can run `npm run build:intl-polyfills`
import "@formatjs/intl-getcanonicallocales/polyfill";
import "@formatjs/intl-locale/polyfill";

import "@formatjs/intl-pluralrules/polyfill";
import "@formatjs/intl-pluralrules/locale-data/en";

import "@formatjs/intl-relativetimeformat/polyfill";
import "@formatjs/intl-relativetimeformat/locale-data/en";
```


scripts/build-intl-polyfills.mjs
```diff
@@ -0,0 +1,118 @@
...
+
+  writer.write('\n');
+
+  // Write lines to load plural rules polyfill
+  writer.write(createImportStatement('@formatjs/intl-pluralrules/polyfill'));
+  for (const locale of locales) {
+    writer.write(
+      createImportStatement(`@formatjs/intl-pluralrules/locale-data/${locale}`),
+    );
+  }
```
ErikSin, line 82: I am not sure if this matters, but perhaps we should check if the [locale needs to be polyfilled](https://formatjs.io/docs/polyfills/intl-pluralrules#dynamic-import--capability-detection), and only write the import statement as necessary.

Its working now, but would creating a polyfill if it is not needed cause a problem?

scripts/build-intl-polyfills.mjs
```diff
@@ -0,0 +1,118 @@
...
+  writer.write(
+    createImportStatement('@formatjs/intl-relativetimeformat/polyfill'),
+  );
+  for (const locale of locales) {
+    writer.write(
+      createImportStatement(
+        `@formatjs/intl-relativetimeformat/locale-data/${locale}`,
+      ),
+    );
+  }
```
ErikSin, line 96: Same comment as above, should we [do a check here](https://formatjs.io/docs/polyfills/intl-pluralrules#dynamic-import--capability-detection)?

ErikSin (changes requested):
I think we should document why we have this intl script. Doesn't have to be extensive, it can be a couple lines, and a link to the format js documents.

I know we haven't discussed this yet, but I think it would be good to document any things that fall outside of the regular dev flow or has logic that does not follow regular conventions. I am compiling a list of things we can document and lets discuss at a later time.

For now, lets just create a `readme` inside the scripts folder, and can you document what this script is for?

Otherwise, 2 related comments about whether we should check if we need to write a polyfill. Im not sure if that is necessary, Im just not sure if it would break if we create a polyfill that is not needed...

scripts/build-intl-polyfills.mjs
achou11: the `shouldPolyfill()` functions provided for each of these won't make sense to use in this context because it would be checking for properties in the development node runtime, and not the runtime that exists on the device. 

https://github.com/formatjs/formatjs/blob/main/packages/intl-pluralrules/should-polyfill.ts

> Its working now, but would creating a polyfill if it is not needed cause a problem?

a properly implemented polyfill is relatively inexpensive and should not cause any side effects if it's not needed. they do a proper job of this based on the implementations i'm seeing (for example, for plural rules: https://github.com/formatjs/formatjs/blob/main/packages/intl-pluralrules/polyfill.ts). writing the import lines into the file even if the runtime eventually gets built-in support is not problematic, as the polyfill will then no-op.

the polyfills check for two things:

1. does the interface in question already exist or not in the runtime?
2. does formatjs provide support for the locale(s) specified

In our case, only (2) really matters when we're running this as a build script. as a result, it makes more sense to only check the supported locales here.

achou11:
> I think we should document why we have this intl script. Doesn't have to be extensive, it can be a couple lines, and a link to the format js documents.
> ...
> For now, lets just create a readme inside the scripts folder, and can you document what this script is for?

done via https://github.com/digidem/CoMapeo-mobile/pull/201/commits/743bf6196c79a70511b40e48a864e379c3b27933 👍

achou11:
Turns out Hermes already supports `Intl.getCanonicalLocales` (https://hermesengine.dev/docs/intl), so will update the PR to not include that polyfill I guess

achou11:
Turns out Hermes already supports `Intl.getCanonicalLocales` (https://hermesengine.dev/docs/intl) so technically could remove it, but think it's fine to leave in given that the polyfill will defer to the existing interface at runtime and no-op. think keeping it is fine since it's inexpensive and will make it more resilient to cases where we might be running on a different engine (e.g. JSC on iOS)

ErikSin (approved):

PR #203, opened 2024-03-27
chore: allow @node-rs/crc32, a dependency of mapeo-core, to be bundled

EvanHahn:
`@mapeo/core` (indirectly) depends on `@node-rs/crc32`, which can't be bundled. Replace it with a pure JavaScript implementation.

### DIFF
--- a/src/backend/src/node-rs-crc32-shim.js
+++ b/src/backend/src/node-rs-crc32-shim.js
@@ -1,5 +1,8 @@
 /**
  * Shim for @node-rs/crc32 written in pure JavaScript.
+ *
+ * Modified from [this sample implementation](https://www.w3.org/TR/2022/WD-png-3-20221025/#D-CRCAppendix).
+ *
  * @module
  */
 
--- a/src/backend/src/node-rs-crc32-shim.js
+++ b/src/backend/src/node-rs-crc32-shim.js
@@ -1,0 +1,37 @@
+/**
+ * Shim for @node-rs/crc32 written in pure JavaScript.
+ * @module
+ */
+
+/** @type {undefined | Uint32Array} */
+let crcTable
+
+/**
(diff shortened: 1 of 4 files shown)


src/backend/src/node-rs-crc32-shim.js
```diff
@@ -0,0 +1,37 @@
...
+
+/** @type {undefined | Uint32Array} */
+let crcTable
+
+/**
+ * @param {string | Buffer} input
+ * @param {number | undefined | null} [initialState]
+ * @returns {number}
+ */
+function crc32(input, initialState = 0) {
```
ErikSin, line 14: did you reference this from somewhere? If so, can you add the link?

EvanHahn: Done in 1b448451d5a4d23a4ecba7d578f02e6cc8ca45f4.

ErikSin:
Hmm, when I stack the [upgrade mapeo-core PR](https://github.com/digidem/CoMapeo-mobile/tree/es/update-mapeo-core) on top of this PR and run it, i am getting this error: `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'original-fs' imported from /data/data/com.comapeo/files/nodejs-project/index.js` and I cannot open the app. 

I ran an `npm ls original-fs` and there is no direct dependency within our tree that uses this

EvanHahn:
That's surprising...does the same happen when you stack atop `main`?

ErikSin:
> That's surprising...does the same happen when you stack atop `main`?

it doesn't build if i stack on main because of the yazul dependency...so im not really sure if it is because of the updated mapeo-core or your shimming work...

ErikSin:
rebased on https://github.com/digidem/CoMapeo-mobile/pull/205 and merged into main

PR #206, opened 2024-03-29
Es/listen for invites

ErikSin:
Updated PR for listening to invites

User can invite another user to their project. The invitee will see a modal pop up asking if they would like to join.

## Architecture Notes
- We are using tanstack query to handle the events `[invite-recieved](https://github.com/digidem/mapeo-core-docs/tree/main/client-docs#invite-received)` and `invites-removed`. Every time one of of these events get fired, we invalidate the query cache that handles the api `mapeoApi.invite.getPending()`. This causes tanstack to re call the api, leaving the front end with the most updated list of all pending invites. These event listeners are initialized at the top of the react tree as the user should always be listening for new invites.
- `useProjectInvite` hook tracks all the pending invites based on the api `mapeoApi.invite.getPending()`. It returns the oldest invite (not yet sorted by date, but has been added to the API), and all the methods associated with an invite (`accept()` and `reject()`)
-`ProjectInviteBottomSheet` is a bottoms sheet that encapsulates the `useProjectInvite` hook. When there is an invite (as returned by the hook), it will open and prompt the user to `accept` or `reject` the invite. It is also at the top of the react tree as the user needs to be able to accept an invite from any screen. If the user is in the middle of editting/creating an observation and there is an invite, the bottoms sheet will not open until after the user has completed editting/creating that observation.

# To do
- [ ] When a invitee declines an invite, the invitor is NOT notified (the API for that is being built)
- [ ] User should leave a project before the join a project, this will be in a follow up PR
- [ ] After accepting an invite, the user should be prompted to go to sync screen (which is not yet complete)


src/frontend/hooks/useProjectInvite.ts
ErikSin, line 1: @gmaclennan I deviated from [your suggested architecture](https://github.com/digidem/CoMapeo-mobile/pull/172#pullrequestreview-1877951847).

I tried with your suggested architecture, but there was a problem with the UI and state of the modal not lining up.

In your implementation the state of the invite was being managed by the hook. Since the states were our own type (eg, "noInvite", "accepted"....etc), it required the states to be manually set based on the state of the API calls. I was imperatively changing the state based on the action of the user. I tried doing it in a declarative way by deriving the state, but it lead to some difficult code to follow.

I separated the state from the hook, and allowed the bottom sheet to deal with that state instead. So now the hook just returns the tanstack query hooks (which encapsulates the state of the api call). The bottom sheet can derive its own ui state based on that, without having to create custom state types. 

I think the code is much cleaner this way, and allows the hook to deal with just the invites, and the bottoms sheet to deal with react state.

src/frontend/hooks/useProjectInvite.ts
```diff
@@ -0,0 +1,33 @@
...
+  const resetState = () => {
+    acceptMutation.reset();
+    rejectMutation.reset();
+  };
+
+  const clearAllInvites = useCallback(() => {
+    invites.forEach(inv => {
+      mapeoApi.invite.reject(inv);
+    });
+  }, [invites, mapeoApi]);
```
ErikSin, line 23: @EvanHahn this is not ideal as everytime we delete an invite react re-renders. Is it possible to create an api call `mapeoApi.invite.clearAllPendingInvites()`

EvanHahn: Sure. Would `rejectAllPendingInvites()` work?

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/index.tsx
```diff
@@ -42,6 +42,7 @@ export const ReviewAndInvite: NativeNavigationComponent<'ReviewAndInvite'> = ({
         }
       })
       .catch(err => {
+        console.log(err);
```
achou11, line 45: remove?

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -31,8 +32,16 @@ export const SelectDevice: NativeNavigationComponent<'SelectDevice'> = ({
...
+  const projectMembers = useProjectMembers();
   const devices = useLocalPeers();
+  const nonMemberDevices = !projectMembers.data
+    ? devices
+    : devices.filter(
+        device =>
+          !projectMembers.data.some(
+            member => member.deviceId === device.deviceId,
+          ),
+      );
```
achou11, line 44: nit but usually a little easier to follow when using the non-negated form in cases like this e.g. `projectMembers.data ? ... : ...`

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
```diff
@@ -62,43 +67,65 @@ export const YourTeam: NativeNavigationComponent<'YourTeam'> = ({
...
-          style={{
-            flexDirection: 'row',
-            justifyContent: 'center',
-            alignItems: 'center',
+      {deviceInfo.isLoading ? (
+        <UIActivityIndicator size={30} />
+      ) : !deviceInfo.data ||
+        !coordinators.some(
+          coordinator => coordinator.deviceId === deviceInfo.data.deviceId,
+        ) ? null : (
```
achou11, line 103: think it's worth extracting this into a variable e.g. `const shouldShowAddDeviceButton = ...`

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
```diff
@@ -1,23 +1,28 @@
...
+} from '../../../../hooks/server/projects';
 import {Loading} from '../../../../sharedComponents/Loading';
+import {UIActivityIndicator} from 'react-native-indicators';
 import {useDeviceInfo} from '../../../../hooks/server/deviceInfo';
-import {
-  COORDINATOR_ROLE_ID,
-  CREATOR_ROLE_ID,
-  MEMBER_ROLE_ID,
-} from '../../../../sharedTypes';
+import {projectSettings} from '@mapeo/schema/dist/validations';
```
achou11, line 23: remove unused import

```suggestion

```

src/frontend/sharedComponents/BottomSheet/Content.tsx
achou11, line 1: what are the styling changes here accounting for? was the original implementation not aligned with the designs? remember being very mindful of the styles implementation here just checking if it's specific to the new prop that was added or a general visual issue

src/frontend/hooks/server/invites.ts
```diff
@@ -0,0 +1,37 @@
+import {useMutation, useSuspenseQuery} from '@tanstack/react-query';
+import {useApi} from '../../contexts/ApiContext';
+
+export const INVITE_KEY = 'invites';
+
+export function useInvites() {
```
achou11, line 6: nit: would make the name here more specific e.g. `usePendingInvites()`

src/frontend/hooks/server/invites.ts
```diff
@@ -0,0 +1,37 @@
+import {useMutation, useSuspenseQuery} from '@tanstack/react-query';
+import {useApi} from '../../contexts/ApiContext';
+
+export const INVITE_KEY = 'invites';
```
achou11, line 4: would use a more specific key name e.g. `PENDING_INVITES_KEY = 'pending_invites'`

src/frontend/hooks/server/invites.ts
```diff
@@ -0,0 +1,37 @@
...
+    queryFn: async () => {
+      return await mapeoApi.invite.getPending();
+    },
+  });
+}
+
+export function useAcceptInvite(inviteId?: string) {
+  const mapeoApi = useApi();
+  return useMutation({
+    mutationKey: [INVITE_KEY, inviteId],
```
achou11, line 19: react query-specific question but given that `mutationKey` doesn't seem to be used anywhere (e.g. no usage of `useMutationState()`), does it make sense to specify the `inviteId` param in the `mutationFn` instead? honestly a little bit shaky on when to use which (pass in the top-level hook vs mutationFn).

the main drawback with the current approach is that you're dealing with a potentially undefined value. not a big deal but just something i noticed when looking at usage

in this case, wondering if there's a meaningful difference between these two (there may not be!):

```js
// Whenever `invite.inviteId` changes, `mutation` changes
const mutation = useAcceptInvite(invite?.inviteId)

// the mutate function holds implicit information about the invite id to work with
// this won't do anything if `invite.inviteId` is undefined
mutation.mutate(undefined, { ... })
```

```js
// `mutation` is agnostic to invite.inviteId
const mutation = useAcceptInvite()

// the mutate function is more "functional" and doesn't hold any implicit information
// also can do a check to make sure the param of interest actually exists
if (invite?.inviteId) mutation.mutate({ inviteId: invite.inviteId }, { ... })
```

achou11 (commented):
Mostly a first pass at just the code itself. Haven't gotten around to testing it and wrapping my head around the more complicated hooks that are introduced

achou11:
Noticing on the device that accepts the invite, if I go to the `Your Team` screen, it's still showing the UI associated with not being on a project (probably a missing query invalidation somewhere?):

on `Your Team` screen:

image

although navigating to the `Create or Join Project` screen, it seems to correctly know that it's on a project already:

image

src/frontend/hooks/useProjectInvite.ts
achou11: Personally struggling to know the best way to approach this is. i did my own implementation of the hooks that's a bit stricter about functional separations, but not sure if other would agree that it's easier to maintain/follow 😅 

<details>
<summary>Lengthy explanation of what I explored, may not be useful info though</summary>

- `useInvites()` = existing hook that provides information about the invites

- `useProjectInviteHandler()` (name is not great...) = new hook i created to surface `accept` and `reject` actions (can add more later on). these actions ideally invalidate `useInvites()` so that it provides updated info and remains the source of truth for that. this hook also provides a `status` field, which is a more semantic description of the async actions occurring (similar to what Gregor suggested). This `status` is derived from the mutations used internally (see above comment). the actions are only available if the `status` is `idle` (i.e. you shouldn't be able to accept or reject an invitation after already responding to it).

then the usage looks approximately like this:

```tsx
const invites = useInvites()
const inviteHandler = useProjectInviteHandler()

// Render UI based on status
if (inviteHandler.status === 'accepted') {
  return <AcceptedView />
}

// Have to check the status to make sure that you can perform an action
if (inviteHandler.status === 'idle') {
  inviteHandler.accept(...)
}
```

can probably combine `useInvites()` and `useProjectInviteHandler()` like Gregor originally suggested, but didn't spend the time exploring how that would work (or if it's easier to use/maintain)

</details>

achou11 (commented):
Some more comments as I poke around.

Overall, the happy path works very well during testing!

src/frontend/hooks/useProjectInvite.ts
```diff
@@ -0,0 +1,33 @@
...
+  // this will eventually sort invite by date
+  const invite = invites[0];
+  const acceptMutation = useAcceptInvite(invite?.inviteId);
+  const rejectMutation = useRejectInvite(invite?.inviteId);
+  const mapeoApi = useApi();
+
+  const resetState = () => {
+    acceptMutation.reset();
+    rejectMutation.reset();
+  };
```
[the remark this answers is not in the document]
ErikSin: Becaues the hook is inside a modal, and the modal is at the top level, the modal and the hook never get unmounted (a modal just gets hidden, it doesn't unmount when closed). Because it never unmounts the state of the modal is not automatically reset. So, when we close the modal we want to make sure that we reset the state so the user does not see the last state

src/frontend/hooks/useProjectInvite.ts
ErikSin: Is there a world where the user accepts an invite and we do NOT want to clear the pending invites? IF there is, then I think it would make sense for the front end to handle clearing the pending invites.

If we want to always clear the pending invites on accept, then I agree with you andrew, this should be handled by the backend. 

I can imagine that maybe in the future we want to list all pending invites...but we can expose that in the future if we do need that. So for now @EvanHahn lets just clear all invites on accept in the backend and not expose an API for it?

src/frontend/hooks/useProjectInvite.ts
ErikSin: To me, it feels clearer to have the modal derive its own UI state based on the states of the mutations. The following logic feels very straightforward:
- if `acceptMutation` or `rejectMutation` is loading, render loading UI
- if `acceptMutation` is successful, render success UI.
- if `rejectMutation` is successful, close modal

Deriving a `Type` and then using that type feels unnecessary when the logic above feels intuitive.

src/frontend/hooks/server/invites.ts
```diff
@@ -0,0 +1,37 @@
...
+  const mapeoApi = useApi();
+  const invites = useInvites().data;
+  return useMutation({
+    mutationKey: [INVITE_KEY, inviteId],
+    mutationFn: async () => {
+      if (!inviteId) return;
+      mapeoApi.invite.reject({inviteId});
+    },
+  });
+}
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/08e8da0a50ed4aab5af65a7381b60bd137ea40f2

src/frontend/hooks/server/invites.ts
```diff
@@ -0,0 +1,37 @@
...
+export function useAcceptInvite(inviteId?: string) {
+  const mapeoApi = useApi();
+  return useMutation({
+    mutationKey: [INVITE_KEY, inviteId],
+    mutationFn: async () => {
+      if (!inviteId) return;
+      mapeoApi.invite.accept({inviteId});
+    },
+  });
+}
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/08e8da0a50ed4aab5af65a7381b60bd137ea40f2

src/frontend/hooks/server/invites.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/98c4cda753be4dbb96c252b51e57f1d1bd031991

src/frontend/hooks/server/invites.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/9da9d474b3fde53e98b7114a8f78d48e6742bf3a

src/frontend/hooks/useProjectInvite.ts
```diff
@@ -0,0 +1,33 @@
+import {Invite} from '@mapeo/core/dist/invite-api';
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/f1ffc6164440a2fac2342b4c9b0e96aae0ba7a73

src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/68830d6ef35767bac401868da18c13a653f25390

src/frontend/hooks/server/invites.ts
ErikSin: Yea, reading though the documentation we dont need to have a query key associated with mutations.

Also it looks like the convention is follow the latter pattern, that is, let the mutation handle the params-not the top level hook

addressed: https://github.com/digidem/comapeo-mobile/pull/206/commits/09003f1bf87bbc97ac776cc15f89c4ab6f4bc4f4

src/frontend/hooks/useProjectInvite.ts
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/e9d05a9472261ba7dd0cb44a8d964dd7bc031237

src/frontend/sharedComponents/BottomSheet/Content.tsx
ErikSin: Hmm yeah. The styling changes were not necessary, but they did make it a little easier. Before, each component had its own spacing, and now the container deals with the spacing. It just made it a little easier to deal with the UI changing (aka, the buttons turning into a loading state).

It didn't seem to effect general styling of the modal. 

But do you remember the specific reasons you chose the original styling choice?

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/206/commits/a9e79e44b113478dc5f6bc3596bdc77bb6882af9

achou11 (approved):
Nice work! Don't think the merge conflicts will introduce any major changes, so should be good to merge after you do that

src/frontend/sharedComponents/BottomSheet/Content.tsx
achou11: > and now the container deals with the spacing

side note: what's been a big win on React Native's side is that they've introduced [flexbox gap](https://reactnative.dev/blog/2023/01/12/version-071#simplifying-layouts-with-flexbox-gap), which makes this much cleaner and easier to do

> But do you remember the specific reasons you chose the original styling choice?

Not really unfortunately, but I just remember that I was pretty considerate about making sure that it was implemented with space-responsiveness being a priority (but maybe i'm just overconfident 😄 ). probably not worth trying to figure out as I think our sheet implementation needs some attention anyways, which could include styling adjustments

Issue #56, opened 2023-09-18
Github Actions


ErikSin:
Create github actions to run ci and start the build process.

## To Do
- [ ] run backend tests
- [ ] run frontend test

achou11:
Already have some basics set up. Closing in favor of a future task that's better scoped

PR #208, opened 2024-04-01
chore: add basic github workflow

achou11:
Towards https://github.com/digidem/CoMapeo-mobile/issues/56

Introduces a basic github action that runs on non-draft PRs. ~Currently only does typechecking for both backend and frontend.~

Does the following checks for now:

- all: formatting checks
- backend: type checks
- frontend: unit tests

ESLint (for all) and typechecking (for frontend) to be done as a follow-up

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,8 +4,8 @@ on:
   push:
     branches: [main]
   pull_request:
-    # By default, a workflow only runs when a pull_request 's activity type is
-    # opened , synchronize , or reopened. Adding ready_for_review here ensures
+    # By default, a workflow only runs when a pull_request's activity type is
+    # opened, synchronize, or reopened. Adding ready_for_review here ensures
     # that CI runs when a PR is marked as not a draft, since we skip CI when a
     # PR is draft
     types: [opened, synchronize, reopened, ready_for_review]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,53 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # By default, a workflow only runs when a pull_request 's activity type is
+    # opened , synchronize , or reopened. Adding ready_for_review here ensures
+    # that CI runs when a PR is marked as not a draft, since we skip CI when a
+    # PR is draft
+    types: [opened, synchronize, reopened, ready_for_review]
+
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          node-version-file: '.nvmrc'
+      - name: Install deps
+        run: npm ci
+      - name: Build translations
+        run: npm run build:translations
+      - name: Build Intl polyfills
+        run: npm run build:intl-polyfills
+      - name: Check types
+        run: npm run lint:types
```
achou11, line 53: could potentially check formatting + eslint too by using `npm run lint` but wanted to check to see what people thought would be worth doing in this workflow first.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # By default, a workflow only runs when a pull_request 's activity type is
+    # opened , synchronize , or reopened. Adding ready_for_review here ensures
+    # that CI runs when a PR is marked as not a draft, since we skip CI when a
+    # PR is draft
```
EvanHahn, line 10: Nitpick to avoid weird spacing:

```suggestion
    # By default, a workflow only runs when a pull_request's activity type is
    # opened, synchronize, or reopened. Adding ready_for_review here ensures
    # that CI runs when a PR is marked as not a draft, since we skip CI when a
    # PR is draft
```

EvanHahn (changes requested):
Totally subjective, but: I think we should merge this such that CI passes, even if the tests are incomplete. I would do this by running `npm run lint:prettier` and `npm test` (which pass) instead of `lint:types`.

Later, we should fix types & ESLint, then enable them in CI.

Just my opinion. I think this approach is also reasonable.

achou11:
> I would do this by running `npm run lint:prettier` and `npm test` (which pass) instead of `lint:types`.

any thoughts on running prettier in a separate job, since it applies to both frontend and backend? currently laid out such that there are separates jobs for backend and frontend checks.

EvanHahn:
> any thoughts on running prettier in a separate job, since it applies to both frontend and backend?

No preference as long as there's no way for Prettier mistakes to slip through. Maybe a separate job for tasks that apply to both?

EvanHahn (approved):

PR #212, opened 2024-04-01
feat: add GPS pill and screen with GPS-related info

### LINKED ISSUE #199: GPS Pill

achou11:
Closes #199

Mostly a direct port of the implementation from Mapeo Mobile, although a few adjustments are made to account for the differences in hooks.

---

Preview:

- Map screen (location services enabled):

  image

- Map screen (location services disabled):
  
  image

- GPS modal (location services enabled):

  image

- GPS modal (location services disabled):

  image

### DIFF
--- a/src/frontend/lib/utils.ts
+++ b/src/frontend/lib/utils.ts
@@ -62,10 +62,10 @@ export function getLocationStatus({
   location?: LocationObject;
   providerStatus?: LocationProviderStatus;
 }): LocationStatus {
-  const gpsNotAvailable = !providerStatus?.gpsAvailable;
-  const locationServicesDisabled = !providerStatus?.locationServicesEnabled;
+  const gpsAvailable = !!providerStatus?.gpsAvailable;
+  const locationServicesEnabled = !!providerStatus?.locationServicesEnabled;
 
-  if (gpsNotAvailable || locationServicesDisabled) return 'error';
+  if (!gpsAvailable || !locationServicesEnabled) return 'error';
 
   const positionStale =
     location && Date.now() - location.timestamp > STALE_TIMEOUT;
--- a/src/frontend/lib/utils.ts
+++ b/src/frontend/lib/utils.ts
@@ -2,6 +2,11 @@
 import {fromLatLon} from 'utm';
 import {SelectOptions, LabeledSelectOption} from '../sharedTypes/PresetTypes';
 import {Preset, Observation} from '@mapeo/schema';
+import {
+  LocationObject,
+  LocationPermissionResponse,
+  LocationProviderStatus,
+} from 'expo-location';
 
 // import type {
 //   ObservationValue,
@@ -19,15 +24,6 @@ import {Preset, Observation} from '@mapeo/schema';
 // export function getDisplayName(WrappedComponent: any) {
 //   return WrappedComponent.displayName || WrappedComponent.name || "Component";
 // }
-
-// // If the current position on the app state is more than 60 seconds old then we
-// // consider it stale and show that the GPS is searching for a new position
-// const STALE_TIMEOUT = 60 * 1000; // 60 seconds
-// // If the precision is less than 10 meters then we consider this to be a "good
-// // position" and we change the UI accordingly
-// const GOOD_PRECISION = 10; // 10 meters
-
-export type LocationStatus = 'searching' | 'improving' | 'good' | 'error';
 
 // Little helper to timeout a promise
 // export function promiseTimeout<T>(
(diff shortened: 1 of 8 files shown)


src/frontend/screens/GpsModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  return (
+    <ScrollView style={styles.container} testID="gpsScreenScrollView">
+      <View style={styles.infoArea}>
+        <Text style={styles.sectionTitle}>
+          <FormattedMessage {...m.lastUpdate} />
+        </Text>
+        <DateDistance
+          style={styles.rowValue}
+          date={locationTimestamp ? new Date(locationTimestamp) : new Date()}
```
achou11, line 94: noting that there's a notable difference in implementation for this between Mapeo Mobile and here

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/GpsModal.tsx#L86

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/GpsModal.tsx#L135-L140

Attempting to directly copy over the implementation caused infinite render loops when turning off location services. Seems to be something about `new Date('None')`, which returns `Invalid Date` (which is confusingly a `Date` type...) and that not playing nicely with React Intl's `FormattedRelativeTime` component (used in `DateDistance`).

src/frontend/screens/GpsModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+const GpsModalRow = ({label, value}: {label: string; value: string}) => (
+  <View style={styles.row}>
+    <Text style={styles.rowLabel}>{label}</Text>
+    <Text style={styles.rowValue}>{value}</Text>
+  </View>
+);
+
+export const GpsModal = () => {
+  const {location} = useLocation({maxDistanceInterval: 15});
```
achou11, line 77: not sure what the number here should be. using the one defined for the MapScreen

src/frontend/sharedComponents/GpsPill.tsx
```diff
@@ -0,0 +1,106 @@
...
+            {text}
+          </Text>
+        </View>
+      </TouchableOpacity>
+    );
+  },
+);
+
+export const GpsPill = ({onPress}: Pick<Props, 'onPress'>) => {
+  const locationState = useLocation({maxDistanceInterval: 15});
```
achou11, line 62: not sure what the number here should be. using the one defined for the MapScreen

src/frontend/lib/utils.ts
```diff
@@ -50,30 +46,40 @@ export type LocationStatus = 'searching' | 'improving' | 'good' | 'error';
...
+
+export function getLocationStatus({
+  location,
+  providerStatus,
+}: {
+  location?: LocationObject;
+  providerStatus?: LocationProviderStatus;
+}): LocationStatus {
+  const gpsNotAvailable = !providerStatus?.gpsAvailable;
+  const locationServicesDisabled = !providerStatus?.locationServicesEnabled;
```
ErikSin, line 66: nit: would call `gpsAvailable` and `locationServicesEnabled` to better match location services API.

src/frontend/sharedComponents/GpsPill.tsx
ErikSin: I think this should have a `maxDistanceInterval` of 0. Setting it to 15 makes sense for the map because we do not want the map to be constantly re-rendering. With the accuracy, we do want the user to know that the observation that they are saving has the most up to date accuracy. So anytime the accuracy changes, we want that to be reflected in the UI

ErikSin: This is also causing the "searching" text to appear when the user has not moved 15 meters due to the `STALE_TIMEOUT`

ErikSin (changes requested):
If you switch between the map screen and the camera screen, the pill flickers to `searching` for a brief moment.

src/frontend/sharedComponents/GpsPill.tsx
achou11: addressed via 2869abd8f2f8b7b412c04a547dc13ec475d73d99

src/frontend/lib/utils.ts
achou11: addressed via https://github.com/digidem/CoMapeo-mobile/pull/212/commits/12a910a06877512719c44ecc0251d9ab40d38f4f

achou11:
> If you switch between the map screen and the camera screen, the pill flickers to searching for a brief moment.

this was a good catch! narrowed down the issue to being my usage of the `useLocation()` hook and how that doesn't play nicely with react navigation's behavior of re-rendering the tab header component on navigation. solution was to move the hook outside the the `HomeHeader` and passing the info as props. see https://github.com/digidem/CoMapeo-mobile/pull/212/commits/7055bf21f51759318ef9c1373e047545d3f8b94e for more context.

src/frontend/screens/GpsModal.tsx
```diff
@@ -0,0 +1,183 @@
...
+          <>
+            <Text style={styles.sectionTitle}>
+              <FormattedMessage {...m.locationSensors} />
+            </Text>
+            {Object.entries(provider).map(([key, value]) => (
+              <GpsModalRow
+                key={key}
+                label={key}
+                value={value != null ? t(m.yes) : t(m.no)}
+              />
```
ErikSin, line 130: Some of these values are showing up as "Yes" when they shouldn't be.

I have my Location Services Turned off turned off in this picture:

image

ErikSin (approved):
Just the one comment about the "yes" showing up when it shouldn't. Otherwise feel free to merge once that fixed

src/frontend/screens/GpsModal.tsx
achou11: i remember running into this - forgot to document as a known issue. will note that i get this same behavior on Mapeo Mobile, so my guess is that there's something about the location hooks that providing stale data?

achou11: > forgot to document as a known issue.

ah technically the last screenshot in my description shows this but i never highlighted it as an issue 😄 but yeah, given that the behavior is consistent with Mapeo Mobile (despite being wrong), I'd be okay with filing an issue and fixing as a follow-up (unless the fix is super easy)

ErikSin: yeah! lets merge this then. Can you create an issue for this?

achou11: yeah best guess is that there's no API from expo-location that will react to location services being turned off (usually done via settings outside the app)

in [`useLocationProviderStatus()`], the provider status is only refetched if the permissions change, which isn't the case in this scenario

achou11: ah wait we have polling set up though, so it should be refetching anyways 🤔 going to dive deeper right now then...

achou11: ah turns out this is just a bug from the ported code 😅

```suggestion
            {Object.entries(provider).map(([key, value]) => (
              <GpsModalRow
                key={key}
                label={key}
                value={value ? t(m.yes) : t(m.no)}
              />
```

achou11: fixed via https://github.com/digidem/comapeo-mobile/pull/212/commits/5ab28c984d005b10091b6c69227d6f52ac0dc832

PR #215, opened 2024-04-02
chore: "fix" lint task

EvanHahn:
`npm run lint` had two problems:

1. An error: "prettier.resolveConfig.sync is not a function".
2. Linting too many files

This fixes that by installing the latest version of `eslint-plugin-prettier` as a direct dependency and by using our existing `.gitignore` to exclude files.

There are 92 errors and 353 warnings to fix, but now we can fix them. (Most of them should be easy to address.)


package.json
```diff
@@ -90,7 +90,7 @@
     "@mapeo/schema": "3.0.0-next.13",
     "@react-native-community/cli": "^12.3.6",
     "@react-native/babel-preset": "^0.73.21",
-    "@react-native/eslint-config": "^0.73.2",
+    "@react-native/eslint-config": "^0.75.0-main",
```
EvanHahn, line 93: Not strictly related, but figured I'd update.

achou11: probably harmless but always a little bit hesitant to update deps that are part of React Native's monorepo, which tend to be tightly tied to specific React Native versions. Wondering if it would make more sense to pin this to a specific version and only update it in tandom with React Native

a template react native app, as as example:

https://github.com/react-native-community/rn-diff-purge/blob/c29cc8066d7249bf0a6be017fb117f991c6be4ba/RnDiffApp/package.json#L21

package.json
```diff
@@ -10,7 +10,7 @@
     "prestart": "npm run build:translations && npm run build:intl-polyfills",
     "start": "react-native start",
     "test": "jest",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx --ignore-path .gitignore --fix",
```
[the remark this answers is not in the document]
achou11: probably worth using the `--cache` and git ignoring the `.eslintcache` file as part of this too

achou11 (changes requested):
Personally, I'd be more keen on completely separating prettier and eslint. Couple of ways to potentially do that:

1. Patch `@react-native/eslint-config` so that it doesn't extend the eslint prettier plugin

2. Add the following to the eslint config to disable prettier rules (can remove if upstream decides to remove them)

   ```js
   rules: {
    'prettier/prettier': 0,
    },
   ```
   
Unrelated, but taking a look at this is reminding me of how much I don't like the Prettier rules I initiated this project with 😄

EvanHahn:
Closing in favor of #217.

PR #219, opened 2024-04-03
feat: implement manual gps coordinate entry screen

### LINKED ISSUE #164: Manual GPS Screen

achou11:
Closes #164 

Mostly ported over from Mapeo Mobile. Had to do a non-trivial amount of refactoring due to different setups for persisted state and differences in hooks though.

Mostly worried about any potential edge cases throughout the app that may be dependent on the `manualLocation` field of the observation. Tried my best to address the more notable areas (such as in the observation creation screen)

There are also a few small changes that are not strictly related to adding this feature but fix subtle bugs that one could encounter by using this.

---

Preview:

image

image

image

image


src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
```diff
@@ -76,6 +76,7 @@ const draftObservationSlice: StateCreator<DraftObservationSlice> = (
           metadata: {
             ...prevValue.metadata,
             position: props.position,
+            manualLocation: props.manualLocation,
```
achou11, line 79: super subtle but found out why `updateObservationPosition()` didn't update the `manualLocation` field when initially implementing this 🥲

src/frontend/hooks/persistedState/usePersistedSettings.ts
```diff
@@ -4,15 +4,22 @@ import {CoordinateFormat} from '../../sharedTypes';
...
   };
 };
 
 const settingsSlice: StateCreator<SettingsSlice> = (set, get) => ({
   coordinateFormat: 'utm',
+  manualCoordinateEntryFormat: 'utm',
   actions: {
     setCoordinateFormat: coordinateFormat => set({coordinateFormat}),
+    setManualCoordinateEntryFormat: coordinateFormat =>
+      set({manualCoordinateEntryFormat: coordinateFormat}),
```
achou11, line 22: in Mapeo Mobile, we use to store an individual value in async storage that pertained to the coordinate format used specifically for manual entry. I believe the context for this is that the user's chosen coordinate format setting for _displaying_  is not necessarily the same as for _manual input_, hence the separation

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  state: Parameters<Parameters<typeof usePersistedSettings>[0]>[0],
+) {
+  return state.manualCoordinateEntryFormat;
+}
+
+function observationValueSelector(
+  state: Parameters<Parameters<typeof usePersistedDraftObservation>[0]>[0],
+) {
+  return state.value;
+}
```
achou11, line 207: probably unnecessary but small easy-to-implement optimization when working with zustand selectors

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -43,11 +51,15 @@ export const LocationView = () => {
...
           <Text style={styles.locationText}>
-            <FormattedCoords format={coordinateFormat} lat={lat} lon={lon} />
+            <FormattedCoords
+              format={coordinateFormat}
+              lat={coordinateInfo.lat}
+              lon={coordinateInfo.lon}
+            />
           </Text>
-          {!accuracy ? null : (
+          {coordinateInfo.accuracy === undefined ? null : (
```
achou11, line 60: subtle bug fix

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -17,20 +18,27 @@ const m = defineMessages({
...
+    : {
+        lat: liveLocation?.coords?.latitude,
+        lon: liveLocation?.coords?.longitude,
+        accuracy: liveLocation?.coords?.accuracy,
+      };
 
   return (
     <View style={styles.locationContainer}>
-      {!lat || !lon ? (
+      {coordinateInfo.lat === undefined || coordinateInfo.lon === undefined ? (
```
achou11, line 41: subtle bug fix

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -184,11 +184,12 @@ export const SaveButton = ({
       return;
     }
 
-    const hasLocation = value.lat && value.lon;
+    const hasLocation = value.lat !== undefined && value.lon !== undefined;
```
achou11, line 187: subtle bug fix

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -184,11 +184,12 @@ export const SaveButton = ({
...
     const locationSetManually = value.metadata.manualLocation;
     if (
-      locationSetManually ||
-      (hasLocation && isGpsAccurate(value.metadata.position?.coords?.accuracy))
+      hasLocation &&
+      (locationSetManually ||
+        isGpsAccurate(value.metadata.position?.coords?.accuracy))
```
achou11, line 192: potentially subtle bug fix? noticed that it differs from Mapeo Mobile implementation:

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L85

src/frontend/screens/ObservationEdit/useMostAccurateLocationForObservation.ts
```diff
@@ -20,13 +20,21 @@ export function useMostAccurateLocationForObservation() {
...
+    value?.metadata.position &&
+    !isLocationManuallySet
+  ) {
     updateObservationPosition({position: undefined, manualLocation: false});
   }
 
   useFocusEffect(
     useCallback(() => {
-      if (!permissions || !permissions.granted) return;
+      if (!permissions || !permissions.granted || isLocationManuallySet) return;
```
ErikSin, line 37: We also should check if the `watchPositionAsync` is running, and if it is turn if off. In most cases, the subscription is turned on, then location is manually set.

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
ErikSin, line 1: it is possible to save when there is no `East` or `North`. This is returning a latLon of `NaN`.

ErikSin: for posterity i saved zone number as `8` and zone letter as 'S'

ErikSin: When trying to save, this throws an error

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  React.useLayoutEffect(() => {
+    function handleSavePress() {
+      if (convertedData.error) {
+        return ToastAndroid.showWithGravity(
+          convertedData.error.message,
+          ToastAndroid.LONG,
+          ToastAndroid.TOP,
+        );
+      }
```
ErikSin, line 98: I think we should also check if `lat` and `lon` are valid here. I know that we attempted to do that in each component BUT having one last check will avoid the issue that I found in UTM form

ErikSin (changes requested):
I know you talked about this before, but the heavy use of `useEffect` to track the global state is not ideal. I know that you more or less just copied this over for time sake, but perhaps we should create an issue to refactor this in the future.

Otherwise lets write the unit tests before we merge this.

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+      <KeyboardAvoidingView>
+        <ScrollView contentContainerStyle={styles.scrollContentContainer}>
+          <View style={styles.formatSelect}>
+            <Text style={styles.inputLabel}>
+              <FormattedMessage {...m.coordinateFormat} />
+            </Text>
+            <Select
+              containerStyles={styles.selectContainer}
+              onChange={value => {
+                if (typeof value === 'number' || !isCoordinateFormat(value)) {
```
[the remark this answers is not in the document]
achou11: hm good question - doesn't seem like it's needed

src/frontend/screens/ManualGpsScreen/DdForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+  ];
+
+  const [latitudeDegrees, setLatitudeDegrees] = React.useState<string>('');
+  const [longitudeDegrees, setLongitudeDegrees] = React.useState<string>('');
+
+  const [selectedLatCardinality, setSelectedLatCardinality] = React.useState(
+    getInitialCardinality('lat', initialCoordinates),
+  );
+  const [selectedLonCardinality, setSelectedLonCardinality] = React.useState(
+    getInitialCardinality('lon', initialCoordinates),
```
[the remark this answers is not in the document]
achou11: oh wow TIL - in this case, i don't need the currying because the intention is to return an initial value (not a function), but will update

src/frontend/screens/ObservationEdit/useMostAccurateLocationForObservation.ts
achou11: not quite following - isn't this handled in the cleanup of this effect?

```ts
      return () => {
        ignore = true;
        locationSubscriptionProm.then(sub => sub.remove());
      }
```

ErikSin: Oh right, i always forget that the cleanup effect gets called on every change not just on dismount.

achou11: yeah in this case i think the behavior is slightly different since it's `useFocusEffect` but i think it's okay to let it handle the subscription removal - can follow up if there turns out to be a major issue

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
achou11: thanks for catching. was hoping to avoid adding fixes for pre-existing bugs to this PR but was able to fix this pretty easily, so will include the changes for that in this PR

src/frontend/screens/ManualGpsScreen/index.tsx
achou11: Addressed via e5057536eaf1ad4944d87886a98d303c7b2ea4ef

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
achou11: addressed via e79fb893085d174b80ebe60849dd7b26f3e35b56

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  );
+
+  const entryCoordinateFormat = usePersistedSettings(
+    entryCoordinateFormatSelector,
+  );
+
+  const {setManualCoordinateEntryFormat} = usePersistedSettingsAction();
+  const {updateObservationPosition} = useDraftObservation();
+
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
achou11: addressed via 5b57199031b22b5dc8a6611d48d6ed0b041185ae

src/frontend/screens/ManualGpsScreen/index.tsx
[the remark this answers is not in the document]
achou11: addressed via 556dabcde6f3a083723a7ad216d48d8a0a322bed

src/frontend/screens/ManualGpsScreen/DdForm.tsx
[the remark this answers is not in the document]
achou11: addressed via 298c24b0d16e07fa576cab0feeab9baa859735c6

achou11 (commented):
Addressed feedback and [added some unit tests](1f3e48ed77f572a0f0b1fc8dfd083e7319e6fdaf), which mostly test the various functions that are used to determine if the inputs in the various forms are valid or not. Not a full solution to testing this but think it's fine for now, especially if this screen sees a major refactor

PR #220, opened 2024-04-04
chore: improve wifi state management

EvanHahn:
We need the device's wifi state at a few points in the app. The way we previously did this had a few disadvantages:

1. If you were connected to wifi and another network (like cellular), we might get the wrong network's IP address. You might also be unable to fetch the SSID.

2. It used two different dependencies, because neither does exactly what we want.

3. These dependencies rely on deprecated APIs.

This change fixes those by implementing a native module, `WifiModule`, to address these problems. It also adds an React hook for using it.

Here's what it looks like in the app:

<p align="center">Screenshot</p>

I think this is a useful change on its own, but will also help us as we try to [improve local peer discovery][0] in upcoming work.

[0]: https://github.com/digidem/mapeo-core-next/issues/474


android/app/src/main/java/com/comapeo/MainApplication.kt
```diff
@@ -23,10 +23,7 @@ class MainApplication : Application(), ReactApplication {
             object : DefaultReactNativeHost(this) {
                 override fun getPackages(): List<ReactPackage> =
                     PackageList(this).packages.apply {
-                        // Packages that cannot be autolinked yet can be added manually
-                        // here,
-                        // for example:
-                        // add(MyReactNativePackage())
+                        add(WifiPackage())
```
EvanHahn, line 26: See the relevant [React Native docs](https://reactnative.dev/docs/next/native-modules-android?android-language=kotlin#register-the-module-android-specific).

android/app/src/main/java/com/comapeo/WifiPackage.kt
```diff
@@ -0,0 +1,16 @@
+package com.comapeo
```
EvanHahn, line 1: See the [relevant React Native docs](https://reactnative.dev/docs/next/native-modules-android?android-language=kotlin#register-the-module-android-specific).

android/app/src/main/java/com/comapeo/WifiModule.kt
```diff
@@ -0,0 +1,183 @@
...
+        numberOfListeners++
+    }
+
+    @ReactMethod
+    fun removeListeners(count: Int) {
+        numberOfListeners -= count
+        if (numberOfListeners == 0) {
+            stopListening()
+        }
+    }
```
EvanHahn, line 94: See the [relevant React Native docs](https://reactnative.dev/docs/next/native-modules-android?android-language=kotlin#sending-events-to-javascript) for sending events from a native module.

EvanHahn:
You're right. I wrongly assumed that `NetInfo.fetch()` wouldn't return wifi info if you were also connected to cellular, but that's what the "wifi" argument is for.

Closing this PR.

Issue #118, opened 2023-12-02
Sync Screen/No Wifi Screen


ErikSin:
Create Sync Screen that toggles into the no wifi screen, if no wifi is detected.

## To Do:
- [ ] [update useWifiStatus, see closed pull request for simpler implementation](https://github.com/digidem/CoMapeo-mobile/blob/dd60e2eab5d972f2d1b5b0772799e1a75b300ead/src/frontend/hooks/useWifiStatus.ts)
- [ ] Create NoWifi Screen (Ported over from [Mapeo Mobile](https://github.com/digidem/mapeo-mobile/blob/56528dcccf291d4248beb6348767ca50a6ce92be/src/frontend/screens/SyncModal/SyncView.js#L51))
- [ ] Create [Sync Screen](https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1068-4024&mode=design&t=bkFW7l1mRiavQbED-4) **ui only is ok for now if useSync hook is not ready
- [ ] If there is no wifi, user sees no wifi screen 
- [ ] If there is only 1 project show the [CreateOrJoinProject Screen](https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1075-5377&mode=design&t=bkFW7l1mRiavQbED-4)
- [ ] Create transitions between the screens using [React Native Reanimated](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/entering-exiting-animations) [optional]

## Technical Notes
- The three screen, noWifi, not on a project, and sync, should all be in one component in the React Navigation Stack.
- Because they are one component, with three different pages, we want to use entering/exiting animation between the screens so it looks like the user is navigating between pages
- Since there is no default project key, and a user cannot delete a project at this time, we know the user is on a default project if there is only one project available. As soon as they create or join a project, there will be at least 2 

### Other technical requirements as noted by @gmaclennan and @achou11 
- Once user has started sync, it continues running as long as the app is in the foreground, even if the user navigates to a different screen.
- Once sync is complete, after a period of time (maybe about a minute?) stop syncing (e.g. sync.stop()), this ensures that any new devices that come online do not start sync.
- When the app moves into the background, continue syncing until the OS (android) shuts us down, but as soon as sync completes, stop sync and disconnect all peers.
[...]

PR #241, opened 2024-04-09
feat: add Sync screen

achou11:
#118 

Intentional omissions/departures:

- no line of text displaying connection to cloud server
- no mention of map sync in the sync progress title
- no animation between different screen states
- no wifi screen being primarily white instead of primarily blue

Important changes:

- refactored the `useSyncState` hook because I was noticing stale closure issues with the implementation. using a class-based approach seemed easier to work with and easier to get working.

TODOs that may block merging:

- [ ] use correct color for project icon in ProjectSyncDisplay
- [ ] figure out what settings button in top right of ProjectSyncDisplay is supposed to do
- [ ] decide if we want to suspend data sync when app is not focused
- [ ] figure out how to calculate progress based on sync state
- [ ] review changes to `useSyncState` hook

---

Pretty difficult to test the sync states right now, so previews are working with mocked situations:

- No project joined/created:

image

- Nothing to sync / sync done state:

image

- Something to sync state:

image

- Syncing state:

image


achou11:
There are some TODOs that I scattered in the PR in the earlier days of working on this. I haven't looked at them in a while but will revisit tomorrow to see if they're still relevant

ErikSin:
Can you rebase/merge from main? It will be easier to test once I can invite someone to a project

achou11:
> Can you rebase/merge from main? It will be easier to test once I can invite someone to a project

done!

ErikSin:
Were you able to make 2 devices sync? I am testing with 2 devices that are both on the same project and network and they are not discovering each other. My phone has this problem with sync in mapeo-mobile, so it could be my device.

achou11:
> Were you able to make 2 devices sync? I am testing with 2 devices that are both on the same project and network and they are not discovering each other. My phone has this problem with sync in mapeo-mobile, so it could be my device.

honestly haven't tried actually syncing yet (needed the invite stuff merged before I could test it). expecting there to be some issues that I need to resolve, so consider this PR "ready to review but probably needs lots of work" 😅

as for your issue, are you talking about discovery when inviting or during sync?

ErikSin:
> as for your issue, are you talking about discovery when inviting or during sync?

I was able to discover the other device and invite them to a project. But on the sync screen, both devices are saying "0 devices nearby" even when they are on the same wifi network

achou11:
converting to a draft until I revisit this and test it out given the recent updates, which are relevant for this

ErikSin:
I did a quick code review (not a functionality review) and no glaring problems. I didn't go to deep into it though because it seems like there will some code changes when trying to make it work

src/frontend/hooks/useSyncState.ts
```diff
@@ -29,68 +38,132 @@ function identity(state: SyncState | undefined) {
...
+      listener();
+    }
   }
 
-  function subscribeInternal() {
-    project.$sync.on('sync-state', onSyncState);
-    isSubscribedInternal = true;
-    project.$sync
+  #onSyncState = (state: SyncState) => {
+    console.log({from: 'listener', ...state});
```
EvanHahn, line 125: Nitpick: should we remove this?

```suggestion
```

src/frontend/hooks/server/projects.ts
EvanHahn, line 1: Not sure why this changed as part of this work, but it's fine.

src/frontend/hooks/useSyncState.ts
```diff
@@ -29,68 +38,132 @@ function identity(state: SyncState | undefined) {
...
+    this.#isSubscribedInternal = true;
+    this.#project.$sync
       .getState()
-      .then(onSyncState)
+      .then(this.#onSyncState)
       .catch(e => {
-        error = e;
-        listeners.forEach(listener => listener());
+        this.#error = e;
+        this.#notifyListeners();
```
EvanHahn, line 153: Should we do anything special if we've unsubscribed before this promise resolves? I imagine the following scenario:

1. `#startSubscription` is called, kicking off `this.#project.$sync.getState()`.
2. `#stopSubscription` is called.
3. The `getState()` promise resolves.

EvanHahn (approved):
LGTM with a small question and a few nitpicks.

src/frontend/sharedComponents/ScreenContentWithDock.tsx
achou11, line 1: i'm hoping this is a useful shared component for us moving forward. basically creates a layout that includes a fixed position "dock" at the bottom and a scrollable content area above it. this is to account for allowing dynamic content area size without needing to scroll down to the bottom of the view to see the content in the dock. the overflow from the content area gets clipped by the fixed dock area (usually this dock area contains big, actionable buttons)

image

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+            // Should not throw but in case it does, no-op
+            console.warn(err);
+          });
+        }}
+      />
+    );
+  }
+
+  // TODO: Replace with proper check of being a part of a shared project
+  if (data && data.length === 1) {
```
achou11, line 61: highlighting this, even though this needs to be addressed later due to core limitations

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+          }}>
+          <SettingsIcon />
+        </IconButton>
+      ),
+    };
+  };
+}
+
+export const SyncScreen = ({navigation}: NativeRootNavigationProps<'Sync'>) => {
+  // TODO: Is this the right field to use?
```
achou11, line 38: can probably remove this?

src/frontend/screens/Sync/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const wifiStatus = useLocalDiscoveryState(state => state.wifiStatus);
+
+  // TODO: Handle error case
```
achou11, line 41: is this something we can address in this PR, or should be leave it for later? not sure what the error state would look like (would it use the bottom sheet modal pattern?)

src/frontend/hooks/useSyncState.ts
achou11: i _think_ it's okay because resolving will eventually call the subscribed listeners, but if `#stopSubscription` is called it's because there are no more listeners, so it should result in a no-op.

that's my initial thinking, anyways

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  const {connectedPeers, data, initial} = syncState;
+  const isSyncDone = !initial.dataToSync && !data.dataToSync;
+
+  React.useEffect(() => {
+    if (isSyncDone) {
+      project.$sync.stop();
+      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY]});
+    }
+  }, [isSyncDone, project, queryClient]);
```
achou11, line 95: discussed elsewhere but this check should be done when leaving the screen (e.g. `navigation.addListener('beforeRemove', ...)`

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  // stops sync when user leaves sync screen. The api allows us to continue syncing even if the user is not on the sync screen, but for simplicity we are only allowing sync while on the sync screen. In the future we can easily enable background sync, there are just some UI questions that need to answered before we do that.
+  useFocusEffect(
+    React.useCallback(() => {
+      return () => {
+        project.$sync.stop();
+        queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY]});
+      };
+    }, [project, queryClient]),
+  );
```
achou11, line 99: this seems to be advised against by React Navigation based on what's described here: https://reactnavigation.org/docs/use-focus-effect/#when-to-use-focus-and-blur-events-instead

seems like the preferred approach is using what's described there (i.e. use `blur`), although i think in this specific case, using [`beforeRemove`](https://reactnavigation.org/docs/navigation-events#beforeremove) makes more sense since we want this to happen where you're definitely leaving the screen (iirc, `blur` can be triggered in cases where you don't actually leave the screen)

PR #244, opened 2024-04-10
chore: remove unused wifi status hook and dependency

EvanHahn:
_This change should have no user impact and is not urgent._

This removes the `useWifiStatus` hook which we never used. It was also the only reason we depended on `react-native-network-info`, which can now be removed.

(I spoke with Gregor about removing the dependency and we don't plan to use it elsewhere, if that means anything.)


achou11 (approved):
Awesome, thanks for spotting!

PR #249, opened 2024-04-11
chore: clean up `convertToUTM`

EvanHahn:
This change:

- Adds a basic test for this function

- Removes an unnecessary `leftPad` function (which caused an ESLint error)

### DIFF
--- a/src/frontend/lib/utils.test.ts
+++ b/src/frontend/lib/utils.test.ts
@@ -1,0 +1,7 @@
+import {convertToUTM} from './utils';
+
+describe('convertToUTM', () => {
+  it('converts latitude + longitude to UTM format', () => {
+    expect(convertToUTM({lat: 12, lon: -34})).toBe('UTM 25P 391136 1326751');
+  });
+});
--- a/src/frontend/lib/utils.ts
+++ b/src/frontend/lib/utils.ts
@@ -151,10 +151,8 @@ function convertToDMS({lat, lon}: {lat: number; lon: number}) {
 
 export function convertToUTM({lat, lon}: {lat: number; lon: number}) {
   try {
-    let {easting, northing, zoneNum, zoneLetter} = fromLatLon(lat, lon);
-    easting = +leftPad(easting.toFixed(), 6, '0');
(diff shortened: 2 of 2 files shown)


achou11 (commented):
Could be missing something, but removing the left pad usage seems like a breaking change. not familiar enough with the UTM format (or the underlying library we use to calculate it), but wouldn't be surprised if the return values of the rounded `northing` and `easting` can be less than 6 digits

Guess the question is how meaningful is the change in behavior in reality? Not entirely sure myself, but wanted to highlight just in case

EvanHahn:
I was confused by this while I was working. I _think_ it's fine...I would say I'm 80% sure.

Here's a summarized version of the diff, as far as breaking changes are concerned:

```diff
-return `UTM ${+leftPad(easting.toFixed(), 6, '0')}`
+return `UTM ${easting.toFixed()}`
```

The old code:

1. Calls `easting.toFixed()`.

   This is similar to `Math.round().toString()` but rounds towards zero according to [the spec](https://262.ecma-international.org/5.1/#sec-15.7.4.5). For example, `Math.round(-1.5) === -1` but `(-1.5).toFixed() === "-2"`. (I had to look this up.)

2. Adds zeroes to the front until the length is 6 or more.

   For example, `"123"` becomes `"000123"`. `"-456"` becomes `"00-456"`.

3. Uses a [unary plus](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus) to convert this to a number.

   `"000123"` becomes `123`. `"00-456"` becomes `NaN`.[^1]

4. Puts the value in the template string.

As far as I can tell, steps 2 and 3 are unnecessary, and this PR removes them.

Does that look right to you?

[^1]: I don't think `easing` or `northing` can ever be negative so I don't think this was ever a bug in practice.

achou11:
> Uses a [unary plus](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus) to convert this to a number.

This part I overlooked 🙃 in that case, my note about potential breaking change doesn't apply.

achou11 (approved):

PR #252, opened 2024-04-11
chore: fix 40 trivial "unused variables" lint violations

EvanHahn:
This fixes several violations of [@typescript-eslint/no-unused-vars][0]. There are still several that I didn't feel comfortable changing, but these ones were easy to fix.

[0]: https://typescript-eslint.io/rules/no-unused-vars/

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,12 @@ module.exports = {
       files: ['*.ts', '*.tsx'],
       rules: {
         '@typescript-eslint/no-shadow': ['error'],
+        '@typescript-eslint/no-unused-vars': [
+          'error',
+          {
+            ignoreRestSiblings: true,
+          },
+        ],
         'no-shadow': 'off',
         'no-undef': 'off',
       },
--- a/src/frontend/hooks/useSyncState.ts
+++ b/src/frontend/hooks/useSyncState.ts
@@ -56,8 +56,8 @@ function createSyncState(project: MapeoProjectApi) {
   const listeners = new Set<() => void>();
   let error: Error | undefined;
 
-  function onSyncState(state: SyncState) {
-    state = state;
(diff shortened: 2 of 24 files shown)


.eslintrc.js
```diff
@@ -8,6 +8,12 @@ module.exports = {
       files: ['*.ts', '*.tsx'],
       rules: {
         '@typescript-eslint/no-shadow': ['error'],
+        '@typescript-eslint/no-unused-vars': [
+          'error',
+          {
+            ignoreRestSiblings: true,
```
EvanHahn, line 14: This means that this is no longer a violation:

```javascript
// `foo` is never used; we only care about `rest`
const { foo, ...rest } = obj
return rest
```

src/frontend/hooks/useSyncState.ts
```diff
@@ -56,8 +56,8 @@ function createSyncState(project: MapeoProjectApi) {
   const listeners = new Set<() => void>();
   let error: Error | undefined;
 
-  function onSyncState(state: SyncState) {
-    state = state;
+  function onSyncState(newState: SyncState) {
+    state = newState;
```
achou11, line 60: this is more related to this rule, right? https://eslint.org/docs/latest/rules/no-shadow

not a big deal (we should eventually look into disabling that rule too)

achou11 (approved):
Thanks for all of these fixes! There are cases where I like to prefix the unused variable with an underscore (seems like ESLint wouldn't complain about it), but not adamant about trying to figure out the convention right now

src/frontend/hooks/useSyncState.ts
EvanHahn: If I recall, it was violating both issues. (Not 100% sure, let me know if you want me to go back and check.)

PR #253, opened 2024-04-11
chore: disable "no inline styles" ESLint rule

EvanHahn:
To quote [the rule's docs][0]:

> It's (subjectively) good practice to separate styles from the view layout, when possible

Though we might want to do this someday, now is not the time. Let's disable this rule to avoid hundreds of warnings.

[0]: https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md


achou11 (approved):
yeah this is actually one of the more helpful rules but happy to turn off for now and revisit later 😄

PR #259, opened 2024-04-18
chore: update Mapeo deps

EvanHahn:
This updates `@mapeo/core`, `@mapeo/schema`, and `@mapeo/ipc`.

It requires a workaround for <https://github.com/rollup/plugins/issues/1709>, which is the most significant change.


package.json
```diff
@@ -29,7 +29,7 @@
     "@formatjs/intl-pluralrules": "^5.2.4",
     "@formatjs/intl-relativetimeformat": "^11.2.4",
     "@gorhom/bottom-sheet": "^4.5.1",
-    "@mapeo/ipc": "^0.2.0",
+    "@mapeo/ipc": "^0.3.0",
```
achou11, line 32: maybe pin this to exact version?

```suggestion
    "@mapeo/ipc": "0.3.0",
```

package.json
achou11, line 1: more tangential, but do you know if there's a way to reference the deps used in the backend package.json here? ideally, the backend one is the source of truth for these mapeo deps and we avoid doing additional work to make sure the top-level is in sync

achou11 (approved):

package.json
EvanHahn: Done in #283.

PR #261, opened 2024-04-18
chore: add function to generate some of the metrics report

### LINKED ISSUE #269: Add basic metrics report JSON object
Create a function called `generateMetricsReport` that returns a JSON-serializable object with a report identifier of `type: 'metrics-v1'`.

It does not need to include any other properties.

EvanHahn:
_I recommend reviewing this one commit at a time._

This introduces a (dormant) function for generating metrics reports. It returns a result like this:

```json
{
  "type": "metrics-v1",
  "appVersion": "0.0.1",
  "os": "android",
  "osVersion": 10,
  "screen": { "width": 1024, "height": 768 },
  "countries": ["CAN", "USA", "MEX"]
}
```

This isn't all the metrics we want to collect but I think it's a useful starting point.

Closes #269 and #270.


src/frontend/metrics/positionToCountries.ts
```diff
@@ -0,0 +1,29 @@
+import borders from '@osm_borders/maritime_10000m';
```
EvanHahn, line 1: This data is accurate within 10 kilometers. There are [other choices with higher resolution][0] but (1) this is small and therefore the best for performance[^1] (2) high accuracy might violate user privacy.

[0]: https://github.com/itisem/osm_borders?tab=readme-ov-file#maritime-borders

[^1]: I assume

src/frontend/metrics/positionToCountries.ts
```diff
@@ -0,0 +1,29 @@
...
+  latitude: number,
+  longitude: number,
+): Set<string> {
+  lookup ??= new GeojsonGeometriesLookup(borders);
+
+  const result = new Set<string>();
+
+  const {features} = lookup.getContainers({
+    type: 'Point',
+    coordinates: [longitude, latitude],
```
EvanHahn, line 16: I thought this was backwards, but I believe it's right (and this function is unit tested).

src/frontend/metrics/generateMetricsReport.ts
```diff
@@ -0,0 +1,38 @@
+import type {ReadonlyDeep} from 'type-fest';
+import type {Observation} from '@mapeo/schema';
+import positionToCountries from './positionToCountries';
+
+export default function generateMetricsReport({
```
EvanHahn, line 5: This function takes a bunch of dependencies. It's more verbose, but I think it has some advantages:

- Easier to port to Electron in the future.
- Clearer to the caller what goes in; data isn't pulled from anywhere unexpected.
- Easier to test.

src/frontend/metrics/generateMetricsReport.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  it('includes a report type', () => {
+    const report = generateMetricsReport(defaultOptions);
+    expect(report.type).toBe('metrics-v1');
+  });
+
+  it('includes the app version', () => {
+    const report = generateMetricsReport(defaultOptions);
+    expect(report.appVersion).toBe('1.2.3');
+  });
```
ErikSin, line 38: Would this test be more meaningful if we actually pulled the version from the package json?

src/frontend/metrics/generateMetricsReport.ts
```diff
@@ -0,0 +1,38 @@
...
+  packageJson,
+  os,
+  osVersion,
+  screen,
+  observations,
+}: ReadonlyDeep<{
+  packageJson: {version: string};
+  os: string;
```
ErikSin, line 13: would it be useful to limit the typeing of os? eg
``` ts
type Os = 'android' | 'ios' | 'win32' //... etc
```

ErikSin (approved):
1 small question but otherwise looks good!

src/frontend/metrics/generateMetricsReport.test.ts
EvanHahn: Done in 9b3b7c09f656aec124cbac6844ea6a2992e60358.

src/frontend/metrics/generateMetricsReport.ts
EvanHahn: Done in c6be4982e9f41ee2327876ac24de7948d5cd7e53.

PR #265, opened 2024-04-21
chore: update @babel/traverse

EvanHahn:
`npm audit` showed a security issue on `@babel/traverse`. I don't think this issue affects us but I ran `npm audit fix @babel/traverse` to fix it in any case.


ErikSin (approved):
Tested on a device and doesn't seem to effect functionality.  Thanks @EvanHahn

PR #266, opened 2024-04-21
chore: fix "react-in-jsx-scope" ESLint errors

EvanHahn:
This fixes all 50 violations of the [`react/react-in-jsx-scope`][0] lint rule.

In the docs for that rule:

> If you are using the [new JSX transform from React 17][1], you should disable this rule by extending `react/jsx-runtime` in your eslint config (add `"plugin:react/jsx-runtime"` to `"extends"`).

We are, so we can do this! It also lets us remove many `React` imports.

[0]: https://github.com/jsx-eslint/eslint-plugin-react/blob/4467db503e38b9356517cf6926d11be544ccf4b1/docs/rules/react-in-jsx-scope.md
[1]: https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   root: true,
-  extends: ['@react-native', 'prettier'],
+  extends: ['@react-native', 'plugin:react/jsx-runtime', 'prettier'],
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint'],
   overrides: [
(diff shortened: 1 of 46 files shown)


ErikSin:
I am not going to merge this right now. There are alot of active branches and this touches over 50 files, so it is going to cause alot of git problems. I think we should do this eventually, but maybe after MVP when some of us have switched to working on desktop (and there is not 6 people with multiple different branches working on this repo)

PR #267, opened 2024-04-22
docs: `npm start` is required to develop

EvanHahn:
We may want to address this in the future. For now, we should document that the `npm start` step is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,13 @@ The next version of Mapeo Mobile
         npm install
         ```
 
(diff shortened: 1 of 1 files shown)


achou11:
alternative (or in-addition-to) thing to document is making sure `npm run build:translations` and `npm run build:intl-polyfills` are run before running `npm run android`. these are run in `prestart` so docs for `npm start` doesn't document it

achou11 (approved):

PR #271, opened 2024-04-22
feat: Tracks saving

### LINKED ISSUE #239: Saving or discarding tracks
## Saving or discarding tracks
- User clicks on `Stop Tracks` button
- User is invited to save track and add details as a New Observation
- User can edit and save track or discard it
- FYI - On a CoMapeo project, an observation refers to a point on the map collected by CoMapeo Mobile. It is based on a geographic location (it has [coordinates](https://docs.mapeo.app/complete-reference-guide/glossary-and-quick-reference-sheets/glossary-of-mapeo-related-terms#gps-georeference-coordinate-formats-location)) and can have associated photos, notes, and other details.

## Saving tracks
Screen Shot 2024-04-08 at 3 02 12 PM

- [x] Display Tracks as an Observation
- [x] Press checkmark icon in the top right to Save Tracks

## Discarding tracks
If the user taps to Exit we show them our warning modal. This action is destructive so we want to double confirm with the user  that this is indeed the action they want to take.

Screen Shot 2024-04-08 at 3 13 16 PM

- [x] Display warning modal
- [x] Display title in bold *Discard Track?* and warning *Your Track will not be saved. This cannot be undone.*
- [x] Create `Discard Tracks` button
- [x] If `Discard Tracks` button clicked, redirect to the map
- [x] Create `Continue editing` button
- [x] If `Continue editing` button clicked, bring back to the Tracks as a New Observation 

## Design
https://www.figma.com/file/1Knqv1NrijUuqMo1pvYyF7/Tracks-and-Audio-(Share)?type=design&node-id=72-1761&mode=design&t=uliZJlzlAUbMNi91-0

CDFN:
This PR adds hooks for backend communication and implements tracks saving.

Closes #239


src/frontend/hooks/server/track.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+  });
+}
+
+export function useTrackQuery(docId: string) {
+  const project = useProject();
+  return useSuspenseQuery({
+    queryKey: [TRACK_KEY, docId],
+    queryFn: async () => {
+      if (!project) throw new Error('Project instance does not exist');
```
achou11, line 41: check is not necessary since `useProject` ensures that project exists

src/frontend/hooks/server/track.ts
```diff
@@ -0,0 +1,59 @@
...
+import {TrackValue} from '@mapeo/schema';
+
+export const TRACK_KEY = 'tracks';
+
+export function useCreateTrack() {
+  const queryClient = useQueryClient();
+  const project = useProject();
+  return useMutation({
+    mutationFn: async (params: TrackValue) => {
+      if (!project) throw new Error('Project instance does not exist');
```
achou11, line 16: check is not necessary since `useProject` ensures that project exists

src/frontend/hooks/server/track.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+  });
+}
+
+export function useTracksQuery() {
+  const project = useProject();
+  return useSuspenseQuery({
+    queryKey: [TRACK_KEY],
+    queryFn: async () => {
+      if (!project) throw new Error('Project instance does not exist');
```
achou11, line 30: check is not necessary since `useProject` ensures that project exists

src/frontend/hooks/server/track.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+  });
+}
+
+export function useDeleteTrackMutation() {
+  const queryClient = useQueryClient();
+  const project = useProject();
+  return useMutation({
+    mutationFn: async (docId: string) => {
+      if (!project) throw new Error('Project instance does not exist');
```
achou11, line 52: check is not necessary since `useProject` ensures that project exists

src/frontend/screens/MapScreen/track/saveTrack/DiscardTrackModal.tsx
achou11, line 1: thinking this should reuse the `BottomSheetModal` component we have implemented in `sharedComponents`. It has properties and behavior that align with design decisions we've agreed on (such as dismissing only by button). the component implementation is not perfect but it works well enough and using it will make it easier to maintain for us later on.

Can elaborate on usage elsewhere if needed, but basically the exports of `sharedComponents/BottomSheetModal/index.tsx` should have what you need to implement this. General usage will look like this:

```tsx
import {useBottomSheetModal, BottomSheetModal, BottomSheetContent} from '...' // wherever sharedComponents/BottomSheetModal lives

function ParentComponent() {
  const {sheetRef, isOpen, openSheet} = useBottomSheetModal({
    openOnMount: false,
  });

  return (
    <>
      { /* Whatever other markup you need for your component */}
      <View>{...}</View>
      <BottomSheetModal isOpen={isOpen} ref={sheetRef}>
        <BottomSheetContent
          titleStyle={}
          descriptionStyle={}
          title={}
          description={}
          buttonConfigs={}
          icon={}
        />
      </BottomSheetModal>
    </>
  )
}
```

achou11 (changes requested):
Mostly a UX review but some feedback about implementation

- fix: indentation of text area should be aligned with category row displayed above

  image

- fix: behavior of discard track bottom sheet modal is not aligned with our design practices around bottom sheet modals. would suggest using the shared component we have to make it easier

src/backend/package.json
```diff
@@ -13,8 +13,8 @@
   "author": "Digital Democracy",
   "license": "MIT",
   "dependencies": {
-    "@mapeo/core": "9.0.0-alpha.7",
-    "@mapeo/ipc": "0.3.0",
+    "@mapeo/core": "^9.0.0-alpha.7",
+    "@mapeo/ipc": "^0.3.0",
```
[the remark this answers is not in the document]
bohdanprog: fixed

src/frontend/screens/MapScreen/track/saveTrack/DiscardTrackModal.tsx
bohdanprog: Hey, I have updated our solution based on your recommendations. Let me know if I need to change something else :)

src/frontend/screens/MapScreen/track/SaveTrackScreen.tsx
```diff
@@ -0,0 +1,101 @@
...
+    defaultMessage: 'Camera',
+    description: 'Button label for adding photo',
+  },
+});
+
+export const SaveTrackScreen: React.FC<{navigation: NavigationProp<any>}> = ({
+  navigation,
+}) => {
+  const {formatMessage: t} = useIntl();
+  const [description, setDescription] = useState('');
```
ErikSin, line 37: We want to persist the description in case the user closes their phone. instead of using `setDescription` use                 [updateTags('notes', newVal);](https://github.com/digidem/comapeo-mobile/blob/dd9101b1765e5d3398b0ab6f5a231a0ba7f049f9/src/frontend/screens/ObservationEdit/DescriptionField.tsx#L26) and instead of description use `[const notes = usePersistedDraftObservation(store => store.value?.tags.notes)](https://github.com/digidem/comapeo-mobile/blob/dd9101b1765e5d3398b0ab6f5a231a0ba7f049f9/src/frontend/screens/ObservationEdit/DescriptionField.tsx#L18C3-L18C80)`

ErikSin (approved):
Besides my 2 comments, and andrew's comments, looks good to me.

I have noticed some architecture changes related to the location provider that I think we need to revert, but I will create an issue for that.

ErikSin:
If the user has stopped "tracking" but is in the middle of saving it (aka they are on `SaveTrackScreen`) and they close the app, we want to make sure it opens again on the saveTrackScreen, and that the tracks are persisted in state. In observations acheive this by [saving observations in persisted state](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts).

We should follow a similar pattern with tracks. Instead of saving [locationHistory to plain zustand state](https://github.com/digidem/comapeo-mobile/pull/271/files#diff-8e3529efd063e458ea9b2c98b258af8cad74a993665ca6390dc59ad823da98bf) we should save it to persisted state.

Also, when the [app open we want to do a check](https://github.com/digidem/comapeo-mobile/blob/dd9101b1765e5d3398b0ab6f5a231a0ba7f049f9/src/frontend/Navigation/AppNavigator.tsx#L83) to see if `locationHistory` exists in persisted state, and if it does, open the SaveTrackScreen. 

^^When we do this  changes the navigation behaviour. When the hit save, they will be [navigated to the map](https://github.com/digidem/comapeo-mobile/blob/8a47fdfaf088fa744f7866aa70c67fdf054e8729/src/frontend/screens/MapScreen/track/saveTrack/SaveTrackHeader.tsx#L60). But the user will now be able to press the back button and go back to the saveTrackScreen. So we need to prevent this by resetting the nav stack on save.

src/frontend/screens/MapScreen/track/SaveTrackScreen.tsx
ErikSin, line 1: We also need to prevent the user from going back by using the back button. When they press the back button it should open the discard track modal

src/frontend/screens/MapScreen/track/saveTrack/DiscardTrackModal.tsx
ErikSin, line 1: We are going to start using this modal for the observations. So can you make the modal `title`, `description` and `discardButton Text` props, so that it can be reused by another component

bohdanprog: Certainly, I have completed that task. Please let me know if I can assist with anything else.

src/frontend/hooks/tracks/useTracking.ts
```diff
@@ -13,32 +14,34 @@ type LocationCallbackInfo = {
...
+            data.locations.map(loc => ({
+              latitude: loc.coords.latitude,
+              longitude: loc.coords.longitude,
+              timestamp: loc.timestamp,
+            })),
+          );
+        }
+      },
+    );
+  }, []);
```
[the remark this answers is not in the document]
bohdanprog: fixed

src/frontend/screens/MapScreen/track/SaveTrackScreen.tsx
bohdanprog: Yeah, good catch, thanks.

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsEnabled.tsx
```diff
@@ -30,13 +32,36 @@ const m = defineMessages({
...
+    }
+
+    cancelTracking();
+
+    if (locationHistory.length <= 1) {
+      clearCurrentTrack();
+      navigation.navigate('Map');
+    } else {
+      navigation.navigate('SaveTrack');
+    }
```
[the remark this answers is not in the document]
CDFN: Frequency is about every 3 meters. However if you stay in one place, there are no updates being sent (you can think of it like some kind of debounce). If you travel at least 3 meters you would see there is another location saved. We discussed this behavior on Slack on April 22nd.

src/frontend/screens/MapScreen/track/saveTrack/SaveTrackHeader.tsx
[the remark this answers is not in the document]
CDFN: Thanks for adding direct link, indeed it didn't show under previous comment for us. So explanation for this is how the header looks in Figma. We will make it use navigator default header if complying 100% with design isn't that important in this case.

achou11:
> If the user has stopped "tracking" but is in the middle of saving it (aka they are on `SaveTrackScreen`) and they close the app, we want to make sure it opens again on the saveTrackScreen, and that the tracks are persisted in state. In observations acheive this by [saving observations in persisted state](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts).
> 
> We should follow a similar pattern with tracks. Instead of saving [locationHistory to plain zustand state](https://github.com/digidem/comapeo-mobile/pull/271/files#diff-8e3529efd063e458ea9b2c98b258af8cad74a993665ca6390dc59ad823da98bf) we should save it to persisted state.
> 
> Also, when the [app open we want to do a check](https://github.com/digidem/comapeo-mobile/blob/dd9101b1765e5d3398b0ab6f5a231a0ba7f049f9/src/frontend/Navigation/AppNavigator.tsx#L83) to see if `locationHistory` exists in persisted state, and if it does, open the SaveTrackScreen.
> 
> ^^When we do this changes the navigation behaviour. When the hit save, they will be [navigated to the map](https://github.com/digidem/comapeo-mobile/blob/8a47fdfaf088fa744f7866aa70c67fdf054e8729/src/frontend/screens/MapScreen/track/saveTrack/SaveTrackHeader.tsx#L60). But the user will now be able to press the back button and go back to the saveTrackScreen. So we need to prevent this by resetting the nav stack on save.

@ErikSin thoughts on this being a follow-up to this PR? The scope of changes required to get the above working might be  large enough to warrant a separate PR to make it easier to review. Not sure if @CDFN @bohdanprog have already started making changes here towards but if they haven't, would put it into consideration

ErikSin:
> @ErikSin thoughts on this being a follow-up to this PR? The scope of changes required to get the above working might be large enough to warrant a separate PR to make it easier to review. Not sure if @CDFN @bohdanprog have already started making changes here towards that but if they haven't, would put it into consideration

Yeah, that sounds good. I do think we should do this sooner, rather than later as it will effect the subsequent Pr, but I think separating it makes sense.

CDFN:
I got some changes but it's fairly easy to move it to another branch. For issue, would be great to summarize it there if you could.

CDFN:
@achou11 does this mean we should merge this one and work on #306 right after? Or just merge #306 into `feat/save-track` and then merge this pr into main?

src/frontend/screens/MapScreen/track/saveTrack/SaveTrackHeader.tsx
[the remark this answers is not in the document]
achou11: yeah i think it's preferred to use the existing header for now, that way it'll be easier for us to convert everything at once without gaps of inconsistency

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsEnabled.tsx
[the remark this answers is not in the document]
achou11: yeah i'm seeing the conversation you're referring to on slack. thanks for pointing that out! seems like i missed it

achou11 (changes requested):
Can this [comment about the navigation header](https://github.com/digidem/comapeo-mobile/pull/271/files/358aa50a66c217596a66c41e3d2f6e31728df326#r1581406217) be addressed before merging?

Otherwise changes seem ready for merging :)

bohdanprog:
> Can this [comment about the navigation header](https://github.com/digidem/comapeo-mobile/pull/271/files/358aa50a66c217596a66c41e3d2f6e31728df326#r1581406217) be addressed before merging?
> 
> Otherwise changes seem ready for merging :)

Hm...but I already changed Header in SaveTrackScreen :D

achou11 (approved):

ErikSin:
Im going to merge this as I have a PR stacked on this, and I would like to create another PR based on this

PR #279, opened 2024-04-22
chore: fix type error in `ObservationMapLayer` component

EvanHahn:
_I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/comapeo-mobile/pull/279/files?w=1)._

Our `onPress` handler code was failing to import a type from `@rnmapbox/maps`.

This fixes that by moving the handler into the component directly, avoiding the need to import the type.

### DIFF
--- a/src/frontend/screens/MapScreen/ObsevationMapLayer.tsx
+++ b/src/frontend/screens/MapScreen/ObsevationMapLayer.tsx
@@ -2,7 +2,6 @@ import {Observation} from '@mapeo/schema';
 import React from 'react';
 import MapboxGL from '@rnmapbox/maps';
 import {useAllObservations} from '../../hooks/useAllObservations';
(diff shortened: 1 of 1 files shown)


src/frontend/screens/MapScreen/ObsevationMapLayer.tsx
```diff
@@ -23,17 +22,15 @@ export const ObservationMapLayer = () => {
...
-    if (!('id' in properties)) return;
-
-    navigate('Observation', {observationId: properties.id});
-  }
-
   return (
     <MapboxGL.ShapeSource
-      onPress={handlePressEvent}
+      onPress={event => {
+        const properties = event.features[0]?.properties;
```
EvanHahn, line 28: I had to add this `?`, which I don't think changes the functionality.

achou11 (approved):

PR #301, opened 2024-04-29
add library expo-splash-screen

bohdanprog:
I have changed the library from `react-native-bootsplash` to `expo-splash-screen`.
It helps us auto-generate the App icon and splash screen.


achou11:
fwiw `react-native-bootsplash` has CLI tooling to autogenerate assets too: https://github.com/zoontek/react-native-bootsplash?tab=readme-ov-file#assets-generation

I vaguely remember that after exploring expo's library for this in https://github.com/digidem/comapeo-mobile/pull/97, i found that bootsplash's approach behaved better for our use case. think it was related to it creating an activity that overlaid the main activity which allowed a more graceful hiding of it.

generally, think if using bootsplash with expo is not problematic, i would prefer to stick with it, unless there's another major benefit to switching to expo's library.

bohdanprog:
| Before | After |
|--------|--------|
| |  |
| | |

bohdanprog:
@achou11  I encountered some issues where the server does not restart when attempting to reload the app using the hotkey R.

achou11:
do you mind rebasing this PR? Github doesn't handle automatically updating stacked PRs very well so hopefully not too much of a pain to fix 😅 

> @achou11 I encountered some issues where the server does not restart when attempting to reload the app using the hotkey R.

yeah this is a pre-existing issue. something to do with react fast refresh not playing nicely with the lifecycle events that the nodejs backend uses to gracefully restart 😬 (and maybe combined with some frontend state initialization setup issues)

package-lock.json
achou11, line 1: looks like this file got malformed due to the rebase. would recommend copying the one from main branch and running install again

.watchmanconfig
achou11, line 1: is this intentional?

app.json
```diff
@@ -9,8 +9,11 @@
     "userInterfaceStyle": "light",
     "splash": {
       "image": "./assets/splash.png",
-      "resizeMode": "cover",
-      "backgroundColor": "#050F77"
+      "backgroundColor": "#050F77",
+      /*resizeMode - native work only for android
+      https://docs.expo.dev/develop/user-interface/splash-screen/#custom-configuration-for-android-and-ios
+      **/
```
achou11, line 15: guess app.json is secretly a jsonc/json5 file? 😄 

```suggestion
      // resizeMode - native work only for android: https://docs.expo.dev/develop/user-interface/splash-screen/#custom-configuration-for-android-and-ios
```

src/frontend/hooks/persistedState/usePersistedLocale.ts
achou11, line 1: revert the change here since it's out of this PR's scope

src/frontend/contexts/IntlContext.tsx
```diff
@@ -90,6 +90,6 @@ a language name and translations in \`src/frontend/languages.json\``,
   if (supportedLanguages.find(lang => lang.locale === locale))
     return locale as keyof typeof languages;
   const nonRegionalLocale = locale.split('-')[0];
-  if (supportedLanguages.find(({locale}) => locale === nonRegionalLocale))
+  if (supportedLanguages?.find(({locale}) => locale === nonRegionalLocale))
```
achou11, line 93: revert this change? doesn't seem to fix any pre-existing issue

package.json
```diff
@@ -93,7 +93,8 @@
     "uint8array-extras": "^0.5.0",
     "utm": "^1.1.1",
     "validate-color": "^2.2.4",
-    "zustand": "^4.4.6"
+    "zustand": "^4.4.6",
+    "expo-splash-screen": "~0.26.5"
```
achou11, line 97: move this so that's placed in the file based on alphabetical order (default when running `npm install ...`)

achou11 (changes requested):
Mostly cleanup requests but otherwise the switch seems fine. 

Tangential, but I remember that I chose `react-native-bootsplash` because getting it to fade was much easier than [Expo's approach to it](https://github.com/expo/examples/blob/master/with-splash-screen/App.js). Mapeo Mobile fades the splash so that was a soft requirement in the context of porting over old behavior, but I see that we're not doing that fading now anyways.

Also seems like the next major version of bootsplash is adding an expo config plugin, but not sure what the timeline for that to be released is (timing can be cruel sometimes 🥲 ):

`https://github.com/zoontek/react-native-bootsplash/pull/578`

app.json
bohdanprog: @achou11 XD what do you mean :D ?

.watchmanconfig
bohdanprog: Hm... something strange happened when I rebased that branch to main.

src/frontend/hooks/persistedState/usePersistedLocale.ts
bohdanprog: reverted.

src/frontend/contexts/IntlContext.tsx
bohdanprog: reverted

package.json
bohdanprog: To be honest, I don't know why, but when I installed that package, it was placed last in the list.

package.json
```diff
@@ -15,10 +15,11 @@
     "lint:eslint": "eslint . --ext .js,.jsx,.ts,.tsx --cache --ignore-path .gitignore",
     "lint:types": "tsc --noEmit",
     "lint": "npm run lint:prettier && npm run lint:eslint && npm run lint:types",
-    "postinstall": "patch-package",
+    "postinstall": "patch-package && npm run build:autogenerated",
     "prepare": "husky install",
     "format": "prettier \"src/**/*.{js,ts,jsx,tsx}\" --write",
     "deep-clean": "./scripts/deep-clean.sh",
+    "build:autogenerated": "npm run build:translations && npm run build:intl-polyfills && npm run build:backend",
```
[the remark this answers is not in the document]
bohdanprog: That changes was reverted :)

app.json
achou11: mostly just joking how the file extension doesn't match how it's treated - comments are not allowed in json files but if expo doesn't care, then it's treating this file as jsonc/json5, which supports comments but technically should be using a different file extension name 😄 

suggestion here is to clean up the comment formatting (basically make it a single line comment)

achou11 (approved):

package-lock.json
bohdanprog: yeah, right.

app.json
bohdanprog: oh :D it's work, that's why I put them :)

CDFN: Eventually this will be moved to app.config.js file anyway 😁 I'll open PR for it tomorrow.

PR #302, opened 2024-04-29
feat: add env variables schema and validation

CDFN:
This PR introduces schema for environment variables that is checked upon app start. It also lists all possible env vars used across the app. This will definitely improve developer experience. It wasn't listed as an issue so if you feel like this is redundant addition feel free to close this PR without merging.
image


ErikSin:
Im going to close this for now. Thanks for doing this work but I dont think this a priorty for this time. But when this work is done later, i will use this PR for reference, so the work was not done in vain.

PR #305, opened 2024-04-30
chore: rename GpsModal to LocationInfoScreen

### LINKED ISSUE #264: Rename `GpsModal` screen to something more specific and accurate
[`GpsModal`](https://github.com/digidem/comapeo-mobile/blob/92fc7f42dd42ab45b267cc0df389f447b97a7880/src/frontend/screens/GpsModal.tsx) name was ported over from Mapeo Mobile, but it's a slightly misleading name and could stand to be more specific. I suggest either `GPSInfoScreen` or `LocationInfoScreen`

achou11:
Closes #264. Should not affect the app in any way

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -268,42 +268,6 @@
   "screens.FatalError.somethingWrong": {
     "message": "Something Went Wrong"
   },
-  "screens.GpsModal.details": {
-    "description": "Section title for details about current position",
-    "message": "Details"
-  },
-  "screens.GpsModal.gpsHeader": {
-    "description": "Header for GPS screen",
-    "message": "Current GPS Location"
-  },
-  "screens.GpsModal.lastUpdate": {
-    "description": "Section title for time of last GPS update",
-    "message": "Last update"
-  },
-  "screens.GpsModal.locationDD": {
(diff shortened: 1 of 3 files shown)


ErikSin:
Do you think that we should actually implement this as a [react-navigation-modal](https://reactnavigation.org/docs/modal/)?

achou11:
> Do you think that we should actually implement this as a [react-navigation-modal](https://reactnavigation.org/docs/modal/)?

hm that's a good question - looks like it was [implemented as a normal screen in Mapeo Mobile](https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/Navigation/ScreenGroups/AppScreens.tsx#L207-L218) but could see it making sense for it to be proper modal. Still think renaming it makes sense because some of the tracks work has introduced naming that uses `GPS` in it, which was getting a little bit confusing for me

ErikSin (approved):
Could possibly make this a modal, but i think its a subtle difference, and since it was already a nav stack page in mapeo mobile, I think this is ok

PR #310, opened 2024-05-01
Es/chore replace toast with modal

ErikSin:
Replaces android native toast with a bottom sheet.

image


src/frontend/screens/PresetChooser.tsx
```diff
@@ -41,17 +44,22 @@ export const PresetChooser: NativeNavigationComponent<'PresetChooser'> = ({
...
         ) : (
-          <CustomHeaderLeftClose headerBackButtonProps={props} />
+          <CustomHeaderLeftClose
+            openDiscardModal={openSheet}
+            headerBackButtonProps={props}
+          />
         ),
     });
-  }, [prevRouteNameInStack, CustomHeaderLeft, CustomHeaderLeftClose]);
+  }, [prevRouteNameInStack, openSheet, navigation]);
```
ErikSin, line 62: eslint told me to take out `CustomHeaderLeft` and `CustomHeaderLeftClose`....?

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
```diff
@@ -61,17 +27,20 @@ interface SharedBackButtonProps {
 
 type CustomHeaderLeftCloseProps = {
   observationId?: string;
+  openDiscardModal: () => void;
```
achou11, line 30: seeing this being added is making me notice the scope creep of this component, but i think addressing that is outside the scope of this PR. can create an issue to provide more details of what a refactor could entail

src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -88,6 +118,15 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> & {
...
       <ErrorModal sheetRef={sheetRef} closeSheet={closeSheet} isOpen={isOpen} />
+      <DiscardModal
+        bottomSheetRef={sheetRef}
+        isOpen={isOpen}
+        closeSheet={closeSheet}
+        title={deleteObservationMessages.discardObservation}
+        description={deleteObservationMessages.discardObservationDescription}
+        discardButtonText={deleteObservationMessages.discardObservationButton}
+        handleDiscard={handleDiscard}
+      />
```
achou11, line 129: won't this cause issues with the error modal that's also used in this component?

achou11 (changes requested):
Think there's a good opportunity here to clean up the way that our custom bottom sheet modal is used. It's never been clearly documented, but the original design of the component when i implemented it long ago is that it the `BottomSheetModal` should (almost) *always* be co-located with the usage of the `useBottomSheetModal()` hook. Instead, i'm noticing that custom bottom sheet modal components are including the `BottomSheetModal` and unnecessarily accepting and passing props that come from the hook in the consuming component. what should be happening is that the *bottom sheet contents* can/should be componentized when appropriate, and then used as a child of the `BottomSheetModal` that's being used in the consuming component.

I added a patch that lives on a different branch to demonstrate the ideal usage i'm describing. Diff between this PR and that patch can be found [here](https://github.com/digidem/comapeo-mobile/compare/es/chore-replace-toast-with-modal...ac/bottom-sheet-content-variants). the summary of changes in that patch are:

1. Create a directory called `contentVariants` that lives in `sharedComponents/BottomSheet/`. This serves as a collection of sheet content implementations that are geared towards specific usages based on common patterns.
2. Implement `contentVariants/Destructive.tsx`, which is a "foundational" sheet content component that is only opinionated about the icons being used to match our design patterns (in this case, the error icon at the top and the delete icon within the destructive button)
3. Implement `contentVariants/ObservationDiscard.tsx`, which uses `Destructive` but is opinionated about the content (e.g. title, description, text in buttons). It accepts callback props that are more semantically meaningful but not functionally different than `Destructive` (just passed along directly). 
4. Update `ObservationEdit/index.tsx` and `PresetChooser.tsx` to use `BottomSheetModal` and `ObservationDiscard`
5. Update `SaveTrackScreen.tsx` to use `BottomSheetModal` and `Destructive` 
6. Remove `DiscardModal` in its entirety. In its current form (aka prior to this PR), it's a non-useful wrapper. Should've caught that in previous PRs

ErikSin:
Closing this to reimplement in a cleaner way.

PR #312, opened 2024-05-01
feat: update header with new UI

ErikSin:
Updated the header in creating/editting an observation to include new save button. Also replaces the native toast with a modal for confirming whether user wants to discard changes when trying to leave the edit screen.

image

image


src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -56,9 +56,9 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> & {
     navigation.navigate('AddPhoto');
   }, [navigation]);
 
-  const handleDetailsPress = React.useCallback(() => {
-    navigation.navigate('ObservationDetails', {question: 1});
-  }, [navigation]);
+  // const handleDetailsPress = React.useCallback(() => {
+  //   navigation.navigate('ObservationDetails', {question: 1});
+  // }, [navigation]);
```
ErikSin, line 61: this will be added really soon, so just commented out for linting purposes

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
```diff
@@ -68,51 +85,80 @@ export const CustomHeaderLeftClose = ({
...
+              dangerous: true,
+              onPress: handleDiscard,
+              text: observationId
+                ? formatMessage(m.discardChangesButton)
+                : formatMessage(m.discardObservationButton),
+              icon: <DiscardIcon />,
+            },
+            {
+              onPress: closeSheet,
+              text: formatMessage(m.discardCancel),
```
achou11, line 142: nit: Seeing the cancel action copy of `Continue editing` seems a little off when you're on the category chooser screen.

src/frontend/screens/Settings/ProjectSettings/DeviceName/EditScreen.tsx
```diff
@@ -147,10 +147,9 @@ export const EditScreen = ({
     [
       handleSubmit,
       navigation,
-      editDeviceInfoMutation.mutate,
-      editDeviceInfoMutation.isPending,
+      editDeviceInfoMutation,
       nameHasChanges,
-      errorModal.openSheet,
```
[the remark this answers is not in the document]
achou11: if you extract the field first and use that, it should work e.g.

```ts
const {openSheet} = useBottomSheetModal({openOnMount: false});
```

i think adding the whole object is not desirable (in theory) because the effect will change when one of its fields changes (e.g. `isOpen`). maybe it's not a big issue in practice though 🤷

achou11 (approved):
Couple of non-blocking comments but otherwise lgtm

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
ErikSin: Im going to leave this because we are planning on updating all the copy at a later time, so that can be dealt with then

PR #315, opened 2024-05-06
feat: Persist tracks between app restarts (closes #306)

### LINKED ISSUE #306: Tracks saving flow should persist
_PR comment with original context located [here](https://github.com/digidem/comapeo-mobile/pull/271#issuecomment-2084322253)_

When in the flow of saving a track, there's a chance that the app may quite unexpectedly or the user leaves it unintentionally. In order to avoid lost information, we want to have a similar pattern to what we do with observations, where we persist information to local storage (we mmkv via zustand persist) and read that on initial app load to determine if there was an "in-progress" track being saved.

Changes needed:

- [ ] Update the zustand store used for tracks to use persist the relevant tracks information (e.g. `locationHistory`). see `createPersistedState.ts`
- [ ] update `AppNavigator.tsx` to load an existing track as the initial route for the navigator if it exists and if the feature flag is enabled. See implementation in that file for observations as an example
    - [ ] in this specific case, the save behavior of the Saving Track screen should *reset* the navigation state so that the screen cannot be returned to afterwards (e.g. back button/gesture).

CDFN:
Closes #306


ErikSin (changes requested):
The `description` in `TrackDescriptionField` should also be persisted. `Description` and `setDescription` should be part of the persisted Tracks hook. This means that the` description` and `setDescription` does not need to be passed as a prop to `TrackDescriptionField` and `SaveButton` as it can be passed through the hook directly in the component

Also, I noticed that you are saving a new `TrackObservation` on [every save of an observation](https://github.com/digidem/comapeo-mobile/blob/c3f4ee2f687f85a5fb59bd943088e4f82084b6c2/src/frontend/screens/ObservationEdit/SaveButton.tsx#L150) and every [Edit of an observation](https://github.com/digidem/comapeo-mobile/blob/c3f4ee2f687f85a5fb59bd943088e4f82084b6c2/src/frontend/screens/ObservationEdit/SaveButton.tsx#L182). As well you are doing the same thing for `AddNewTrackLocation` in both the [create](https://github.com/digidem/comapeo-mobile/blob/c3f4ee2f687f85a5fb59bd943088e4f82084b6c2/src/frontend/screens/ObservationEdit/SaveButton.tsx#L141) and [edit](https://github.com/digidem/comapeo-mobile/blob/c3f4ee2f687f85a5fb59bd943088e4f82084b6c2/src/frontend/screens/ObservationEdit/SaveButton.tsx#L174).

A user can take an observation separate from tracks. So not all observation are saved or editted as tracks.

When a user is recording tracks, then any observation are included in the tracks. But when they are not recording the observation should NOT be saved as part of tracks. Does that make sense?

CDFN:
Good catch with saving observations/points when recording is off, thanks! It makes sense obviously, for some reason I just missed this while implementing the feature. Everything should be sorted now. Thanks for review! 😅

CDFN:
Sorry for requesting another review before the PR was ready, meant to do it for other PR 😅 Was going to add description persistence today 😁 Should be all sorted now.

src/frontend/screens/MapScreen/track/saveTrack/SaveTrackButton.tsx
```diff
@@ -40,7 +36,7 @@ export const SaveTrackButton: FC<SaveTrackButton> = ({description}) => {
       },
       {
         onSuccess: () => {
-          navigation.navigate(TabName.Map);
+          navigation.navigate('Home', {screen: 'Map'});
```
ErikSin, line 39: We actually need to reset the stack when we navigate to the map. 

So normally when we navigate to the 'map', the map is the first thing in the stack. React navigation handles this by going `back` to the map, instead of adding it to the stack => this mean after a user saves a track, they are navigated `back` to the map. And if they were to press the back button, then they would not be able to go back to the tracks screen

With the persisted state, the tracks screen is sometimes the first screen (because of setting the initial route name). So if that is the case, maps is not part of the stack, and when they are navigated to maps, it is added to the stack. So the user is able to press the back button and go to the tracks screen, which is undesired behaviour.

In order to avoid this behavior we need to [reset the navigation](https://reactnavigation.org/docs/navigation-actions/#reset) to make maps the only thing in the stack.

src/frontend/screens/MapScreen/track/SaveTrackScreen.tsx
```diff
@@ -18,20 +18,16 @@ import {TrackDescriptionField} from './saveTrack/TrackDescriptionField';
...
-import {useNavigationFromHomeTabs} from '../../../hooks/useNavigationWithTypes.ts';
+import {usePersistedTrack} from '../../../hooks/persistedState/usePersistedTrack';
+import {useNavigationFromHomeTabs} from '../../../hooks/useNavigationWithTypes';
 import {useFocusEffect} from '@react-navigation/native';
-import {SaveTrackButton} from './saveTrack/SaveTrackButton.tsx';
+import {SaveTrackButton} from './saveTrack/SaveTrackButton';
 import Close from '../../../images/close.svg';
 
 export const SaveTrackScreen = () => {
   const navigation = useNavigationFromHomeTabs();
```
ErikSin, line 28: This is the wrong hook. Should be using `useNavigationFromRoot()`.

src/frontend/screens/MapScreen/track/SaveTrackScreen.tsx
```diff
@@ -55,7 +51,7 @@ export const SaveTrackScreen = () => {
 
   const handleDiscard = () => {
     closeSheet();
-    navigation.navigate(TabName.Map);
+    navigation.navigate('Map');
```
ErikSin, line 54: Your navigation action needs to be a reset here as well (same comment as in `saveTrackButton`). The user needs to be prevented from going back to this screen

src/frontend/screens/ObservationEdit/SaveButton.tsx
```diff
@@ -170,16 +169,18 @@ export const SaveButton = ({
...
+            addNewTrackObservation(observationId);
+            if (value.lat && value.lon) {
+              addNewTrackLocation([
+                {
+                  timestamp: new Date().getTime(),
+                  latitude: value.lat,
+                  longitude: value.lon,
+                },
+              ]);
+            }
```
ErikSin, line 182: Sorry I didn't catch this earlier. But we actually don't want to save an observation in a track when they are editting an observation, only creating an observation. If is a user is `tracking` and goes to edit a previous observation, that observation will be unrelated to the track, so we don't want to add it to the track

ErikSin (changes requested):
The persisted tracks looks good!

There is a slight issue with navigation that I only know from experience with persisted observations. 

Also I realize that we shouldn't be saving observations to track when editting an observation, only on creating an observation.

Lastly, `SaveTrackScreen` shouldn't be in `screen/MapScreen`. It should be its own file in the screens folder. And to remain consistent, can we rename the file, component, and navigation typescript `name` to "SaveTrack".  We only use "Screen" when it is ambiguous, such as the maps where there are some maps that aren't screens.

ErikSin:
Also 1 more thing because we switched to expo in "ScreenGroups/AppScreen" the `process.env.FEATURE_TRACKS` has to be converted to `process.env.FEATURE_TRACKS === 'true'`

CDFN:
Should be sorted now, thanks for catching those! Let me know if location of SaveTrack screen looks fine now.

ErikSin (approved):
Looks good. Can you rebase/update from main. I think there are quite a lot of changes, so can you make sure to just quickly test the app on a device after your rebase before merging. 

Thanks for all the hard work on this.

CDFN:
Thanks for approval, is already merged with main containing all changes from that branch. Everything looked good to me when I tested. When we merge to main we need to make sure to squash.

ErikSin:
@CDFN sorry, you are correct so its good merge. Also i made squash merging the only type merging possible.

PR #319, opened 2024-05-06
fix: missing icon import for save button in manual GPS entry screen

achou11:
Icon was removed in https://github.com/digidem/comapeo-mobile/pull/312 but this screen wasn't updated to import the replacement one

---

Preview:

image


ErikSin (approved):
Looks good!

PR #320, opened 2024-05-06
chore: observation won't sync warning

### LINKED ISSUE #317: No Project/Sync
## No project/ Sync
-  User first opens the app, they are on an unnamed unofficial project by themselves. 
-  let them know that they won’t be able to sync their observations until they create or join a project.

Screen Shot 2023-10-04 at 2 46 54 PM

## Todo
- [ ] Display text box with copy **You won’t be able to share observations collected yet. To sync and share with other devices**
- [ ] Create `Create or Join a Project` button

## Design
https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1200-69&mode=design&t=xfOwkkUA7kBhB7vK-0

## Copydeck
https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1197-17493&mode=design&t=ojmacHaDmI1ARJ5k-0

ErikSin:
Adds a warning, warning the user that they will not be able to sync their observations if the user is not part of a project.

Closes #317 

image

### DIFF
--- a/src/frontend/screens/ObservationsList/NoProjectWarning.tsx
+++ b/src/frontend/screens/ObservationsList/NoProjectWarning.tsx
@@ -1,0 +1,69 @@
+import * as React from 'react';
+import {StyleSheet, View} from 'react-native';
+import Warning from '../../images/Warning.svg';
+import {Text} from '../../sharedComponents/Text';
+import {defineMessages, useIntl} from 'react-intl';
+import {TouchableOpacity} from 'react-native-gesture-handler';
+import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
+import {BLACK, DARK_BLUE, LIGHT_GREY} from '../../lib/styles';
+import {ViewStyleProp} from '../../sharedTypes';
+
+const m = defineMessages({
+  cantShare: {
+    id: 'screens.ObservationsList.NoProjectWarning',
+    defaultMessage: 'You won’t be able to share observations collected yet.',
+  },
+  createOrJoin: {
+    id: 'screens.ObservationsList.createOrJoin',
+    defaultMessage: 'Create or Join a Project.',
+    description:
+      "The full sentence is 'To sync and share with other devices, Create or Join a Project.' The sentence needs to be seperated for styling purposes",
+  },
+  toSync: {
+    id: 'screens.ObservationsList.toSync',
+    defaultMessage: 'To sync and share with other devices,',
+    description:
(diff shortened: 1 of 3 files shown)


src/frontend/screens/ObservationsList/NoProjectWarning.tsx
```diff
@@ -0,0 +1,69 @@
...
+          onPress={() => navigate('CreateOrJoinProject')}>
+          <Text style={{flexShrink: 1}}>
+            {formatMessage(m.toSync) + ' '}
+            <Text
+              style={{
+                textDecorationLine: 'underline',
+                color: DARK_BLUE,
+              }}>
+              {formatMessage(m.createOrJoin)}
+            </Text>
```
EvanHahn, line 49: We probably want to change the order of this for RTL languages like Arabic. Is that possible?

src/frontend/screens/ObservationsList/index.tsx
```diff
@@ -63,11 +68,19 @@ export const ObservationsList: React.FC<
 
   return (
     <View style={styles.container} testID="observationsListView">
+      {isLoading ? (
+        <Loading />
+      ) : data && data.length <= 1 ? (
+        <NoProjectWarning style={{margin: 20}} />
```
ErikSin, line 74: this `data.length <= 1` is the temporary solution for determining whether the user is on the default project. Since there is not support in the front end for multi-projects, the user is always going to be in the latest project they made. So if they have only 1 project, we know they are in the default project.

src/frontend/screens/ObservationsList/NoProjectWarning.tsx
ErikSin: Right. So this is the only place in the app that has one sentence with 2 parts of it differently styled. RN doesnt handle that well. You have to have 2 separate text boxes to make it work. 

I think this actually brings up a bigger issue that this might not work in terms of translations. We cannot guarantee the gramatical orderbof these 2 lines, or like you point out, RTL will be a problem. 

Ill flag this as an issue with design, but i think its ok to ignore for now.

src/frontend/screens/ObservationsList/index.tsx
EvanHahn: This isn't ideal but it's fine. I'd probably file an issue for this and refer to this line so we don't forget to fix it once we add multi-project support.

I'd also add _some_ explanation for what this is doing. I'd probably use a variable like `isUserOnDefaultProject`.

EvanHahn (approved):

PR #324, opened 2024-05-07
chore: organize navigation types

ErikSin:
creates a `sharedTypes/navigation.ts` file, and updates all the types related to navigation to live in this file.

PR was created with atomic commits, so can be reviewed by commit.

### DIFF
--- a/src/frontend/sharedTypes/navigation.ts
+++ b/src/frontend/sharedTypes/navigation.ts
@@ -35,7 +35,7 @@ export type HomeTabsList = {
 
 export type TabName = keyof HomeTabsList;
 
-export type AppList = {
+export type RootStackParamList = {
   Home: NavigatorScreenParams<HomeTabsList>;
   GpsModal: undefined;
   Settings: undefined;
@@ -100,13 +100,13 @@ export type AppList = {
   Sync: undefined;
 };
 
-export type DeviceNamingParamList = {
+export type DeviceNamingParamsList = {
   IntroToCoMapeo: undefined;
   DeviceNaming: undefined;
   Success: {deviceName: string};
 };
 
-export type AppStackList = AppList & DeviceNamingParamList;
+export type AppStackList = RootStackParamList & DeviceNamingParamsList;
 
 export type NativeRootNavigationProps<ScreenName extends keyof AppStackList> =
   NativeStackScreenProps<AppStackList, ScreenName>;
--- a/src/frontend/sharedTypes/navigation.ts
+++ b/src/frontend/sharedTypes/navigation.ts
@@ -1,5 +1,11 @@
-import {NavigatorScreenParams} from '@react-navigation/native';
+import {
+  CompositeScreenProps,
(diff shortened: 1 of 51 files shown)


src/frontend/sharedTypes/navigation.ts
```diff
@@ -0,0 +1,124 @@
...
+  Sync: undefined;
+};
+
+export type DeviceNamingSceens = {
+  IntroToCoMapeo: undefined;
+  DeviceNaming: undefined;
+  Success: {deviceName: string};
+};
+
+export type AppStackList = AppList & DeviceNamingSceens;
```
gmaclennan, line 109:
```suggestion
export type AppStackList = AppList & DeviceNamingScreens;
```

src/frontend/sharedTypes/navigation.ts
```diff
@@ -0,0 +1,124 @@
...
+export type HomeTabsList = {
+  Map: undefined;
+  Camera: undefined;
+  Tracking: undefined;
+  ObservationsList: undefined;
+};
+
+export type TabName = keyof HomeTabsList;
+
+export type AppList = {
```
gmaclennan, line 38: Nit: It took me a while to figure out what this is. I think a name like `RootStackParamList` would be clearer, and then calling `HomeTabsList` -> `HomeTabsParamList` and `DeviceNamingScreens` -> `DeviceNamingParamsList` would be clearer, as I understand that these are defining the prop types for each screen. A code comment would be helpful for a future maintainer.

gmaclennan (approved):
This looks ok to me. As a nit-pick, I would personally prefer the screen props to be defined alongside the component, and import them from there into sharedTypes/navigation.ts, but I don't think the change is worth the time and effort right now.

There are a few eslint errors in these files, which I think existed prior to this PR. It would be good to address those in a separate PR. Most of the errors are related to not importing React in files where JSX is used (because JSX components are shorthand for `React.createElement()` which uses the React import). This would be an issue if unit testing and importing these components on their own. When bundled it doesn't really matter.

gmaclennan (approved):
This looks ok to me. As a nit-pick, I would personally prefer the screen props to be defined alongside the component, and import them from there into sharedTypes/navigation.ts, but I don't think the change is worth the time and effort right now.

There are a few eslint errors in these files, which I think existed prior to this PR. It would be good to address those in a separate PR. Most of the errors are related to not importing React in files where JSX is used (because JSX components are shorthand for `React.createElement()` which uses the React import). This would be an issue if unit testing and importing these components on their own. When bundled it doesn't really matter.

ErikSin:
> As a nit-pick, I would personally prefer the screen props to be defined alongside the component, and import them from there into sharedTypes/navigation.ts

I am happy to change this down the line if we want to do that. And generally I would agree, but the file was becoming huge. So I thought separating was nice to cut down the rootstack file by 70 lines. But im going to merge, and happy to change it on a follow up PR if anyone feels strongly about it.

PR #326, opened 2024-05-08
fix/ci: build backend post install, create directory when building translations

CDFN:


package.json
```diff
@@ -93,7 +94,8 @@
     "uint8array-extras": "^0.5.0",
     "utm": "^1.1.1",
     "validate-color": "^2.2.4",
-    "zustand": "^4.4.6"
+    "zustand": "^4.4.6",
+    "expo-updates": "~0.24.12"
```
achou11, line 98: Can you move this so that it's alphabetically sorted? usually doing npm install directly does this

achou11 (approved):
left a comment about fixing a small thing but otherwise lgtm

package.json
achou11: maybe this is an expo thing? like when installing via expo it doesn't do the sorting 🤔

CDFN: `npm install` alone didn't fix the order, but removing entry manually and adding it via `npm install` worked fine

PR #327, opened 2024-05-08
fix: await accept/reject invites

EvanHahn:
**⚠️ I didn't test this because I can't get to this part of the app right now. Probably my fault. Opening in any case!**

*I recommend reviewing this with whitespace changes disabled.*

Projects weren't available right after accepting an invite. This was because we weren't awaiting it.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -32,4 +32,10 @@ module.exports = ({config}) => ({
     package: 'com.comapeo' + SUFFIX,
     googleServicesFile: process.env.GOOGLE_SERVICES_JSON,
   },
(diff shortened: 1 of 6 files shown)


ErikSin:
@EvanHahn I have tries this in a previous iteration, it is not the problem. Tanstack query just expects a promise, and it internally deals with resolving and rejecting.

EvanHahn:
Cool, thanks for taking a look!

achou11:
@ErikSin even if it doesn't fix the issue, I would still consider this to be a valid fix

PR #334, opened 2024-05-10
fix: Initialize background task in correct place

### LINKED ISSUE #304: Fix usage of `TaskManager.defineTask()`
According to Expo's [docs](https://docs.expo.dev/versions/latest/sdk/task-manager/#taskmanagerdefinetasktaskname-taskexecutor) for the method:

> Defines task function. It must be called in the global scope of your JavaScript bundle. In particular, it cannot be called in any of React lifecycle methods like componentDidMount. This limitation is due to the fact that when the application is launched in the background, we need to spin up your JavaScript app, run your task and then shut down — no views are mounted in this scenario.

[Docs for `Expo Location`](https://docs.expo.dev/versions/latest/sdk/location/#background-location-methods) also seem to confirm this:

> Background location task must be defined in the top-level scope, using [TaskManager.defineTask](https://docs.expo.dev/versions/latest/sdk/task-manager#taskmanagerdefinetasktaskname-taskexecutor).

CDFN:
This PR goal is to start background location task in correct place. Current implementation was against Expo documentation.

Fixes #304


package.json
```diff
@@ -59,7 +60,7 @@
     "expo-status-bar": "~1.11.1",
     "expo-system-ui": "^2.9.4",
     "expo-task-manager": "~11.7.3",
-    "expo-updates": "^0.25.9",
+    "expo-updates": "~0.24.12",
```
CDFN, line 63: For reviewer: This was due to incorrect version being pushed into main. For current Expo version, 0.24.12 is desired one.

src/frontend/hooks/useTracking.ts
```diff
@@ -2,11 +2,10 @@ import * as Location from 'expo-location';
 import * as TaskManager from 'expo-task-manager';
 import {useCallback, useState} from 'react';
 import {usePersistedTrack} from './persistedState/usePersistedTrack';
-import React from 'react';
 import {FullLocationData} from '../sharedTypes/location';
-import {useGPSModalContext} from '../contexts/GPSModalContext';
-
-export const LOCATION_TASK_NAME = 'background-location-task';
+import {useInterval} from 'react-native-confirmation-code-field/esm/useTimer';
```
CDFN, line 6: For reviewer: this import may be unstable, let me know if we should create our own useInterval in utils

achou11 (commented):
haven't delved into the code but wondering why async storage is being introduced here? is it serving a different purpose from the persisted state setup that tracks is using via zustand + mmkv?

package.json
achou11: this has been resolved on main branch already, would recommend rebasing/updating this PR

CDFN:
Thanks for comment, Andrew :) This solution is little "hacky" or "ugly", you name it 😄 Basically TaskManager requires us to register outside of App component. This makes it impossible to use zustand store (as zustand store is accessed via hook, and you can use hooks only in components). So we had to find a way to store all the data in zustand store while not being able to do it directly from task. Only thing that comes to my mind to overcome this problem is piping all the data through AsyncStorage into Zustand store. This solution is probably far from perfect in terms of cleanliness but I couldn't think of anything better. Let me know what do you think about it and whether you have any idea on how to improve it.

achou11:
> Thanks for comment, Andrew :) This solution is little "hacky" or "ugly", you name it 😄 Basically TaskManager requires us to register outside of App component. This makes it impossible to use zustand store (as zustand store is accessed via hook, and you can use hooks only in components). So we had to find a way to store all the data in zustand store while not being able to do it directly from task. Only thing that comes to my mind to overcome this problem is piping all the data through AsyncStorage into Zustand store. This solution is probably far from perfect in terms of cleanliness but I couldn't think of anything better. Let me know what do you think about it and whether you have any idea on how to improve it.

right. i think what you want is to use Zustand's [vanilla store API](https://github.com/pmndrs/zustand?tab=readme-ov-file#using-zustand-without-react) so that you can have access to the store instance (which lives as a global variable).

Instead of describing the changes in detail, i created a branch from this branch with some commits to show how to go about this (not a complete implementation, but most of what's needed). to summarize:

1. Create a zustand _store_ for the tracks state.
2. Use `useStore` to create React hooks for that store.
3. Update the background task executor to use the store directly

I haven't tested this at all but in theory, this should do what you want. Things to test:

- does persistence with this setup actually work? there is this note that looks relevant, but don't think is actually applicable given our usage of the store in step (3) because we're not using the store instance's `setState()` or `getState()` directly.

    > ⚠️ Note that middlewares that modify set or get are not applied to getState and setState.

CDFN:
Thanks for helping me out, @achou11 😄 Your way of doing it is way better than what I had initially. I've merged your branch into this PR's branch and made adjustments on top of it. I've tested it and tracks behavior stayed the same. It is ready for re-review, lmk if you have more suggestions.

src/frontend/sharedTypes/index.ts
```diff
@@ -1,6 +1,8 @@
 import {ImageStyle, StyleProp, TextStyle, ViewStyle} from 'react-native';
 import {Observation, ObservationValue} from '@mapeo/schema';
 import type {RoleId, RoleIdForNewInvite} from '@mapeo/core/dist/roles';
+import {FullLocationData} from './sharedTypes/location.ts';
+import * as TaskManager from 'expo-task-manager';
```
achou11, line 5: seems like only the type's needed

```
import {type TaskManagerError} from 'expo-task-manager';
```

src/frontend/sharedTypes/index.ts
achou11, line 1: would place these changes to `sharedTypes/location.ts`

achou11 (approved):
couple of non-blocking suggestions but otherwise lgtm!

PR #336, opened 2024-05-10
chore: useLocationStore

ErikSin:
Implementation of a location store that use `useSyncExternalStore`

### DIFF
--- a/src/frontend/hooks/useLocationStore.ts
+++ b/src/frontend/hooks/useLocationStore.ts
@@ -1,0 +1,151 @@
+import {
+  watchPositionAsync,
+  type LocationObject,
+  Accuracy,
+  LocationSubscription,
+  useForegroundPermissions,
+} from 'expo-location';
+import {useCallback, useRef, useSyncExternalStore} from 'react';
+import CheapRuler from 'cheap-ruler';
+import {useLocationProviderStatus} from './useLocationProviderStatus';
+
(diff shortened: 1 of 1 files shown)


src/frontend/hooks/useLocationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+  // Stop watching for location changes
+  private stopWatching(): void {
+    if (this.subscription) {
+      this.subscription.then(sub => sub.remove());
+      this.subscription = null;
+    }
+  }
+}
+
+const locationStore = LocationStore.getInstance();
```
ErikSin, line 141: i think this is a singleton instance, but react life cycle confuse me so im not sure if that is correct.

src/frontend/hooks/useLocationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+  maxTimeInterval?: number;
+}
+
+export function useLocationStore({
+  maxDistanceInterval,
+  maxTimeInterval = 1000,
+  minTimeInterval = 200,
+}: LocationOptions) {
+  const [permissions] = useForegroundPermissions();
+  const providerStatus = useLocationProviderStatus();
```
ErikSin, line 27: this hook could also be optimized using useSyncExternalStore

ErikSin:
Closing as Andrew is working on this

Issue #65, opened 2023-09-25
Observation Edit


ErikSin:
Bring over the observation edit screen, and update the architecture to use the mmkv/zustand store

To do:
- [ ] update `useDraftObservations` to use new architecture
- [ ] update `locationView`
- [ ] bring over screens
- [ ] using mock api, save an observation

PR #344, opened 2024-05-14
feat: display photos in observations

### LINKED ISSUE #229: Show photos in observation
Currently photos made during and observation are being saved to the db, but we are not showing them in the front end when an observation is opened. This is being ported over from [mapeo-mobile](https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/Observation/ObservationView.js#L206). The api for retrieving a photo will be different from mapeo mobile but the UI should be as similar as possible.

Image

https://github.com/digidem/comapeo-mobile/assets/67773827/34c7d379-8a00-4821-9771-3b9a73ff76a1

## To Do
- [x] Create a tanstack query that retrieves photos via the [blob api](https://github.com/digidem/mapeo-core-next/blob/main/docs/guides/mapeo-http-server.md#working-with-blobs). Use a [suspenseQuery](https://tanstack.com/query/latest/docs/framework/react/guides/suspense) so you do not have to deal with the loading state.
- [x] In an [observation, retrieve photos](https://github.com/digidem/CoMapeo-mobile/blob/626e116dfa6ee7fbfab5c0f77e38ab01c70ec174/src/frontend/screens/Observation/index.tsx#L64) using tantack query. 
- [x] In the observation [show photos](https://github.com/digidem/CoMapeo-mobile/blob/626e116dfa6ee7fbfab5c0f77e38ab01c70ec174/src/frontend/screens/Observation/index.tsx#L90) in the thumbnail scroll view:
- [x] `onPress` should navigation to modal which shows the photo (see video above). Create modal based on [mapeo-mobile's photo-modal](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/screens/PhotosModal/index.tsx). Note that this modal will also be used when a user is creating an observation.

## To note
- this photos modal is the same photos modal used in #230. The user should not be able to see the "delete photo" button in this component.
- In #230 the source of the photo is different then the source of the photo used for this ticket.

CDFN:
Closes #229

### DIFF
--- a/src/frontend/screens/Observation/index.tsx
+++ b/src/frontend/screens/Observation/index.tsx
@@ -14,6 +14,11 @@ import {InsetMapView} from './InsetMapView';
 import {ButtonFields} from './Buttons';
 import {NativeNavigationComponent} from '../../sharedTypes/navigation';
 import {ObservationHeaderRight} from './ObservationHeaderRight';
+import {ThumbnailScrollView} from '../../sharedComponents/ThumbnailScrollView.tsx';
+import {
+  useAttachmentsBase64Query,
+  useAttachmentUrlQueries,
+} from '../../hooks/server/media.ts';
 
 const m = defineMessages({
   deleteTitle: {
@@ -61,7 +66,13 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
   const {lat, lon, createdBy} = observation;
   const isMine = deviceId === createdBy;
   // Currently only show photo attachments
-  const photos = [];
+  const attachmentUrls = useAttachmentUrlQueries(
+    observation.attachments,
+    'original',
+  );
+  const base64Uris = useAttachmentsBase64Query(
+    attachmentUrls.filter(r => !!r.data).map(r => r.data!),
+  );
 
   return (
     <ScrollView
@@ -86,15 +97,16 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
               <Text style={styles.textNotes}>{observation.tags.notes}</Text>
             </View>
           ) : null}
-
-          {/* {!!photos.length && (
-            <ThumbnailScrollView
-              photos={
-                photos
-              }
-              onPressPhoto={() => {}}
-            />
-          )} */}
+          <ThumbnailScrollView
+            photos={base64Uris
+              .filter(uri => !!uri.data)
+              .map(({data}) => {
+                return {
+                  thumbnailUri: data!.base64Uri,
+                  id: data!.driveDiscoveryId,
+                };
+              })}
+          />
         </View>
         {fields && fields.length > 0 && <FieldDetails fields={fields} />}
         <View style={styles.divider}></View>
@@ -112,7 +124,9 @@ const styles = StyleSheet.create({
     flex: 1,
     flexDirection: 'column',
   },
-  scrollContent: {minHeight: '100%'},
+  scrollContent: {
+    minHeight: '100%',
+  },
   divider: {
     backgroundColor: LIGHT_GREY,
     paddingVertical: 15,
--- a/src/frontend/sharedComponents/ThumbnailScrollView.tsx
+++ b/src/frontend/sharedComponents/ThumbnailScrollView.tsx
@@ -38,6 +38,7 @@ export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
     log('Error loading image:\n', e.nativeEvent && e.nativeEvent.error);
     setError(true);
   }
+
   return (
     <TouchableOpacity
       style={[styles.thumbnailContainer, {width: size, height: size}, style]}
@@ -60,6 +61,7 @@ export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
 export const ThumbnailScrollView = (props: {photos: Photo[]}) => {
   const scrollViewRef = React.useRef<ScrollView>(null);
   const {photos} = props;
+
   React.useLayoutEffect(() => {
     scrollViewRef.current && scrollViewRef.current.scrollToEnd();
   }, [photos.length]);
--- a/src/frontend/sharedComponents/ThumbnailScrollView.tsx
+++ b/src/frontend/sharedComponents/ThumbnailScrollView.tsx
@@ -38,7 +38,6 @@ export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
     log('Error loading image:\n', e.nativeEvent && e.nativeEvent.error);
     setError(true);
   }
-
   return (
     <TouchableOpacity
       style={[styles.thumbnailContainer, {width: size, height: size}, style]}
@@ -58,10 +57,9 @@ export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
   );
 };
 
-export const ThumbnailScrollView = () => {
+export const ThumbnailScrollView = (props: {photos: Photo[]}) => {
   const scrollViewRef = React.useRef<ScrollView>(null);
-  const photos = usePersistedDraftObservation(store => store.photos);
-
+  const {photos} = props;
   React.useLayoutEffect(() => {
     scrollViewRef.current && scrollViewRef.current.scrollToEnd();
   }, [photos.length]);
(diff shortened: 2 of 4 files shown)


src/frontend/screens/Observation/index.tsx
```diff
@@ -61,7 +66,13 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
   const {lat, lon, createdBy} = observation;
   const isMine = deviceId === createdBy;
   // Currently only show photo attachments
```
EvanHahn, line 68: This comment isn't respected. Can we make sure to only show photo attachments?

src/frontend/screens/Observation/index.tsx
```diff
@@ -61,7 +66,13 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
-  const photos = [];
+  const attachmentUrls = useAttachmentUrlQueries(
+    observation.attachments,
+    'original',
```
EvanHahn, line 71: Shouldn't we be requesting the `'thumbnail'`?

src/frontend/screens/Observation/index.tsx
```diff
@@ -61,7 +66,13 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
+  );
+  const base64Uris = useAttachmentsBase64Query(
+    attachmentUrls.filter(r => !!r.data).map(r => r.data!),
+  );
```
EvanHahn, line 75: Why do you need to convert things to base64 data URIs? Why can't you pass the thumbnail URLs directly?

src/frontend/screens/Observation/index.tsx
```diff
@@ -86,15 +97,16 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
-
-          {/* {!!photos.length && (
-            <ThumbnailScrollView
-              photos={
-                photos
-              }
-              onPressPhoto={() => {}}
-            />
-          )} */}
+          <ThumbnailScrollView
```
EvanHahn, line 100: Should we show this view if there are no photos?

EvanHahn (changes requested):
I think this is a good change but three additional things are needed:

1. This doesn't completely address #229 because tapping a photo doesn't open a modal. See the screen recording in the issue for a reference. (I assume you'll need to fill in `handlePhotoPress` in `ThumbnailScrollView.tsx` to accomplish this.)
1. `ThumbnailScrollView` bug: when photos take a long time to load, 
1. `ThumbnailScrollView` bug: when photos take a long time to load, there is no UI for this. A grayed-out square, or some other placeholder, should be shown.

All of these can be done in a followup PR.

src/frontend/screens/Observation/index.tsx
CDFN: This is weird one. I was struggling for good hour or so to make it work just by passing URL to a blob. However every time I did that the backend would return 400 so this can be considered as some kind of "workaround". If you have any idea why the backend may return 400 when requesting image via `<Image source={{uri: <ATTACHMENT URI>}}`, all suggestions are very welcome because I don't really like current solution cleanliness either.

src/frontend/screens/Observation/index.tsx
CDFN: This will return empty scroll view which AFAIK renders as "nothing" if it has no items but I'll change in nonetheless to keep previous (commented out) behavior.

src/frontend/screens/Observation/index.tsx
CDFN: Good point, didn't know this is a thing to be honest. That's changed now.

CDFN:
Thanks for quick review, Evan 😃 You're right, this PR doesn't implement opening modal when you click a photo. I should have noted we're going to do this in another PR which also implements delete button in spoken modal. I'll address #3 (and #2 perhaps as I guess it's just typo) probably tomorrow morning for us.

Also, is it okay for us to include modal changes along with observation editing (#65) or you'd like to see that in completely separate PR?

src/frontend/screens/Observation/index.tsx
CDFN: Good point, that's sorted now 👍

src/frontend/screens/Observation/index.tsx
EvanHahn: We discussed this offline and it seems like this is a necessity.

src/frontend/screens/Observation/index.tsx
```diff
@@ -86,15 +97,18 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
-          {/* {!!photos.length && (
+          {base64Uris.length > 0 && (
             <ThumbnailScrollView
-              photos={
-                photos
-              }
-              onPressPhoto={() => {}}
+              photos={base64Uris.map(({data}) => {
+                return !data
+                  ? undefined
```
EvanHahn, line 104: When does this happen?

src/frontend/sharedComponents/ThumbnailScrollView.tsx
```diff
@@ -58,10 +64,9 @@ export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
   );
 };
 
-export const ThumbnailScrollView = () => {
+export const ThumbnailScrollView = (props: {photos: (Photo | undefined)[]}) => {
```
EvanHahn, line 67: It feels a little weird that we'd allow an array with `undefined` in it. What's the use case for that?

src/frontend/sharedComponents/ThumbnailScrollView.tsx
```diff
@@ -32,7 +32,11 @@ type ThumbnailProps = {
 export const Thumbnail = ({photo, style, size, onPress}: ThumbnailProps) => {
   const [error, setError] = React.useState(false);
 
-  const uri = 'thumbnailUri' in photo ? photo.thumbnailUri : undefined;
+  const uri =
+    (photo && 'thumbnailUri' in photo && photo.thumbnailUri) || undefined;
+  const isCapturing =
+    (photo && 'capturing' in photo && (photo.capturing as boolean)) ||
+    undefined;
```
EvanHahn, line 39: Nitpick: I would expect `isCapturing` to be a boolean, not `undefined | boolean`. Can you make sure it's always a bool?

src/frontend/sharedComponents/ThumbnailScrollView.tsx
```diff
@@ -16,14 +16,14 @@ import debug from 'debug';
...
-import {usePersistedDraftObservation} from '../hooks/persistedState/usePersistedDraftObservation';
+import * as Progress from 'react-native-progress';
 
 const spacing = 10;
 const minSize = 150;
 const log = debug('Thumbnail');
 
 type ThumbnailProps = {
-  photo: Photo;
+  photo?: Photo;
```
EvanHahn, line 26: When would this happen?

EvanHahn (changes requested):
> 2) perhaps as I guess it's just typo

Whoops, sorry for the typo.

When photos take a long time to load, `ThumbnailScrollView` will jump around unexpectedly. I reproduced this by adding 10 photos to an observation.

Again, this can probably be fixed in a followup PR.

> Also, is it okay for us to include modal changes along with observation editing (#65) or you'd like to see that in completely separate PR?

I'd prefer a separate, smaller PR. Does that work for you?

src/frontend/screens/Observation/index.tsx
CDFN: Whenever a query isn't resolved yet (image is loading)

src/frontend/sharedComponents/ThumbnailScrollView.tsx
CDFN: it's so we can keep placeholders for still loading images

src/frontend/sharedComponents/ThumbnailScrollView.tsx
CDFN: Yeah good point 👍

CDFN:
This PR will most likely need to be put on hold until #335 is approved as points made there are also applicable for this PR.

EvanHahn:
Sounds good. Let me know when I should give this another look.

EvanHahn (approved):

PR #348, opened 2024-05-14
Feat/show icons associated with a preset

### LINKED ISSUE #340: Show Icons Associated with a preset
Update [CategoryIcon Component](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/sharedComponents/icons/CategoryIcon.tsx) to display the appropriate icon. Also Rename the "CategoryIcon" to "PresetIcon". 

The `CategoryIcon` should display the icon associated with a preset (hence the name change). A Preset is a category associated with each observation. When a preset is created, and png or svg icon is saved in the database as well. Using the [blobs api](https://github.com/digidem/mapeo-core-next/blob/main/docs/guides/mapeo-http-server.md#working-with-icons), update the `CategoryIcon` component to retrieve the appropriate icon. 

## To note
- the current implementation is from the previous architecture. 

## To do:
- [x] Rename  "CategoryIcon" to "PresetIcon".
- [x] Remove `color` prop
- [x] Create a tanstack query to retrieve the icon
- [x] Use the tanstack query inside of `PresetIcon`
- [x] Create a loading state that shows the user a [UIActivityIndicator](https://github.com/n4kz/react-native-indicators?tab=readme-ov-file#uiactivityindicator-properties) in place of the preset while the query is being run
- [x] if there is a query error or there is no `iconId`, show < MaterialIcon name="place" / >

bohdanprog:
Add ability to show preset icon, by iconId

Closes #340


src/frontend/sharedComponents/icons/PresetIcon.tsx
```diff
@@ -0,0 +1,56 @@
+import React, {memo} from 'react';
+import {Image} from 'react-native';
+
+import {Circle} from './Circle';
+import {IconSize} from '../../sharedTypes';
+import {UIActivityIndicator} from 'react-native-indicators';
+import {useGetPresetIcon} from '../../hooks/server/presets';
+import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
+
+interface CategoryIconProps {
```
ErikSin, line 10: change to `PresetIconProps`

src/backend/package-lock.json
ErikSin, line 1: Do you know why this is being changed? Can you just run an `npm i` in the backend and see if gets rid of this

ErikSin (approved):
Looks good. One small name change, and a weird thing with the package-lock.

Otherwise can be merged

src/frontend/sharedComponents/icons/PresetIcon.tsx
bohdanprog: done

src/backend/package-lock.json
bohdanprog: IDK, I remove `package-lock.json` and node_modules, and run `npm i` but still has some diff

ErikSin: Thats fine, its probably just related to your node/npm versioning (or if your using yarn). package lock in the backend is not actually included in the bundle, so you can merge

achou11: please do not merge this - the lockfile is important when building the backend that eventually gets run in the app. There shouldn't be significant changes to it based on what you have here. Would prefer you revert the changes to that file entirely and look into it

achou11: > I remove package-lock.json and node_modules, and run npm i but still has some diff

generally this is not a recommended practice and will result in a complete re-resolving of the dependencies, leading to tons of upgrades/changes in dependencies. most of the time that is not desirable.

what i recommended when you're trying to figure out why obscure lockfile changes have happened: copy the lockfile from a branch that has the expected lockfile before you ran `npm install`. after that, run `npm i` (or `npm ci`). this will properly make more granulat adjustments to the lockfile if there is an actual issue.

achou11 (changes requested):
Please see my comments in in https://github.com/digidem/comapeo-mobile/pull/348#discussion_r1602218117

you shouldn't have any lockfile changes based on the changes this PR introduces

src/backend/package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mapeo-mobile-node-next",
   "version": "1.0.0",
-  "lockfileVersion": 2,
```
achou11, line 4: this indicates that you're using a npm version that's older than the one this project expects, hence the massive changes. you should be using npm v10 for this project.

https://stackoverflow.com/a/67874161

bohdanprog: Okay, I got package-lock.json from main branch.

achou11 (approved):
approving the removal of the lockfile changes. deferring to Erik's approval for that app changes

PR #349, opened 2024-05-14
Feat default preset added to project with no preset

ErikSin:
Default Preset in project creation


EvanHahn (approved):

EvanHahn:
Looks like tests are failing because versions aren't exact.

PR #353, opened 2024-05-15
improve ErrorModal

### LINKED ISSUE #232: Update and rename Error Modal
Update the name, functionality, and images used in the [ErrorModal](https://github.com/digidem/CoMapeo-mobile/blob/main/src/frontend/sharedComponents/ErrorModal.tsx). Error modal is actually a bottom sheet, so it should be renamed accordingly. Also, we want to replace the CoMapeo logo with a more generic icon (see image below). Lastly we want to update the functionality of the Error bottom sheet to be declarative and be integrated with #224 work.

### Declarative Bottom Sheet :
Currently the `ErrorBottomSheet` is opened imperatively. The parent component uses the `openSheet()` function when there in error. This means that the parent component has to detect an error and imperatively call openSheet(). As well, because the `openSheet` function comes from` useBottomSheet` it needs to pass the appropriate props into the bottom sheet.

```ts
// Currently implementation of the bottom sheet, which is imperative:

function SomeComponent(){
     const {data, error}  = someTanstackQuery()
     // most of these props are not being used by the parent component
     const {isOpen, closeSheet, openSheet, sheetRef} = useBottomSheet()

     // We check on every render whether we should manually open the bottom sheet
     if(error && !isOpen){
          openSheet()

          return null
     }
     
     return( 
          <>
               <ThisComponent {...data}/>
               <ErrorBottomSheet isOpen={isOpen} closeSheet={closeSheet} ref={sheetRef} error={error} />
          </>
      )
}


// Declarative

function SomeComponent(){
     const {mutatate, isError, reset}  = someTanstackMutation()
    
     return( 
          <>
               <ThisComponent onPress={()=>mutate()}/>
               <ErrorBottomSheet error={isError} clearError={reset} />
          </>
      )
}

function ErrorBottomSheet({error, clearError?}){
     const {isOpen, closeSheet, openSheet, sheetRef} = useBottomSheet()

     // This is technically still imperative BUT it is not exposed to the parent component
     if(error && !isOpen){
          openSheet()
     }

     function handleDismiss(){
          if(clearError){
               clearError()  
          }
          // if there is no way to clear the error, the user should be navigated away from the origianl page
          else { 
                navigation.goBack()
           }

          closeSheet()
     }    
     return( 
          <BottomSheetModal isOpen={isOpen}  onDismiss={handleDismiss} closeSheet={closeSheet} ref={sheetRef} >
               <BottomSheetModalContent/>
          </>
      )
}

```

We can make the bottom sheet declarative by just passing it an error. On close of the bottom sheet the error is cleared (so the user can try again) or if there is no way to clear the error, navigate the user back.

### Images

Image

## To do:
- [x] Rename `ErrorModal` to `ErrorBottomSheet`
- [x] Replace CoMapeo logo with map icons
- [x] Make `ErrorBottomSheet` more declarative. Bottom sheet should take some sort of Error as a prop (see example above). 
- [x] Inside the `ErrorBottomSheet` pass the error to Sentry.

### To note
- @ErikSin has asked Aldo for the updated SVGs and will update this ticket once those become avaialble

bohdanprog:
Closes #232 

- [x] We have to update the asset for `ErrorBottomSheet` component


src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -87,7 +85,7 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> & {
         <ThumbnailScrollView />
       </ScrollView>
       <BottomSheet items={bottomSheetItems} />
-      <ErrorModal sheetRef={sheetRef} closeSheet={closeSheet} isOpen={isOpen} />
+      <ErrorBottomSheet error={error} clearError={() => setError(null)} />
```
ErikSin, line 88: I think we should only pass a meaningful clear error to the bottom sheet, such as [resettingMutationState](https://tanstack.com/query/latest/docs/framework/react/guides/mutations#resetting-mutation-state). If we can't pass a way to meaningfully clear the error, than we should just pass a go back function, which navigates the user back so they can start the process again.

ErikSin (changes requested):
Some changes in regards to `goBack` and `clearError`

src/frontend/sharedComponents/ErrorBottomSheet.tsx
```diff
@@ -0,0 +1,112 @@
...
+});
+
+type ErrorModalProps = {
+  error: Error | null;
+  clearError?: () => void;
+};
+
+export const ErrorBottomSheet = ({error, clearError}: ErrorModalProps) => {
+  const {formatMessage} = useIntl();
+  const navigation = useNavigationFromRoot();
```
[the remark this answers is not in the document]
bohdanprog: I think maybe it is better to have goBack props and pass where we want to navigate the user.

src/frontend/sharedComponents/ErrorBottomSheet.tsx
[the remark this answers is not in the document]
bohdanprog: yeah, unfortunately, I didn't know what what's the difference between clear and go back buttons.

src/frontend/screens/ObservationEdit/index.tsx
bohdanprog: It makes sense for me know :) Thank you

bohdanprog: Let me know if current solution is okay :)

src/frontend/sharedComponents/ErrorBottomSheet.tsx
ErikSin, line 1: The props should be:

```ts
type ErrorModalProps = {
  error: Error | null;
  retryFunction?:()=>unknown 
  clearError: () => void; // This cannot be optional as it is needed to close the bottom sheet
};

Comment (ErikSin) [inline src/frontend/sharedComponents/ErrorBottomSheet.tsx:1]: And then we should also have 2 functions here:

```ts
function handleGoBack() {
      clearError();
     closeSheet();
  }

```

And then have another function:
```ts
// this function retries the query/Mutation and should ONLY be available if `!!retryFunction`
function handleTryAgain(){
     clearError()
     closeSheet()
     retryFunction()
}
```

src/frontend/sharedComponents/ErrorBottomSheet.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const {openSheet, sheetRef, isOpen, closeSheet} = useBottomSheetModal({
+    openOnMount: false,
+  });
+
+  useEffect(() => {
+    if (error && !isOpen) {
+      Sentry.captureMessage(error.message, 'error');
+      openSheet();
+    }
+  }, [error, isOpen, openSheet]);
```
ErikSin, line 44: I don't think we need a useEffect here. I think we can do an inline function that can get caught during the render.

```ts
    if (error && !isOpen) {
      Sentry.captureMessage(error.message, 'error');
      openSheet();
    }

```

ErikSin (changes requested):
Sorry, I just going through the notes about how this modal should work and I realize that I am giving you slightly wrong information.

There are 2 possibilities:
1. Try Again: This should close the modal AND try to re-run the query/mutation.
2. Go back: this should simply clear the error and close the modal (it has to clear the error, otherwise error would be true and the modal wouldn open up immediately again).

So we don't actually need deal with any navigation. And the bottomsheet should take 3 props:

```ts
type ErrorModalProps = {
  error: Error | null;
  retryFunction?:()=>unknown 
  clearError: () => void; // This cannot be optional as it is needed to close the bottom sheet
};

```

ErikSin:
Also just noting that we don't actually need to deal with navigation anymore. `Go Back` just implies closing the bottoms sheet, which is not part of navigation.

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/index.tsx
```diff
@@ -61,12 +54,25 @@ export const ReviewAndInvite: NativeNavigationComponent<'ReviewAndInvite'> = ({
...
+  }
+
+  function tryAgain() {
+    if (sendInviteMutation.isIdle) {
+      sendInvite();
+    } else {
+      cancelInvite();
+    }
+  }
+
```
ErikSin, line 75: Should be
```ts
  function tryAgain() {
    if (sendInviteMutation.isError) {
      sendInviteMutation.reset()
      sendInvite();
     return
    } 
     if(requestCancelInviteMutation.isError){
          requestCancelInviteMutation.reset()
          cancelInvite()
     }
    
  }

```

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/index.tsx
```diff
@@ -61,12 +54,25 @@ export const ReviewAndInvite: NativeNavigationComponent<'ReviewAndInvite'> = ({
...
     });
   }
 
+  function clearError() {
+    if (sendInviteMutation.isIdle) {
+      sendInviteMutation.reset();
+    } else {
+      requestCancelInviteMutation.reset();
+    }
+  }
```
ErikSin, line 66: should be:

if (sendInviteMutation.isError) {
      sendInviteMutation.reset()
    } 
     if(requestCancelInviteMutation.isError){
          requestCancelInviteMutation.reset()
     }

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/index.tsx
```diff
@@ -80,11 +86,14 @@ export const ReviewAndInvite: NativeNavigationComponent<'ReviewAndInvite'> = ({
...
-        sheetRef={sheetRef}
-        closeSheet={closeSheet}
-        isOpen={isOpen}
-        clearError={() => navigation.navigate('YourTeam')}
+      <ErrorBottomSheet
+        error={
+          sendInviteMutation.isIdle
+            ? sendInviteMutation.error
+            : requestCancelInviteMutation.error
+        }
```
ErikSin, line 94: should be:
```ts
 error={ sendInviteMutation.error || requestCancelInviteMutation.error }
```

ErikSin (approved):
3 small changes, but can be merged afterwards

PR #354, opened 2024-05-15
pass device status to DeviceCard, add disconnected icon

### LINKED ISSUE #235: Project Settings/ Your Team/ deviced disconnected
When inviting other devices to a project, user can see a list of available devices on the network. When one of those available devices disconnect from the network, we show that device as disconnected. 

We do not want the device that disconnected to be removed from the list, nor do we want the order of the list to change (new devices should added to the bottom). 

This screen can be found [here](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx)

## User flow of deviced disconnected
- Project Coordinator was on Project Settings and clicked on **Your team**
- Project Coordinator is on a project and can see the project and available devices
- 1 or more devices disconnect : Use case -  bad internet connection.
- Project Coordinator and others see the disconnect devices on the **Your Team** screen and **Select Device to Invite** screen

## Todo
- [x] Add the orange icon and display *Disconnected* under the disconnected devices
- [x] If the user is disconnected, the `onPress` should be disabled.
- [x] Icon should be removed and button re-enabled if the user re-connects

Screen Shot 2024-04-06 at 12 55 18 PM

Screen Shot 2024-04-06 at 12 53 51 PM
Screen Shot 2024-04-06 at 12 54 07 PM

## Design
https://www.figma.com/file/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?type=design&node-id=1075-5095&mode=design&t=1zbIlqd9eRkTVFTz-0

bohdanprog:
Closes #235


ErikSin (changes requested):
The device should update when it disconnects [here](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/hooks/useLocalPeers.ts). I have noticed with some playing around that it is working pretty inconsistently. Its registering the device as disconnected only sometimes.

@bohdanprog if you could make that change i was referring to about the boolean.

@EvanHahn can you look into the `api.on('local-peers', onPeers);` and see if you can find any problems in the backend

src/frontend/sharedComponents/DeviceCard.tsx
```diff
@@ -1,20 +1,15 @@
...
 import {DeviceNameWithIcon} from './DeviceNameWithIcon';
 
-const m = defineMessages({
-  thisDevice: {
-    id: 'sharedComponents.DeviceCard.ThisDevice',
-    defaultMessage: 'This Device!',
-  },
-});
-
+export type DeviceStatus = 'connected' | 'disconnected';
```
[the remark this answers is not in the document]
bohdanprog: fixed

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
```diff
@@ -52,6 +59,14 @@ export const DeviceNameWithIcon = ({
             {formatMessage(m.thisDevice)}
           </Text>
         )}
+        {isConnected && (
```
EvanHahn, line 62: This looks backwards to me. Should this instead be `!isConnected`?

```suggestion
        {!isConnected && (
```

src/frontend/sharedComponents/DeviceCard.tsx
```diff
@@ -30,16 +24,16 @@ export const DeviceCard = ({
   deviceId,
   dateAdded,
   onPress,
+  isConnected,
```
EvanHahn, line 27: I assume the default here is `true`. Could we set that?

```suggestion
  isConnected = true,
```

EvanHahn (changes requested):

EvanHahn:
> @EvanHahn can you look into the `api.on('local-peers', onPeers);` and see if you can find any problems in the backend

_tl;dr: I think `useLocalPeers` looks fine to me._

I only see two small issues in `useLocalPeers`, and I would only expect these bugs to manifest at the very beginning (if at all).

1. During the initial subscription, it's possible that a `local-peers` event will fire _before_ `api.listLocalPeers()` resolves. Here's the code:

   https://github.com/digidem/comapeo-mobile/blob/c64cdf06e08f424bf3da1ba15366b62526460a5b/src/frontend/hooks/useLocalPeers.ts#L30-L37

   Imagine that `api.listLocalPeers()` starts, then a `local-peers` event fires, then `api.listLocalPeers()` resolves. The data might be stale.

1. If `api.listLocalPeers()` fails after we're unsubscribed, it might fire listeners unnecessarily.

   Imagine that `api.listLocalPeers()` starts, then we unsubscribe. This would cause stale listeners to be called.

Again, not sure if these are real issues; I would need to test.

Also, I might have missed something!

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
bohdanprog: Oh, my mistake. I believe I forgot to make that change after testing it earlier.

EvanHahn (commented):
LGTM but we should wait for @ErikSin before merging, because it sounds like he was seeing some strange behavior.

ErikSin:
@EvanHahn on testing, i did the followng: `api.on('local-peers', peers => {console.log(peers )}).`. And i was getting very inconsistent results when my device was leaving or joining a network. It was not the expected behaviour. so i think it is the backend? But lmk if you want to hop on a call and look at it together

EvanHahn:
I'd love to hop on a call and look at this, as I can't see any obvious problems with the code.

EvanHahn:
Looks like there's a merge conflict here. @bohdanprog  Could you fix?

EvanHahn (approved):
If there's a bug here, it seems unlikely that it's in this PR. LGTM.

PR #362, opened 2024-05-16
fix: fix location-related imports used by TaskManager.defineTask()

achou11:
Follow up to #334 which implemented a suggestion but forgot to update imports 😅


CDFN:
Woops, sorry 👀

PR #363, opened 2024-05-16
fix: refactor active project setup to fix potential issues related to using stale client

achou11:
Seeing the comment [here](https://github.com/digidem/comapeo-mobile/pull/351/files#diff-90bcb86592b756e9aad55bd5061a3d637466f8276c9958d3839df507057a3b27R21) highlighted issues around the active project setup where one could be working with a stale project client instance. These changes were initially applied to that branch to confirm that they fixed the issue noted there, which it did.

Instead of relying on useEffects and manually fetching new project instances, we now use React Query to handle the fetching. More importantly, this makes it easier to properly invalidate caches when needed, which I believe is the root of the issue that initiated this refactor.

Took some liberties to also clean up some low-hanging fruit in other existing query hooks. Apologies for the extra noise if distracting


src/frontend/hooks/server/invites.ts
```diff
@@ -21,23 +24,30 @@ export function usePendingInvites() {
...
-  const switchActiveProject = useUpdateActiveProjectId();
+  const switchActiveProject = usePersistedProjectId(
+    state => state.setProjectId,
+  );
 
   return useMutation({
     mutationFn: async ({inviteId}: {inviteId: string}) => {
       if (!inviteId) return;
-      mapeoApi.invite.accept({inviteId});
+      return mapeoApi.invite.accept({inviteId});
```
achou11, line 34: it's important to actually return this value here. otherwise a couple of things will happen:

1. You won't get the return value of `invite.accept()` (doesn't currently break anything)
2. You'll potentially run into race conditions because you may try to do work after the mutation call occurs under the assumption that `invite.accept()` has resolved. However, without this change, the accept resolves on the next iteration in the event loop, breaking that assumption

src/frontend/hooks/server/invites.ts
```diff
@@ -21,23 +24,30 @@ export function usePendingInvites() {
...
     },
     onSuccess: () => {
       // This is a workaround. There is a race condition where the project in not available when the invite is accepted. This is temporary and is currently being worked on.
       setTimeout(() => {
-        queryClient
```
achou11, line 34: Wondering if the changes here might help with the comment above this call? wasn't too sure so didn't attempt to make any changes to fix it in this PR

ErikSin (commented):
There is actually a slight problem (that was also present in the previous implementation, but is a little more obvious in your implementation because of your use of tanstack).

tanstack uses `stale while revalidate` architecture. So when a user changes to a new project, while it is fetching the new project, the old project is returned. Meaning that the user is able to interact with the old project while the new project in being re-fetched. This is not going to be a problem on most phones because it is really fast. But it might be a problem for slower phones.

We could possibly use the [background fetching indicator](https://tanstack.com/query/latest/docs/framework/react/guides/background-fetching-indicators), to block the UI, but the context lives outside the navigator, so conditionally rendering a loader will unmount the navigator which is not ideal. There are other ways we can block the ui without unmounting but im not sure of the most elegant way of doing that.

achou11:
> We could possibly use the [background fetching indicator](https://tanstack.com/query/latest/docs/framework/react/guides/background-fetching-indicators), to block the UI, but the context lives outside the navigator, so conditionally rendering a loader will unmount the navigator which is not ideal. There are other ways we can block the ui without unmounting but im not sure of the most elegant way of doing that.

Not sure what the preferred solution is, but we can probably use an absolutely positioned translucent loader that overlays the whole app and blocks pointer events, at least to start.

achou11:
do you think i should try to tackle that in this PR or do it as a follow up?

ErikSin (approved):
I added a small fix (i think it was missed in the rebase).

I'm going to merge becasue this is blocking my work in #351

Issue #364, opened 2024-05-16
Cryptic deserialization error after accepting invite


achou11:
Have very little idea of the cause of this. Guess is that some API call is returning an error instance, which RPC reflector (or React Native) can't deserialize?

only tried this in development

<img width="483" alt="image" src="https://github.com/digidem/comapeo-mobile/assets/18542095/7bc8faaa-5df4-4d64-aa9c-6d76248d2321">

achou11:
Context from Gregor:

> The error is coming from RPC reflector, so you are not able to see the actual error. If you look at the earlier thread in the SWM I explain to SWM why this is happening because of the metro config. There is a setting in metro config that tells it to not resolve dependencies correctly, so deps are resolved with incorrect versions. I think they have done this because the bundled was maybe trying to bundle the backend code. I would think the solution is to blocklist the backend folder in metro config - see Mapeo-mobile for the config options. The current metro config in comapeo will cause weird bugs if it’s not changed. 

Will fix the deserialization part, but not the underlying error

achou11:
Closing this for now since the PR to fix tge deserialization error has been merged. there might be an error at the app level that needs fixing, but will create a separate issue if needed

PR #367, opened 2024-05-20
fix: Fix metro config hierarchical lookup option

gmaclennan:
This PR removes the custom metro config option `disableHierarchicalLookup: true` option. This option is problematic because it appears to break correct package resolution according to semver. In normal behaviour, if two different dependencies have the same sub-dependency, but incompatible semver versions, then the dependency will be saved in nested `node_modules`, to ensure each dependency gets the sub-dependency that they are expecting. Disabling this behaviour seems to cause dependencies to be incorrectly resolved, in this case for the package `serialize-error`. `metro` depends on `serialize-error@2.1.0` and `rpc-reflector` depends on `serialize-error@8.1.0`. With `disableHierarchicalLookup: true`, `rpc-reflector` tries to use the version of `serialize-error` installed in the top-level `node_modules` folder, which is version @2.1.0. This causes errors in rpc-reflector because the API is different across major versions of this package.

I believe that this option will result in subtle bugs down the line in other modules. It's going to be rare that things actually crash, because in general two versions of a module are compatible, but behaviour might be different and dependencies might not be tested on these different versions.

I initially got build errors upon making this change, but I think they might have been due to the debug version of the app installed on my phone. The error I got was `Error: Cannot find native module 'ExpoSplashScreen'`. I fixed this with these steps:

1. Uninstalled the debug build of my app from my development phone.
2. Ran `npx expo prebuild --clean`
3. Rebuilt and installed the app via `npm run android`

After these steps the app was working without error for me.

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -2,6 +2,9 @@ const {getSentryExpoConfig} = require('@sentry/react-native/metro');
 
 /** @type {import('expo/metro-config').MetroConfig} */
 const config = getSentryExpoConfig(__dirname);
+const defaultBlockList = Array.isArray(config.resolver.blockList)
+  ? config.resolver.blockList
+  : [config.resolver.blockList];
 
 module.exports = {
   ...config,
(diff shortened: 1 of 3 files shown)


achou11 (approved):

achou11:
Just adding that this should fix #364 too

bohdanprog (approved):

PR #369, opened 2024-05-20
feat: Localize App

ErikSin:
Creates language screen in settings

image

### DIFF
--- a/src/frontend/screens/Settings/AppSettings/Languages.tsx
+++ b/src/frontend/screens/Settings/AppSettings/Languages.tsx
@@ -1,0 +1,42 @@
+import * as React from 'react';
+import {ScrollView} from 'react-native';
+import {defineMessages} from 'react-intl';
+import {usePersistedLocale} from '../../../hooks/persistedState/usePersistedLocale';
+import {supportedLanguages} from '../../../lib/intl';
+import {SelectOne} from '../../../sharedComponents/SelectOne';
+import {NativeNavigationComponent} from '../../../sharedTypes/navigation';
+
+const m = defineMessages({
+  title: {
+    id: 'screens.LanguageSettings.title',
+    defaultMessage: 'Language',
+    description: 'Title language settings screen',
+  },
+});
+
+export const Languages: NativeNavigationComponent<'Languages'> = () => {
+  const [locale, setLocale] = usePersistedLocale(store => [
+    store.locale,
+    store.setLocale,
+  ]);
+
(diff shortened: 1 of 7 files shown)


src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -25,7 +25,11 @@ export const HomeHeader: FC<BottomTabHeaderProps> = ({navigation}) => {
         <SyncIconCircle />
       </IconButton>
       <GPSPill navigation={navigation} />
-      <IconButton onPress={() => {}} testID="observationListButton">
+      <IconButton
+        onPress={() => {
+          navigation.navigate('Settings');
+        }}
+        testID="observationListButton">
```
ErikSin, line 32: I updated this a temporary button to show the setting screen (otherwise this would be impossible to test)

src/frontend/sharedTypes/navigation.ts
```diff
@@ -35,7 +35,7 @@ export type RootStackParamsList = {
   Settings: undefined;
   ProjectConfig: undefined;
   AboutMapeo: undefined;
-  LanguageSettings: undefined;
```
achou11, line 38: nit: think keeping this as the navigation name is a little clearer and less prone to ambiguity in the future. not sure if there would be any other kind of language-related screens in the future but you never know 🤷

src/frontend/screens/Settings/AppSettings/Languages.tsx
```diff
@@ -0,0 +1,42 @@
...
+    defaultMessage: 'Language',
+    description: 'Title language settings screen',
+  },
+});
+
+export const Languages: NativeNavigationComponent<'Languages'> = () => {
+  const [locale, setLocale] = usePersistedLocale(store => [
+    store.locale,
+    store.setLocale,
+  ]);
```
achou11, line 21: not entirely sure, but i _think_ returning an array here is not advised by zustand because you lose reference equality in the return value each time the selector is called. the safer option would be to have two invocations instead

```suggestion
  const locale = usePersistedLocale(store => store.locale);
  const setLocale = usePersistedLocale(store => store.setLocale);
```

achou11: ah yeah here's where i got this from: https://tkdodo.eu/blog/working-with-zustand#prefer-atomic-selectors

alternative is to use specify the `shallow` option, but i think having the two separate invocations works fine

PR #371, opened 2024-05-21
feat: add screen for importing config file

### LINKED ISSUE #161: Import Project Config
Import a config file (Api takes path to the zip file).

1. User Selects file
2. Front end copies file into temp storage
3. front end passes path to api
4. When resolved or rejected clear the copy in the

achou11:
Closes #161 

As I understand it, config imports in Mapeo are treated differently from config imports in CoMapeo. In Mapeo, imports are entities that have metadata such as the name and version, so we surfaced that information in this screen. In CoMapeo, we do not have access to this metadata, as imports are solely treated as a method of updating information about a project. Due to this, it didn't make sense to directly port what exists in Mapeo.

Open questions to address, maybe as a follow-up:

- [ ] what can/do we want to show in this screen? as currently implemented, i only have the button to import the config and an alert depending on the result of the import.

  - would it be helpful to surface an import history? maybe persist a history of the name of the file and the timestamp for when the import was attempted. how would we persist this history? (e.g. local storage, or in the actual backend)
  
- [ ] how should we surface errors and success? the ticket provides no information about this and i couldn't find the design for this screen in Figma. right now I'm just using alerts, which is what Mapeo more or less did.

---

Preview:

- Screen: 

  image

- Success flow:

  https://github.com/digidem/comapeo-mobile/assets/18542095/df117de0-ca0c-4929-ae85-0e1cb5fdfc4e

- Failure flow (happens if bad file type, or some error on backend):

  https://github.com/digidem/comapeo-mobile/assets/18542095/d4b4aed2-e7fc-4ec7-bdb1-e588f9bd8d84

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -765,6 +765,30 @@
   "screens.Settings.CreateOrJoinProject.whatIsAProject": {
     "message": "What is a Project"
   },
+  "screens.Settings.ProjectSettings.ProjectConfiguration.alertOkButton": {
+    "description": "Button to dismiss alert after attempting to import a config file",
+    "message": "OK"
+  },
+  "screens.Settings.ProjectSettings.ProjectConfiguration.configErrorTitle": {
+    "description": "Title of error dialog when there is an error importing a config file",
+    "message": "Import Error"
+  },
+  "screens.Settings.ProjectSettings.ProjectConfiguration.configImportAlertSuccessMessage": {
+    "description": "Message for alert after successful config import",
+    "message": "Successfully imported config"
+  },
+  "screens.Settings.ProjectSettings.ProjectConfiguration.configImportErrorMessage": {
+    "description": "Description of error dialog when there is an error importing a config file",
+    "message": "There was an error trying to import this config file"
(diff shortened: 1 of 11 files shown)


achou11:
> I haven't reviewed the code, but I think we need to bring this to codesign. This is one of the blindspots where we thought we could just port it over from mapeo mobile.
> 
> In the mean time, I think we should get rid of that page, and just add the flow for [importing a config when creating a project](https://www.figma.com/design/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?node-id=1072-4940&t=NQZ40AU46PvTaLEC-4).
> 
> Also, we can handle errors with the < ErrorBottomSheet/ >
> 
> As far as the success state, i think in this case of creating a project, we need to tie the api call with create a project...So clicking the create project button will handle both importing the config and creating the project, and the success state of that is already handled

yeah makes sense. might just make a new PR in that case 😄

ErikSin:
Im going to close this as your redoing the work on #161

PR #379, opened 2024-05-23
feat: add config importing to project creation flow

achou11:
Related to #161 

Would say that this is maybe 80% done. Didn't want to go further until answering some of the following questions:

- ~what's the success state after selecting the file? so far, i just plainly show the name of the file below the import button~ will be addressed later
- ~what's the error state if the file selected is invalid? should we use the error modal that's already part of this screen?~ uses existing error bottom sheet modal
- ~should there be a way for the user to remove the imported file? (e.g. they decide they don't want to use an imported file)~ will be addressed later

---

Preview:

After successful import:

image


ErikSin:
> * what's the success state after selecting the file? so far, i just plainly show the name of the file below the import button

Aldo is going to take this to the codesign team. But we can leave it as is for now

ErikSin:
> what's the error state if the file selected is invalid? should we use the error modal that's already part of this screen?

Yes, lets just use the error modal that is there

ErikSin (approved):
Tested (with the default config-but i took out the default config being added to mapeo manager by default) and seems to be working!

PR #385, opened 2024-05-28
feat: deleting photos in preview screen

### LINKED ISSUE #230: Show and Delete Photos when editting an observation
When editing or creating an observation, user should be able to click on photos, which opens a modal and allows them to delete the photo from an observation. This photos modal can be ported over from mapeo mobile. 

https://github.com/digidem/comapeo-mobile/assets/67773827/0490c4c3-ea76-452d-b581-19c562661529

## To Do
- [x] `onPress` of a photo thumbnail should open photos modals
- [x] Create a PhotosModal. This can be ported over from [MapeoMobile](https://github.com/digidem/mapeo-mobile/tree/develop/src/frontend/screens/PhotosModal). Note that this photos modal is also being used for ticket #229. When used by this component it should show a header button asking the user if they would like to delete the photo
- [x] User is able toggle hiding the header in the photos modal by clicking anywhere on the screen while the photos modal is open
- [x] `onPress` of "Delete Image" should open a bottom sheet, confirming the user would like to delete photo
- [x] User is able to delete photo

## To note
- the UI of the button is changed from mapeo-mobile (it is now rounded instead of square): use [Shared Button Component](https://github.com/digidem/comapeo-mobile/blob/626e116dfa6ee7fbfab5c0f77e38ab01c70ec174/src/frontend/sharedComponents/Button.tsx)
- this photos modal is the same photos modal used in #229 
- In #229 the source of the photo is different then the source of the photo used for this ticket.

CDFN:
Closes #230 

This PR adds feature enabling people to delete photo attachments from observation draft. 

Also corrected parameter name in draft observation store, as it was `id` previously which actually was original uri of a photo. 

By accident I did little more than scope of the issue, as photos in already created observations are possible to delete too (including deleting them from attachments on backend). However everything that was unasked for is hidden, to enable deleting photos in already created observations just pass `deletable: true` when navigating to `PhotoPreviewModal` screen in `Observation` screen. Worth nothing there is no permission check, as currently `deviceId` everywhere is mocked, so this would need to be added too before enabling deleting in existing observations.


src/frontend/screens/Observation/PhotoPreviewModal.tsx
```diff
@@ -27,17 +71,75 @@ export const PhotoPreviewModal: FC<
...
+  deleteButtonWrapper: {
+    borderStyle: 'solid',
+    borderWidth: 1,
+    borderColor: WHITE,
+    borderRadius: 3,
+    flexDirection: 'row',
+    justifyContent: 'center',
+    alignItems: 'center',
+    paddingHorizontal: 4,
+    paddingVertical: 6,
```
ErikSin, line 138: Can you make the button more similar to the button component. Rounded edges mainly

src/frontend/screens/Observation/PhotoPreviewModal.tsx
```diff
@@ -27,17 +71,75 @@ export const PhotoPreviewModal: FC<
           onPress={() => navigation.goBack()}
         />
       ),
+      // eslint-disable-next-line react/no-unstable-nested-components -- it's correct syntax
+      headerRight: () =>
+        deletable ? (
+          <Pressable onPress={openSheet} style={styles.deleteButtonWrapper}>
```
ErikSin, line 77: Can you Make this a TouchableOpacity instead

src/frontend/screens/Observation/PhotoPreviewModal.tsx
ErikSin, line 1: this file should not be in the `Observation` folder. It should be its own file in `Screens`

src/frontend/screens/Observation/PhotoPreviewModal.tsx
```diff
@@ -27,17 +71,75 @@ export const PhotoPreviewModal: FC<
...
-      />
+      {observationId && attachmentId ? (
+        <PhotoUnpreparedView
+          onPress={() => setShowHeader(!showHeader)}
+          observationId={observationId}
+          attachmentId={attachmentId}
+          variant={'original'}
+        />
+      ) : originalPhotoUri ? (
+        <PhotoPreparedView photoUri={originalPhotoUri} />
```
ErikSin, line 100: This should also have  `onPress={() => setShowHeader(!showHeader)}`

ErikSin (approved):
This still has that flickering problem in #383.  But otherwise looks good.

So i am approving, but don't merge until that flickering has been fixed in 383 and this is rebased off that

PR #387, opened 2024-05-28
chore: refactor location watching

### LINKED ISSUE #295: adjust implementation for location watching setup
The implementation of location watching at the time of writing lives in the [`useLocation` hook](https://github.com/digidem/comapeo-mobile/blob/ec280d48bcb158cbe8ac16915ba03c11d2014ac5/src/frontend/hooks/useLocation.ts). As surfaced by the [tracks work](https://github.com/digidem/comapeo-mobile/blob/ec280d48bcb158cbe8ac16915ba03c11d2014ac5/src/frontend/hooks/useLocation.ts), one of the limitations in the implementation is that _each usage of the hook manages its own internal state and creates a separate subscription for watching position changes_. Due to this, there were some parts of the app that one would expect to have the same location information showing at a specific moment but didn't. Each had their own loading state and the time difference in reaching the same state was notable (depending on when initial rendering occurred), making the visual experience feel disjointed.

To solve this issue, the ideal solution is to have one source of truth that manages a _single_ subscription and is responsible for providing the same state to any subscribers. We have something similar for the hooks that use `useSyncExternalStore()` (e.g. sync state and local discovery state). Ideally we can follow a similar pattern but...

The main trickiness in the case of location updates is that depending on what screen/component is subscribed to updates, we want to be able to _limit the update frequency_ if desired based on some parameters that have changed since the last accepted location update (thinking in terms of React re-renders, basically). This could be desirable for a couple of reasons:

- a re-render could be costly in resource intensive UIs such as a map. 
- getting frequent updates may not be essential for UX reasons (e.g. the location marker on a map screen may not need to update every time you move meter, whereas the GPS info screen probably should be as up-to-date as possible).

In the immediate, not entirely sure what the best implementation approach would be, but my guess is that it would be ~easier~ cleaner to do with `useSyncExternalStore()` than it would be with Zustand, as I'm not not really sure how Zustand plays with subscriptions handling or internal state in general (might be possible if using a closure works when calling `create()` 🤔 )

---

I've explored a couple of approaches for implementing what I described above. I haven't fully tested them but at least based on my understanding of React, they seem reasonably sound.

### `LocationStore`

Manages the subscription created by `ExpoLocation.watchPositionAsync()` and holds the position state that's provided to any subscribers. Although `watchPositionAsync()` accepts a limited number options for adjusting the update frequency, it is not a goal here to control this at the native subscription level (although it would be ideal). In other words, we configure the subscription to fire off as frequently as the native implementation is able to. Filtering happens at the consumer level, as we'll see below in the hook implementation.

<details>
<summary>General implementation</summary>

<br />

```ts
class LocationStore {
  #locationSubscription: LocationSubscription | null = null;
  #subscribers = new Set<() => void>();
  #state: LocationState = {
    location: null,
    error: null,
  };

  /**
   * Initialize position watching. Permissions handling should be done before calling this.
   */
  init = async () => {
    if (this.#locationSubscription) return;

    try {
      this.#locationSubscription = await watchPositionAsync(
        {accuracy: Accuracy.BestForNavigation},
        location => {
          this.#state = {error: null, location};

          for (const s of this.#subscribers) {
            s();
          }
        },
      );
    } catch (err) {
      if (err instanceof Error) {
        this.#state = {location: this.#state.location, error: err};
      }

      for (const s of this.#subscribers) {
        s();
      }
    }
  };

  getSnapshot = (): LocationState => {
    return this.#state;
  };

  subscribe = (cb: () => void) => {
    this.#subscribers.add(cb);

    if (!this.#locationSubscription) {
      this.init().catch(err => {
        console.log(err);
      });
    }

    return () => {
      this.#subscribers.delete(cb);
      if (this.#subscribers.size === 0) {
        this.#locationSubscription?.remove();
        this.#locationSubscription = null;
      }
    };
  };
}
```
</details>

### The `useLocation()` hook (v1)

At first, it wasn't clear to me how to implement this using `useSyncExternalStore()` because of the filtering requirement. Basically the challenge is making sure we have access to the previous "accepted" state so that the filtering options provided by the hook consumer can be acknowledged. How to implement this was not immediately apparent when first starting, but then I figured that maybe plain old `useState()` could work well enough! With that, I implemented a version of this refactored hook that uses `useState()` and `useEffect()`:

<details>
<summary>General implementation</summary>

<br />

```ts
export function useLocation(
  shouldLocationUpdate: (changes: ParameterChanges) => boolean = () => true,
) {
  const [state, setState] = useState(locationStore.getSnapshot());

  useEffect(() => {
    const unsubscribe = locationStore.subscribe(() => {
      setState(prevState => {
        const newState = locationStore.getSnapshot();

        // 1. If `error` field value is different, apply new state
        if (prevState.error !== newState.error) return newState;

        // 2. If `location` field is non-existent for previous or new state, apply new state
        // Former can technically happen on initialization (although unlikely)
        // Latter should not ever happen (but to make TypeScript happy)
        if (!(prevState.location && newState.location)) return newState;

        const distanceChange = getDistance(
          prevState.location,
          newState.location,
        );

        const timeElapsed =
          newState.location.timestamp - prevState.location.timestamp;

        // 3. Provide the necessary information to let hook consumer decide if new state should be used
        return shouldLocationUpdate({
          distance: distanceChange,
          time: timeElapsed,
        })
          ? newState
          : prevState;
      });
    });

    return () => {
      unsubscribe();
    };
  }, [shouldLocationUpdate]);

  return state;
}
```
</details>

It feels a bit "old school" in terms of managing the hook subscription via `useEffect()`, but using`useState()` and its functional update parameter made it clear in terms of how to fulfill the filtering requirement. The hook accepts a `shouldLocationUpdate` parameter, which is a callback that is provided the changes to parameters that are relevant for our filtering requirements (distance and time), and returns a boolean to indicate that the changes fit the filtering requirements of the hook consumer. This provides flexibility for the consumer to decide how the filtering is done based on the context of usage. If omitted, it's equivalent to saying "whenever the location subscription provides a new value, I want to use that new value". If provided, one could decide "I only want to acknowledge new updates if the distance traveled is greater than X meters", as an example.

For the sake of convenience, we can define some custom `shouldLocationUpdate` functions to handle our specific use cases more easily. For example, we can create a distance-based filter, where usage could look like this:

```ts
import {createDistanceFilter, useLocation} from 'hooks/...' // wherever this hook is implemented

// It's useful to create this outside the component because of function stability and how that plays into React render cycle
// Similar to Zustand's limitation around selectors
const atLeastFifteenMetersFilter = createDistanceFilter({
  minimumDistance: 15,
})

function MapScreen() {
  // This state value now only updates when change in distance is at least 15 meters
  const locationState = useLocation(atLeastFifeenMetersFilter)
  
  // ...
}
```

Once I had this implemented, I had a feeling that this _should_ be feasible to do with `useSyncExternalStore()`, since what i have implemented here is kind of a rougher version of what it does anyways. Another attempt led me to...

### `useLocation()` (v2)

The implementation using `useSyncExternalStore()` is pretty close to the v1 implementation, but relies on `useRef` to create a reference for the previous accepted state value, which is then used for comparison when new updates are emitted by the location subscription. Just need to make that you update it the ref whenever you decide to return the new state value instead of preserving the old one.

<details>
<summary>General implementation</summary>

<br />

```ts
export function useLocation(
  shouldLocationUpdate: (changes: ParameterChanges) => boolean = () => true,
) {
  const prevState = useRef(locationStore.getSnapshot());

  const getSnapshot = useCallback(() => {
    const newState = locationStore.getSnapshot();

    // 1. If `error` field value is different, apply new state
    if (prevState.current.error !== newState.error) {
      prevState.current = newState;
      return newState;
    }

    // 2. If `location` field is non-existent for previous or new state, apply new state
    // Former can technically happen on initialization (although unlikely)
    // Latter should not ever happen (but to make TypeScript happy)
    if (!(prevState.current.location && newState.location)) {
      prevState.current = newState;
      return newState;
    }

    const distanceChange = getDistance(
      prevState.current.location,
      newState.location,
    );

    const timeElapsed =
      newState.location.timestamp - prevState.current.location.timestamp;

    // 3. Provide the necessary information to let hook consumer decide if new state should be used
    if (
      shouldLocationUpdate({
        distance: distanceChange,
        time: timeElapsed,
      })
    ) {
      prevState.current = newState;
      return newState;
    } else {
      return prevState.current;
    }
  }, [shouldLocationUpdate]);

  return useSyncExternalStore(locationStore.subscribe, getSnapshot);
}
```
</details>

---

To reiterate, I have NOT tested the above implementations yet but I'm hoping they're closer to what I want our location subscription setup to look like. Being able to have a single subscription and preserve the flexibility for restricted React updates at the component-level is the goal for me. Open to thoughts or suggestions!

achou11:
Closes #295 

TODO:

- [x] handle edge cases related to permissions 
- [x] add `accuracy` field to `ParameterChanges` type


ErikSin:
On review of `useLocationProviderStatus`, we used it EVERY TIME we used `useLocation` The idea was that we did not want the user to see a stale location. We were manually checking  whether gps was enabled. And if it was not enabled, we would conditionally render the UI to show an undefined value, and if it was enabled we were conditionally rendering the location value. So I just coupled the 2 into one hook. Now when the gps is not enabled, the `useLocation` hook simply returns `location===null`, no need to check every time before returning the location value.

In the future we may want to provide the location, even it the gps is not enabled. But I think it would be better to add that as an option to this hook as there would be less maintenance overall.

src/frontend/contexts/LocationStoreContext.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  checkProviderStatus() {
+    getProviderStatusAsync().then(status => {
+      if (!status.locationServicesEnabled && this.#state.location !== null) {
+        this.#state.location = null;
+        for (const s of this.#subscribers) {
+          s();
+        }
+      }
+    });
```
ErikSin, line 98: we only want to update the react state of location services is disabled (not when it is enabled). When it is enabled, the location should update, and the location updating will update the react state.

src/frontend/hooks/useLocation.ts
```diff
@@ -118,3 +142,24 @@ export function getCoords(location: LocationObject): [number, number] {
...
+  newAccuracy: number | null,
+  distance: number,
+): Accuracy {
+  if (previousAccuracy === null) {
+    return newAccuracy === null ? 'unchanged' : 'better';
+  }
+
+  if (distance > previousAccuracy) {
+    return 'newBounds';
+  }
```
ErikSin, line 157: It could be the case that the user has travelled a large distance, but the accuracy has not gotten better. If the user is asking for the most "accurate" location, we would return the value with the lowest accuracy radius (which in this case is not actually the most accurate). This check will allow the user to know if they have travelled outside of the original accuracy bounds.

src/frontend/hooks/useLocation.ts
```diff
@@ -118,3 +142,24 @@ export function getCoords(location: LocationObject): [number, number] {
   const {longitude, latitude} = location.coords;
   return [longitude, latitude];
 }
+
+export function getAccuracy(
```
gmaclennan, line 146: What is this being used for? If only for setting the location on an observation, then I think it might be better to keep this logic along with updating the observation location. It feels fragile to rely on hook updates to decide to update something. If using this in other places, I think it might lead to unexpected behaviour for users. Even if accuracy is worse, and the distance moved less than the accuracy, then a user would expect to see their location move.

src/frontend/hooks/useLocation.ts
```diff
@@ -1,112 +1,136 @@
...
+ */
+export function useLocation(
+  shouldAcceptUpdate: ShouldAcceptUpdateCheck = DEFAULT_SHOULD_ACCEPT_UPDATE,
+) {
+  const locationStore = useLocationStore();
+  const prevState = useRef<LocationState | undefined>(
+    locationStore.getSnapshot(),
+  );
+
+  const getSnapshot = useCallback(() => {
```
gmaclennan, line 58: Logic here looks sound. Note that you could have also kept the existing debounce to avoid the arguably more confusing `shouldAcceptUpdate` function e.g.

```
const getSnapshot = useCallback(() => {
  return debounceLocation(options)(location => location)
}, Object.values(options))
```

Although would need new logic for minDistanceInterval, and could change the signature so you can just `return debounceLocation(options)`.

src/frontend/lib/utils.ts
```diff
@@ -54,23 +54,20 @@ const GOOD_PRECISION = 10; // 10 meters
...
-  if (!gpsAvailable || !locationServicesEnabled) return 'error';
+  if (!locationState?.location || locationState.error) return 'error';
 
   const positionStale =
-    location && Date.now() - location.timestamp > STALE_TIMEOUT;
+    locationState &&
+    Date.now() - locationState.location.timestamp > STALE_TIMEOUT;
 
-  if (positionStale) return 'searching';
+  if (positionStale || !locationState.location.coords.accuracy)
```
gmaclennan, line 67: nit-pick: accuracy being `null` is not an indication that the GPS is searching, it is `null` if the accuracy is unavailable (which is only on web I think, so hence this is only a nit-pick)

src/frontend/screens/LocationInfoScreen.tsx
```diff
@@ -74,7 +74,7 @@ const InfoRow = ({label, value}: {label: string; value: string}) => (
 );
 
 export const LocationInfoScreen = () => {
-  const {location} = useLocation({maxDistanceInterval: 0});
+  const {location} = useLocation();
```
gmaclennan, line 77: Can't leave a comment below, but on line #94 I don't think we should display the current time if we don't have a location - it suggests to the user that there is a location

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -32,18 +34,23 @@ const DEFAULT_ZOOM = 12;
 Mapbox.setAccessToken(config.mapboxAccessToken);
 const MIN_DISPLACEMENT = 3;
 
+function locationFilter({distance, time}: ParameterChanges) {
+  return time >= 1000 || distance >= 15;
```
gmaclennan, line 38: Note that this logic is different from what it was previously. The previous logic written for this function would be:

```
const MIN_TIME_INTERVAL = 200
const MAX_TIME_INTERVAL = 1000
const MIN_DISTANCE_INTERVAL = 1
const MAX_DISTANCE_INTERVAL = 15

// ...

return time > MIN_TIME_INTERVAL && (
  distance > MAX_DISTANCE_INTERVAL ||
  (time > MAX_TIME_INTERVAL && distance > MIN_DISTANCE_INTERVAL)
)
```

src/frontend/screens/MapScreen/track/TrackPathLayer.tsx
```diff
@@ -3,13 +3,13 @@ import {usePersistedTrack} from '../../../hooks/persistedState/usePersistedTrack
...
 import {LineString} from 'geojson';
-import {useLocation} from '../../../hooks/useLocation';
 import {LocationHistoryPoint} from '../../../sharedTypes/location';
+import {useLocation} from '../../../hooks/useLocation';
 
 export const TrackPathLayer = () => {
   const locationHistory = usePersistedTrack(state => state.locationHistory);
   const isTracking = usePersistedTrack(state => state.isTracking);
-  const {location} = useLocation({maxDistanceInterval: 3});
+  const {location} = useLocation();
```
gmaclennan, line 12: You will need some kind of filter on here or this is going to be a battery hog with very frequent map updates (location updates can come in at 60 times a second I think on some phones)

src/frontend/screens/MapScreen/track/UserTooltipMarker.tsx
```diff
@@ -3,11 +3,11 @@ import {StyleSheet, Text, View} from 'react-native';
 import React from 'react';
 import {usePersistedTrack} from '../../../hooks/persistedState/usePersistedTrack';
 import {useTrackTimerContext} from '../../../contexts/TrackTimerContext';
-import {useSharedLocationContext} from '../../../contexts/SharedLocationContext';
+import {useLocation} from '../../../hooks/useLocation';
 
 export const UserTooltipMarker = () => {
   const {timer} = useTrackTimerContext();
-  const {locationState} = useSharedLocationContext();
+  const locationState = useLocation();
```
gmaclennan, line 10: This needs a filter to avoid too frequent map-redraws.

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -19,66 +19,95 @@ const m = defineMessages({
...
-        accuracy: liveLocation?.coords?.accuracy,
-      }
-    : {
-        lat: liveLocation?.coords?.latitude,
-        lon: liveLocation?.coords?.longitude,
-        accuracy: liveLocation?.coords?.accuracy,
-      };
+    updateObservationPosition({
+      position: {
+        mocked: false,
```
gmaclennan, line 59: The value of `mocked` should come from the location object.

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -19,66 +19,95 @@ const m = defineMessages({
...
+  //only update the location if the accuracy increases OR the user has moved outside of the accuracy's radius of uncertainty.
+  const {location} = useLocation(
+    ({accuracy}) => accuracy === 'better' || accuracy === 'newBounds',
+  );
+  const {updateObservationPosition} = useDraftObservation();
+
+  useEffect(() => {
+    const newCoord = !location
+      ? undefined
+      : Object.entries(location.coords).map(
```
gmaclennan, line 53: Why are we converting the coordinates into entries and then back again?

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -19,66 +19,95 @@ const m = defineMessages({
...
-        lat: observationValue.lat,
-        lon: observationValue.lon,
-        accuracy: liveLocation?.coords?.accuracy,
-      }
-    : {
-        lat: liveLocation?.coords?.latitude,
-        lon: liveLocation?.coords?.longitude,
-        accuracy: liveLocation?.coords?.accuracy,
-      };
+    updateObservationPosition({
```
gmaclennan, line 57: I think updating the observation location just based on whether there has been a new location update, and relying on renders from the hook, seems possibly flakey, and for a future maintainer reading this there is some indirection to find the logic. Consider moving the logic here, and comparing with the accuracy of the location saved to the observation, rather than the accuracy change in the location state.

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -19,66 +19,95 @@ const m = defineMessages({
...
+      position: {
```
gmaclennan, line 58: Previously (Mapeo legacy) we had been saving all the location information from location state to the observation. Is there a reason to stop saving this info?

src/frontend/screens/ObservationEdit/LocationView.tsx
```diff
@@ -19,66 +19,95 @@ const m = defineMessages({
...
+    <LocationViewInner
+      lat={observationValue?.lat}
+      lon={observationValue?.lon}
+      accuracy={null}
+    />
   );
-  const coordinateFormat = usePersistedSettings(coordinateFormatSelector);
+};
+
+const LocationViewUpdatePosition = () => {
```
gmaclennan, line 42: I haven't read all the surrounding code, but it seems like we are replicating the bug from Mapeo legacy where observation position will update during edits as well as on initial creation? Also it seems future-confusing to have the updates happening within this location view component - it requires someone to remember that if a future UX change decided to not display this component then it would stop the observation from having a saved location. It also means that the observation location is not updated when the user is choosing a category, filling in details, or taking photos.

gmaclennan (commented):
Phew, lots in here! There a quite a few places to review and update - this is a tricky part of the app and it's important to get it right.

gmaclennan (changes requested):
As a more general higher-level suggestion, I don't think the location provider should be setting the location to `null` based on the provider - I think the consumer component should decide what to display with the location based on that state.

src/frontend/contexts/LocationStoreContext.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+const LocationStoreContext = createContext<LocationStore | null>(null);
+
+// TODO: Return undefined when provider status changes to unavailable
+export function LocationStoreProvider({children}: {children: React.ReactNode}) {
+  const [store] = useState(() => new LocationStore());
+
+  const [locationPermissions] = useForegroundPermissions();
+
+  if (locationPermissions?.granted) {
```
gmaclennan, line 19: Suggestion: if you move this state into the LocationStore, then you can remove the restriction about only calling `init` after checking permissions, e.g. `store.setPermissions(locationPermissions)` - then the store can handle what it needs to do based on the permissions, and we can include the location permissions state in the LocationStore state, so it's clearer why the location might be null.

src/frontend/hooks/useLocation.ts
ErikSin: In mapeo mobile, when making an observation we only updated the location when the accuracy got better. This is just a selector (it doesn't need to be exported). So this implementation allows the user to decide whether or not they want to account for accuracy when updating

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: it because expo location returns `{value} | null` and our backend expects `{value} | undefined`

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: A problem with that is the hook will cause a new re-render on every change, even if we are not updating the UI. Because we would not be filtering the accuracy, it would received a new state value on every accuracy change but we would still keep the UI the same if the accuracy didn't increase.

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: Currently the location View is only mounted if there no `observationId` (aka observations that have yet to be saved to the backend). That avoids the observation being updated if the observation is being edited (as opposed to being created).

In mapeo mobile the location did not update when choosing a category, filling in details, or taking photos. Ir has a similar thing where the `LocationField` updated the location, and that was only [mounted when the observation edit screen was mounted](https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/ObservationEditView.js#L190).

ErikSin:
> As a more general higher-level suggestion, I don't think the location provider should be setting the location to `null` based on the provider - I think the consumer component should decide what to display with the location based on that state.

let talk about this tomorrow. My reasoning behind this was that every time we used the location hook we also checked for the provider status. There was never a time we didn't do that. It felt like an unnecessary maintenance burden to have 2 hooks even though they were never being used separately. I thought a good compromise was to create a third hook that combine the 2 hooks. So if we want location to be tied to the provider status (which is currently at all time) we can just use the one hook. But the other 2 hooks are still available if we ever need them separately

ErikSin:
I resolved a couple comments related to having the 2 subscription in one class. I separated them into 2 implementations of `useSyncExternalStore`

src/frontend/contexts/LocationStoreContext.tsx
```diff
@@ -0,0 +1,125 @@
...
+  #subscribers = new Set<() => void>();
+  #state: LocationState = {
+    location: null,
+    error: null,
+  };
+
+  /**
+   * Initialize position watching. Permissions handling should be done before calling this.
+   */
+  init = async () => {
```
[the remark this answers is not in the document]
achou11: no specific reason, but vaguely remember running into `this` issues when doing `async init()`, but maybe that was for some other external sync store work 😄

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387/commits/e3d3d89dd25f39826a29ad249b5bb9b6481388a2

src/frontend/screens/LocationInfoScreen.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387/commits/9b28ba6ce7068d30088de5ff991e378f7a915066

src/frontend/screens/MapScreen/track/TrackPathLayer.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387/commits/fbedb98d066c2fdf3a3b4bd4f6fcade8d1a9a834 I changed the default (i chose completely arbitrary values, but it should throttle it now)

src/frontend/screens/MapScreen/track/UserTooltipMarker.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387#discussion_r1630203254

src/frontend/hooks/useLocation.ts
ErikSin: Will be addressed in [this](https://github.com/digidem/comapeo-mobile/issues/402)

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: Will be addressed in [this](https://github.com/digidem/comapeo-mobile/issues/402)

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: Will be addressed in [this](https://github.com/digidem/comapeo-mobile/issues/402)

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: Will be addressed in [this](https://github.com/digidem/comapeo-mobile/issues/402)

src/frontend/screens/ObservationEdit/LocationView.tsx
ErikSin: Will be addressed in [this](https://github.com/digidem/comapeo-mobile/issues/402)

src/frontend/contexts/LocationStoreContext.tsx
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387/commits/19659ea6e5fb363c6f47430daf1978eb107bc491

src/frontend/lib/utils.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/387/commits/0a7a4e5551fc0876d50c6d2675cb82c274c702b7

src/frontend/hooks/useLocationWithProviderStatus.ts
achou11, line 1: this hook seems to be unnecessary and contrary to Gregor's reasoning for having the provider status be separated from the location state. Would suggest having the consumers of this hook to use `useLocationProviderStatus` and `useLocation` directly and have the logic live at the usage sites.

src/frontend/lib/utils.ts
```diff
@@ -320,3 +320,22 @@ export function isEditingScreen(
...
+  )
+    return false;
+  for (const key in a) {
+    if (!(key in b) || a[key] !== b[key]) return false;
+  }
+  for (const key in b) {
+    if (!(key in a)) return false;
+  }
+  return true;
+}
```
achou11, line 341: is this from some existing implementation? if not, would lean towards using a well-tested module instead

src/frontend/lib/utils.ts
```diff
@@ -320,3 +320,22 @@ export function isEditingScreen(
 
   return !!EDITING_SCREEN_NAMES.find(val => val === routeName);
 }
+
+export function shallowEqual(a: any, b: any) {
+  console.log('in function');
```
achou11, line 325: remove

src/frontend/screens/MapScreen/index.tsx
achou11: @ErikSin - should implement this as suggested above

src/frontend/contexts/LocationStoreContext.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  /**
+   * Initialize position watching. Permissions handling should be done before calling this.
+   */
+  async init() {
+    try {
+      if (this.#locationSubscription) return;
+      this.#locationSubscription = await watchPositionAsync(
+        {accuracy: Accuracy.BestForNavigation},
```
achou11, line 61: based on the call we had, should update this so that it fires only once per second at most

```suggestion
        {accuracy: Accuracy.BestForNavigation, timeInterval: 1000},
```

achou11 (commented):
Can't leave a proper review, but left a few blocking comments but feeling like it's pretty close! Didn't get a chance to test the provider status changes you introduced though.

Also, would be helpful to update this with main branch

achou11:
Closing this as we've decided to address this post-MVP. Still useful to have as a reference for whenever we return it

PR #390, opened 2024-05-29
fix: update extracted messages

achou11:
Looks like the messages were a bit stale so ran extract-messages script again to update


achou11:
Closing since this has been resolved with recent merges

PR #396, opened 2024-06-04
feat: send metrics reports to server once a day

### LINKED ISSUE #332: Send metrics report to server

EvanHahn:
Once a day, a metrics report will be generated and sent to our metrics server. These are governed by two new environment variables, documented in the README.

Most of the magic happens in the `useSendMetrics` hook.

There were a few parts of this work that were kinda unrelated:

- I added several small helper functions (`assert`, `sleep`, and `throwIfAborted`).
- `@rnmapbox/maps` has a bug where its Jest helper breaks `AbortController`, which we use here. I had to patch the module, and I [submitted the patch upstream][0]. That patch was approved but not yet deployed.

Closes #332.

[0]: https://github.com/rnmapbox/maps/pull/3506


src/frontend/metrics/sendMetricsReport.ts
```diff
@@ -0,0 +1,53 @@
...
+  metricsApiKey,
+  metricsReport,
+  signal,
+}: ReadonlyDeep<{
+  metricsUrl: string;
+  metricsApiKey: string;
+  metricsReport: Jsonifiable;
+  signal: AbortSignal;
+}>): Promise<void> {
+  const body = new Blob([JSON.stringify({data: metricsReport})]);
```
ErikSin, line 33: Is this working for you. I'm pretty sure you can't just use the browser Blob Api in RN and you will need to export an external library.

src/frontend/hooks/useSendMetrics.ts
```diff
@@ -0,0 +1,129 @@
...
+        await sendMetricsReport({
+          metricsUrl,
+          metricsApiKey,
+          metricsReport,
+          signal,
+        });
+      }
+
+      setLastMetricsReportSentAt(Date.now());
+    })().catch(noop);
```
ErikSin, line 123: Instead of `noop', we should pass the errors to Sentry (at least the server errors, but probably also the `getenv` errors)

ErikSin (changes requested):
2 main comments:

1. We should handle the errors with Sentry
2. I think you can use the same `blob` api as the browser. You will need to install a external library to properly handle a blob (or not use a blob at all)

src/frontend/lib/throwIfAborted.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Ponyfill of `AbortSignal.prototype.throwIfAborted`, which is not supported
```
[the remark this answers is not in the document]
EvanHahn: [Ponyfill!](https://github.com/sindresorhus/ponyfill)

src/frontend/hooks/useSendMetrics.ts
```diff
@@ -0,0 +1,129 @@
...
+  const projectSettings = await api.listProjects();
+  return (
+    await Promise.all(
+      projectSettings.map(async ({projectId}) => {
+        const project = await api.getProject(projectId);
+        return project.observation.getMany();
+      }),
+    )
+  ).flat();
+};
```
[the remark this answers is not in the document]
EvanHahn: I did this because `useObservations` only works with the active project. Maybe that's fine because we only support one project right now, but that could be an issue later on when we add multi-project. Maybe I could add a hook like `useObservationsFromAllProjects`?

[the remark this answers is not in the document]
ErikSin: ohhh, I see. No, I don't think using a react-query hook makes sense, we don't need to cache all the observations. I think this implementation makes sense

src/frontend/hooks/useSendMetrics.ts
```diff
@@ -0,0 +1,129 @@
...
+> = set => ({
+  lastMetricsReportSentAt: null,
+  setLastMetricsReportSentAt: lastMetricsReportSentAt =>
+    set({lastMetricsReportSentAt}),
+});
+
+const useLastMetricsReportSentAt = createPersistedState(
+  lastMetricsReportSentAtSlice,
+  'LastMetricsReportSentAt',
+);
```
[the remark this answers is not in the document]
EvanHahn: Done in 3b744f43151ba951847f5c43475833ccb57d6031.

src/frontend/hooks/useSendMetrics.ts
EvanHahn: Done in e7b3f922d1267ef2c6f7b2b32a5a053c711d6a8c.

src/frontend/metrics/sendMetricsReport.ts
EvanHahn: It was working for me but I think it's cleaner without it anyway. Removed in 6a12ca145f2eb855fe289b2972047e0731aa0f7d.

src/frontend/hooks/useSendMetrics.ts
```diff
@@ -0,0 +1,114 @@
...
+          metricsApiKey,
+          metricsReport,
+          signal,
+        });
+      }
+
+      setLastMetricsReportSentAt(Date.now());
+    })().catch(err => {
+      if (err === EFFECT_CANCELED_REASON) return;
+      Sentry.captureException(err);
```
ErikSin, line 107: I think it would also be useful to catch and report any errors from `sendMetricsReport`

EvanHahn: This `catch` already does that (I think). Do you want additional reporting in `sendMetricsReport` itself, in case we call it from elsewhere?

ErikSin (approved):

PR #397, opened 2024-06-05
chore: tell `@mapeo/core` whether the app is in the foreground

EvanHahn:
We need to tell `@mapeo/core` whether the app is in the foreground so it can start/stop sync. This does that with a React hook.

See also: [the related `@mapeo/core` commit][0].

[0]: https://github.com/digidem/mapeo-core-next/commit/bb9b16cdf3a4eebfc1910d1c889df2b6d9afe0a5


src/frontend/useOnBackgroundedAndForegrounded.ts
ErikSin, line 1: I think this should do in the `src/frontend/hooks` folder

src/frontend/useOnBackgroundedAndForegrounded.ts
```diff
@@ -0,0 +1,16 @@
+import {type MapeoClientApi} from '@mapeo/ipc';
+import {useEffect} from 'react';
+import {useAppState} from '@react-native-community/hooks';
+
+export default function useOnBackgroundedAndForegrounded(
```
ErikSin, line 5: Also, we stay away from default exports in the front end (because we are civilized people...)

ErikSin (approved):
Just a comment about location and use of default exports

src/frontend/useOnBackgroundedAndForegrounded.ts
EvanHahn: 🔥 🔥 🔥

EvanHahn:
@achou11 Yes, it's blocked on #433.

PR #403, opened 2024-06-10
feat: generate report of network availability

### LINKED ISSUE #275: Add cell phone coverage info to metrics report
cellphone coverage could be grab from the `Metadata.networkAvailable` value in `Observation`. 
The idea would be, grab this value from every observation and generate a percentage as the metric

tomasciccola:
should close #275


src/frontend/metrics/getPercentageOfNetworkAvailability.ts
```diff
@@ -0,0 +1,10 @@
+import type {Observation} from '@mapeo/schema';
+
+export default function (observations: ReadonlyArray<Observation>) {
```
EvanHahn, line 3: 1. Could you change this to a named export? That's the code style for the rest of this repo, I learned. 
2. I could see moving this into `generateMetricsReport.ts`, as it's really only useful there. Feel free to ignore.

src/frontend/metrics/getPercentageOfNetworkAvailability.ts
```diff
@@ -0,0 +1,10 @@
...
+  const networkAvailability = observations
+    .map(obs => obs.metadata?.positionProvider?.networkAvailable)
+    .filter(networkAvailable => networkAvailable !== undefined);
+
+  const trues = networkAvailability.filter(Boolean);
+  return (trues.length / networkAvailability.length) * 100;
```
EvanHahn, line 9: This could result in a division by 0. Could we check for this case and return 0 or maybe `null`?

src/frontend/metrics/generateMetricsReport.ts
```diff
@@ -22,17 +24,18 @@ export default function generateMetricsReport({
...
     osVersion,
     screen: {width: screen.width, height: screen.height},
     ...(countries.size ? {countries: Array.from(countries)} : {}),
+    percentageOfNetworkAvailability:
+      getPercentageOfNetworkAvailability(observations),
   };
 }
 
-function addToSet<T>(set: Set<T>, toAdd: Iterable<T>): void {
+export function addToSet<T>(set: Set<T>, toAdd: Iterable<T>): void {
```
EvanHahn, line 39: Nitpick: IMO, this makes more sense as a separate utility. Could you move it to a separate file, or avoid using it in the test?

src/frontend/metrics/generateMetricsReport.test.ts
```diff
@@ -1,29 +1,32 @@
...
     packageJson,
     os: 'android',
     osVersion: 123,
     screen: {width: 12, height: 34, ignoredValue: 56},
-    observations: [
-      // Middle of the Atlantic
-      {lat: 10, lon: -33},
-      // Mexico City
-      {lat: 19.419914, lon: -99.088059},
-      // Machias Seal Island, disputed territory
```
EvanHahn, line 20: Does the mock data include any disputed territories? I think that's useful to test.

EvanHahn (approved):
LGTM other than some nitpicks.

src/frontend/metrics/generateMetricsReport.test.ts
tomasciccola: disputed in what sense? Politically? Everything is kinda random generated. But I can spread generated data with manual data (like the example here)

src/frontend/metrics/generateMetricsReport.ts
tomasciccola: yeah, I fixed it like this cause it seem weird that `positionToCountries` returns a `Set` instead of only one country. But seeing your `disputed territory` comment, now it makes sense...

src/frontend/metrics/generateMetricsReport.ts
```diff
@@ -22,17 +23,25 @@ export default function generateMetricsReport({
...
-
   return {
     type: 'metrics-v1',
     appVersion: packageJson.version,
     os,
     osVersion,
     screen: {width: screen.width, height: screen.height},
     ...(countries.size ? {countries: Array.from(countries)} : {}),
+    percentageOfNetworkAvailability:
+      getPercentageOfNetworkAvailability(observations),
```
tomasciccola, line 34: I'm having a type error here (gh doesn't show it). I won't paste it since its too long, but the gist of it is:
` is 'readonly' and cannot be assigned to the mutable type`

Any clues @EvanHahn ? (probably if you run it locally you'd have a better clue...

src/frontend/metrics/generateMetricsReport.test.ts
```diff
@@ -1,38 +1,46 @@
...
   const packageJson = readPackageJson();
+  const count = 10;
+  const n = Math.floor(Math.random() * count);
+  const observations: ReadonlyArray<Observation | Record<string, never>> = [
+    ...generate('observation', {count}).map((obs, idx) =>
+      // Manually add Machias Seal Island, disputed territory
+      idx === n ? {...obs, lat: 44.5, lon: -67.101111} : obs,
+    ),
+    {},
+  ];
```
EvanHahn, line 24: Would you mind adding Machias Seal Island to the end, rather than mixing it in?

```suggestion
  const observations: ReadonlyArray<Partial<Observation>> = [
    ...generate('observation', {count: 10}),
    // Machias Seal Island, disputed territory
    {lat: 44.5, lon: -67.101111},
    {},
  ];
```

src/frontend/metrics/generateMetricsReport.test.ts
```diff
@@ -1,38 +1,46 @@
...
+  const generatedCountries = new Set<string>();
+
+  for (const {lat, lon} of observations) {
+    if (typeof lat === 'number' && typeof lon === 'number') {
+      addToSet(generatedCountries, positionToCountries(lat, lon));
+    }
+  }
+
+  const generatedNetworkAvailabilityPercentage =
+    getPercentageOfNetworkAvailability(observations);
```
EvanHahn, line 34: I'm a little uneasy about using this in tests. If `getPercentageOfNetworkAvailability` is wrong, the test will still pass.

Would you mind adding tests for `getPercentageOfNetworkAvailability` specifically, that don't use mock data? I think that would make me more comfortable with this part of the test.

src/frontend/metrics/generateMetricsReport.ts
```diff
@@ -22,17 +23,25 @@ export default function generateMetricsReport({
...
   };
 }
 
-function addToSet<T>(set: Set<T>, toAdd: Iterable<T>): void {
-  for (const item of toAdd) set.add(item);
+export function getPercentageOfNetworkAvailability(
```
EvanHahn, line 38: Nitpick: could you mention that this is only exported for tests? Alternatively, moving it to its own file with tests would be great.

src/frontend/metrics/generateMetricsReport.test.ts
```diff
@@ -1,38 +1,46 @@
...
-      {lat: 12},
-      {lon: 34},
-    ],
+    observations,
   } as MetricsReportOptions;
 
-  it('can be serialized and deserialized as JSON', () => {
-    const report = generateMetricsReport(defaultOptions);
-    const actual = JSON.parse(JSON.stringify(report));
-    const expected = removeUndefinedEntries(report);
```
EvanHahn, line 32: `removeUndefinedEntries` can now be removed.

src/frontend/metrics/generateMetricsReport.ts
EvanHahn: `getPercentageOfNetworkAvailability`'s argument should change to:

```diff
 export function getPercentageOfNetworkAvailability(
-  observations: ReadonlyArray<Observation>,
+  observations: ReadonlyDeep<Array<Partial<Observation>>>,
 ): number {
```

EvanHahn (approved):
Again, LGTM other than some comments.

src/frontend/metrics/generateMetricsReport.ts
tomasciccola: I don't understand this. You mean, having a test file only for this function? Or, two files? (one with the implementation and a test file)

src/frontend/metrics/generateMetricsReport.test.ts
tomasciccola: Do you think this is also true with `generatedCountries`? I'm basically generating mock-data and matching tests with that

tomasciccola: I've created individual tests for networkAvailability without mock data. I think maybe thats enough?

EvanHahn (approved):
LGTM! Thanks for answering all of my comments.

As a followup, we should make sure we're actually setting this property.

tomasciccola:
> As a followup, we should make sure we're actually setting this property.

What do you mean by this? The `generateMetrics` will always return a object with `percentageOfNetworkAvailability` setted. If there's no stats it will return 0. Do you think this is misleading?

EvanHahn (approved):
Sorry, I was unclear. We aren't populating `observation.metadata.positionProvider.networkAvailable` right now, so `percentageOfNetworkAvailability` will _always_ be 0.

I think this is good to merge for now.

PR #404, opened 2024-06-10
fix: fix navigation error after responding to invite

### LINKED ISSUE #391: navigation error in ProjectInviteBottomSheet
When responding to a project invite that results in a navigation action, I get the following error:

```
 ERROR  Error: Couldn't find a navigation object. Is your component inside NavigationContainer?

This error is located at:
    in InviteSuccessBottomSheetContent (at ProjectInviteBottomSheet/index.tsx:81)
```

This is because the `ProjectInviteBottomSheet` is trying to access the react-navigation hooks, but it can only do so if it's a child of a _navigator_ (not just a navigation container). Right now, it lives as a child of the app navigation container, but as a sibling of the DrawerNavigator.

See https://github.com/digidem/comapeo-mobile/blob/7cda44ed6f099781c485898887f39e27e56d43c1/src/frontend/AppNavigator.tsx for an example of how to fix it.

achou11:
Fixes #391 

Also fixes an issue with determining the name of the current route. The `isEditingScreen()` util (now removed completely) did not properly extract the route name for screens within the tab container (e.g. going to camera or observations list). This would've caused unexpected behavior in some cases, such as the invite sheet showing up while in the camera screen (which is not desired).

### DIFF
--- a/src/frontend/AppNavigator.tsx
+++ b/src/frontend/AppNavigator.tsx
@@ -1,21 +1,57 @@
-import {NavigationContainer} from '@react-navigation/native';
+import {
+  NavigationContainer,
+  createNavigationContainerRef,
+} from '@react-navigation/native';
 import * as React from 'react';
 import * as SplashScreen from 'expo-splash-screen';
 import {DrawerNavigator} from './Navigation/Drawer';
 import {ProjectInviteBottomSheet} from './sharedComponents/ProjectInviteBottomSheet';
 import {Loading} from './sharedComponents/Loading';
+import {AppStackParamsList} from './sharedTypes/navigation';
+import {EDITING_SCREEN_NAMES} from './constants';
+
+export const rootNavigationRef =
+  createNavigationContainerRef<AppStackParamsList>();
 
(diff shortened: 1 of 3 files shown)


src/frontend/AppNavigator.tsx
```diff
@@ -1,21 +1,57 @@
...
 import * as React from 'react';
 import * as SplashScreen from 'expo-splash-screen';
 import {DrawerNavigator} from './Navigation/Drawer';
 import {ProjectInviteBottomSheet} from './sharedComponents/ProjectInviteBottomSheet';
 import {Loading} from './sharedComponents/Loading';
+import {AppStackParamsList} from './sharedTypes/navigation';
+import {EDITING_SCREEN_NAMES} from './constants';
+
+export const rootNavigationRef =
+  createNavigationContainerRef<AppStackParamsList>();
```
ErikSin, line 14: Does this need to be exported?

ErikSin (approved):
Looks good! thanks for fixing that

src/frontend/AppNavigator.tsx
achou11: ah no, not for our case. will fix

PR #415, opened 2024-06-18
fix: fix visual issues with shared content component used in bottom sheet modal

achou11:
changes to the original implementation from Mapeo Mobile (which did not have many of the issues) when porting it to here led to a few visual issues when trying to use the `BottomSheetContent` component:

- the "fullscreen" variant did not justify content properly. as a result, some usages of the bottom sheet modal implemented their own content, which becomes prone to visual inconsistencies.

the original intent of this content component was to unify the visual and functional behavior of the content inside of the bottom sheet modals, which is designed very intentionally.

this PR attempts to update screens where using the shared BottomSheetContext makes sense (which is most of them). the current exception is the `GPSPermissionsModal`, which is not using the shared BottomSheetModal component so would require some more work to convert (can be done as a follow-up).

also fixes an issue where we were using a context-based hook without the provider being set up (`react-native-safe-area`)

important implementation notes:

- ~both `BottomSheetModal` and `BottomSheetContent` now accept a `fullScreen` prop. this is a little unfortunate but was the simplest way to get the sizing behavior we need for various use cases, without coupling the content and the modal too closely. in practice, that means whenever you want a full screen bottom sheet modal, you'll have to remember to specify the prop for the bottom sheet content component as well e.g.~ EDIT: No longer the case, see https://github.com/digidem/comapeo-mobile/pull/415/commits/b8389d5548a685c2b91f0e5267231d00defca184

    ```tsx
    <BottomSheetModal fullScreen ...>
      <BottomSheetContent fullScreen .../>
    </BottomSheetModal>
    ```

- the minimum sheet height is set to 400. when not full screen, maximum is set to 80% of the window height

---

- `CustomHeaderLeftClose`

    custom-header-left-close

- `ErrorBottomSheet`

    error-bottom-sheet

    error-bottom-sheet-with-try-again

- `PhotoPreviewModal`

    photo-preview-modal

- `PermissionAudio`

    permission-audio

- Receiving invite flow (forgot to screenshot success)

    new-invite-received

    invite-cancelled

- `SaveTrackScreen`

    save-track

- Loading state

    loading


ErikSin:
Is this desired design (with the space where the description belongs)? It feels a little akward to me, it seems better to get rid of that space an dynamically size the modal (Which I think we were doing before)

image

achou11:
> Is this desired design (with the space where the description belongs)? It feels a little akward to me, it seems better to get rid of that space an dynamically size the modal (Which I think we were doing before)

yeah this is an unfortunate consequence of adding a minimum height for the sheet content in the case of the non-fullscreen variant. without it, the re-introduced flex behavior that fixes the fullscreen variant of the content doesn't play well with the bottom sheet modal dep's dynamic sizing behavior 😓 

basically, super tedious to figure out! i have a couple of quick ideas i can experiment with, but may punt on this and instead just fix the cases where the fullscreen variant is broken...

ErikSin (changes requested):
Instead of passing the full screen prop to both the `Modal` and the `content` could we not just use a context? We know that the `content` will always be wrapped by the modal (and if is not, it should throw an error). I think have a full screen prop in two places is not the best design choice.

achou11:
> Instead of passing the full screen prop to both the `Modal` and the `content` could we not just use a context? We know that the `content` will always be wrapped by the modal (and if is not, it should throw an error). I think having a full screen prop in two places is not the best design choice.

can give it a try! Agree that having to remember to specify the prop in two separate places isn't ideal

achou11:
@ErikSin good suggestion with the context! see https://github.com/digidem/comapeo-mobile/pull/415/commits/b8389d5548a685c2b91f0e5267231d00defca184 for changes, but to summarize:

- creates a context called `BottomSheetModalPropertiesContext` (very open to naming suggestions). Provider is set up in the `BottomSheetModal/index.tsx` and the hook is used in the Content component (which has been moved, see next note)

- moves the `Content` component from `sharedComponents/BottomSheet/Content.tsx` to `sharedComponents/BottomSheetModal/Content.tsx`. this component is now closely tied to the modal sheet so it made more sense to co-locate it. I think in Mapeo, the original reasoning for having this component live in `BottomSheet/` was because it technically could be used for bottom sheets that weren't modals. in practice, it's never used like that, so it felt appropriate to make the change.
  - given this change, I also updated `sharedComponents/BottomSheet/BottomSheet.tsx` to just `sharedComponents/BottomSheet.tsx`

src/frontend/contexts/AppProviders.tsx
```diff
@@ -39,26 +42,32 @@ export const AppProviders = ({
...
-                        </PhotoPromiseProvider>
-                      </BottomSheetModalProvider>
-                    </ActiveProjectProvider>
-                  </ApiProvider>
-                </LocalDiscoveryProvider>
-              </ServerLoading>
-            </GPSModalContextProvider>
-          </TrackTimerContextProvider>
-        </GestureHandlerRootView>
+        <SafeAreaProvider>
```
achou11, line 45: this fixes an issue where we were using `useSafeAreaInsets()` without a provider set up

src/frontend/sharedComponents/BottomSheetModal/Content.tsx
```diff
@@ -0,0 +1,160 @@
...
+  buttonConfigs: ActionButtonConfig[];
+  description?: React.ReactNode;
+  descriptionStyle?: TextStyle;
+  icon?: React.ReactNode;
+  loading?: boolean;
+  title: React.ReactNode;
+  titleStyle?: TextStyle;
+}
+
+export const Content = ({
```
achou11, line 45: part of me thinks that maybe we should have 2 separate components exported: a fullscreen and a dynamic variant. can't imagine cases where a consuming component would switch between the two.

can be done as a follow-up though

src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -74,34 +80,47 @@ interface Props extends React.PropsWithChildren<{}> {
...
 export const BottomSheetModal = React.forwardRef<RNBottomSheetModal, Props>(
-  ({children, isOpen, onBack, fullHeight, onDismiss}, ref) => {
+  ({children, isOpen, onBack, onDismiss, fullScreen}, ref) => {
     useBackHandler(isOpen, onBack);
 
+    const {top} = useSafeAreaInsets();
+
+    const contentContextValue = React.useMemo(() => {
+      return {fullScreen: !!fullScreen};
+    }, [fullScreen]);
```
ErikSin, line 94: nit: I don't think this memoization is necessary. Each modal creates it own instance of a context, so there is no need to worry about changes in state causing unnecessary re-renders of other components

achou11: addressed via a51326570fa820ddbfda19d3a5cc4d4202c15e46

achou11: to clarify the initial reasoning: i was worried that having the provider being a child of the BottomSheetModal would potentially be problematic. based on reading the documentation, using the dynamic sizing can cause several re-renders, which i _think_ would affect the children and hence, create a new context value each time.

probably not a big issue in general though

ErikSin (approved):
Looks good! Thanks for doing that.

PR #416, opened 2024-06-18
chore: Re-useable edit observation/track screen

ErikSin:
Create a new reusable component that holds the UI for the observation edit/observation create (and eventually track edit and track create) screens.

Previously the observation Edit and Observation create screen were the same screen. While they shared the same UI, their functionalities were quite different (eg, different API calls, different back button behaviour). This difference in functionality was handled by detecting the presence of an observationId (the presence of an observationId implied that the observation had already been created, and therefore the user was editing, not creating, it). This meant quite a lot of overhead with trying to determine which actions to take in the screen.

With the introduction of tracks, the same screens would also be used for editing and creating a track. Having 1 component, with 4 different states was going to be difficult to maintain.

This PR introduces a shared UI component. This is designed to be imported and used within a screen, but all the functionality is separated and taken as props, and therefore is controlled by the parent screen. This allows us to have the same UI between several different screens, but have the functionality be modularized and controlled by the parent screen.

In order to keep this PR small, I have only used this shared UI component in a new CreateObservation screen. The previous implementation of the observationEdit screen is still intact and untouched, and that will be refactored in a follow up PR.

image


src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: WHITE,
+    flexDirection: 'column',
+    alignContent: 'stretch',
```
achou11, line 69: probably not needed

src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  notes,
+  updateNotes,
+  photos,
+  audioRecordings,
+  location,
+  fieldIds,
+}: EditScreenProps) => {
+  return (
+    <View style={styles.container}>
+      <ScrollView contentContainerStyle={styles.scrollViewContent}>
```
achou11, line 48: doing this will not provide the desired scroll behavior when the content overflows (try adding something like `<View style={{height: 1000, backgroundColor: 'cyan'}} />` as a child here to see). what's needed is to make the scrollview itself stretch i.e. `style={{flex: 1}}` and then remove the `flex: 1` from `styles.scrollViewContent`, which will then properly handle the overflow

src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    flex: 1,
+    backgroundColor: WHITE,
+    flexDirection: 'column',
+    alignContent: 'stretch',
+  },
+  scrollViewContent: {
+    flex: 1,
+    flexDirection: 'column',
+    alignContent: 'stretch',
+  },
```
achou11, line 75: see other comment, but this should be applied to the `style` prop of the `ScrollView` instead

src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  container: {
+    flex: 1,
+    backgroundColor: WHITE,
+    flexDirection: 'column',
+    alignContent: 'stretch',
+  },
+  scrollViewContent: {
+    flex: 1,
+    flexDirection: 'column',
+    alignContent: 'stretch',
```
achou11, line 74: probably not needed

src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <ActionButtons fieldIds={fieldIds} showAudio={!!showAudio} />
+      <ErrorBottomSheet error={error} clearError={clearError} />
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: WHITE,
```
achou11, line 67: is this needed?

src/frontend/sharedComponents/EditScreen/ActionButtons.tsx
```diff
@@ -0,0 +1,110 @@
...
+      navigation.navigate('Home', {screen: 'Map'});
+    } else {
+      openAudioPermissionSheet();
+    }
+  }, [
+    navigation,
+    openAudioPermissionSheet,
+    permissionResponse?.granted,
+    shouldShowAudio,
+  ]);
```
achou11, line 69: pressing this causes the app to crash for me

```
Error: Couldn't find a navigation object. Is your component inside NavigationContainer?
```

src/frontend/sharedComponents/EditScreen/ActionButtons.tsx
```diff
@@ -0,0 +1,110 @@
...
+    description: 'Button label to add details',
+  },
+});
+
+interface ActionButtonsProps {
+  showAudio: boolean;
+  fieldIds?: string[];
+}
+
+export const ActionButtons = ({showAudio, fieldIds}: ActionButtonsProps) => {
```
achou11, line 35: due to the issues I'm encountering when testing this, might make sense to remove all of the audio-related code here (and in `EditScreen/index.tsx`) for now, and let that get re-introduced with the audio work I'm doing. reason is that I don't think the issues are related to this PR, but could be mistaken.

thoughts?

achou11 (changes requested):
overall looking good! couple of blocking comments and issues that may not be related to this PR

src/frontend/sharedComponents/EditScreen/ActionButtons.tsx
achou11: i also end up in this weird state when clearing/recovering from the error. may be related to the issues being addressed via #415

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+import {CommonActions} from '@react-navigation/native';
+import {Photo, DraftPhoto} from '../../contexts/PhotoPromiseContext/types';
+import {useCreateBlobMutation} from '../../hooks/server/media';
+import {usePersistedTrack} from '../../hooks/persistedState/usePersistedTrack';
+import {SaveButton} from '../../sharedComponents/SaveButton';
+import {useMostAccurateLocationForObservation} from '../ObservationEdit/useMostAccurateLocationForObservation';
+
+const m = defineMessages({
+  observation: {
+    // Keep id stable for translations
```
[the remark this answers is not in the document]
ErikSin: fun fact, the translations are shared between repos. We should have done a better job at this, but if they have the same id and the same strings as mapeo, it will just take the mapeo translations. Hence this comment

src/frontend/sharedComponents/EditScreen/index.tsx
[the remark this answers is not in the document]
ErikSin: Its going to be used by tracks as well. Will rename to Editor

src/frontend/sharedComponents/EditScreen/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <PresetAndLocationView
+          onPressPreset={onPressPreset}
+          presetName={presetName}
+          PresetIcon={PresetIcon}
+          location={location}
+        />
+        <DescriptionField notes={notes} updateNotes={updateNotes} />
+      </ScrollView>
+      <MediaScrollView photos={photos} audioRecordings={audioRecordings} />
+      <ActionButtons fieldIds={fieldIds} showAudio={!!showAudio} />
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d3b0e6a5cd848490aa0959df5967749366a92e28

src/frontend/sharedComponents/EditScreen/index.tsx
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/3d264e511e55ff5c35a58fcba6fb4f0fccde918b

src/frontend/screens/ObservationCreate/index.tsx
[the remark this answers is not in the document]
achou11: ah good to know, although that doesn't seem ideal. maybe i'm missing some context for why it's useful to have it set up that way

src/frontend/sharedComponents/EditScreen/ActionButtons.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/9e33768e294d1bc9e5fa9cdd2bddc6ed3e433dac

src/frontend/screens/ObservationCreate/index.tsx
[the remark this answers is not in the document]
ErikSin: Its just avoids us having to do the translations twice. Most of the strings in CoMapeo and Mapeo are identical, so there are many translations that we can just sync up.

src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -269,5 +272,14 @@ export const createDefaultScreenGroup = ({
...
       options={{headerTitle: intl(LanguageSettings.navTitle)}}
     />
+    <RootStack.Screen
+      name="ObservationCreate"
+      component={ObservationCreate}
+      options={{
+        headerTitle: intl(ObservationCreate.navTitle),
+        headerRight: () => <SaveButton onPress={() => {}} isLoading={false} />,
+        headerLeft: props => <HeaderLeft headerBackButtonProps={props} />,
+      }}
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d4a7b65f4b5ef991b23b4ea7fa473cc945a8cf21

achou11:
@ErikSin can you rebase your PR? it's needs changes from main branch for the PermissionAudio component that prevents the weird UI i ran into in my PR review earlier

src/frontend/screens/ObservationCreate/index.tsx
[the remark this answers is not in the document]
achou11: seems kind of fragile to me, especially since this isn't documented yet and hasn't been mentioned until now. wondering if crowdin provides tooling for copying over translations or something like that so that the burden of re-doing translations is on us and not actual translators, if we wanted to create a separate project for this

src/frontend/sharedComponents/SaveButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+import SaveCheck from '../images/CheckMark.svg';
+export const SaveButton = ({
+  onPress,
+  isLoading,
+}: {
+  onPress: () => void;
+  isLoading: boolean;
+}) => {
+  return isLoading ? (
+    <View style={{marginRight: 10}}>
```
achou11, line 14: nit: consumer of this component should be responsible for the spacing here

src/frontend/screens/ObservationCreate/index.tsx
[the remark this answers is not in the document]
ErikSin: Im actually making it seem more fragile than it is. So mapeo (legacy) and CoMapeo share a [translation-memory](https://support.crowdin.com/translation-memory/). Which means that any translated strings (separate of the ID) that are the same, will automatically be detected and translated.

When i talked to crowdin the support, they also told me that if id's are the same, it would make it easier for the system to match them. They are not actually tied to the same translation, the id are just used to help match any translations. But if we overwrite the ids or strings, it won't effect the other app unless they are identical 'defaultMessages`. And in that case it is up to the final reviewer (an AD team member ) to determine whether the translation is overwritten or both translations should exist in either app.

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  photos,
+  location,
+  fieldIds,
+  actionButtonChildren,
+}: EditorProps) => {
+  return (
+    <ScreenContentWithDock
+      dockContainerStyle={{padding: 0}}
+      dockContent={
+        actionButtonChildren || <ActionButtons fieldIds={fieldIds} />
```
achou11, line 40: not keen on this interface. the `fieldIds` prop being conditionally applied based on another prop's existence is a smell to me. i would suggest removing the `fieldIds` prop and then renaming `actionButtonChildren` to something like `actionsRow` and making it a required prop:

```tsx
export const Editor = ({
  presetName,
  onPressPreset,
  PresetIcon,
  notes,
  updateNotes,
  photos,
  location,
  actionsRow,
}: EditorProps) => {
  return (
    <ScreenContentWithDock
      dockContainerStyle={{padding: 0}}
      dockContent={actionsRow}>
      <PresetAndLocationView
        onPressPreset={onPressPreset}
        presetName={presetName}
        PresetIcon={PresetIcon}
        location={location}
      />
      <DescriptionField notes={notes} updateNotes={updateNotes} />
      <MediaScrollView photos={photos} />
    </ScreenContentWithDock>
  );
};
```

and then the usage of it looks like this:

```tsx
<Editor actionsRow={<ActionButtons fieldIds={fieldIds} />} />
```

achou11: on a related note, although perhaps outside the desired scope of this PR, the `ActionTab` component (which existed prior to this) feels like an overabstraction, which is causing some of what's being introduced here to be more complicated than necessary.

probably more changes that you'd like to implement, but I'd prefer the usage of the Editor to ultimately look more like this: 

```tsx
<Editor
  actionsRow={
    <EditorActions>
      <CameraActionButton />
      <DetailsActionButton />
    </EditorActions>
  }
/>
```

src/frontend/sharedComponents/Editor/PresetAndLocationView.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const m = defineMessages({
+  searching: {
+    id: 'sharedComponents.EditScreen.PresetAndLocationView.seaching',
+    defaultMessage: 'Searching…',
+    description: 'Shown in new observation screen whilst looking for GPS',
+  },
+});
+
+export const PresetAndLocationView = ({
```
achou11, line 18: this component is unnecessary. suggest having the `Editor` use the `PresetView` and `LocationView` directly so:

1. Create separate files for the `PresetView` and `LocationView`
2. Basically transfer the content of the body in this component to `Editor`

achou11 (changes requested):
Ran into a strange bug that seems to be introduced by this PR but I can't quite figure out why:

1. Create observation. For the sake of reproducibility, choose the "Animal" category.
2. On editor screen, press the "Details" button.
3. On observation fields screen, press either the header back button or the save button.

After (3), I end up navigating to the map screen instead of back to the editor screen. I would think the bug lives in the `ObservationField` component, but this PR doesn't touch that 🤔

src/frontend/sharedComponents/ConfirmDiscardBottomSheetContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+          text: discardButtonText,
+          icon: <DiscardIcon />,
+        },
+        {
+          onPress: closeSheet,
+          text: discardButtonCancel,
+          variation: 'outlined',
+        },
+      ]}
+      icon={<ErrorIcon width={80} height={80} style={{marginBottom: 15}} />}
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/0d66b3dcf8140c92f410ceaa5501fdcd60f06d65

src/frontend/sharedComponents/ConfirmDiscardBottomSheetContent.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import {BottomSheetContent} from './BottomSheet';
```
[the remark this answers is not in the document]
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/0d66b3dcf8140c92f410ceaa5501fdcd60f06d65

src/frontend/sharedComponents/Editor/index.tsx
ErikSin: I think the abstraction actually makes sense because of the keyboard view. We customize whats on the top part of the keyboard based on what actions are available. I think if the keyboard didn't change I would agree though!

ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/d4dbe937cb9adcb6e6cabb36dde7d11bf613c683

src/frontend/sharedComponents/Editor/PresetAndLocationView.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/416/commits/a04566a0651a9d93d55539fee881b1adf9277449

src/frontend/sharedComponents/SaveButton.tsx
ErikSin: the consuming component doesn't have the granularity control of changing the styling of only the loading state. Unless we refactor it, I'm not sure there is a way to have the header change the styling of only the loading state.

This actually is a problem with how the icon button was implemented (which was ported over from mobile). It has a width of 60, and then is centered. In this case it makes it look like it has a margin right of 10. So if we don't add the margin right of 10 to the loader, it causes it be in a different place from the button. So this margin was a hacky way of fixing that.

I think in an ideal world we would refactor the icon button, so that the consumer can deal with the margin/width. But the icon button is used in many places, so I think it is out of scope to do that in this PR.

achou11:
hm let me try again - possible that i ran into some weird app state when switching branches and whatnot

ErikSin:
On another note, the screen is glitchy when moving from `observationFields` back to the 'observationCreate`....

achou11:
hm yeah i can't seem to reproduce that issue anymore - must've ran into some edge case related to switching between various branches

src/frontend/screens/ObservationCreate/index.tsx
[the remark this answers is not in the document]
achou11: this is very helpful context - thanks for explaining! can this be documented somewhere? (whether it's in the repo docs or an issue)

achou11 (approved):
good work on this! looking forward to it being used for other similar screens 👍

PR #448, opened 2024-06-27
fix: Introduces CleartextTraffic permission

ErikSin:
On production builds the map was not rendering and showing a black screen. The error was identified as [useClearTextTraffic](https://developer.android.com/guide/topics/manifest/application-element#usesCleartextTraffic) needed to be enabled.

This PR uses expo [BuildProperties](https://docs.expo.dev/versions/latest/sdk/build-properties/) to enable it/

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "eventemitter3": "^5.0.1",
     "expo": "~50.0.8",
     "expo-av": "~13.10.6",
+    "expo-build-properties": "~0.11.1",
     "expo-camera": "~14.1.3",
     "expo-crypto": "~12.8.1",
     "expo-dev-client": "~3.3.11",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -104,7 +104,8 @@
     "uint8array-extras": "^0.5.0",
     "utm": "^1.1.1",
     "validate-color": "^2.2.4",
-    "zustand": "^4.4.6"
+    "zustand": "^4.4.6",
+    "expo-build-properties": "~0.11.1"
```
achou11, line 108: nit: move this so that it's alphabetically sorted (an unfortunate behavior of `npx expo install ...`)

achou11 (approved):

package.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/448/commits/4fc064286e0af9563e98d2a4adbd6271f7c4c7c0

PR #453, opened 2024-07-01
fix: properly fetch "is data sync enabled?" boolean

EvanHahn:
`data.syncing` is undefined, and should probably be `data.isSyncEnabled`.


achou11 (changes requested):
Nice catch! Should include the fix for this line too:
 
https://github.com/digidem/comapeo-mobile/blob/6c28fbebd53b008f2fe3701aa2cdb76ad1c0eb0f/src/frontend/hooks/useSyncState.ts#L126

achou11 (approved):

PR #455, opened 2024-07-01
chore: remove unnecessary select option label helper function

EvanHahn:
We have a helper function, `convertSelectOptionsToLabeled`, which has two problems:

1. It gives type errors because neither its parameter nor return types exist any more.
2. It is not necessary.

This removes the helper function.


src/frontend/sharedComponents/FormattedData.tsx
```diff
@@ -154,8 +154,9 @@ function getValueLabel(
     // Look up label from field options. This is not necessary for presets
     // created with mapeo-settings-builder@^3.1.0, which will have these options
     // in the translation file, but is needed for older versions of presets
-    const options = convertSelectOptionsToLabeled(field.options);
-    const matchingOption = options.find(option => option.value === value);
+    const matchingOption = field.options?.find(
```
achou11, line 157: is the comment above no longer relevant then? my understanding is that any presets we use for CoMapeo are generated using more recent versions of settings builder. also, just to verify: we're sure that we won't be dealing with trying to import older configs and needing to handle any of that client-side (as the helper originally seemed to needed for?)

achou11 (approved):
one comment that i'd like a response to but otherwise changes seem reasonable to me

PR #457, opened 2024-07-01
chore: only target arm-based architectures (for now)

### LINKED ISSUE #447: Only build for arm architectures
Extracted from #401 and #446.

We decided that for the time being, we can solely build for arm and not for intel architectures. This will notably improve the compilation speed and resulting apk size.

In terms of code changes, we should:

1. Write a custom config plugin that inserts the `ndk` option in the `defaultConfig` field of the android/app/build.gradle
    
    ```sh
    defaultConfig {
        // Other stuff preceding this...
        ndk {
            abiFilters = project(":app").android.defaultConfig.ndk.abiFilters ?: ["armeabi-v7a", "x86_64", "arm64-v8a"]
        }
    }
    ```
    
2. Remove `android-x64` from here

    https://github.com/digidem/comapeo-mobile/blob/23dc4dd1175c1c4af78fdc8bb205be63c7c5ee95/scripts/download-prebuilds.mjs#L7

achou11:
Closes #447 

Reduces the debug apk's size by almost half(!)

image

apk internals comparison (white left is before, white right is after, right green is differences):

image

### DIFF
--- a/app.json
+++ b/app.json
@@ -50,12 +50,16 @@
           "fonts": ["./public/Rubik-Regular.ttf", "./public/Rubik-Medium.ttf"]
         }
       ],
-      ["./expo-config-plugins/removeExpoInputStyles.js"],
-      ["expo-build-properties",{
-        "android":{
-          "usesCleartextTraffic":true
+      [
+        "expo-build-properties",
+        {
+          "android": {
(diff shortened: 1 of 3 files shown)


ErikSin (approved):
spent a couple hours diving into how plugins work which was very informative!

PR #458, opened 2024-07-02
chore: fix missing dependency in "on backgrounded/foregrounded" hook

EvanHahn:
I forgot to add this dependency in bb201e0f80a4a3efab0e2f109278a48febbd4ffd.

We don't ever change this argument in practice, but this fixes a lint error. And if we ever *do* change it in the future, we'll be ready.


cimigree (approved):
I guess that is partially what linting is for, right?

EvanHahn:
Agreed! Not sure why it didn't show up in my editor at the time.

PR #462, opened 2024-07-03
chore: send device diagnostic metrics

### LINKED ISSUE #459: Gather device diagnostic metrics once a month

EvanHahn:
We've changed our plan for metrics.

This implements one part of the new plan: **device diagnostic metrics**. This includes device manufacturer, device type, total RAM, and other data about the device itself.

These metrics should be sent once a month, as they are unlikely to change.

_Notes for reviewers:_

- I don't think it's really worth looking at the deleted files. They represent our old plan.
- I would start in `useSendMetrics.ts`, then go to `useSendDeviceDiagnosticMetrics.ts`, and follow the code from there.

Closes #459.


src/frontend/__mocks__/expo-crypto.ts
EvanHahn, line 1: Jest can't import `expo-crypto`. This mocks out the parts we use so tests pass.

src/frontend/hooks/persistedState/usePersistedMetricsDeviceId.ts
EvanHahn, line 1: This is a persisted "metrics device ID", stored for the lifetime of the app. I couldn't figure out a better way to do this, but maybe there's some stable device ID we can grab from the OS somehow.

(I found something for Android but not for iOS.)

src/frontend/metrics/getMonthlyHash.ts
EvanHahn, line 1: This is currently just used for device diagnostic metrics, but we also plan to use it in a few other places.

src/frontend/metrics/getMonthlyHash.ts
```diff
@@ -0,0 +1,46 @@
...
+  writeStringToBytes(result, 3 + salt.length, metricsDeviceId);
+  return result;
+}
+
+export async function getMonthlyHash(
+  salt: string,
+  metricsDeviceId: string,
+  now: Date,
+): Promise<string> {
+  const hashedBuffer = await digest(
```
gmaclennan, line 39: What benefit does this have over:

```ts
return await digestStringAsync(
  CryptoDigestAlgorithm.SHA512,
  `${now.getUTCFullYear()}-${now.getUTCMonth()}-${metricsDeviceId}`
)
```
?

src/frontend/metrics/deviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,68 @@
+import pProps from 'p-props';
+import {Dimensions, PixelRatio, Platform} from 'react-native';
+import DeviceInfo from 'react-native-device-info';
```
gmaclennan, line 3: Just highlighting that [expo-device](https://docs.expo.dev/versions/latest/sdk/device/) does, I think, more normalization of this data between platforms, and also provides more data as constants rather than async. We are however already using react-native-device-info in a few places, so fine to use here for now. As we add the daily app data, it may be more convenient to switch to expo-device.

src/frontend/metrics/getMonthlyHash.ts
EvanHahn: Good point. Simplified in d08e69d8727f4b7ff6114317688ec42fdc5f8ff2.

src/frontend/metrics/deviceDiagnosticMetrics.ts
EvanHahn: The fact that it returns data synchronously makes things much easier. Switched in a5b55f246c9a7f5eecd2ace1fdeb2ceb25d986e8.

EvanHahn:
Rewrote this to use a singleton.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+      return 'TV';
+    case null:
+    case Device.DeviceType.UNKNOWN:
+    default:
+      return 'UNKNOWN';
+  }
+}
+
+function getMetricsDeviceId(): string {
+  const result = storage.getString('MetricsDeviceId');
```
gmaclennan, line 58: I think this should be read/set at initialization, and kept in memory. Storage is fast, but it still has a cost, and there is no need to re-read this every time.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+  if (isEmulator) result.isEmulator = true;
+
+  return result;
+}
+
+const isAppActive = (appState: AppStateStatus): boolean =>
+  appState === 'active';
+
+const hasEnoughTimeElapsed = (): boolean => {
+  const lastSentAtMs = storage.getNumber(STORAGE_KEY) ?? -Infinity;
```
gmaclennan, line 113: As above, read at initialization, and keep the result in memory.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.#update();
+  }
+
+  async #update(): Promise<void> {
+    this.#updateSubscriptions();
+
+    const shouldSendMetrics =
+      this.#isEnabled &&
+      this.#isOnline &&
+      this.#isAppActive &&
```
gmaclennan, line 145: Read this from AppState.currentState rather than storing it as state - there is no additional cost to reading this, and keeping the two states in sync just makes things difficult for ourselves (e.g. currently there are edge-cases where this could not be the same as `AppState.currentState` if setEnabled() is called at certain times in the app lifecycle, or the class is initialized at certain times).

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  #subscriptionCleanupFns: null | (() => void)[] = null;
+
+  setEnabled(isEnabled: boolean): void {
+    this.#isEnabled = isEnabled;
+    this.#update();
+  }
+
+  async #update(): Promise<void> {
+    this.#updateSubscriptions();
```
gmaclennan, line 140: Personal opinion: Updating these as a side-effect of setting enabled (vs. just subscribing / unsubscribing within setEnabled) creates more scope for introducing bugs (e.g. circular code). I don't see the advantage of having these as a side-effect rather than within enable/disable action.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+      const appStateSubscription = AppState.addEventListener(
+        'change',
+        appState => {
+          this.#isAppActive = isAppActive(appState);
+          this.#update();
+        },
+      );
+      subscriptionCleanupFns.push(() => appStateSubscription.remove());
+
+      const nextMonthTimeout = setLongTimeout(() => {
```
gmaclennan, line 187: I can't find any info online about leaving long timeouts when the app is in the background, however I feel like we should avoid doing that, in case it results in either the timeout being cleared (which will lead to unexpected behaviour), or the app being closed by the OS. I think it's safest to just run this while the app is in the foreground (active state).

Also, setLongTimeout is very clever, however if we are only handling the use-case of the app being in the foreground, the main edge-case we are trying to catch with the timeout is the user having the app open and connected over the month-end. It may be worth just removing this, and acknowledging that if a user keeps the app open for 25 days without turning off their phone or switching to another app, then metrics will fail to send, and that avoids adding additional code to maintain.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -0,0 +1,198 @@
...
+  );
+  setIfNotNull(result, 'totalMemory', Device.totalMemory);
+
+  const isEmulator = !Device.isDevice;
+  if (isEmulator) result.isEmulator = true;
+
+  return result;
+}
+
+const isAppActive = (appState: AppStateStatus): boolean =>
```
gmaclennan, line 109: Complete personal preference: just write `appState === 'active'` in the code. Adding this helper function means that a reviewer / maintainer needs to look up what this helper does, and it doesn't really save many characters.

src/frontend/hooks/useSendMetrics.ts
gmaclennan, line 1: Is this used anywhere any more?

src/frontend/hooks/persistedState/createPersistedState.ts
gmaclennan, line 1: Are these changes still necessary?

src/frontend/metrics/DeviceDiagnosticMetrics.ts
EvanHahn: Removed in 7deb5d846d7365b24a46cac61005bed10c2174e8.

src/frontend/hooks/useSendMetrics.ts
EvanHahn: Nope, thanks for catching. Removed in 94c925e93ff5e9ab56fd5fddda5eb74fabcee009.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
EvanHahn: Changed in f94bdc53aca2fee09b3144baffd35805233f051b.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
EvanHahn: Removed in f94bdc53aca2fee09b3144baffd35805233f051b.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
EvanHahn: I'd rather read it every time, especially given that this is called infrequently. Keeping the two states in sync is error-prone; for example, if someone sets the key outside this class, the in-memory value will be wrong.

I don't feel strongly, but I'm going to leave things as is. Let me know if you want to profile this or change it.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
EvanHahn: Moved in d8830d5cfedede2a8116ccd35b1619355fa6b2aa.

src/frontend/hooks/persistedState/createPersistedState.ts
EvanHahn: No. Reverted in 8db0c255cb326a6f8252c8621b9be422724cc33c.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
gmaclennan: agreed about the out-of-sync risk, and for deviceId I guess it's not called often, however one would be called every time the user gets a connection, and every time the app opens, and modifying this value outside this class would break things anyway, however the perf impact is relatively small, just want to flag this to keep an eye on it if we do start seeing perf issues - I think mmkv is fairly fast, but I've seen issues in the past with storage causing slowdowns in react native apps on older / lower cost devices.

gmaclennan (approved):

PR #468, opened 2024-07-04
chore: update file name for @mapeo/ipc patch

achou11:
to get rid of warnings from patch-package 😄


cimigree (approved):
I blasted node modules, reinstalled, and ran e2e tests and all worked perfectly

PR #470, opened 2024-07-08
fix: fix submit button when editing device name

### LINKED ISSUE #434: Save button when editing device name not working
**Describe the bug**
If someone tries to edit the device name, the check mark in the top left of the screen does not successfully save the changes. Then any edits have to be discarded to leave the screen.

**To Reproduce**
1. User has gone through onboarding and has named their device
2. User wishes to edit their device name
3. User goes to **Settings - Project Settings - Device Name**
4. User clicks on the pen icon to edit
5. User can edit in the field
6. User can click on the `Save` icon (a blue checkmark in the top right of the screen) but nothing happens

**Expected behavior**
User should be able to click on the save button to save the edited device name.

**Screenshots**
https://github.com/digidem/comapeo-mobile/assets/20581639/d9e0fc7d-349b-4dcc-b945-8d8fbf82c452

 - Device: Galaxy A25
 - OS: Android 14

**Additional context**
Also visible testing on an emulator

achou11:
Fixes #434 

Root issue was not properly creating a press handler via react hook form, which meant that pressing the button essentially did nothing.

unrelated, but `handleSubmit` is kind of an unfortunate naming from them - probably would've benefited from being called something like `createSubmitHandler`

### DIFF
--- a/src/frontend/screens/Settings/ProjectSettings/DeviceName/EditScreen.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/DeviceName/EditScreen.tsx
@@ -108,10 +108,8 @@ export const EditScreen = ({
     [t, navigation, getValues, nameHasChanges],
   );
 
-  const onPress = React.useCallback(() => {
-    if (isPending) return;
-
-    handleSubmit(async value => {
+  const onSubmit = React.useMemo(() => {
+    return handleSubmit(value => {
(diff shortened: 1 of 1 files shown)


cimigree (approved):
I was able to slightly modify the maestro test and test it out, so that was fun. Your naming and fixes make a lot of sense.

PR #472, opened 2024-07-08
feat: color mapped observations by preset

### LINKED ISSUE #419: Add Color to PresetIcon and Observation Map Layer

achou11:
Closes #419 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "@rnmapbox/maps": "^10.1.27",
         "@sentry/react-native": "~5.20.0",
         "@tanstack/react-query": "^5.12.2",
+        "@turf/helpers": "^7.0.0",
         "@types/luxon": "^3.4.2",
         "assert": "^2.0.0",
         "buffer": "^6.0.3",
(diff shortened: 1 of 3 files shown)


cimigree (approved):
Everything seems to be working great and the code makes sense to me (even including the comments, I suppose), but in the future it would be helpful to have more information in PRs. Such as, 1, how to test the code (in this case, I needed to figure out how to mock observations), and 2. if you add packages, why? how are they best used? I am sure you and Erik know this stuff inside and out, but it is new for me.

Also, just curious and for my own edification, when you are using arrays to create new arrays, why are you using for loops instead of mapping?

achou11:
> Everything seems to be working great and the code makes sense to me (even including the comments, I suppose), but in the future it would be helpful to have more information in PRs. Such as, 1, how to test the code (in this case, I needed to figure out how to mock observations), and 2. if you add packages, why? how are they best used? I am sure you and Erik know this stuff inside and out, but it is new for me.

my apologies, should've been more mindful of this! regarding the question about packages: I've installed `@turf/helpers`, which is a collection of utilities for working with data that conforms to the GeoJSON spec (used in mapping applications). Essentially it's lodash but for geospatial data. There are a couple places benefit from using this:

- creating data that uses the geojson format. in most cases, it's easy enough to hardcode this in some cases, but it's also easy to make mistakes. using the turf helpers makes the code more straightforward. this PR makes use of the relevant helpers for this purpose.
- geospatial calculations. there are some spots where calculating a bounding box based on a collection of points is needed. right now we have a custom implementation (which was deliberate choice by Erik), but i would prefer to lean on turf for this kind of thing moving forward.

> Also, just curious and for my own edification, when you are using arrays to create new arrays, why are you using for loops instead of mapping?

aside from performance advantages, using a `for` loop is usually mostly necessary if you want to create a collection of data that doesn't necessarily operate on every item in the original collection. in this case, it's useful because we don't necessarily want to include every observation in the case that it's missing relevant data. you could achieve the same outcome using something like `observations.filter().map()` but this has performance overhead for larger collections, as you're looping over the original data twice (length of observations array can eventually be large too), or more clever approaches such as `flatMap()`, which are generally harder to follow compared to a `for` loop.

in general, i prefer using `for` loops where I can if the functional approaches have drawbacks based on the context

PR #473, opened 2024-07-08
chore: specify deviceType field when setting device info

achou11:
the `deviceType` field was added as part of the `deviceInfo` that can be retrieved and saved.

### DIFF
--- a/src/frontend/hooks/server/deviceInfo.ts
+++ b/src/frontend/hooks/server/deviceInfo.ts
@@ -1,4 +1,6 @@
 import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query';
+import {isTablet} from 'react-native-device-info';
+
 import {useApi} from '../../contexts/ApiContext';
 
 export const DEVICE_INFO_KEY = 'deviceInfo';
@@ -21,7 +23,10 @@ export const useEditDeviceInfo = () => {
   return useMutation({
     mutationKey: ['device'],
     mutationFn: async (name: string) => {
-      return mapeoApi.setDeviceInfo({name});
+      return mapeoApi.setDeviceInfo({
+        name,
+        deviceType: isTablet() ? 'tablet' : 'mobile',
+      });
     },
     onSuccess: () => {
       queryClient.invalidateQueries({queryKey: [DEVICE_INFO_KEY]});
--- a/src/frontend/hooks/server/deviceInfo.ts
+++ b/src/frontend/hooks/server/deviceInfo.ts
@@ -1,4 +1,4 @@
-import {DeviceInfo} from '@mapeo/schema';
+import {type DeviceInfo} from '@mapeo/schema';
 import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query';
 import {deviceType, DeviceType} from 'expo-device';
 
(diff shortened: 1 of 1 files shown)


EvanHahn (approved):
LGTM.

You may wish to use `expo-device`'s [`Device.deviceType` property](https://docs.expo.dev/versions/latest/sdk/device/#devicedevicetype) instead, but I don't know which is better.

achou11:
> LGTM.
> 
> You may wish to use `expo-device`'s [`Device.deviceType` property](https://docs.expo.dev/versions/latest/sdk/device/#devicedevicetype) instead, but I don't know which is better.

ah good suggestion. think it's worth doing that if we're interested in moving over to just using expo-device

src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -21,10 +24,33 @@ export const useEditDeviceInfo = () => {
...
+function expoToCoreDeviceType(d: DeviceType): DeviceInfo['deviceType'] {
+  switch (d) {
+    case DeviceType.PHONE: {
+      return 'mobile';
+    }
+    case DeviceType.TABLET: {
+      return 'tablet';
+    }
+    case DeviceType.TV: {
+      return undefined;
```
achou11, line 47: leaving this as `undefined` seemed to make more sense than `UNRECOGNIZED`, but not super sure

src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -21,10 +24,33 @@ export const useEditDeviceInfo = () => {
...
+      });
     },
     onSuccess: () => {
       queryClient.invalidateQueries({queryKey: [DEVICE_INFO_KEY]});
     },
   });
 };
+
+function expoToCoreDeviceType(d: DeviceType): DeviceInfo['deviceType'] {
+  switch (d) {
```
achou11, line 39: i ordered these based on likelihood given that the app probably only supports phones and tablets. gotta get that 0.0000000001% speed boost 😎

src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -1,4 +1,7 @@
+import {DeviceInfo} from '@mapeo/schema';
```
EvanHahn, line 1: If we're only using this as a type, maybe change this to `import type`?

Either of these:

```suggestion
import type {DeviceInfo} from '@mapeo/schema';
```

```suggestion
import {type DeviceInfo} from '@mapeo/schema';
```

src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -21,10 +24,33 @@ export const useEditDeviceInfo = () => {
...
+    case DeviceType.TV: {
+      return undefined;
+    }
+    case DeviceType.DESKTOP: {
+      return 'desktop';
+    }
+    case DeviceType.UNKNOWN: {
+      return 'UNRECOGNIZED';
+    }
+  }
```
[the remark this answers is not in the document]
achou11: does default return `undefined` here or throw an error?

src/frontend/hooks/server/deviceInfo.ts
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/473/commits/48890d891f0e42d358474ad155916d70c1f329c8

src/frontend/hooks/server/deviceInfo.ts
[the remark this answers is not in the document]
achou11: opted for throwing an error (see ff87d0a2125758316c7237982c84e2e24231238e)

PR #476, opened 2024-07-08
fix: update invites api usage to fix invite accept flow

achou11:
a recent upgrade to `@mapeo/core` changed a couple of things about how invites are handled, most notably replacing the `projectPublicId` field with the `projectInviteId` field. This wasn't accounted for when upgrading to the core version that introduced that change, causing some invite-related code to be broken.

### DIFF
--- a/src/frontend/hooks/server/invites.ts
+++ b/src/frontend/hooks/server/invites.ts
@@ -21,7 +21,7 @@ export function usePendingInvites() {
   });
 }
 
-export function useAcceptInvite(projectId?: string) {
+export function useAcceptInvite() {
   const mapeoApi = useApi();
   const queryClient = useQueryClient();
   const switchActiveProject = usePersistedProjectId(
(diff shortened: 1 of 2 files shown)


src/frontend/hooks/server/invites.ts
```diff
@@ -30,25 +30,17 @@ export function useAcceptInvite(projectId?: string) {
...
-          }),
-          queryClient.invalidateQueries({
-            queryKey: [ALL_PROJECTS_KEY],
-          }),
-        ]).then(() => {
-          if (projectId) {
-            switchActiveProject(projectId);
-          }
-        });
-      }, 5000);
```
achou11, line 51: the workaround described here does not seem to be relevant anymore. did a basic test with the replacement implementation and it seemed to behave as expected. the only thing that _may_ need to be adjusted is potentially invalidating the caches before switching projects e.g.

```ts
// Invalidate the queries first
await Promise.all([...])

// Then switch projects
switchActiveProject(...)
```

src/frontend/hooks/server/invites.ts
```diff
@@ -30,25 +30,17 @@ export function useAcceptInvite(projectId?: string) {
...
+    onSuccess: projectPublicId => {
```
achou11, line 35: `invite.accept()` resolves with the `projectPublicId`, so no need for the hook to accept that parameter anymore, which makes usage simpler

ErikSin (approved):
Thanks for fixing this. I am going to merge right away as this is the fix for users not being able to properly join a project, which was blocking QA sync

PR #478, opened 2024-07-09
chore: Adds test for saving an edited device name and updates docs.

cimigree:
No issue for the test itself, so nothing to put here.

### Instructions:
maestro test e2e

If you really want, you could build an apk and install it on an emulator and then test it

... LMK and I can give further instructions.

### Screen Recording
editdevicename

### DIFF
--- a/e2e/project/edit-device-name.yaml
+++ b/e2e/project/edit-device-name.yaml
@@ -55,9 +55,8 @@ env:
     text: ${output.names.editdevice}
 - tapOn:
     id: 'save-icon'
-- tapOn:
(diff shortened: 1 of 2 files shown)


e2e/project/edit-device-name.yaml
achou11, line 1: nit: think it would be worth adding a step at the end to re-enter this screen and make sure that the edited name is displayed, to ensure that the persistence is actually working

achou11 (approved):

e2e/project/edit-device-name.yaml
cimigree: Ok done. Want a new screen recording?

PR #480, opened 2024-07-09
chore: send app diagnostic metrics

### LINKED ISSUE #460: Gather app diagnostic metrics once a day

EvanHahn:
This sends app diagnostic metrics, which includes things like app and OS versions, locale, and more.

Reports are queued up for the current month and sent when the user is online.

I tried to test as much as I could, but some of it is really tricky to check automatically.

See also: 39773c54826ed1f13a679c835d0bfd81a24e719d

Closes #460.


src/frontend/lib/OneAtATimeQueue.ts
EvanHahn, line 1: I wanted to use `PQueue({ concurrency: 1 })` but couldn't get it to import properly. After 15 minutes of trying, I gave up and wrote a smaller class.

src/frontend/lib/setIfNotNull.ts
EvanHahn, line 1: This used to be in `DeviceDiagnosticMetrics.ts` but is now used in multiple places, so I pulled it out and wrote some quick tests.

src/frontend/metrics/AppDiagnosticMetrics.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  return stored as AppDiagnosticMetricsQueue;
+}
+
+async function getCountry(): Promise<undefined | string> {
+  const lastKnownPosition = await getLastKnownPositionAsync().catch(() => null);
+  if (!lastKnownPosition) return undefined;
+  const {latitude, longitude} = lastKnownPosition.coords;
+  const countries = positionToCountries(latitude, longitude);
+  return first(countries);
```
EvanHahn, line 49: It's likely that this only has one value, but 

- it could be empty if the report is in the middle of the ocean
- it could have multiple values if the coordinates are in disputed territory

I decided to just pick the first one. Does that seem reasonable?

src/frontend/metrics/AppDiagnosticMetrics.ts
```diff
@@ -0,0 +1,200 @@
...
+const CHECK_INTERVAL = 5 * MINUTE_MS;
+
+function loadQueueFromStorage(): AppDiagnosticMetricsQueue {
+  const storedString = storage.getString(STORAGE_KEY);
+  if (!storedString) return {reports: []};
+
+  const stored = maybeJsonParse(storedString);
+  if (!stored) return {reports: []};
+
+  return stored as AppDiagnosticMetricsQueue;
```
EvanHahn, line 41: This cast is not type-safe. The stored value could be of a type other than `AppDiagnosticMetricsQueue`. I considered writing a bunch of code to check the type (possibly using something like [Zod](https://www.npmjs.com/package/zod)) but I chose not to.

src/frontend/metrics/AppDiagnosticMetricsQueue.ts
EvanHahn, line 1: A lot of this PR is managing the queue. I tried to pull that out into an isolated, testable module.

src/frontend/metrics/DeviceDiagnosticMetrics.ts
```diff
@@ -43,23 +43,6 @@ function deviceTypeToString(deviceType: null | Device.DeviceType): string {
...
-}
-
-function setIfNotNull<ObjT, KeyT extends keyof ObjT>(
-  obj: ObjT,
-  key: KeyT,
-  value: null | ObjT[KeyT],
-): void {
-  if (value !== null) obj[key] = value;
-}
-
```
EvanHahn, line 62: Moved to shared helpers.

EvanHahn:
Simplified this to only send metrics for the current month.

EvanHahn:
We discussed merging this during a call. Merging now...

src/frontend/lib/date.ts
```diff
@@ -1,3 +1,19 @@
...
+const ISO_DATE_REGEXP = /^(?<year>\d{4,})-(?<month>\d{2})-(?<day>\d{2})$/;
+
+export const beginningOfMonthUtc = (date: Readonly<Date>): Date =>
+  new Date(Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), 1));
+
+export const formatIsoUtc = (date: Readonly<Date>): string =>
+  [
+    date.getUTCFullYear().toString().padStart(4, '0'),
+    (date.getUTCMonth() + 1).toString().padStart(2, '0'),
+    date.getDate().toString().padStart(2, '0'),
```
gmaclennan, line 14: I think this should be `date.getUTCDate()`?

gmaclennan: Maybe a test case needs added to catch this?

src/frontend/lib/date.ts
```diff
@@ -11,3 +27,28 @@ export const isSameUtcMonthAndYear = (
...
+  if (!year) return null;
+
+  const monthIndex = parseMaybeInt(matchGroups.month) - 1;
+  if (monthIndex < 0 || monthIndex > 11) return null;
+
+  const day = parseMaybeInt(matchGroups.day);
+  if (!day || day > getDaysInMonth(year, monthIndex)) return null;
+
+  const result = new Date(
+    Date.UTC(year, monthIndex, parseMaybeInt(matchGroups.day)),
```
gmaclennan, line 51: nit: `Date.UTC(year, monthIndex, day)

src/frontend/lib/first.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Get the first value from an iterable, or `undefined` if the iterable is empty.
+ *
+ * Like `myArray[0]`, but works for other iterables like `Set`s.
+ */
+export function first<T>(iterable: Iterable<T>): undefined | T {
+  for (const value of iterable) return value;
```
gmaclennan, line 7: I wonder if this is any different (performance, memory) to `iterable.next().value`? Anyway, inconsequential for this code.

src/frontend/metrics/AppDiagnosticMetrics.ts
```diff
@@ -0,0 +1,204 @@
...
+            // send metrics right at the end of the month which would generate
+            // a new monthly hash, letting us track users across months.
+            newestReportDate,
+          ),
+          reports: queue.reports,
+        },
+      });
+
+      queue = updateQueueHighWatermark(queue);
+      saveQueueToStorage(queue);
```
gmaclennan, line 199: Why keep reports in the queue after a successful send?

src/frontend/metrics/AppDiagnosticMetrics.ts
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * Should only be run once at a time. See `this.#update` for a queued alternative.
+   */
+  async #doUpdate(): Promise<void> {
+    if (!this.#isEnabled) {
+      storage.delete(STORAGE_KEY);
+      return;
+    }
+
+    let queue = loadQueueFromStorage();
```
gmaclennan, line 148: This _could_ get slow as reports build up and/or we add things to reports. We could load this once at app initialization, and then only write when it changes. This is a low-priority optimization. I think it is how the persistent storage that we already use works.

gmaclennan (commented):
All looks fine. The only small bug I see is using `getDate()` instead of `getUTCDate()`.

In terms of architecture, the `update()` method is doing two things:

1. Checking if there is data to be sent, and if there is, and the app is online, trying to send it.
2. Checking if a report needs to be generated and queued.

I think it might be easier to follow if these two responsibilities were separated in the future, since there are slightly different reasons to take these two actions.

This does not need to be address urgently. As we add more metrics, we may want to separate the "send queue" from metrics generation code completely, so that all these classes need to worry about is `metricsApi.send(report)`, and then the metrics API class can deal with queuing and sending as a separate concern.

src/frontend/lib/first.ts
EvanHahn: Did a quick benchmark comparing this to `return iterable[Symbol.iterator]().next().value` and it was ~0.01ns different. My benchmark was _very_ simplistic but I suspect they're comparable.

src/frontend/metrics/AppDiagnosticMetrics.ts
EvanHahn: We don't. `updateQueueHighWatermark` returns an object like `{highWatermark: '2024-08-12', reports: []}`. That function probably needs to be renamed for clarity.

src/frontend/lib/date.ts
EvanHahn: See #593 and #595.

Issue #449, opened 2024-07-01
Deal with schema errors more gracefully


EvanHahn:
If someone is on a newer version of the schema, we shouldn't error unless it's backwards-incompatible.

For example, if there's a new field on `Observation` that we don't recognize, we shouldn't error.

gmaclennan:
This mainly applies to enums. The main issue to deal with is future attachment types.

achou11:
A quick look through the existing codebase and I'm not sure if there's anything that stands out as needing changes to account for this right now. The only potential one is maybe this, but this is done at the query level, so not sure if that strictly needs changing for now.

https://github.com/digidem/comapeo-mobile/blob/470a568b0552f4f41e876dc10c936af8b6c59e36/src/frontend/hooks/server/media.ts#L64

PR #491, opened 2024-07-10
feat: more thoroughly implement attachments list in existing observation screen

achou11:
~Towards #449~ EDIT: see PR comment - not totally accurate

Updates the existing observation screen to better handle valid attachment types that it may be unfamiliar with. At the moment, any non-photo attachments are essentially rendering a placeholder thumbnail with some copy about what it is. when saving other attachment types is more extensively implemented, updating the corresponding placeholders should be straightforward.

Open to ideas about what the placeholders should look like. Tried to use my best judgment but I'm sure it can be improved upon.

--- 

To support this, this PR introduces more flexible and reusable components with the intention of replacing the existing `MediaScrollView` component. reasons for a new component:

- the `MediaScrollView` implementation only works with photos. in addition to that, there's a good amount code that's attempting to unify the interface for a photo attachment and a draft photo. i found that mixing these together instead of separating the implementations made the implementation difficult to follow, and I have a strong idea of how to go about this (that i think is easier to reason about)

- moves the attachment blob url query logic closer to where it's needed. the existing implementation had a more complicated approach of using `useQueries` and mapping over the observation attachments at the list level. it also did not fully handle error or loading states for any of these queries. my change introduces a query that's meant to be used at the singular level i.e. when the attachment thumbnail component is rendered (not at the list level).

- uses a `FlatList` instead of `ScrollView`, which may not be completely necessary but felt like an appropriate change (see differences [here](https://reactnative.dev/docs/0.73/scrollview))

Follow up PRs will update the other callsites of the `MediaScrollView` component (editor component, which is planned to be used in the observation edit screen)

There may be some design help needed but I did what felt reasonable for now, especially since we don't have a non-experimental way of creating other attachment types just yet.

---

Preview

Success (not scrollable):

image

Success (scrollable):

image

image

Loading (when fetching blob url, not the image itself):

image
  
Error (when trying to fetch blob url, or if the image can't load):

image


achou11:
Misunderstood the ticket that this is supposedly addressing, so this PR is relatively non-urgent. will keep as a draft for now in case we are interested in potentially merging since it's mostly done. what's missing is implementation of some of the thumbnails based on ready-to-go designs.

gmaclennan:
Let's maybe hold off on this until we have some UX time available to think through whether these should be pressable and when the modal should look like, and also make sure that we have the different states correctly included.

PR #496, opened 2024-07-11
fix: hide "disconnected" for unknown device connection status

### LINKED ISSUE #493: convert `isConnected` prop in `DeviceNameWithIcon` to `isDisconnected`
DeviceNameWithIcon has a prop `isConnected`. This is an optional prop, but we are [showing the disconnected icon and message on a falsy check of isConnected](https://github.com/digidem/comapeo-mobile/blob/54be6258d27ea8eea162e8c6268039c3a179a8da/src/frontend/sharedComponents/DeviceNameWithIcon.tsx#L62). This is causing the disconnected icon and message to show up when isConnected is undefined and the user is not necessarily disconnected.

## To Do
- [ ] Convert `isConnected` prop to `isDisconnected`
- [ ] convert [isConnected falsy check](https://github.com/digidem/comapeo-mobile/blob/54be6258d27ea8eea162e8c6268039c3a179a8da/src/frontend/sharedComponents/DeviceNameWithIcon.tsx#L62) to isDisconnected truth check
- [ ] convert all instances of `DeviceNameWithIcon` to do isDisconnected check instead.

EvanHahn:
The `<DeviceNameWithIcon>` component takes an optional `isConnected` prop. If it's missing, we treat the device as disconnected, which is not necessarily right.

This change inverts that, hiding the "Disconnected" text in these cases.

Fixes #493.

### DIFF
--- a/src/frontend/sharedComponents/DeviceCard.tsx
+++ b/src/frontend/sharedComponents/DeviceCard.tsx
@@ -1,14 +1,19 @@
 import React from 'react';
 import {StyleSheet} from 'react-native';
 import {LIGHT_GREY} from '../lib/styles';
-import {DeviceType, ViewStyleProp} from '../sharedTypes';
+import {ExhaustivenessError} from '../lib/ExhaustivenessError';
+import type {
+  DeviceConnectionStatus,
+  DeviceType,
+  ViewStyleProp,
+} from '../sharedTypes';
 import {TouchableOpacity} from 'react-native-gesture-handler';
 import {DeviceNameWithIcon} from './DeviceNameWithIcon';
 
 type DeviceCardProps = {
   deviceType: DeviceType;
   name: string;
-  isConnected?: boolean;
+  deviceConnectionStatus?: DeviceConnectionStatus;
   thisDevice?: boolean;
   deviceId?: string;
   dateAdded?: Date;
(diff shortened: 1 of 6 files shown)


achou11 (approved):
fixes the issue but not particularly thrilled with the solution (mostly around default values for optional props). how about something like a `connectionState?: 'connected' | 'disconnected'` so that the check is more explicit i.e. `connectionState === 'disconnected' && ...`?

EvanHahn:
Done in 86acc928c1306609a87f875b8fd8d6323e334fa3.

src/frontend/sharedComponents/DeviceCard.tsx
```diff
@@ -24,16 +29,29 @@ export const DeviceCard = ({
...
+      isDisconnected = true;
+      break;
+    default:
+      throw new ExhaustivenessError(deviceConnectionStatus);
+  }
+
   return (
     <TouchableOpacity
-      disabled={!onPress || !isConnected}
+      disabled={!onPress || isDisconnected}
```
achou11, line 49: nit: seems sufficient to do `!onPress || deviceConnectionStatus === 'disconnected'` in this case

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
```diff
@@ -59,7 +77,7 @@ export const DeviceNameWithIcon = ({
             {formatMessage(m.thisDevice)}
           </Text>
         )}
-        {!isConnected && (
+        {isDisconnected && (
```
achou11, line 80: again, seems sufficient to do `deviceConnectionStatus === 'disconnected'`

achou11 (approved):
the exhaustiveness check seems unnecessary - what's the rationale for it?

EvanHahn:
The exhaustiveness check is primarily stylistic. I like it because it requires a decision if we change the possible enum values.

For example, imagine we add a new status, `'connecting'`. Without the check, the code will type-check just fine. With the check, it will not, requiring an explicit decision in these parts of the code.

Personally, I prefer to require a decision from future developers, but it's stylistic. Happy to change it.

achou11:
> The exhaustiveness check is primarily stylistic. I like it because it requires a decision if we change the possible enum values.
> 
> For example, imagine we add a new status, `'connecting'`. Without the check, the code will type-check just fine. With the check, it will not, requiring an explicit decision in these parts of the code.
> 
> Personally, I prefer to require a decision from future developers, but it's stylistic. Happy to change it.

ah i see. yeah stylistically not a fan of it but the reasoning makes sense. fine to keep as is then

PR #499, opened 2024-07-13
fix: Drawer only closes on touch of menu icon

### LINKED ISSUE #486: Setting Menu to only close on press of hamburger menu
When the setting drawer menu is open (accessible through the map screen), the drawer should only close when the hamburger icon is open. Currenly pressing anywhere along the top of the menu closes the drawer

## To Do
- [ ] Close drawer only on when user presses the hamburger icon

### Notion issue
https://www.notion.so/digidem/Touch-anywhere-on-top-to-close-menu-505982b9acb045e68e6290aa42c1c3b9

ErikSin:
closes #486 

On the drawer menu, the hamburger icon's `onPress` was being activated when touching around the icon. I originally though it was the hitslop, but setting it to null did not did fix the problem. Converting the `TouchableNativeFeedback` to a `TouchableOpacity` seemed to solved the problem.

I also converted the styling so the consumer can control it. The two components have added styling to adjust for this.

### DIFF
--- a/src/frontend/Navigation/Drawer.tsx
+++ b/src/frontend/Navigation/Drawer.tsx
@@ -106,7 +106,10 @@ const DrawerContent = ({navigation}: DrawerContentComponentProps) => {
         style={{
           paddingBottom: 40,
         }}>
-        <DrawerMenuIcon onPress={navigation.closeDrawer} />
+        <DrawerMenuIcon
+          style={{alignSelf: 'flex-end', marginRight: 20}}
+          onPress={navigation.closeDrawer}
+        />
         <Text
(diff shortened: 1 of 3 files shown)


cimigree (approved):
How did you figure that out?! Definitely works as expected now.  Looks and works the same on different screen sizes and text sizes. The positioning is a little different from originally, but it actually seems like an improvement to me (the one on the bottom is the old version).

Screenshot 2024-07-15 at 10 27 51 AM

ErikSin:
This PR took me way too long for such a simple fix. I got to the point where I just decided to try random things to see what would work, which lead me to trying out the touchable opacity...

PR #500, opened 2024-07-14
fix: stores previous project

### LINKED ISSUE #489: Project Created Success Screen
After the creation of a project, user should be navigated to the [Project Created Screen](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx).

## To do
- [ ] Navigate user to project created screen after the creation of a project.

### Notion Issue
https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=9ec4a5637952435cbcef39849b273c15&p=c396ff6ed8d2425aacea1d774e29de6d&pm=s

ErikSin:
closes #489

### DIFF
--- a/src/frontend/contexts/ActiveProjectContext.tsx
+++ b/src/frontend/contexts/ActiveProjectContext.tsx
@@ -20,16 +20,26 @@ export const ActiveProjectProvider = ({
 
   const activeProjectQuery = useProject(activeProjectId);
   const [currentProject, setCurrentProject] = React.useState<MapeoProjectApi>();
+  const [previousProject, setPreviousProject] =
+    React.useState<MapeoProjectApi>();
   const {mutate: createProject} = useCreateProject();
 
-  // This guarantees that a project is always mounted in between creation of a new project
-  // Otherwise, a new project is created, the cache is reset, activeProjectQuery.data becomes undefined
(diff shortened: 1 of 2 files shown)


achou11:
My understanding was that React Query has functionality to handle this kind of need pretty elegantly. I was able to achieve the same thing by solely using their [query placeholder data](https://tanstack.com/query/latest/docs/framework/react/guides/placeholder-query-data) functionality. Curious if you think it makes sense to use this in this situation.

```ts
export function useProject(projectId?: string) {
  const api = useApi();

  return useQuery({
    queryKey: [PROJECT_KEY, projectId],
    queryFn: async () => {
      if (!projectId) throw new Error('Active project ID must exist');
      return api.getProject(projectId);
    },
    enabled: !!projectId,
    // This is the part I added
    placeholderData: previousData => {
      return previousData;
    },
  });
}
```

ErikSin:
Sweet! Yeah that worked. way too many hours went into this for such a simple fix....

achou11 (approved):
nit: i suggest updating the PR title to be a little clearer about the specific issue that's being fixed with this change

PR #501, opened 2024-07-15
fix: update map scale bar when zoom changes

### LINKED ISSUE #487: Scale bar on map not adjusting based on zoom
The [scale bar on the map screen](https://github.com/digidem/comapeo-mobile/blob/171fea9e8f43b36a052c28f66016ae4c8d37de20/src/frontend/screens/MapScreen/index.tsx#L117) does not display the right distance as it always remain the same

## To Do
- [ ] Fix scale bar so it adjust based on the zoon

### Notion Issue
- [ ] https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=9ec4a5637952435cbcef39849b273c15&p=2b691fb5e6c44975b71adaffbaf15eae&pm=s

achou11:
Fixes #487 

Test instructions:

1. Open app to main map screen
2. Zoom in/out using pinch gestures
3. Observe scale bar change to account for zoom change when map is idle.

---

Preview:

https://github.com/user-attachments/assets/1b2700e5-e347-4848-9786-f69527a61f6b

### DIFF
--- a/src/frontend/screens/MapScreen/index.tsx
+++ b/src/frontend/screens/MapScreen/index.tsx
@@ -79,6 +79,9 @@ export const MapScreen = () => {
         compassEnabled={false}
         scaleBarEnabled={false}
         styleURL={styleUrlQuery.data}
+        onMapIdle={event => {
+          setZoom(event.properties.zoom);
+        }}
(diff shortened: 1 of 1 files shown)


cimigree (approved):
Cool! Seems more accurate but no idea if it is truly accurate...

PR #503, opened 2024-07-15
feat: expose app passcode setting

### LINKED ISSUE #490: Add security settings
Add "Security" to settings list, which navigates to the [Security Page](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Security/index.tsx). Security was ported directly from Mapeo-Mobile.

## To Do
- [ ] Add [SecurityContext](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/contexts/SecurityContext.tsx) to app provider
- [ ] Add "Security" as an option is the setting drawer. Use MaterialIcon with iconName="security"
- [ ] Only show "Security" on the settings drawer, if the [authState](https://github.com/digidem/comapeo-mobile/blob/171fea9e8f43b36a052c28f66016ae4c8d37de20/src/frontend/contexts/SecurityContext.tsx#L18) does NOT equal "obscured"

### Notion issue
- https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=9ec4a5637952435cbcef39849b273c15&p=b3ab7d03fc19468abf223868f6f24240&pm=s&pvs=31
-https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=9ec4a5637952435cbcef39849b273c15&p=dfd122b235b44472b2f71f5599f44bad&pm=s

### To note
This should probably be refactored using zustand, but for the sake of time lets just use the architecture used in mapeo mobile

achou11:
Fixes #490

Seems that getting the security context set up was already done prior to this.

Also adds UI fixes for:

- bottom sheet modal that appears when confirming the app passcode to set
- bottom sheet modal that appears when confirming to turn off app passcode

Testing instructions:

1. Open side drawer
2. Press `Security` list item
3. Press `App Passcode` list item
4. Press `Continue` button
5. Use input to create app passcode. Press `Next` button
6. Repeat (5)
7. Press `Save App Passcode` in bottom sheet modal
8. Unfocus - but do not close - the app (e.g. go to a different app, do gesture/press button to go to phone home screen, etc). Re-focus the app. The auth screen should appear and require inputting passcode. After entering passcode successfully, authscreen should disappear and you should land on the same screen that you were on when unfocusing the app
9. Close the app and then re-open. The auth screen should appear and require input. After succeeding, you should be taken to the expected initial screen (which is usually the map screen, or if you had a draft observation, the appropriate observation flow screen).
10. To turn off app passcode, repeat (1), (2), and (3).
11. Enter passcode
12. Press `Use App Passcode`. A bottom sheet modal confirming your choice should appear. Press `Turn Off` button to confirm.

Open question:

- Should we hide the obscure passcode setting for now? None of the app code is actually using that setting, so unless we plan to follow up on it immediately, it seems misleading to have it available via this PR. My suggestion would be to hide it for now, and then reintroduce it when we make the appropriate changes throughout the app to actually use it.

---

Preview:

(note that the obscure passcode setting has been removed, was just too lazy to recreate the video for now 😅 )

https://github.com/user-attachments/assets/5ad47185-cd2b-4230-ab73-bf4a2736356b

### DIFF
--- a/src/frontend/Navigation/Stack/index.tsx
+++ b/src/frontend/Navigation/Stack/index.tsx
@@ -48,14 +48,6 @@ export function useDrawerNavigation() {
 export function RootStackNavigator({
   navigation,
 }: DrawerScreenProps<DrawerScreens, 'DrawerHome'>) {
-  const security = useSecurityContext();
-
-  React.useEffect(() => {
-    if (security.authState === 'unauthenticated') {
-      navigation.navigate('DrawerHome', {screen: 'AuthScreen'});
-    }
-  }, [security.authState, navigation]);
-
   return (
     <DrawerNavigationContext.Provider value={navigation}>
       <React.Suspense fallback={<Loading />}>
@@ -82,25 +74,21 @@ function RootStackNavigatorChild() {
 
   const security = useSecurityContext();
 
-  // const navigation = useNavigationFromRoot();
+  const {navigate} = useNavigationFromRoot();
 
-  // React.useEffect(() => {
-  //   if (security.authState === 'unauthenticated') {
-  //     navigation.navigate('AuthScreen');
-  //   }
-  // }, [security.authState, navigation]);
+  React.useEffect(() => {
+    if (security.authState === 'unauthenticated') {
+      navigate('AuthScreen');
+    }
+  }, [security.authState, navigate]);
 
   return (
     <RootStack.Navigator
-      initialRouteName={
-        security.authState === 'unauthenticated'
-          ? 'AuthScreen'
-          : getInitialRouteName({
-              hasDeviceName: !!deviceInfo.data.name,
-              existingObservation,
-              presets,
-            })
-      }
+      initialRouteName={getInitialRouteName({
+        hasDeviceName: !!deviceInfo.data.name,
+        existingObservation,
+        presets,
+      })}
       screenOptions={NavigatorScreenOptions}>
       {deviceInfo.data?.name
         ? createDefaultScreenGroup({
--- a/src/frontend/Navigation/Stack/index.tsx
+++ b/src/frontend/Navigation/Stack/index.tsx
@@ -19,6 +19,8 @@ import {createDeviceNamingScreens} from './DeviceNamingScreens';
 import {useSuspenseQuery} from '@tanstack/react-query';
 import {useApi} from '../../contexts/ApiContext';
 import {Loading} from '../../sharedComponents/Loading';
+import {useSecurityContext} from '../../contexts/SecurityContext';
+import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
 
 export const RootStack = createNativeStackNavigator<AppStackParamsList>();
 
@@ -46,6 +48,14 @@ export function useDrawerNavigation() {
 export function RootStackNavigator({
   navigation,
 }: DrawerScreenProps<DrawerScreens, 'DrawerHome'>) {
+  const security = useSecurityContext();
+
+  React.useEffect(() => {
+    if (security.authState === 'unauthenticated') {
+      navigation.navigate('DrawerHome', {screen: 'AuthScreen'});
+    }
+  }, [security.authState, navigation]);
+
   return (
     <DrawerNavigationContext.Provider value={navigation}>
       <React.Suspense fallback={<Loading />}>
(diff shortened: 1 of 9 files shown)


cimigree (dismissed):
This works and looks fine and everything is straightforward, but probably for a different ticket, when I set a passcode, I am not actually required to enter it to open the app. And there is some weird UI within the security feature.

achou11:
hm yeah i didn't really test the actual implementation of those security screens. i might hold off on merging this and add fixes for those things if they're straightforward. primary reasoning is that i don't want to expose features that are not working

what do you think?

achou11:
@cimigree I fixed the styling issue that you most likely saw with the app passcode settings screen in https://github.com/digidem/comapeo-mobile/pull/503/commits/79df1a18110bfd4ee7ad05cde81854b822b0e2c8

image

It's occurred to me that the app isn't actually making use of that setting once it's enabled though (i.e. forcing you to auth when opening the app). i'm thinking it makes sense to fix in this PR but let me know if you have other thoughts.

cimigree:
That is exactly the UI I was talking about. 
I agree about not exposing features that aren't working. Makes sense to me to fix in this PR to me

achou11:
> Should we hide the obscure passcode setting for now? None of the app code is actually using that setting, so unless we plan to follow up on it immediately, it seems misleading to have it available via this PR. My suggestion would be to hide it for now, and then reintroduce it when we make the appropriate changes throughout the app to actually use it.

@ErikSin thoughts on this?

achou11:
Discussed elsewhere, but decided to remove the obscure passcode setting for now. See https://github.com/digidem/comapeo-mobile/pull/503/commits/347dcb76ede455442f71d1801ddecbcc4291626e for relevant changes

achou11:
Outside the scope of this PR, but noticed some UI improvements for the app passcode flow that can be done. will have a separate PR for this ready shortly

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -70,6 +72,16 @@ function RootStackNavigatorChild() {
     },
   });
 
+  const security = useSecurityContext();
+
+  const {navigate} = useNavigationFromRoot();
+
+  React.useEffect(() => {
```
ErikSin, line 79: nit: convert to `useLayoutEffect` so this happens slightly sooner in the react life cycle

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
ErikSin, line 1: Im not sure why, but the bottoms sheet is slightly cutting off the button at the bottom:
image

src/frontend/screens/AuthScreen.tsx
```diff
@@ -74,53 +76,42 @@ export const AuthScreen = ({
...
-        )}
-      </KeyboardAvoidingView>
-    </ScrollView>
+  return (
+    <ScreenContentWithDock
+      contentContainerStyle={[styles.contentContainer, {paddingTop: top}]}
+      dockContent={
+        error && <Text style={styles.wrongPass}>{t(m.wrongPass)}</Text>
+      }>
+      <Image source={require('../images/icon_mapeo_pin.png')} />
```
ErikSin, line 87: This should now be the CoMapeo logo!

src/frontend/screens/AuthScreen.tsx
```diff
@@ -74,53 +76,42 @@ export const AuthScreen = ({
...
-        {error && (
-          <Text style={[styles.wrongPass]}>
-            <FormattedMessage {...m.wrongPass} />
-          </Text>
-        )}
-      </KeyboardAvoidingView>
-    </ScrollView>
+  return (
+    <ScreenContentWithDock
+      contentContainerStyle={[styles.contentContainer, {paddingTop: top}]}
```
ErikSin, line 83: There doesn't seem to be any padding at the top. Can't see it in this screenshot, but it is right up against the bevel

image

ErikSin (approved):
Some minor styling issues, but can be merged when fixed.

src/frontend/Navigation/Stack/index.tsx
achou11: I think using `useLayoutEffect` causes issues here because what's happening inside this unfortunately needs the navigator to mount first. My understanding is that `useLayoutEffect` will attempt to run before that happens, which unfortunately doesn't play well with some of the navigation complexities this relies on.

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
achou11: ugh yeah this is an issue with the `BottomSheetModal/Content` implementation. Currently, it assumes that the title text usually won't take up a large amount of space, which isn't ideal but made the implementation a little more straightforward if i remember correctly. can probably fix it but would rather do that outside of this PR.

will argue in this case that most of that title should belong as the description, but also outside the scope of this PR and more of co-design concern. for example, the issue goes away if moving that second sentence to the description:

image

ErikSin: I think just put it in the second sentence in the description. I can't imagine anyone vehemently disagreeing and there needs to be an entire review of copy anyways, so that can be fixed then....

src/frontend/screens/AuthScreen.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/503/commits/3cdbf5089f64437e9956714a8dbf553974a42bd7

src/frontend/screens/AuthScreen.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/503/commits/3cdbf5089f64437e9956714a8dbf553974a42bd7

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/503/commits/d05439d699b2b48ec364873a222e37e054d2173a

PR #507, opened 2024-07-15
chore: fix type errors related to intl

achou11:
There were a couple of type errors related to deriving values from the intl-related json files. This addresses those issues and also updates the implementation for deriving the supported languages to be a little more efficient (fewer loops over potentially large in-memory data)

### DIFF
--- a/src/frontend/lib/intl.ts
+++ b/src/frontend/lib/intl.ts
@@ -1,10 +1,10 @@
-import messages from '../../../translations/messages.json';
-import languages from '../languages.json';
+import MESSAGES from '../../../translations/messages.json';
+import LANGUAGES from '../languages.json';
 
-export type TranslatedLocale = keyof typeof messages;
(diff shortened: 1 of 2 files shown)


src/frontend/lib/intl.ts
```diff
@@ -13,40 +13,58 @@ export interface LanguageName {
...
+      Object.keys(MESSAGES[locale]).length > 0;
+
+    if (!hasAtLeastOneTranslatedString) continue;
+
+    if (!isSupportedLanguageLocale(locale)) {
+      if (process.env.APP_VARIANT === 'development') {
+        console.warn(
+          `Locale "${locale}" is not available in CoMapeo (see \`src/frontend/languages.json\`)`,
+        );
+      }
```
achou11, line 36: figured it would be fine to update the message here a bit and also only have it logged in development builds. open to changing though

ErikSin (approved):

PR #508, opened 2024-07-15
chore: fix type issues related to non-existent audioRecordings prop

achou11:
Addresses a couple of type errors related to specifying a prop that doesn't exist for the `MediaScrollView` component.


cimigree (approved):
I noticed this when I was adding test ids... Thanks for cleaning this up for now.

PR #522, opened 2024-07-17
fix: navigate to app settings after changing language

### LINKED ISSUE #485: Navigate to App Setting after language change
Currently when a user selects a new language, they are navigated back to the Map screen. Instead of navigating the user to the map screen, navigate them back one screen (App Settings)

## To Do
- [ ] Navigate user to app settings after selecting a language

### Notion Issue
https://www.notion.so/digidem/jump-from-language-selection-to-main-screen-083962fa0aea439e9032d8eaea043936

achou11:
Fixes #485 

---

Preview:

https://github.com/user-attachments/assets/5cec9f71-f616-477f-9d44-fb68b55d6231

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import {createMapeoClient} from '@mapeo/ipc';
 import {AppNavigator} from './AppNavigator';
-import {IntlProvider} from './contexts/IntlContext';
 import {MessagePortLike} from './lib/MessagePortLike';
 import {initializeNodejs} from './initializeNodejs';
 import {PermissionsAndroid} from 'react-native';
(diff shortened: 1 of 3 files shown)


src/frontend/App.tsx
```diff
@@ -78,14 +77,12 @@ const App = () => {
   }, []);
 
   return (
-    <IntlProvider>
```
achou11, line 81: this is an extra intl provider that isn't needed (already set up in AppProviders)

src/frontend/contexts/IntlContext.tsx
```diff
@@ -30,7 +30,6 @@ export const IntlProvider = ({children}: {children: React.ReactNode}) => {
 
   return (
     <IntlProviderOrig
-      key={appLocale}
```
achou11, line 33: this was causing essentially the whole app to remount when switching languages. a fun one to debug and eventually find 😄

ErikSin (approved):
Thanks for helping me figure this out

PR #531, opened 2024-07-18
chore: make UI for app passcode settings screens more consistent with other parts of app

achou11:
Stacked on #503 

Noticed that the UI in these screens was very different compared to other parts of the app with similar designs. Believe this is an artifact of porting over directly from Mapeo Mobile, which may have had slightly different design specifications, but I would like it to be more consistent with how other similar screens are implemented (namely action buttons docked to the bottom of the screen, consistent spacing, etc).

In terms of technical changes:

- uses the shared `ScreenContentWithDock` component for relevant screens to ensure uniformity and consistency
- fixes an issue with injecting rich formatting into specific parts of a translation string. this requires using the [defaultRichTextElements](https://formatjs.io/docs/react-intl/components#defaultrichtextelements) functionality built into react-intl. currently only defines a `<bold>` tag.

---

Preview:

### Before

image

image

image

image

image

image

image

image

image

image

### After

image

image

image

image

image

image

image

image

image

image

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -259,7 +259,7 @@
     "message": "App Passcode allows you to add an additional layer of security by requiring that you enter a passcode in order to open the Mapeo app. You can define your own 5-digit passcode by turning on the feature below."
   },
   "screens.AppPasscode.PasscodeIntro.warning": {
-    "message": "**Please note that forgotten passcodes cannot be recovered!** Once this feature is enabled, if you forget or lose your passcode, you will not be able to open Mapeo and will lose access to any Mapeo data that has not been synced with other project participants."
+    "message": "<bold>Please note that forgotten passcodes cannot be recovered!</bold> Once this feature is enabled, if you forget or lose your passcode, you will not be able to open Mapeo and will lose access to any Mapeo data that has not been synced with other project participants."
   },
   "screens.AppPasscode.TurnOffPasscode.cancel": {
     "message": "Cancel"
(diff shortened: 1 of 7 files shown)


messages/en.json
```diff
@@ -256,7 +256,7 @@
     "message": "App Passcode allows you to add an additional layer of security by requiring that you enter a passcode in order to open the Mapeo app. You can define your own 5-digit passcode by turning on the feature below."
   },
   "screens.AppPasscode.PasscodeIntro.warning": {
-    "message": "**Please note that forgotten passcodes cannot be recovered!** Once this feature is enabled, if you forget or lose your passcode, you will not be able to open Mapeo and will lose access to any Mapeo data that has not been synced with other project participants."
+    "message": "<bold>Please note that forgotten passcodes cannot be recovered!</bold> Once this feature is enabled, if you forget or lose your passcode, you will not be able to open Mapeo and will lose access to any Mapeo data that has not been synced with other project participants."
```
achou11, line 259: may need to document this for translators (maybe in the`description` field?). not sure how it looks when it ends up in crowdin

ErikSin (approved):
Perhaps a little outside the scope of this PR. But we should consider wrapping `ScreenContentWithDock` in a `keyboardAvoidingView`. Luckily it is already acting like that, and pushing the buttons up when the keyboard is shown, but it might be good explicitly use the `keyboardAvoidingView` so it is more predictable in the future.

achou11:
> Perhaps a little outside the scope of this PR. But we should consider wrapping `ScreenContentWithDock` in a `keyboardAvoidingView`. Luckily it is already acting like that, and pushing the buttons up when the keyboard is shown, but it might be good explicitly use the `keyboardAvoidingView` so it is more predictable in the future.

yeah i was thinking about this recently and tbh, i have no idea why it works fine without it 😄

PR #533, opened 2024-07-22
fix: fix extension name allowed for project config file when importing

### LINKED ISSUE #518: Import Config not working
Importing config is not working

## To Do
- [x] Determine why importing config is not working
- [x] Fix config import

### Notion Issue
https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=c58a7b022e084c2bae2c4379bf0c67e7&p=4e09eecabb5047f7bac289ddd9b901a0&pm=s

achou11:
Fixes #518 

Forgot that we're updating the extension name of the config file when I initially implemented this 😅

Testing instructions:

1. Download config file mentioned in the notion issue onto the device.
2. Initiate project creation flow in app
3. Toggle advanced project settings and press import config button
4. Select config file from file picker
5. Press create project
6. Create observation. Notice categories are aligned with the imported config

---

Preview:

https://github.com/user-attachments/assets/e18907c9-541a-460e-91e7-45fce996f554

### DIFF
--- a/src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
+++ b/src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
@@ -130,7 +130,7 @@ export const CreateProject: NativeNavigationComponent<'CreateProject'> = ({
     if (!asset) return;
 
     // Only allow importing files with the desired extension
-    if (asset.name.endsWith('.mapeoconfig')) {
+    if (asset.name.endsWith('.mapeosettings')) {
       setConfigFileResult({type: 'success', file: asset});
     } else {
       setConfigFileResult({


src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
```diff
@@ -130,7 +130,7 @@ export const CreateProject: NativeNavigationComponent<'CreateProject'> = ({
     if (!asset) return;
 
     // Only allow importing files with the desired extension
-    if (asset.name.endsWith('.mapeoconfig')) {
+    if (asset.name.endsWith('.mapeosettings')) {
```
achou11, line 133: my understanding is that we're not supporting the old configurations (ending with `.mapeoconfig`) so I figured it made sense to only check for the new one instead of supporting both

cimigree (approved):
Everything works fine. Seems both simple and brittle. Hope we are communicating the file extension necessity effectively somewhere, but that is outside the scope of this PR.

PR #535, opened 2024-07-22
chore: enables editing of an observation

### LINKED ISSUE #461: Enable Editing of an observation
In an observation, the `createdBy` field does not reference a 'deviceId'. In otherwords, we cannot simply determine if a user created an observation by checking if `createdBy === deviceId`. We need to convert the `createdBy` field using [$createdByToDeviceId](https://github.com/digidem/mapeo-core-next/commit/14f27c4902cd7da5538eede2e4afb593a50bfbc3)

## To Do
- [x] In [deviceInfo](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/hooks/server/deviceInfo.ts) create `useCreatedByToDeviceId` hook as a tanstack wrapper
- [x]  In [Observation](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Observation/index.tsx) update [deviceId](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/Observation/index.tsx#L60) to be derived from [useDeviceInfo](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/hooks/useDeviceInfo.ts#L16)
- [x] In [Observation](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Observation/index.tsx) convert [createdBy](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/Observation/index.tsx#L61C20-L61C29) to a deviceId, using hook above
- [x] In [ObservationHeaderRight](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/Observation/ObservationHeaderRight.tsx#L39) convert [createdBy](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/Observation/ObservationHeaderRight.tsx#L39) to device Id
- [x] convert [isLoading to isPending](https://github.com/TanStack/query/discussions/6297) in [usage for useDeviceInfo](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/Observation/ObservationHeaderRight.tsx#L20)
- [x] Derive [loading state](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/Observation/ObservationHeaderRight.tsx#L29) from `isPending` of both `useDeviceInfo` and `useCreatedByToDeviceId`
- [x] In [ObservationList](https://github.com/digidem/comapeo-mobile/blob/main/src/frontend/screens/ObservationsList/ObservationListItem.tsx), there is a typo in [line 49](https://github.com/digidem/comapeo-mobile/blob/8a129c630a6a8d63e88f3217577c1df81671e751/src/frontend/screens/ObservationsList/ObservationListItem.tsx#L49). should be `isMine && styles.syncedObservation` not `!isMine`

cimigree:
closes #461 
Description:
- Adds a hook using tanstack for useCreatedByToDeviceId 
- Uses the hook in three screens/ components to convert the created by (wherever it happens to come from) to device Id
- Gets the device id from the useDeviceInfo hook in three observation files. Note, I could not use the 'react-native-device-info' deviceId as was mentioned in the issue, as that is the wrong format and instead used the one from our API
 - uses isPending instead of isLoading for both hook calls
 - fixed the typo in the ObservationListItem
 - Although it is not asked for, I used the deviceId and conversion of createdBy to check for isMine in the ObservationListItem for wherever it ends up being used. Not sure if this is desired and can remove it is not the right thing to do

Note:
These changes work well to edit an observation, but it leaves the button in place to change an observation category (preset) which comes from this file: src/frontend/screens/ObservationEdit/PresetView.tsx
If I attempt to change the preset, I get all kinds of errors. In this PR, should I make it so that Change is not there on an existing observation that is being edited? Or is that as expected? Or should it be a different issue/ PR?

Screen recording demoing editing:

Screen Recording demoing error:

### DIFF
--- a/src/frontend/hooks/server/deviceInfo.ts
+++ b/src/frontend/hooks/server/deviceInfo.ts
@@ -3,10 +3,8 @@ import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query';
 import {deviceType, DeviceType} from 'expo-device';
 
 import {useApi} from '../../contexts/ApiContext';
-import {useActiveProject} from '../../contexts/ActiveProjectContext';
 
 export const DEVICE_INFO_KEY = 'deviceInfo';
-export const CREATED_BY_TO_DEVICE_ID_KEY = 'createdByToDeviceId';
 
 export const useDeviceInfo = () => {
   const mapeoApi = useApi();
@@ -33,17 +31,6 @@ export const useEditDeviceInfo = () => {
     },
     onSuccess: () => {
       queryClient.invalidateQueries({queryKey: [DEVICE_INFO_KEY]});
-    },
-  });
-};
-
-export const useCreatedByToDeviceId = (createdBy: string) => {
-  const project = useActiveProject();
-
-  return useQuery({
-    queryKey: [CREATED_BY_TO_DEVICE_ID_KEY, createdBy],
-    queryFn: async () => {
-      return await project.$createdByToDeviceId(createdBy);
     },
   });
 };
--- a/src/frontend/hooks/server/deviceInfo.ts
+++ b/src/frontend/hooks/server/deviceInfo.ts
@@ -3,8 +3,10 @@ import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query';
 import {deviceType, DeviceType} from 'expo-device';
 
 import {useApi} from '../../contexts/ApiContext';
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
 
 export const DEVICE_INFO_KEY = 'deviceInfo';
+export const CREATED_BY_TO_DEVICE_ID_KEY = 'createdByToDeviceId';
 
 export const useDeviceInfo = () => {
   const mapeoApi = useApi();
@@ -31,6 +33,17 @@ export const useEditDeviceInfo = () => {
     },
     onSuccess: () => {
       queryClient.invalidateQueries({queryKey: [DEVICE_INFO_KEY]});
+    },
+  });
+};
+
+export const useCreatedByToDeviceId = (createdBy: string) => {
+  const project = useActiveProject();
+
+  return useQuery({
+    queryKey: [CREATED_BY_TO_DEVICE_ID_KEY, createdBy],
+    queryFn: async () => {
+      return await project.$createdByToDeviceId(createdBy);
     },
   });
 };
--- a/src/frontend/screens/Observation/ObservationHeaderRight.tsx
+++ b/src/frontend/screens/Observation/ObservationHeaderRight.tsx
@@ -8,11 +8,9 @@ import {useDraftObservation} from '../../hooks/useDraftObservation';
 import {EditIcon} from '../../sharedComponents/icons';
 import {SyncIcon} from '../../sharedComponents/icons/SyncIconCircle';
 import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
-import {
-  useDeviceInfo,
-  useCreatedByToDeviceId,
-} from '../../hooks/server/deviceInfo';
+import {useDeviceInfo} from '../../hooks/server/deviceInfo';
 import {UIActivityIndicator} from 'react-native-indicators';
+import {useCreatedByToDeviceId} from '../../hooks/server/projects.ts';
 
 export const ObservationHeaderRight = ({
   observationId,
@@ -41,8 +39,7 @@ export const ObservationHeaderRight = ({
     );
   }
 
-  const {deviceId} = deviceInfo || {};
-  const canEdit = createdByDeviceId === deviceId || false;
+  const canEdit = createdByDeviceId === deviceInfo?.deviceId;
 
   return canEdit ? (
     <IconButton onPress={handlePress} testID="editButton">
--- a/src/frontend/screens/Observation/ObservationHeaderRight.tsx
+++ b/src/frontend/screens/Observation/ObservationHeaderRight.tsx
@@ -8,7 +8,10 @@ import {useDraftObservation} from '../../hooks/useDraftObservation';
 import {EditIcon} from '../../sharedComponents/icons';
 import {SyncIcon} from '../../sharedComponents/icons/SyncIconCircle';
 import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
-import {useDeviceInfo} from '../../hooks/server/deviceInfo';
+import {
+  useDeviceInfo,
+  useCreatedByToDeviceId,
+} from '../../hooks/server/deviceInfo';
 import {UIActivityIndicator} from 'react-native-indicators';
 
 export const ObservationHeaderRight = ({
@@ -17,7 +20,10 @@ export const ObservationHeaderRight = ({
   observationId: string;
 }) => {
   const observationWithPreset = useObservationWithPreset(observationId);
-  const {data, isLoading} = useDeviceInfo();
+  const {data: createdByDeviceId, isPending: isCreatedByDeviceIdPending} =
+    useCreatedByToDeviceId(observationWithPreset.observation.createdBy);
+
+  const {data: deviceInfo, isPending: isDeviceInfoPending} = useDeviceInfo();
   const {editSavedObservation} = useDraftObservation();
   const navigation = useNavigationFromRoot();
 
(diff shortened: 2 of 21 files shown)


src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -33,20 +37,25 @@ function ObservationListItemNotMemoized({
...
+  const isMine = createdByDeviceId === deviceId;
+
   return (
     <TouchableHighlight
       onPress={() => onPress(observation.docId)}
       testID={testID}
       style={{flex: 1, height: 80}}>
       <View
-        style={[styles.container, style, !isMine && styles.syncedObservation]}>
+        style={[styles.container, style, isMine && styles.syncedObservation]}>
```
achou11, line 58: the `syncedObservation` style should be applied to observations from others, not your own

```suggestion
        style={[styles.container, style, !isMine && styles.syncedObservation]}>
```

src/frontend/hooks/server/deviceInfo.ts
```diff
@@ -35,6 +37,17 @@ export const useEditDeviceInfo = () => {
...
+export const useCreatedByToDeviceId = (createdBy: string) => {
+  const project = useActiveProject();
+
+  return useQuery({
+    queryKey: [CREATED_BY_TO_DEVICE_ID_KEY, createdBy],
+    queryFn: async () => {
+      return await project.$createdByToDeviceId(createdBy);
+    },
+  });
+};
```
achou11, line 49: nit: not sure if this belongs here, as device info is more about application-level information about a member's device. this is more about identity within a project, so i'd lean towards placing this in `server/projects`

src/frontend/screens/Observation/ObservationHeaderRight.tsx
```diff
@@ -35,8 +41,9 @@ export const ObservationHeaderRight = ({
     );
   }
 
-  const canEdit =
-    observationWithPreset.observation.createdBy === data?.deviceId || false;
+  const {deviceId} = deviceInfo || {};
+  const canEdit = createdByDeviceId === deviceId || false;
```
achou11, line 45: nit

```suggestion
  const canEdit = createdByDeviceId === deviceInfo?.deviceId;
```

src/frontend/screens/Observation/index.tsx
```diff
@@ -57,9 +61,15 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
   const {lat, lon, createdBy} = observation;
-  const isMine = deviceId === createdBy;
+  const {data: deviceInfo, isPending: isDeviceInfoPending} = useDeviceInfo();
+  const {deviceId} = deviceInfo || {};
+  const {data: convertedDeviceId, isPending: isCreatedByDeviceIdPending} =
+    useCreatedByToDeviceId(createdBy);
+  const isMine =
+    !isDeviceInfoPending &&
+    !isCreatedByDeviceIdPending &&
+    deviceId === convertedDeviceId;
```
achou11, line 72: nit: ideally this page would make use of the query statuses to wait before rendering relevant parts or passing data to components. for example, don't render the `ButtonFields` until data from both queries is available.

aside from that, i think this check can be more explicit to make sure that `deviceId` and `convertedDeviceId` are not undefined (at the TS level, at least):

```suggestion
  const {data: deviceInfo} = useDeviceInfo();
  const {data: convertedDeviceId} = useCreatedByToDeviceId(createdBy);
  const isMine =
    deviceInfo?.deviceId !== undefined &&
    convertedDeviceId !== undefined &&
    deviceInfo.deviceId === convertedDeviceId;
```

src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -33,20 +37,25 @@ function ObservationListItemNotMemoized({
   onPress = () => {},
 }: ObservationListItemProps) {
   const {preset} = useObservationWithPreset(observation.docId);
-  const deviceId = '';
+  const {data: deviceInfo} = useDeviceInfo();
+  const {deviceId} = deviceInfo || {};
```
achou11, line 41: nit: use optional chaining wherever `deviceId` is needed

```suggestion
```

achou11 (changes requested):
One blocking issue related to how items are rendered in the observation list screen

src/frontend/hooks/server/deviceInfo.ts
cimigree: Ok. I am ambivalent about it, so I moved it as you specified.

src/frontend/screens/Observation/index.tsx
cimigree: The only thing I found to conditionally render was the button fields, which I did. Is that what you had in mind? Were there other parts or components that were not being handled as you wished?

achou11: that was the only one that seemed relevant, thanks!

achou11 (commented):
Might be outside the scope of this PR but I'm noticing that once I enter the Edit Observation screen, the attachments do not load:

image

Seems to be an issue with how we're deriving photo draft state upon entering that screen ([link](https://github.com/digidem/comapeo-mobile/blob/de2d49e66d40e72350e1682c7cfd9c470ba918c1/src/frontend/hooks/persistedState/usePersistedDraftObservation/photosMethods.ts)). To put it shortly, we're [not keeping the blob uri for the photo](https://github.com/digidem/comapeo-mobile/blob/de2d49e66d40e72350e1682c7cfd9c470ba918c1/src/frontend/hooks/persistedState/usePersistedDraftObservation/photosMethods.ts#L52-L55) (see [`SavedPhoto`](https://github.com/digidem/comapeo-mobile/blob/de2d49e66d40e72350e1682c7cfd9c470ba918c1/src/frontend/hooks/persistedState/usePersistedDraftObservation/photosMethods.ts#L8-L14) type), which is the data that's eventually passed to that thumbnail previewer. without the uri, the thumbnail stays stuck in the loading state.

initial thought is to address that issue in this PR to be consistent with "not exposing broken features", but not sure how much work that would take. personally okay with it being a follow-up as long as it's addressed soon after.

achou11:
> If I attempt to change the preset, I get all kinds of errors. In this PR, should I make it so that Change is not there on an existing observation that is being edited? Or is that as expected? Or should it be a different issue/ PR?

haven't tested this myself yet, but wondering if this is because that ObservationEdit screen is a little outdated in terms of implementation? Erik introduced a shared `Editor` component that we ideally use for this screen (was an eventually planned follow-up task) - maybe using that would fix the issue you're noticing?

src/frontend/screens/Observation/index.tsx
```diff
@@ -57,9 +60,15 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
       }, defaultAcc)
     : [];
 
-  const deviceId = '';
   const {lat, lon, createdBy} = observation;
-  const isMine = deviceId === createdBy;
+  const {data: deviceInfo, isPending: isDeviceInfoPending} = useDeviceInfo();
+  let {data: convertedDeviceId, isPending: isCreatedByDeviceIdPending} =
```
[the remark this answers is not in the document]
cimigree: Oops! Good catch! Leftover from some testing

cimigree:
- Fixed the let
- Fixed the rendering of that style that shouldn't be flickering
- Fixed (as best as I could figure out, very happy to take feedback) the rendering of photos in the ObservationEdit screen. 
Here it rendering instead of the loader


Note and frustration:
If I go to create an observation, take a photo, add details, and then save the details or go back to the Observation screen, then I get the loader. But that is not fixed if I navigate to the ObservationCreate Screen or do navigation.goBack() from the ObservationFields screen instead of navigating to the ObservationEdit screen. Ugh. Any chance I could leave that edge case for a different issue? Once you save it, it works fine.

achou11 (approved):

PR #540, opened 2024-07-24
fix: detail navigation

### LINKED ISSUE #515: Fix Navigation from Details Screen
The navigation from the details screen is unpredictable. As seen in [this issue](https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=c58a7b022e084c2bae2c4379bf0c67e7&p=48f182ae20884e81985dff81581833fe&pm=s), pressing "done" navigates the user back to the ma, but I was unable to reproduce the error on my device

## To Do
- [ ] Determine and fix error

## Notion Issue
https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=c58a7b022e084c2bae2c4379bf0c67e7&p=48f182ae20884e81985dff81581833fe&pm=s

cimigree:
closes #515 

### Description
- adds a [package](https://docs.expo.dev/debugging/devtools-plugins/#react-navigation) to enable navigation debugging, which can make navigation which is sometimes obtuse, much clearer.
- Turns out when I was testing I was granting *all* permissions, even audio. 
- Because of the way the audio permissions file was set up, which is called from the thumbnail and action tab, if audio permission was granted, as soon as I got to the edit screen, I was being navigated back to the home screen
- Because we don't yet have audio, and I am not sure what is supposed to be happening, I just put comments in the code so that when we actually have that functionality, we can call it as needed. 
- In the meantime, this fixes the problem I was having, and the problem anyone who had granted permission for audio would have
- Because I am now on the ObservationEdit screen in the ObservationDetails test, I had to adjust a few things within the test

### DIFF
--- a/src/frontend/screens/ObservationEdit/ThumbnailAndActionTab.tsx
+++ b/src/frontend/screens/ObservationEdit/ThumbnailAndActionTab.tsx
@@ -67,11 +67,15 @@ export const ThumbnailAndActionTab: FC<ThumbnailAndActionTab> = ({
 
   const handleAudioPress = useCallback(() => {
     if (permissionResponse?.granted) {
-      navigation.navigate('Home', {screen: 'Map'});
+      // handle audio recording start
     } else {
       openAudioPermissionSheet();
     }
-  }, [navigation, openAudioPermissionSheet, permissionResponse?.granted]);
+  }, [openAudioPermissionSheet, permissionResponse?.granted]);
+
+  const handleAudioPermissionGranted = useCallback(() => {
+    // handle audio recording start after permission is granted
+  }, []);
 
   const bottomSheetItems = [
     {
@@ -108,6 +112,7 @@ export const ThumbnailAndActionTab: FC<ThumbnailAndActionTab> = ({
         closeSheet={closeAudioPermissionSheet}
         isOpen={isAudioPermissionSheetOpen}
         sheetRef={audioPermissionSheetRef}
+        onPermissionGranted={handleAudioPermissionGranted}
       />
     </>
   );
--- a/e2e/observations/add-details.yaml
+++ b/e2e/observations/add-details.yaml
@@ -9,8 +9,7 @@ env:
 - tapOn:
     id: '${SUBCAT}.add-observation-btn'
 - tapOn: '.*Gathering Site.*'
-- tapOn:
-    id: 'OBS.add-details-btn'
+- tapOn: '.*Details.*'
 - assertVisible:
     label: 'Add details button opens Details screen'
     id: 'OBS.add-details-scrn'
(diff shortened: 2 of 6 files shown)


src/frontend/screens/ObservationEdit/ThumbnailAndActionTab.tsx
```diff
@@ -108,6 +112,7 @@ export const ThumbnailAndActionTab: FC<ThumbnailAndActionTab> = ({
         closeSheet={closeAudioPermissionSheet}
         isOpen={isAudioPermissionSheetOpen}
         sheetRef={audioPermissionSheetRef}
+        onPermissionGranted={handleAudioPermissionGranted}
```
achou11, line 115: nit: i would lean towards not specifying this for now instead of having a no-op placeholder, especially since we don't have an immediate plan to address anytime soon.

achou11:
reading through the issue, i'm not convinced that this fully fixes it. i think they're reproducing that issue without touching audio permissions at all

achou11:
yeah I remember discussing this (with you i believe): the issue stems from this:

https://github.com/digidem/comapeo-mobile/blob/c2dcc5f98d3b36fce51dadb380309fa9d6996443/src/frontend/screens/ObservationFields/index.tsx#L99

if you're going through the observation creation flow, I think this will "break" and react navigation will defer to going back to "default" tab's screen, which is the map

achou11:
still think it's worth merging the PR here, but I would update the PR title to be more specific about the audio case and update the description so that the issue is not auto-closed

cimigree:
> Can you elaborate on this? Where are you granting all of the permissions? (e.g. test setup or in the app). 

It's in the test setup
```
onFlowStart:
  - launchApp:
      clearState: true
      stopApp: false
      permissions: {all: allow}
```

cimigree:
> reading through the issue, i'm not convinced that this fully fixes it. i think they're reproducing that issue without touching audio permissions at all

I am fully confident this fixes this specific issue. Who is "they"? I am the one who created this issue and that is screen recording from me.

cimigree:
> yeah I remember discussing this (with you i believe). i _think_ the issue stems from these:
> 
> https://github.com/digidem/comapeo-mobile/blob/c2dcc5f98d3b36fce51dadb380309fa9d6996443/src/frontend/screens/ObservationFields/index.tsx#L46
> 
> https://github.com/digidem/comapeo-mobile/blob/c2dcc5f98d3b36fce51dadb380309fa9d6996443/src/frontend/screens/ObservationFields/index.tsx#L99
> 
> if you're going through the observation creation flow, I think this will "break" and react navigation will defer to going back to "default" stack screen, which is the map

No, that is not what is happening. It may be the case that we don't want to navigate back to the ObservationEdit screen from the ObservationFields screen, but that is not why it is bouncing from the ObservationEdit screen to the home, map screen. It is definitely coming from AudioPermissions. I am extremely certain that is what caused this specific error. However, I am not extremely certain that the desired behavior is to navigate to the ObservationEdit screen from the details screen if someone is in the midst of creating an observation.

achou11:
> I am fully confident this fixes this specific issue. Who is "they"? I am the one who created this issue and that is screen recording from me.

sorry, didn't realize you had created the issue. i had assumed that this was caught by someone else during QA testing.

> No, that is not what is happening. It may be the case that we don't want to navigate back to the ObservationEdit screen from the ObservationFields screen, but that is not why it is bouncing from the ObservationEdit screen to the home, map screen. It is definitely coming from AudioPermissions. I am extremely certain that is what caused this specific error. However, I am not extremely certain that the desired behavior is to navigate to the ObservationEdit screen from the details screen if someone is in the midst of creating an observation.

yeah i see. I thought i was able to reproduce this without touching the audio permissions, which is why i thought that it was something else causing the issue. my apologies for not confirming before asserting.

the thing i'm highlighting is definitely a separate issue and something that should be addressed because it _could_ lead to weird navigation behavior, but not the one highlighted in the issue.

Feel free to merge as is whenever ready!

PR #544, opened 2024-07-25
chore: update @mapeo/core to 9.0.0-alpha.13

EvanHahn:
I tested this by launching the app.


ErikSin (approved):
Just for posterity, the `quickbit dynamic require` is no longer required (The app seems to be working so i suppose thats the case, but i dont remember if it was breaking the app)?

achou11:
> Just for posterity, the `quickbit dynamic require` is no longer required (The app seems to be working so i suppose thats the case, but i dont remember if it was breaking the app)?

it's in the readme entry that's been removed with the patch, but it was breaking the backend build process because rollup was complaining about it. otherwise had no effect on the app

PR #555, opened 2024-07-26
fix: MediaScrollView shows saved photos

### LINKED ISSUE #549: Edit Observation / Photos not loading
**Describe the bug**
Problem: When a user has a photo (or photos) in there observation and they go to edit them, only a spinning loader appears
User Goal: A user should be able to see their photos when they are editing an observation

See details : https://www.notion.so/digidem/Photos-not-loading-edit-observation-91aba555a63645dfb68626b9dbb03e9c

ErikSin:
Closes #549 

Updated the architecture of the photos and the media scroll view.

Previously, Mapeo did not have access to the types used in the back end. So the front end created the `SavedPhoto` Type. We ported this type over, but using this outdated type added a lot of complexity as we had to convert `Obervation.attachments` to this `SavedPhoto` type. Now that we have access to the backend types, I updated SavedPhoto to be `Observation["Attachments"]`. Now we no longer have to convert the type back an forth, which was causing confusing errors ans spaghetti code.

But this means that we had to convert all the component which used `Photo` (`type Photo = SavedPhoto & DraftPhoto`) or `SavedPhoto`. This PR converts those component, mainly media scroll view.

Media Scroll View now simply takes a `Photo[]` and maps each of these "photos" in the `PhotoThumbnail` component. The `PhotoThumbnail` component simply checks whether the photos is `SavedPhoto` or `DraftPhoto`. If it is a `DraftPhoto` the uri is accessible in memory (as the photo has just been taken), and produces an image with that uri. If it is a saved photo, it fetches the URI using `useAttachmentUrlQuery`. This `PhotoThumbnail` handles the loading and error state.

### DIFF
--- a/src/frontend/contexts/PhotoPromiseContext/types.ts
+++ b/src/frontend/contexts/PhotoPromiseContext/types.ts
@@ -1,17 +1,15 @@
 import {LocationObject} from 'expo-location';
+import {Attachment} from '../../sharedTypes';
 
 export const THUMBNAIL_SIZE = 400;
 export const THUMBNAIL_QUALITY = 30;
 export const PREVIEW_SIZE = 1200;
 export const PREVIEW_QUALITY = 30;
 
-export interface SavedPhoto {
-  // id of the photo in the Mapeo database
-  id: string;
-  type?: 'photo';
-  // If an image is to be deleted
+export type SavedPhoto = Omit<Attachment, 'type'> & {
+  type: 'photo';
   deleted?: boolean;
-}
+};
 
 export type UnprocessedDraftPhoto = {
   type: 'unprocessed';
--- a/src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
+++ b/src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
@@ -102,7 +102,7 @@ const UnsavedPhotoThumbnail = ({photo}: {photo: DraftPhoto}) => {
 };
 
 type PhotoThumbnailContextProps = {
-  onPress: () => any;
+  onPress?: () => void;
   style?: StyleProp<ViewStyle>;
   size?: number;
 };
--- a/src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
+++ b/src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
@@ -1,5 +1,4 @@
-import React, {FC} from 'react';
-import {Photo} from '../../contexts/PhotoPromiseContext/types';
+import React, {FC, createContext, useContext} from 'react';
 import {
   ActivityIndicator,
   Image,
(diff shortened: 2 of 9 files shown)


src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
```diff
@@ -13,38 +12,62 @@ import {
...
-  const photoProcessingError = photo && 'error' in photo && !!photo.error;
-  const isCapturing = photo && photo.type === 'unprocessed';
+  return (
+    <PhotoThumbnailContext.Provider value={{style, size, onPress}}>
+      {photo.type === 'photo' ? (
+        <SavedPhotoThumbnail photo={photo} />
+      ) : (
+        <UnsavedPhotoThumbnail photo={photo} />
+      )}
+    </PhotoThumbnailContext.Provider>
```
achou11, line 38: nit: personally, would lean towards not using context for this and instead either:

1. settle for a small amount of prop drilling
2. update the internal components so that they're easier to use without passing props excessively (probably more refactor-like work, similar to what i did in https://github.com/digidem/comapeo-mobile/pull/491)

src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
```diff
@@ -56,6 +79,46 @@ export const PhotoThumbnail: FC<PhotoThumbnailProps> = props => {
...
+    <PhotoThumbnailImage
+      isLoading={photo.type === 'unprocessed' && !('error' in photo)}
+      error={'error' in photo ? photo.error : undefined}
+      uri={'thumbnailUri' in photo ? photo.thumbnailUri : undefined}
+    />
+  );
+};
+
+type PhotoThumbnailContextProps = {
+  onPress: () => any;
```
achou11, line 105: adding the optional here makes more sense in the context of my suggestion for cleaning up the `MediaScrollView` component:

```suggestion
  onPress?: () => void;
```

src/frontend/sharedComponents/MediaScrollView/index.tsx
achou11, line 1: would take the opportunity to clean up the press handler code in this component as well by removing `handlePhotoPress()` entirely and instead inline the value for that `onPress` prop in the render body when mapping over the photos:

```tsx
photos
  ?.filter(photo => photo?.deleted == null)
  ?.map((photo, index) => {
    const onPress =
      photo.type === 'photo'
        ? () => {
            navigation.navigate('PhotoPreviewModal', {
              attachmentId: photo.driveDiscoveryId,
              observationId: observationId,
              deletable: false,
            });
          }
        : photo.type === 'processed'
          ? () => {
              navigation.navigate('PhotoPreviewModal', {
                deletable: true,
                originalPhotoUri: photo.originalUri,
              });
            }
          : undefined;

    return (
      <PhotoThumbnail
        key={index}
        photo={photo}
        style={styles.thumbnail}
        size={size}
        onPress={onPress}
      />
    );
  });
```

src/frontend/contexts/PhotoPromiseContext/types.ts
```diff
@@ -1,17 +1,15 @@
...
 export const PREVIEW_SIZE = 1200;
 export const PREVIEW_QUALITY = 30;
 
-export interface SavedPhoto {
-  // id of the photo in the Mapeo database
-  id: string;
-  type?: 'photo';
-  // If an image is to be deleted
+export type SavedPhoto = Omit<Attachment, 'type'> & {
+  type: 'photo';
```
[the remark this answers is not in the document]
achou11: tangential and outside the scope of this PR, but maybe the draft photo types should be renamed to `processed_draft` and `unprocessed_draft`, to be consistent with my suggestion above

achou11 (approved):
A few suggestions but otherwise helpful changes 👍

src/frontend/contexts/PhotoPromiseContext/types.ts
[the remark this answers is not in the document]
ErikSin: Generally I agree that we should have descriptive name. But the reason I think it makes sense to keep `type = "photo"` is to align with the attachment `type`. The main purpose of this PR was to get rid of the overhead of changing the types from the backend to the fronted. I wonder if there is an alternative way to being descriptive, without changing the `type`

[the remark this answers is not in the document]
achou11: ah right, forgot about that note. yeah then I think keep what you have here. my alternative suggestion would've been to rename the typescript type to indicate that it's a saved photo _attachment_, so that `type: photo` makes sense with that, but not worth sweating the details for now

src/frontend/sharedComponents/MediaScrollView/PhotoThumbnail.tsx
ErikSin: Also, I agree that contexts are a little unnecessary when only prop drilling 2 levels. And i actually originally designed it to use prop drilling. But doing a comparison of prop drilling and context, in this case, I actually think the context simplifies things. Because of the `useAttachmentUrlQuery` hook, we have to break each thumbnail into its own component. Both have the same shared prop. prop drilling is doubled since there are 2 component, even though the are the exact same thing. Since we are drillling it 2 levels, they are passed 4 times. It made the code feel unnecessarily verbose.

src/frontend/sharedComponents/MediaScrollView/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/555/commits/f792c751c563f743abd8e8cdf67788c71e59348c & https://github.com/digidem/comapeo-mobile/pull/555/commits/6bb37b5ca9400f0761e66908089cdda1c099749c

Issue #464, opened 2024-07-03
e2e: App Settings


ErikSin:
Create e2e for App Settings

## To Do:
- [ ] Implement any tests in [manual QA](https://www.notion.so/digidem/QA-Ft-Test-cebbb7aaf6bd404aa1ad5caf4a66bf6b)
- [ ] Document, in `docs/e2e`, any scripts that could not be implemented in e2e tests

PR #556, opened 2024-07-29
chore: E2e for app settings

cimigree:
addresses [this issue](https://github.com/digidem/comapeo-mobile/issues/464), but does not close. 

### Description
- Adds tests for several items in the settings screen: coordinates, about, and language. Based on this script:
https://www.notion.so/digidem/QA-Ft-Test-cebbb7aaf6bd404aa1ad5caf4a66bf6b
- Uses text wherever possible as selectors for the tests
- Adds documentation
- Tests for passcode will be in a separate PR and will complete the issue

Also, some tweaks for other things
- table of contents for observations that I forgot earlier
- Fixes the project creation test because there was a bug that when you created a project, the user did not used to go to a success screen, but now it does, so I adjusted the test and documentation to reflect that
- There is still a little bug there related to invite device, so that is commented out for now but I am working on it next, so hopefully can un comment it out soon

### Screen Recording

https://github.com/user-attachments/assets/409e2ea3-bde6-4c81-811f-328dceee04d2


docs/EndToEndTests/Settings.md
```diff
@@ -0,0 +1,56 @@
+## Settings
+
+- [Settings](#settings)
+  - [Overview](#overview)
```
achou11, line 4: this section seems to be missing in this document

docs/EndToEndTests/Settings.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Test Objectives
+
+- Clicking on App Settings shows option to change coordinate system
+- There are three coordinate systems visible
+- Back buttons work to return to settings screen
+
+### Special Considerations
+
+- It is not possible within Maestro to test the functioning of the Radio buttons, although theoretically I should have been able to. Settings is spelled setttings in the screen...
```
achou11, line 41: > Settings is spelled setttings in the screen

this seems like it's worth fixing in this PR, given that we're writing tests that check that.

achou11 (changes requested):
Changes overall looks good! Requesting changes mostly to fix the pre-existing spelling mistake, but let me know if you'd rather address that in a separate PR

docs/EndToEndTests/Settings.md
cimigree: Good catch! I think I copied over it accidentally when I added the table of contents.

docs/EndToEndTests/Settings.md
cimigree: There will be several file changes!

achou11: ok then a separate dedicated PR is preferable?

achou11: > There will be several file changes!

generally not worried about this if the nature of the changes are straightforward (i would imagine they are when it comes to a copy fix)

achou11 (approved):

Issue #547, opened 2024-07-26
Edit Observation / No picture saved


ximenabb:
**Describe the bug**
Testing July25-APK

- Editing an observation
- Adding a new picture while editing, see the image appear on the observation
- Click on save
- Go Back to the observation - no photo saved
- Click on edit again - no photo saved

See details :

https://www.notion.so/digidem/New-picture-in-edit-observation-doesn-t-save-7a6f65be70d54e04bfb59295c1258ec4

Issue #548, opened 2024-07-26
Edit Observation / Changing category wont saved


ximenabb:
**Describe the bug**
**Problem:** If I am editing an existing observation, and I change the category, then when I got to save it, I get a something went wrong screen and can’t do anything else. Even if I go back and change it to the original category.

**User Goal:** A user should be able to change the category associated with an observation
See details : https://www.notion.so/digidem/Changing-category-when-editing-an-observation-won-t-save-dc03512c43c54b89a75a5dd0c2f381b6

PR #557, opened 2024-07-29
fix: Saved observations properly editing

ErikSin:
closes: #547 and #548

This PR updates the `ObservationEdit` screen to use the Editor Component. This allows for the exact same UI as `observationCreate` with all its logic encapsulated inside the component itself. I also updated the photoPreviewModal as it was cuasing issues with Saved Observation. I refactored it to simply take a `Photo`.

### Notable changes:
- Added `existingObservationToDraft` function to `useDraftObservation`. Previously we were using "newDraft", where new draft would take an existing Observation or `undefined` as props. If props was an existing observation it would load the exisiting observation to persisted state. If it was `undefined` it would create a new 'blank' draft in persisted state. This `existingObservationToDraft` seperates the concerns
- `ObservationEdit` screen loads an exisiting observation to persisted draft. Previously when navigating to `ObservationEdit` Screen, we programatically loaded the exisiting observation to persisted state THEN navigated to the screen. This required the developer to make sure they were loading the observation everytime they navigated. Now the developer just needs to navigate to the `ObservationEdit` Screen and the component will retrieve the proper observation, and load it to persisted state, while also handling the loading state.
- PhotoPreviewModal simply takes a `SavedPhoto` or `ProcessedDraftPhoto`. In the case of a `SavedPhoto` (aka saved to the db) it retrieves the photo using useAttachmentURL hook, and deals with the loading and error state. In the case of a ProcessedDraftPhoto, it simply just show it the photo URL since it is saved in memory as the user has just take the photo


ErikSin:
Also, of the 22 "file changed", 6 are just deleted files, and 1 is a moved file

src/frontend/screens/ObservationFields/index.tsx
```diff
@@ -89,17 +101,22 @@ const DetailsHeaderRight = ({questionNumber}: {questionNumber: number}) => {
   const navigation = useNavigationFromRoot();
   const {usePreset} = useDraftObservation();
   const preset = usePreset();
+  const observationId = usePersistedDraftObservation(
```
cimigree, line 104: Is there a reason you can't pass the observationId as a prop from above?
And the preset?

cimigree (approved):
I did the best I could to be critical, but everything looks very nice and is so readable. The only issues I had is with some repeated code, that I marked. I am also wondering about the HeaderLeft in ObservationCreate and ObservationEdit that are just about identical. Is there any way to share that code? Or is that too complicated given things that may happen in the future?

I also tested thoroughly using my device, trying out all of the things I did before that were broken, and now they all work very smoothly!

src/frontend/screens/ObservationFields/index.tsx
ErikSin: No reason. I just used the hook for convenience!

PR #559, opened 2024-07-30
chore: e2e tests passcode

### LINKED ISSUE #464: e2e: App Settings

cimigree:
With, #556, closes #464 

### Description
- Adds tests, documentation, and test ids for the security feature within settings to address this notion script: https://www.notion.so/digidem/QA-Ft-Test-cebbb7aaf6bd404aa1ad5caf4a66bf6b
- Other settings stuff in a different PR
- Uses app text as much as possible
- Created subflows for repeated code
- I broke the different pieces of testing the passcode into different files to make the testing more manageable and readable

### Screen recordings
I made two because it was too long and too much to do just one
setting the passcode:
https://github.com/user-attachments/assets/cd06ee31-0422-476d-97f3-3ce38712bdd9

Testing having to use the passcode, and the features related to once you have a passcode:

https://github.com/user-attachments/assets/fc8a8e34-3185-459b-b255-0115934c3e7c


e2e/passcode/check-passcode-requirements.yaml
```diff
@@ -0,0 +1,40 @@
...
+    label: 'Passcode entry screen appears when user closes Mapeo app completely, and re-opens CoMapeo app'
+    text: 'Enter your passcode'
+- runFlow: './subflows/check-number-keyboard.yaml'
+- pressKey: Home
+- swipe:
+    direction: UP
+- tapOn: 'Clock'
+- pressKey: Home
+- swipe:
+    direction: UP
```
achou11, line 16: pretty cool that you can simulate this! the `tapOn: 'Clock'` is probably not necessary but not adamant about removing

e2e/passcode/post-passcode-setup.yaml
```diff
@@ -0,0 +1,71 @@
...
+- tapOn: 'App Passcode'
+- assertVisible:
+    label: 'From Security screen, [On click] App Passcode button prompts password entry'
+    text: 'Enter Passcode'
+- runFlow: './subflows/check-number-keyboard.yaml'
+- tapOn:
+    id: '${CAT}.header-back-btn'
+- assertVisible:
+    label: 'Back button returns to the Security Screen'
+    id: '${CAT}.security-scrn'
```
achou11, line 17: nit: is there a way to assert to a specific part of the screen? for example, think it could be preferable to strictly check the navigation header and make sure it says `Security` to know that you're on the correct screen.

e2e/passcode/post-passcode-setup.yaml
```diff
@@ -0,0 +1,71 @@
...
+- inputText: ${output.passcode}
+- assertVisible:
+    label: 'Next with correct passcode opens app passcode screen'
+    text: 'App Passcode'
+- assertVisible:
+    label: 'Screen displays abbreviated feature description'
+    text: '.*App passcode adds an additional layer of security.*'
+- assertVisible:
+    label: 'Use App Passcode checkbox in checked state'
+    id: 'SETTINGS.passcode-checked'
```
achou11, line 34: guessing the answer is no, but could these tests use the `checked` selector?

https://maestro.mobile.dev/api-reference/selectors

I'm guessing it doesn't because we're not using "native" radio elements in various cases (mostly just conditionally rendering icons, which is unfortunate 😬 )

src/frontend/sharedComponents/PasscodeInput.tsx
```diff
@@ -77,6 +77,7 @@ export const PasscodeInput = React.forwardRef<TextInput, PasscodeInputProps>(
     return (
       <CodeField
         {...codeFieldProps}
+        testID="SETTINGS.passcode-inp"
```
achou11, line 80: think this component should specify a prop and pass that here since this is a shared component and theoretically could be used on any screen, not just settings-related screens

achou11 (changes requested):

e2e/passcode/post-passcode-setup.yaml
achou11: seems like this could be possible using relative position selectors: https://maestro.mobile.dev/api-reference/selectors#relative-position-selectors

would require adding an selectable id to the navigation header i guess, if possible

e2e/passcode/post-passcode-setup.yaml
cimigree: Yes exactly. I wish...

e2e/passcode/check-passcode-requirements.yaml
cimigree: Ok. It was in the script that another app was opened, so that is why I did it...

e2e/passcode/post-passcode-setup.yaml
cimigree: I tried to use those with the radio button and they didn't work, but I can try again

cimigree: I am not really sure why you wanted this since I put a test id on the one screen and that is all it could be (but maybe there will be more screens in the future?) and it is a little wonky, but if that is what you want... It is now done.

src/frontend/sharedComponents/PasscodeInput.tsx
cimigree: Ok passed a prop.

e2e/passcode/post-passcode-setup.yaml
achou11: the primary motivation is to have the e2e test emulate the user experience more closely. using this case as an example, the user can know that they're returning to the security screen because they see the `Security` text in the header after the interaction of going back. asserting that the navigation header has that text is more reflective of that experience than asserting the presence of an opaque code-only id.

there are definitely cases where this doesn't work or is difficult to do, but i think it's worth considering the approach for the cases where it could be done.

achou11 (approved):

PR #566, opened 2024-07-31
chore: update @mapeo/core to 9.0.0-alpha.14

EvanHahn:


achou11:
Seeing the device type enum change that this introduces prompted me to look into code that potentially needs to handle this enum better. Seems like there are some changes that are needed in the app to handle this better (although they're not outright broken)

ErikSin (approved):
Thanks Evan!

PR #582, opened 2024-08-05
chore: update translations compilation script and use AST output

### LINKED ISSUE #539: Potentially address warning from  `@formatjs/intl` about not pre-compiling messages
I introduced the usage of `defaultRichTextElements` via #531. Since then, we get the following warning:

```sh
[@formatjs/intl] "defaultRichTextElements" was specified but "message" was not pre-compiled. 
Please consider using "@formatjs/cli" to pre-compile your messages for performance.
For more details see https://formatjs.io/docs/getting-started/message-distribution
```

We process messages for usage in the app but do so without using `@formatjs/cli` (via `build-translations.mjs`). Think there's a path for using the CLI with a [custom formatter](https://formatjs.io/docs/getting-started/message-distribution/#translation-management-system-tms-integration) (if one for crowdin + our specific needs doesn't exist), but would need to look further into it.

For now, doesn't seem to be a huge issue but would be nice to address if it's straightforward.

achou11:
Closes https://github.com/digidem/comapeo-mobile/issues/539

Updates the `build-translations` script in the following ways:

- rely more on `@formatjs/cli` and `@format/cli-lib` to turn extracted messages into app-ready translations
- output an AST instead of the normal message format. reasons for this are:
  
    1. it allows react-intl to skip work parsing the data at runtime, improving performance (https://formatjs.github.io/docs/guides/advanced-usage#pre-compiling-messages)
    
    2. ~it allows us to avoid including the relevant parser in the bundled app, which reduces bundle size (https://formatjs.io/docs/guides/advanced-usage#react-intl-without-parser-40-smaller)~ EDIT: had to punt on this because if you're missing message keys for a language, it will fail to parse the code-defined defaultMessage (which is not in AST form) verbatim and thus not actually interpolate placeholders values, which is not desired. Removed in a subsequent commit but would be nice to take advantage of this eventually.
    
   the downside of using the AST is that it's larger in size compared to regular string messages, although it can be efficiently compressed. This means that given the current approach of loading the translations into the app (i.e. all at once, not lazily), this will probably result in slightly more memory being used.
      
   I'm open to removing the change to using the AST, in which case the issue referenced would not be closed and this PR would solely be making our custom script rely more on `@formatjs` tooling.
  
- less boilerplate code to make it more linear in terms of reading (at the small cost of functional purity/modularity)

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -6,6 +6,7 @@ const defaultBlockList = Array.isArray(config.resolver.blockList)
   ? config.resolver.blockList
   : [config.resolver.blockList];
 
+/** @type {import('expo/metro-config').MetroConfig} */
 module.exports = {
   ...config,
   transformer: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -104,6 +104,7 @@
         "@eslint/eslintrc": "3.3.0",
         "@eslint/js": "9.21.0",
         "@formatjs/cli": "6.6.1",
+        "@formatjs/cli-lib": "6.6.6",
         "@mapeo/mock-data": "2.1.5",
         "@react-native/metro-babel-transformer": "0.76.6",
         "@react-native/metro-config": "0.76.6",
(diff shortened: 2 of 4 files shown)


achou11:
Something i'm unsure about is this [note](https://formatjs.io/docs/guides/advanced-usage/#pre-compiling-messages) in the docs:

```
Since this approach uses AST as the data source, changes to @formatjs/icu-messageformat-parser's AST will require cache invalidation.
```

not sure if that means we need some code to handle this or if that's just done internally and they're just letting us know 🤔 

EDIT: I don't think this affects us because we generate the AST at build time and it's never persisted beyond the lifetime of the app. This would be more of a problem if we were storing the AST and loading it into memory in the case that the app uses some updated AST format introduced by updating the library

achou11:
~CI error due to incorrectly formatted translation that made it into develop~ EDIT: resolved after latest rebase on develop

https://github.com/digidem/comapeo-mobile/blob/bcca1fb716d75a765c8f05c4dcf878edad6475e0/messages/es.json#L725

PR #576, opened 2024-08-01
fix: prevent queries using project api from using stale instance

achou11:
Fixes #545 

Before, all queries that were using the project api instance (provided by the `useProject()` hook and accessed via the `useActiveProject()` hook) were prone to:

- working with an outdated instance, which could lead to incorrect results when fetching updates. 
- not refetching after a new project instance is being used. This is due to many of the queries using a `queryKey` that provided no indication for react-query to trigger refetches after setting a new project instance (e.g. after creating a project). 

The main change that this PR introduces is updating the return type of `useProjects()` and `useActiveProjects()` hooks such that they provide the associated project id, which in turn is mostly used for specifying the `queryKey` option for various relevant queries. Refrained from any major adjustments to existing implementations, aside from some types-enabled changes.

Haven't done a thorough testing of the app, nor any additional work to see if anything else can take advantage of the updated return type (which is probably possible, but can be done as a follow-up). 

---

Preview:

https://github.com/user-attachments/assets/59a8b2d7-3006-465d-90fc-2dd6607fe347

PR #584, opened 2024-08-06
fix: fix query cache invalidation after editing observation

achou11:
I introduced a bug via #576 where after editing an observation, the changes are not refetched afterwards. This is due to the query key of the query cache of interest being multiple levels deep, as opposed to the other invalidations being one level. Checked other calls of cache invalidation to make sure that this wouldn't be an issue in other cases.

More of an aside, but I wonder if the query keys should technically be ordered so that the project id is the first path component (similar to an HTTP API) e.g. `queryKey: [projectId, 'observations', observationId]`. Perhaps something to do as a follow-up (or handle this in `@comapeo/react`)

### DIFF
--- a/src/frontend/hooks/server/observations.ts
+++ b/src/frontend/hooks/server/observations.ts
@@ -50,7 +50,7 @@ export function useCreateObservation() {
 
 export function useEditObservation() {
   const queryClient = useQueryClient();
-  const {projectApi} = useActiveProject();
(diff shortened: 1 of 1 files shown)


ErikSin:
Hmm, Im actually a little confused why this matters. When editting an observation, why should the projectId be used as a query key? The project isn't changing, right?

PR #590, opened 2024-08-08
chore: handle sync fetching in UI

### LINKED ISSUE #585: Viewing stale data after syncing
We currently don't invalidate any query caches when data is updated (in the background) via syncing. This leads to cases where certain parts of UI are outdated because they are displaying stale data. An example of this happening is navigating to screens that are part of the same stack navigation (e.g. the map) as the sync screen after syncing

Short-term suggestion is to invalidate all relevant queries when leaving the sync screen.

### LINKED ISSUE #586: Expose observation fetching state

If the user syncs a large amount of data and navigates to the observation list screen, they will see "stale" data (e.g. data from before the sync) with no indication that new data from the sync is loading. The data will appear after the new data from the sync is indexed.

When the user navigates to the observation list screen, the screen mounts, which triggers a refetch. Because we are using Suspense queries, refetching will not throw to the suspense boundary, so the user will not know that a refetch is in progress. Whenever new data is received from sync, it needs to be indexed. We don't yet have measurements about how long this can take, but for large amounts of data it could be several seconds. Any refetch will wait for indexing to complete before returning, so the user can be left looking at stale data for a while, not realizing a refetch is occuring in the background.

Display some kind of progress indicator during background refetch, based on the `isFetching` value from `useQuery`. This could be a spinner at the top of the observation list, or we could replace the entire list with a loading indicator (e.g. we could throw to the suspense boundary based on `isFetching` to show the same indicator displayed during initial load).

ErikSin:
closes #586 
touches on #585 

Properly invalidates observations cache when sync is complete (now uses `projectId` as part of the query key that is being invalidated)

Shows a loader at the top of the observations list when observations are being fetched

image

### DIFF
--- a/src/frontend/screens/Sync/ProjectSyncDisplay.tsx
+++ b/src/frontend/screens/Sync/ProjectSyncDisplay.tsx
@@ -83,7 +83,7 @@ export const ProjectSyncDisplay = ({
 }) => {
   const {formatMessage: t} = useIntl();
 
-  const {projectApi} = useActiveProject();
+  const {projectApi, projectId} = useActiveProject();
   const queryClient = useQueryClient();
   const navigation = useNavigationFromRoot();
   const {connectedPeers, data, initial} = syncState;
@@ -93,11 +93,11 @@ export const ProjectSyncDisplay = ({
   React.useEffect(() => {
     const unsubscribe = navigation.addListener('beforeRemove', () => {
       projectApi.$sync.stop();
-      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY]});
+      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY, projectId]});
     });
 
     return unsubscribe;
-  }, [navigation, projectApi, queryClient]);
+  }, [navigation, projectApi, queryClient, projectId]);
 
   const isDataSyncEnabled = data.isSyncEnabled;
 
--- a/src/frontend/screens/ObservationsList/index.tsx
+++ b/src/frontend/screens/ObservationsList/index.tsx
@@ -16,6 +16,7 @@ import {Loading} from '../../sharedComponents/Loading';
 import {TrackListItem} from './TrackListItem';
 import {useObservations} from '../../hooks/server/observations';
 import {useTracks} from '../../hooks/server/track';
+import {UIActivityIndicator} from 'react-native-indicators';
 
 const m = defineMessages({
   loading: {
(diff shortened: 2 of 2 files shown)


achou11:
Not sure if this PR closes #585 - that issue is also referring to syncable data that isn't just related to observations e.g information about presets, other project members, etc

EDIT: will try testing the specific case of newly synced observations showing up on the map when navigating there from the sync screen, since that was one of the notable examples.

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -93,11 +93,11 @@ export const ProjectSyncDisplay = ({
   React.useEffect(() => {
     const unsubscribe = navigation.addListener('beforeRemove', () => {
       projectApi.$sync.stop();
-      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY]});
+      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY, projectId]});
```
achou11, line 96: > Properly invalidates observations cache when sync is complete (now uses projectId as part of the query key that is being invalidated)

this is already happening without this change because of how react query matches query keys i.e `[OBSERVATION_KEY]` will include queries with `[OBSERVATION_KEY, projectId]`. doesn't hurt to keep this change though since being more specific is generally safer, but wouldn't expect it to change anything unless we had queries that had query keys starting with `OBSERVATION_KEY` but had some other path components following it (currently don't based on a quick check of existing codebase).

ErikSin:
So I was in the understanding was that the observations weren't properly showing up on the map and that it was thought that it was a cache issue.

Right now, when a user syncs and navigates back to the map, all the new observations that were synced are showing up on the map. It seems like my changes may not be the reason that was happening as I believe that it was happening before hand (as you pointed out that the cache invalidation for observations was already correct)

achou11:
> So I was in the understanding was that the observations weren't properly showing up on the map and that it was thought that it was a cache issue.
> 
> Right now, when a user syncs and navigates back to the map, all the new observations that were synced are showing up on the map. It seems like my changes may not be the reason that was happening as I believe that it was happening before hand (as you pointed out that the cache invalidation for observations was already correct)

hm yeah I think maybe the issue about the observations on the map being outdated was already solved then? don't remember the timeline of when that issue was reported vs the cache invalidation that exists when leaving the sync screen (I think that was added recently?)

happy to update the issue in question to be clearer if that's helpful, but still think it's talking about something of greater scope than what this PR is addressing, so think it'd still be reasonable to keep open

achou11 (approved):

ErikSin:
> on't remember the timeline of when that issue was reported vs the cache invalidation that exists when leaving the sync screen (I think that was added recently?)

yeah, it was the latest build that this was happening...but I flagged it in the QA channel as something I am not able to replicate....so i guess we can just leave it for now.

Ill update the description so #585 is not closed. But could you list the things that you had in mind for us to invalidate?

PR #592, opened 2024-08-12
feat: add setting to generate fake observations

### LINKED ISSUE #579: Proposal/Idea: Create mock data for testing

As I understand it, a time-consuming aspect of QA testing CoMapeo is collecting data. Actually testing data collection is important, but as I understand it, collecting data is a pre-requisite for other tests such as sync, editing and viewing data on the map.

Because installing test versions of CoMapeo often involves deleting existing data, this can be time-consuming for a tester. They need to collect lots of data before being able to test sync.

I would like a new menu item under "Project Settings" that only appears for test builds of the app. The menu item should contain a button "generate mock observations", possibly with an input box to enter a number of observations to create (maybe default to 20?).

This menu item should internally call `project.observation.create()`. Ideally:

- Location is randomized within a 10-50km radius/bbox of the users current location
- Categories are randomized
- Device name is added to the "notes" field to aid debugging
- Observations have a random (0 to 3) number of photos attached. Suggest: download ~10 random images from unsplash.com or similar service and then attach these (we might hit rate limits, so maybe we host some random images ourselves).

QA testers can keep collecting data as part of testing.

achou11:
Closes #579

- Adds screen in settings that provides interface for generating fake observations.
  - **~note that this is only accessible for test and development builds of the app.~ You need to have an env variable named `EXPO_PUBLIC_FEATURE_TEST_DATA_UI` specified to enable**. 
- Input parameters are:
  - count (required)
  - bounding distance in kilometers (optional, default is 50 km)
- randomized categories based on active config's presets
- adds note for each observation detailing which device created it
- at the moment, none of the displayed copy is marked as translatable. Wasn't sure if this is something that would be used by people outside the org, so didn't want these strings to end up in crowdin if external users will never see these. open to changing if we think otherwise.
- getting "proper" number inputs in React Native are a little tedious, hence the somewhat odd code 😄 
- DOES NOT implement attachments as described in the issue for now. plan is for that to be a follow-up

TODOs:

- [ ] Not sure what to include in the observation metadata. right now I only have the `mocked` field specified. might make sense to include the other information about the position, but wanted to get feedback before doing so

---

https://github.com/user-attachments/assets/fd29dfe6-c008-43b7-bfe1-12352eacdcef

### DIFF
--- a/src/frontend/Navigation/Drawer.tsx
+++ b/src/frontend/Navigation/Drawer.tsx
@@ -194,6 +194,17 @@ const DrawerContent = ({navigation}: DrawerContentComponentProps) => {
           <DrawerListItemIcon iconName="info-outline" />
           <ListItemText primary={<FormattedMessage {...m.aboutCoMapeo} />} />
         </ListItem>
+        {(process.env.APP_VARIANT === 'development' ||
+          process.env.APP_VARIANT === 'test') && (
+          <ListItem
+            onPress={() => {
+              navigate('CreateTestData');
+            }}
+            testID="settingsCreateTestDataButton">
+            <DrawerListItemIcon iconName="auto-fix-high" />
+            <ListItemText primary="Create Test Data" />
+          </ListItem>
+        )}
       </List>
     </DrawerContentScrollView>
   );
--- a/src/frontend/Navigation/Drawer.tsx
+++ b/src/frontend/Navigation/Drawer.tsx
@@ -194,8 +194,7 @@ const DrawerContent = ({navigation}: DrawerContentComponentProps) => {
           <DrawerListItemIcon iconName="info-outline" />
           <ListItemText primary={<FormattedMessage {...m.aboutCoMapeo} />} />
         </ListItem>
-        {(process.env.APP_VARIANT === 'development' ||
-          process.env.APP_VARIANT === 'test') && (
+        {process.env.EXPO_PUBLIC_FEATURE_TEST_DATA_UI && (
           <ListItem
             onPress={() => {
               navigate('CreateTestData');
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -65,6 +65,7 @@ import {
   createNavigationOptions as createObservationCreateNavigationOptions,
 } from '../../screens/ObservationCreate';
 import {AboutSettings} from '../../screens/Settings/About';
+import {CreateTestDataScreen} from '../../screens/Settings/CreateTestData';
 
 export const TAB_BAR_HEIGHT = 70;
 
@@ -272,5 +273,14 @@ export const createDefaultScreenGroup = ({
       component={AboutSettings}
       options={{headerTitle: intl(AboutSettings.navTitle)}}
     />
+
+    {(process.env.APP_VARIANT === 'development' ||
+      process.env.APP_VARIANT === 'test') && (
+      <RootStack.Screen
+        name="CreateTestData"
+        component={CreateTestDataScreen}
+        options={{headerTitle: 'Create Test Data'}}
+      />
+    )}
   </RootStack.Group>
 );
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -274,8 +274,7 @@ export const createDefaultScreenGroup = ({
       options={{headerTitle: intl(AboutSettings.navTitle)}}
     />
 
-    {(process.env.APP_VARIANT === 'development' ||
-      process.env.APP_VARIANT === 'test') && (
+    {process.env.EXPO_PUBLIC_FEATURE_TEST_DATA_UI && (
       <RootStack.Screen
         name="CreateTestData"
         component={CreateTestDataScreen}
--- a/src/frontend/screens/Settings/CreateTestData.tsx
+++ b/src/frontend/screens/Settings/CreateTestData.tsx
@@ -242,7 +242,7 @@ function useCreateFakeObservationsMutation() {
         projectApi.preset.getMany(),
       ]);
 
-      const deviceName = deviceInfo.name;
+      const notes = deviceInfo.name ? `Created by ${deviceInfo.name}` : null;
 
       const distanceBufferDegrees = lengthToDegrees(distance, 'kilometers');
 
@@ -254,8 +254,6 @@ function useCreateFakeObservationsMutation() {
         longitude + distanceBufferDegrees,
         latitude + distanceBufferDegrees,
       ] satisfies BBox;
-
-      const notes = deviceName ? `Created by ${deviceName}` : null;
 
       const promises = [];
 
--- a/src/frontend/screens/Settings/CreateTestData.tsx
+++ b/src/frontend/screens/Settings/CreateTestData.tsx
@@ -1,0 +1,305 @@
+import {Preset} from '@mapeo/schema';
+import {useMutation, useQueryClient} from '@tanstack/react-query';
+import {lengthToDegrees} from '@turf/helpers';
+import {randomPosition} from '@turf/random';
+import {LocationObject} from 'expo-location';
+import {type BBox} from 'geojson';
+import React, {forwardRef} from 'react';
+import {Controller, useForm} from 'react-hook-form';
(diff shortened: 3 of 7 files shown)


src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+        const value = {
+          attachments: [],
+          lon: fakeCoordinates[0],
+          lat: fakeCoordinates[1],
+          metadata: {
+            position: {
+              mocked: !!location.mocked,
+            },
+          },
```
achou11, line 277: open question: what else should be specified here?

src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -0,0 +1,305 @@
...
+import {UIActivityIndicator} from 'react-native-indicators';
+
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
+import {useDeviceInfo} from '../../hooks/server/deviceInfo';
+import {OBSERVATION_KEY} from '../../hooks/server/observations';
+import {usePresetsQuery} from '../../hooks/server/presets';
+import {useLocation} from '../../hooks/useLocation';
+import {LIGHT_GREY, RED, WHITE} from '../../lib/styles';
+import {Button} from '../../sharedComponents/Button';
+import {LocationView} from '../../sharedComponents/Editor/LocationView';
```
achou11, line 17: not ideal but thought it would be useful to have this info on this screen and didn't want to re-implement this

src/frontend/Navigation/Drawer.tsx
```diff
@@ -194,6 +194,17 @@ const DrawerContent = ({navigation}: DrawerContentComponentProps) => {
           <DrawerListItemIcon iconName="info-outline" />
           <ListItemText primary={<FormattedMessage {...m.aboutCoMapeo} />} />
         </ListItem>
+        {(process.env.APP_VARIANT === 'development' ||
+          process.env.APP_VARIANT === 'test') && (
```
ErikSin: I think we should make this its own flag. I can imagine scenarios where we want a test/development build that does not have this feature

src/frontend/Navigation/Drawer.tsx
```diff
@@ -194,6 +194,17 @@ const DrawerContent = ({navigation}: DrawerContentComponentProps) => {
...
+          <ListItem
+            onPress={() => {
+              navigate('CreateTestData');
+            }}
+            testID="settingsCreateTestDataButton">
+            <DrawerListItemIcon iconName="auto-fix-high" />
```
ErikSin, line 203: I like you choice of icon!

src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+const DISTANCE_BUFFER_KM = 50;
+
+const BASE_NUMBER_INPUT_RULES = {
+  min: 1,
+};
+
+export function CreateTestDataScreen() {
+  const presetsQuery = usePresetsQuery();
+  const locationState = useLocation({maxDistanceInterval: 0});
```
ErikSin, line 28: i think it would be better to`useLastKnowLocation` here. Having a subscription seems a little overkill here.

src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -0,0 +1,305 @@
...
+import {Text} from '../../sharedComponents/Text';
+
+const DISTANCE_BUFFER_KM = 50;
+
+const BASE_NUMBER_INPUT_RULES = {
+  min: 1,
+};
+
+export function CreateTestDataScreen() {
+  const presetsQuery = usePresetsQuery();
```
ErikSin: nit: this query is only being used by `useCreateFakeObservations`. Instead of using the query, just use the `projectApi.preset.getMany()` directly inside the `useCreateFakeObservations`

src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -272,5 +273,14 @@ export const createDefaultScreenGroup = ({
       component={AboutSettings}
       options={{headerTitle: intl(AboutSettings.navTitle)}}
     />
+
+    {(process.env.APP_VARIANT === 'development' ||
+      process.env.APP_VARIANT === 'test') && (
```
ErikSin: same comment as above, i think we should make this its own flag

ErikSin (changes requested):
As discussed with you on a call, i am unable to see the feature in the drawer. But i think changing the feature flag to be manually set by the developer should fix this problem.

src/frontend/Navigation/Drawer.tsx
achou11: haha thank you! i spent way too much time choosing 😂

src/frontend/screens/Settings/CreateTestData.tsx
achou11: what should be done in the case that it returns `null`? unlike `useLocation()`, we'll be stuck with that value since it's not subscription-based. in that case we'd potentially have to allow generating observations that don't have a location, which is definitely valid but will require some more work around how the implementation currently works.

achou11: maybe i should already support observations without location anyways? just didn't want add more complex UI than necessary if that wasn't going to be useful initially. maybe can be done as a follow-up (in addition to the potentially using `useLastKnownLocation()` instead)

src/frontend/screens/Settings/CreateTestData.tsx
achou11: ah good point. can probably do the same and not use `useDeviceInfo()`

achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/592/commits/fc6e45eb5437185ddf60babd0410ae083ca2a514

src/frontend/Navigation/Drawer.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/592/commits/cc0467ea90a4068ece7310789faae6896dc07555

src/frontend/Navigation/Stack/AppScreens.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/592/commits/cc0467ea90a4068ece7310789faae6896dc07555

achou11:
Updated the feature flagging such that you need to have `EXPO_PUBLIC_FEATURE_TEST_DATA_UI` specified (whether in `.env` or `eas.json` depending on context). Open to suggestions about the naming 😄

src/frontend/screens/Settings/CreateTestData.tsx
ErikSin: do we not have this problem with useLocation() as well? Can `useLocation()` also be null? Its not the biggest deal, i think if there is too much work in changing it, happy to just leave it as is

achou11: yeah we could run into the same issue in that case, hence the note about potentially supporting creating observations without a location. at least in that case, there's a chance that the location will become a valid value though since it's subscription-based.

> Its not the biggest deal, i think if there is too much work in changing it, happy to just leave it as is

yeah i think for now it's fine to keep as is but could definitely be a follow-up!

src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -0,0 +1,292 @@
...
+                count: data.count,
+                location: locationState.location,
+                distance:
+                  data.distance === undefined
+                    ? DISTANCE_BUFFER_KM
+                    : data.distance,
+              },
+              {
+                onSuccess: () => {
+                  ToastAndroid.show('Observations created', ToastAndroid.SHORT);
```
ErikSin, line 65: i think we should also just navigate back to the home screen here as well

ErikSin (approved):
Looks good!

src/frontend/screens/Settings/CreateTestData.tsx
achou11: going to leave out for now but can easily follow up if QA asks for it

PR #593, opened 2024-08-12
fix: use UTC date, which could be different from local date

EvanHahn:
`formatIsoUtc` formats a `Date` in UTC. It's possible that the day (such as the 12th of the month) is different in local time versus UTC time. This makes it use UTC, as expected.


EvanHahn:
See #595.

PR #594, opened 2024-08-12
feat: onboarding screens with data and privacy info

### LINKED ISSUE #542: Onboarding Screen / Update welcome flow
## Description
With metrics, we are now introducing a Data Privacy screen at onboarding. Once passed this screen, user will be taken to the usual Device Naming screen during the onboarding.

## Prototype
https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2081-1527&t=AsoPG0QgjEqOFiQI-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=2081%3A1527

## Screen 1 - Onboarding

Screen Shot 2024-07-25 at 10 24 09 AM

- [x] Update screen as designed
- [x] Integrate copy as designed
- [x] (Icon) _Map anywhere and everywhere_
- [x] (Icon) _Collaborate with others_
- [x] (Icon) _Own and control your data_
- [x] (Icon) _Designed with and for Indigenous peoples & frontline communities_

## Screen 2 - Data Privacy
Screen Shot 2024-07-25 at 10 24 59 AM

- [x] Create screen as designed
- [x] Integrate copy as designed
- [x] _Data & Privacy_ - _CoMapeo allows you and your collaborators to map offline without needing servers._ 
- [x] _Your data stays on your devices._
- [x] _All data stays fully encrypted_
- [x] _Easily manage and control sharing and collaboration._ 
- [x] _Private by default — diagnostic information is fully anonymized and you can opt-out any time._
- [x] Create `Learn More` and `Next` button
- [x] When clicked `Learn More` button directs to **Screen 3 - Privacy Policy**
- [x] When clicked `Next` button directs to Device Naming https://github.com/digidem/mapeo-mobile/issues/1117

## Screen 3 - Privacy Policy
Screen Shot 2024-07-25 at 10 25 19 AM
Screen Shot 2024-07-25 at 10 25 38 AM

- [x] Integrate copy as designed in the prototype https://docs.google.com/document/d/1YZOWoS747h4RFl7Jpq8KmnsXd4jFZFz7UAqPT9Kxlcs/edit#heading=h.15by51eu9vfj
- [x] Integrate checkmark, optional to click
- [x] If clicked, collect data
- [x]  When clicked back arrow directs to the Data Privacy screen

## Tech notes
@EvanHahn 's notes https://digidem.slack.com/archives/C01GJ3V8D62/p1722275211691429

## Design
https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/%22Ready-for-Dev%22-Hand-off-Assets?node-id=934-273&t=eiLnXjdi133AXR2g-0

cimigree:
closes #542 

### Description
- Adds new onboarding screens that describe the collection of data and optionally allows someone to opt into having it collected
- Just FYI 11 files are just svgs... so only 15 real files to review ;0)
- I am doing the Zustand storage for the permissions in the next PR because this one is already so big...
- Text comes from this [google doc](https://docs.google.com/document/d/1YZOWoS747h4RFl7Jpq8KmnsXd4jFZFz7UAqPT9Kxlcs/edit#heading=h.740x84mgpr7w), not the prototype per Ximena's direction
- Here is the [prototype](https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2280-1517&t=KwuNJZ5DlKYqJx2d-0&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=2081%3A1527) for quick reference
- The casing and punctuation comes from the prototype and google doc and are not my choice
- Makes the last screen reusable because it is also going to be accessible from the settings (but not the checkmark at the bottom)
- I moved the onboarding screens to a new folder
- I put the PrivacyPolicy in a folder but not a shared component because that just didn't seem right...
- I put the styles in their own file, which I know you don't usually do, but there were sooo many...
- I took the logic out of App.tsx and put it in with the OnboardingPrivacyPolicy screen temporarily (will deal with this in next PR) 

### Screenshots
**New Intro screen**
<image src="https://github.com/user-attachments/assets/ab2c3623-c293-4476-9ce7-bb47f8bd4de4" width="250" />
**Next Screen**
<image src="https://github.com/user-attachments/assets/fa4d6c97-2256-415a-9a27-416830785384" width="250" />
**Scrollable Screen** (just included first view)
<image src="https://github.com/user-attachments/assets/a0416616-a27c-4dcd-be22-ef0ecdcb0e94" width="250" />


src/frontend/screens/Onboarding/DataPrivacy.tsx
```diff
@@ -0,0 +1,188 @@
...
+        </Button>
+      </View>
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: width * 0.05,
```
achou11: nit: usually would lean towards constant values for this kind of styling. what's the reasoning for doing this based on the window/screen dimensions?

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export const IntroToCoMapeo = ({
+  navigation,
+}: NativeStackScreenProps<DeviceNamingParamsList, 'IntroToCoMapeo'>) => {
+  const {formatMessage} = useIntl();
+  return (
+    <View style={styles.container}>
+      <TopoBackground style={styles.background} />
+      <View style={styles.content}>
+        <CoMapeoTextAsSVG width={'95%'} height={height * 0.12} />
```
achou11: nit: not fully sure based on the designs, but wondering if this should be vertically centered in relation to the whole screen.

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
```diff
@@ -0,0 +1,144 @@
...
+      'Designed with and for Indigenous peoples & frontline communities',
+  },
+});
+
+export const IntroToCoMapeo = ({
+  navigation,
+}: NativeStackScreenProps<DeviceNamingParamsList, 'IntroToCoMapeo'>) => {
+  const {formatMessage} = useIntl();
+  return (
+    <View style={styles.container}>
```
achou11, line 50: this should be a scroll view or else you could potentially run into cases where there's overflow but you can't scroll down to the button (e.g. large system font size or small device screen):

image

We also have the `ScreenContentWithDock` component that's used in various screens, which accounts for this but under the assumption that the buttons should be pinned to the bottom of the screen. Not sure if that's desired in this case though, so probably opting for using a scrollview here seems sufficient.

src/frontend/App.tsx
```diff
@@ -33,8 +31,6 @@ Sentry.init({
 const messagePort = new MessagePortLike();
 const mapeoApi = createMapeoClient(messagePort, {timeout: Infinity});
 const localDiscoveryController = createLocalDiscoveryController(mapeoApi);
-const appDiagnosticMetrics = new AppDiagnosticMetrics();
-const deviceDiagnosticMetrics = new DeviceDiagnosticMetrics();
```
achou11, line 37: these need to stay here because they need to exist for the entirety of the application lifecycle.

src/frontend/screens/Onboarding/OnboardingPrivacyPolicy.tsx
```diff
@@ -0,0 +1,118 @@
...
+  },
+  shareDiagnostics: {
+    id: 'screens.OnboardingPrivacyPolicy.shareDiagnostics',
+    defaultMessage: 'Share Diagnostic Information',
+  },
+});
+
+export const OnboardingPrivacyPolicy = () => {
+  const appDiagnosticMetrics = new AppDiagnosticMetrics();
+  const deviceDiagnosticMetrics = new DeviceDiagnosticMetrics();
```
achou11: these shouldn't be instantiated here. see comment in `App.tsx`. 

```suggestion
  const appDiagnosticMetrics = new AppDiagnosticMetrics();
  const deviceDiagnosticMetrics = new DeviceDiagnosticMetrics();
```

if you need access to the instances in this component, you can either:

1. import the instance(s) created in `App.tsx`
2. create a app-level context + hook and use the hook in this component (see `ApiContext` as an example

I think (2) is preferable, mostly for reasons that don't fully apply to our codebase just yet (testing)

src/frontend/screens/Onboarding/OnboardingPrivacyPolicy.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [isDiagnosticsEnabled, setIsDiagnosticsEnabled] = React.useState(false);
+
+  const toggleDiagnostics = () => {
+    setIsDiagnosticsEnabled(prev => !prev);
+  };
+
+  React.useEffect(() => {
+    appDiagnosticMetrics.setEnabled(isDiagnosticsEnabled);
+    deviceDiagnosticMetrics.setEnabled(isDiagnosticsEnabled);
+  }, [isDiagnosticsEnabled]);
```
achou11: nit: don't think you need the effect here and generally should be able to inline that code in the press handler directly. i see the challenge is that these diagnostics instances don't expose their internal enabled state so you need to have your own state, which is used for UI purposes.

in an ideal situation, these metrics classes would provide a way to subscribe to their enabled state (via useSyncExternalStore), but given that's probably more work than it's worth, i have a different suggestion - not really sure if it's much better but in case it's helpful:

```tsx
onPress={() => {
  setIsDiagnosticsEnabled(enabled => {
    const newValue = !enabled
    appDiagnosticMetrics.setEnabled(newValue);
    deviceDiagnosticMetrics.setEnabled(newValue);
    return newValue
  })
}}
```

achou11: To fix an ESLint error:

```suggestion
  React.useEffect(() => {
    appDiagnosticMetrics.setEnabled(isDiagnosticsEnabled);
    deviceDiagnosticMetrics.setEnabled(isDiagnosticsEnabled);
  }, [isDiagnosticsEnabled, appDiagnosticMetrics, deviceDiagnosticMetrics]);
```

src/frontend/screens/Onboarding/DataPrivacy.tsx
```diff
@@ -0,0 +1,188 @@
...
+    },
+    learnMore: {
+      id: 'screens.DataPrivacy.learnMore',
+      defaultMessage: 'Learn More',
+    },
+    next: {
+      id: 'screens.DataPrivacy.next',
+      defaultMessage: 'Next',
+    },
+  });
```
achou11: this should be defined outside the component

src/frontend/screens/Onboarding/DataPrivacy.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  return (
+    <View style={styles.container}>
```
achou11: Should use a scrollview here

src/frontend/screens/PrivacyPolicy/styles.js
achou11, line 1: why js instead of ts here?

src/frontend/screens/Onboarding/OnboardingPrivacyPolicy.tsx
```diff
@@ -0,0 +1,118 @@
...
+  );
+};
+
+OnboardingPrivacyPolicy.navTitle = m.navTitle;
+
+const styles = StyleSheet.create({
+  scrollContent: {
+    backgroundColor: 'white',
+    padding: 20,
+    marginHorizontal: 10,
```
achou11: nit: generally avoid using `margin` (especially horizontal ones) as they can occasionally result in overflow not being detected properly. here's a somewhat extreme example since i'm using a rather large system font size to demonstrate this, but it could also happen on narrower devices:

image

My typical recommendation is to almost always avoid margin and instead use container elements to group spaced items and specify the `gap` styling

src/frontend/screens/PrivacyPolicy/index.tsx
```diff
@@ -0,0 +1,307 @@
...
+  NativeStackScreenProps<DeviceNamingParamsList, 'OnboardingPrivacyPolicy'>
+>;
+
+export const PrivacyPolicy: React.FC<PrivacyPolicyProps> = () => {
+  const {formatMessage} = useIntl();
+  const [awanaExpanded, setAwanaExpanded] = React.useState(false);
+  const [openSourceExpanded, setOpenSourceExpanded] = React.useState(false);
+
+  const toggleAboutAwana = () => setAwanaExpanded(!awanaExpanded);
+  const toggleOpenSource = () => setOpenSourceExpanded(!openSourceExpanded);
```
achou11: use the functional update syntax

```suggestion
  const toggleAboutAwana = () => setAwanaExpanded(prev => !prev);
  const toggleOpenSource = () => setOpenSourceExpanded(prev => !prev);
```

achou11: nit: not really necessary to define named functions for this as opposed to inlining in the relevant press handlers (since they're each only used once)

src/frontend/screens/Onboarding/OnboardingPrivacyPolicy.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <TouchableOpacity
+          onPress={toggleDiagnostics}
+          style={[
+            styles.checkBox,
+            isDiagnosticsEnabled && styles.checkBoxChecked,
+          ]}>
+          {isDiagnosticsEnabled && (
+            <MaterialIcons name="check" size={18} color={WHITE} />
+          )}
+        </TouchableOpacity>
```
achou11: this is pretty difficult to press. would suggest either:

1. asking design whether it would make sense to make the whole row pressable
2. increasing the hit slop of this

src/frontend/Navigation/Stack/DeviceNamingScreens.tsx
```diff
@@ -1,16 +1,33 @@
...
-import {Success} from '../../screens/Success';
+import {IntroToCoMapeo} from '../../screens/Onboarding/IntroToCoMapeo';
+import {DataPrivacy} from '../../screens/Onboarding/DataPrivacy';
+import {OnboardingPrivacyPolicy} from '../../screens/Onboarding/OnboardingPrivacyPolicy';
+import {DeviceNaming} from '../../screens/Onboarding/DeviceNaming';
+import {Success} from '../../screens/Onboarding/Success';
+import {MessageDescriptor} from 'react-intl';
 
-export const createDeviceNamingScreens = () => (
+export const createDeviceNamingScreens = ({
```
achou11: nit: related to just naming stuff, but this should probably be using `onboarding` instead of `deviceNaming`. same for other places that are using `deviceNaming` naming

achou11 (changes requested):

src/frontend/screens/Onboarding/OnboardingPrivacyPolicy.tsx
achou11: > My typical recommendation is to almost always avoid margin and instead use container elements to group spaced items and specify the gap styling

Sorry, might've been confusing to see this in the context of the line that this is commenting on. I should've mentioned this part more at the top-level or in more relevant parts of the PR

cimigree:
@achou11 I updated to the best of my ability. Thanks for all the specific feedback.

src/frontend/contexts/MetricsContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type MetricsProviderProps = {
+  children?: React.ReactNode;
+};
+
+export const MetricsProvider = ({
+  children,
+}: MetricsProviderProps): JSX.Element => {
+  const appMetrics = new AppDiagnosticMetrics();
+  const deviceMetrics = new DeviceDiagnosticMetrics();
```
achou11: these initializations shouldn't live in the component render body because then they'll get re-instantiated whenever this component re-renders (which will happen due to our architecture + React context limitations). Either:

1. Instantiate "globally" in App.tsx. Then you can do something similar to what we do with other instances like the api and message port i.e. provide the instances as [props to the `AppProviders` component](https://github.com/digidem/comapeo-mobile/blob/ebaf02b9e8ea8e4806a87ed364e5ed77d416300d/src/frontend/contexts/AppProviders.tsx#L27-L32) and pass those props to this `MetricsProvider` component (similarly to what we do [here](https://github.com/digidem/comapeo-mobile/blob/ebaf02b9e8ea8e4806a87ed364e5ed77d416300d/src/frontend/contexts/AppProviders.tsx#L49-L51))

2. Move these outside the component

I think 1 is preferable because it ensures that they're instantiated as soon as possible, which I think is desirable?

src/frontend/contexts/AppProviders.tsx
```diff
@@ -52,7 +53,9 @@ export const AppProviders = ({
                       <ActiveProjectProvider>
                         <BottomSheetModalProvider>
                           <PhotoPromiseProvider>
-                            <SecurityProvider>{children}</SecurityProvider>
+                            <SecurityProvider>
+                              <MetricsProvider>{children}</MetricsProvider>
```
achou11: this should probably live further up in this component tree. would say as long as it's higher up than `ActiveProjectProvider`, that should be sufficient.

achou11 (changes requested):
Didn't re-review the UI yet but noticed a couple non-UI things that stood out to me

achou11 (approved):
Great stuff! Having these screens makes the app feel much more polished ✨

PR #596, opened 2024-08-13
fix: fix app variant env variable exposure

achou11:
Based on Expo's documentation around env variables, I _think_ we're not exposing the `APP_VARIANT` env variable properly, despite things kind of just working anyways???

https://docs.expo.dev/build-reference/variables/
https://docs.expo.dev/guides/environment-variables/

Basically, if you want to read an env variable from `process.env` in your **app** (i.e. not at build time), you're supposed to prefix it with `EXPO_PUBLIC_` so that it gets inlined when bundling. We reference `process.env.APP_VARIANT` directly in a few places, which technically shouldn't be working based on their docs, but maybe I'm missing something?

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -1,11 +1,11 @@
 const SUFFIX =
   {development: '.dev', production: '', test: '.test'}[
-    process.env.APP_VARIANT
+    process.env.EXPO_PUBLIC_APP_VARIANT
   ] ?? '';
 
 const NAME =
   {development: ' (DEV)', production: '', test: ' (TEST)'}[
-    process.env.APP_VARIANT
+    process.env.EXPO_PUBLIC_APP_VARIANT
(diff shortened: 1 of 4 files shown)


achou11:
ah based on https://docs.expo.dev/tutorial/eas/multiple-app-variants/#configure-easjson, looks like `APP_VARIANT` is supposed to only be used as a **build-time** variable. My guess is that instead, we should update the `eas.json` profile to specify a different env variable (prefixed with `EXPO_PUBLIC_`) that can be used in app code.

What I _think_ is currently happening is that the `process.env.APP_VARIANT` that's referenced in our app code is referring to what's set in our `eas.json`, which just also happens to be an env variable called `APP_VARIANT`.

achou11:
Gonna close this for now and detail in an issue instead

PR #601, opened 2024-08-13
Updated dependencies

ErikSin:
Updated:
- Mapeo Core Next
- Mapeo Schema
- Mapeo ipc

Its currently broken, the presets are not showing up, but I asked for help in the tech channel.

I am going to assign it to @cimigree, as I would like you to run e2e when the preset problem gets fixed.


cimigree (approved):
Tests passed.

PR #607, opened 2024-08-14
feat: settings data privacy

### LINKED ISSUE #543: Menu Update
## Description
With metrics, we are now introducing a Data Privacy screen at onboarding. User can also access that Data Privacy screen via the right Menu

What we are tracking : https://www.notion.so/digidem/Metrics-MVP-70d4d2990de8407fbe599218ad5cf155

## Menu Update
Screen Shot 2024-07-25 at 10 37 35 AM 

- [x]  Update screen as designed, by added a _Data & Privacy_ button under _App Settings_ (see #565)
- [x] When clicked direct to the Data Privacy screen

## Screen 2 - Data Privacy
Screen Shot 2024-07-25 at 10 37 47 AM

- [x] Display **CoMapeo Respects Your Privacy & Autonomy**
- [x] Display _Learn More_ button, when clicked direct to Privacy Policy screen
- [x] Display check box next to S**hare Diagnostic Information** , when clicked collect data

## Screen 3
Screen Shot 2024-07-25 at 10 38 00 AM

- [x]  When clicked back arrow directs to the Data Privacy screen

## Design
https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/%22Ready-for-Dev%22-Hand-off-Assets?node-id=934-273&t=eiLnXjdi133AXR2g-0

cimigree:
closes #543 

### Description
- This adds the screens that come when someone clicks on Settings and then Data & Privacy (which Evan already added)
- Here is the [prototype](https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2081-1671&t=gop2fYV51SRMyhXi-0&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=2081%3A1527)
- This adds the logic in the hooks for Zustand to set the value of whether or not they have enabled permissions through storage
- This PR adds a reusable toggle permissions component and hook because it is used in two screens
- This PR includes an update to the onboarding privacy policy screen that uses the toggle permissions
- I have reused the Privacy Policy screen from the previous PR

### Screen Shots
**If I have checked the permissions during onboarding, it is already checked on this screen:**

<image src="https://github.com/user-attachments/assets/d201d812-3283-47d6-a8d2-12d72ea4e8c2" width="250" />

**This is the privacy policy screen but does not have the check mark at the bottom as it does during onboarding**

<image src="https://github.com/user-attachments/assets/d812990b-48c3-49f6-b2a4-d1d6b3c4408d" width="250" />


ErikSin:
Can you update this from main? @cimigree

cimigree:
> Can you update this from main?

@ErikSin, I think you mean develop? All set.

ErikSin:
> > Can you update this from main?
> 
> @ErikSin, I think you mean develop? All set.

yup develop, which it looks like you have done! thanks.

ErikSin:
There are 2 problem with flow of logic here that I think will make it hard to maintain in the future

1. This logic of setting persisted state is a little roundabout
2. We are state syncing persisted state of isEnabled to the isEnabled state of context. AKA there are 2 values, both representing the same thing, which we are attempting to keep in sync

## This logic of setting persisted state is a little roundabout

Were setting a context value, then passing that context to persisted state. The persisted state (is technically) a parent of the context value, so we are passing the child value to the parent, so the parent can pass that child value to the rest of the app.

Ideally the parent sets the value, and passes to the child, which then passes it to its children.

this is related to problem # 2

## State Syncing
Metrics internally has its own 'isEnabled` value. And then we also have a persisted enabled value. Now we need to maintain both of them, instead of just having one shared value. This is seen in the state syncing we do where everytime we change one value, we also change the other value

Unfortunantly with the `class AppDiagnosticMetrics` we can just simply pass the persisted state of `isEnabled`. The class does have to manage its own state. We could pass a callback/subsctiption...but there is still state syncing happening.

There is a way to do it so that there is minimal stateSyncing.

We can pass the 'usePersistedPermissionStore', directly to the class. `usePersistedPermissionStore` has a subscribe function.

So we can do:

```ts

export class AppDiagnosticMetrics {

  constructor() {
    this.#update = () => {
      this.#updateQueue.add(() => this.#doUpdate());
    };
    this.unsubscribe = usePersistedPermissionStore.subscribe(
          (state) => state.isEnabled,
          (isEnabled) => this.setEnabled(isEnabled)
        );
   }
   
   setEnabled(isEnabled: boolean): void {
    this.#isEnabled = isEnabled;

    if (isEnabled) {
      this.#addListenersIfNecessary();
    } else {
      this.#removeListeners();
    }

    this.#update();
  }
  
  
  ...rest
}
```
And then the react code simply uses `usePersistedPermissionStore` to set and get permission, we dont need to worry about the internal state of `AppDiagnosticMetrics`

ALTERNATIVELY we could avoid state syncing entirely and have the metrics be tied to react state

It would look like this:

```ts
// this is called at the top level of the app
function useMetrics(){

   const isEnabled = usePersistedPermissionStore(store => store.isEnabled)
   
   useEffect(()=>{
      if(isEnabled){
         // start listners and subscriptions
      }
      
      if(!isEnabled){
         //unsubscribe
      }

      return ()=>{//unsubscribe}
   },[])

}


```
@EvanHahn Im curious what you think. I do see that the class is technically external to react state. In the sense that it is just a bunch of subscriptions happening in the background. And we could just have the persisted state pass a subscription/call back to class, and the class and internally manage its own state based on that callback. Still state syncing, but minimal

Tieing it to react state maybe feel a little unnecessary, but I like that there is no state syncing.

But i am not tied to either idea, so would love to get your insight @cimigree and @EvanHahn.

ErikSin:
On second thought, i think we can get away with the subscription approach without worry about the downfalls of "state syncing"

`isEnabled` in the class is just an internal state, not accessible to anything else.

If we make `setEnabled` private, then the only place it can be set is through that subscription. So it is technically syncing an external to internal state. But if that is the only place that it can be set, then we don't ever have to worry about those states being out of sync. 

If thats the case, its not actually a problem to "state sync" because there is no way for them to be out of sync.

EvanHahn:
First of all, I am _not_ attached to my implementation. If `{App,Device}DiagnosticMetrics` need to change, let's do it.

Second, I imagined that we'd have a normal persisted boolean just like our other persisted values. Then there's an effect that keeps those classes up to date—don't manage our own subscriptions when React will it for us. I sketched this out in #612 (doesn't work, but hopefully shows my idea). This is the meat of it:

```js
const {isEnabled} = usePersistedMetricDiagnosticsPermission();

React.useEffect(() => {
  appMetrics.setEnabled(isEnabled);
  deviceMetrics.setEnabled(isEnabled);
}, [appMetrics, deviceMetrics, isEnabled]);
```

If we don't like my idea (which is fine!), I think passing some kind of subscription into the constructors for `{App,Device}DiagnosticMetrics` feels like a good approach. I also think the way @cimigree has it here is also good.

src/frontend/hooks/persistedState/createPersistedState.ts
```diff
@@ -15,7 +15,8 @@ type PersistedStoreKey =
   | 'MapeoTrack'
   | 'Passcode'
   | 'ActiveProjectId'
-  | 'Settings';
+  | 'Settings'
+  | 'Permissions';
```
EvanHahn: Nit: I could see giving this another name because we plan to add additional permissions in the future. For example, for other kinds of opt-in metrics.

ErikSin:
@cimigree I think we should either follow the useEffect model here that Evan has pointed out of passing the sucbcription directly to the class. The way it is right now works, but I think will be a little harder to maintain in the future.

cimigree:
> @cimigree I think we should either follow the useEffect model here that Evan has pointed out of passing the sucbcription directly to the class. The way it is right now works, but I think will be a little harder to maintain in the future.

@ErikSin  I think you are right! I had just come to the same conclusion after way too much deliberation and trying to understand a few different concepts that are new to me using a non api app and am now in the process of implementing. Will have revised PR up soon. Thanks for all the feedback and input everyone.

ErikSin:
@cimigree can you update from develop now? Your local environment was using a cached dependency that is now unavailable. I cleared my gradle cache on thursday to debug a CI problem, but it meant I couldnt test this PR. so my recent PR should have just fixed that

cimigree:
> @cimigree can you update from develop now? Your local environment was using a cached dependency that is now unavailable. I cleared my gradle cache on thursday to debug a CI problem, but it meant I couldnt test this PR. so my recent PR should have just fixed that

Ok done, @ErikSin !

src/frontend/contexts/AppProviders.tsx
```diff
@@ -56,9 +56,7 @@ export const AppProviders = ({
                 <ServerLoading messagePort={messagePort}>
                   <LocalDiscoveryProvider value={localDiscoveryController}>
                     <ApiProvider api={mapeoApi}>
-                      <MetricsProvider
-                        appMetrics={appMetrics}
-                        deviceMetrics={deviceMetrics}>
+                      <MetricsProvider>
```
ErikSin: Initializing at `appMetrics` and `deviceMetrics` at the app level still makes sense. You don't have to worry about memoizing it and it is unattached from the react-lifecycle

src/frontend/hooks/useMetricsDiagnosticsPermissionsToggle.ts
ErikSin, line 1: this wrapper feels a little unnecessary. Its only being used by one component, so we could just remove the hook, and encapsulate this logic inside `MetricsDiagnosticsPermissionToggle`. Unless you think this will be used somewhere else in the future?

ErikSin (changes requested):
just 2 edits! Thanks Cindy

src/frontend/contexts/AppProviders.tsx
cimigree: All set I think

src/frontend/hooks/useMetricsDiagnosticsPermissionsToggle.ts
cimigree: Yes you are right. Sorry I didn't combine those earlier.

ErikSin (approved):
Looks good!

PR #608, opened 2024-08-14
feat: show config details and upload new config

ErikSin:
Exposes the config name and created at date to the user. 

Also allows for the user to upload a new config

image

### DIFF
--- a/src/frontend/screens/Settings/Config.tsx
+++ b/src/frontend/screens/Settings/Config.tsx
@@ -1,20 +1,15 @@
 import * as React from 'react';
-import * as DocumentPicker from 'expo-document-picker';
-import * as FileSystem from 'expo-file-system';
 import {defineMessages, useIntl} from 'react-intl';
 import {StyleSheet, View} from 'react-native';
 import {Text} from '../../sharedComponents/Text';
-import {
-  useImportProjectConfig,
-  useProjectSettings,
-} from '../../hooks/server/projects';
+import {useProjectSettings} from '../../hooks/server/projects';
 import {Loading} from '../../sharedComponents/Loading';
 import {NativeNavigationComponent} from '../../sharedTypes/navigation';
 import {COMAPEO_BLUE, MEDIUM_GREY} from '../../lib/styles';
 import {Button} from '../../sharedComponents/Button';
-import {convertFileUriToPosixPath} from '../../lib/file-system';
 import {UIActivityIndicator} from 'react-native-indicators';
 import {ErrorBottomSheet} from '../../sharedComponents/ErrorBottomSheet';
+import {useSelectFileAndImportConfig} from '../../hooks/useSelectFileAndImportConfig';
 
 const m = defineMessages({
   navTitle: {
@@ -42,14 +37,13 @@ const m = defineMessages({
 export const Config: NativeNavigationComponent<'Config'> = ({navigation}) => {
   const {formatMessage} = useIntl();
   const {data, isPending} = useProjectSettings();
-  const [importError, setImportError] = React.useState<Error | null>(null);
 
-  const importProjectConfigMutation = useImportProjectConfig();
+  const selectAndImportConfigMutation = useSelectFileAndImportConfig();
 
   React.useEffect(() => {
     // Prevent back navigation while project creation mutation is pending
     const unsubscribe = navigation.addListener('beforeRemove', event => {
-      if (!importProjectConfigMutation.isPending) {
+      if (!selectAndImportConfigMutation.isPending) {
         return;
       }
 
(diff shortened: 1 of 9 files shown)


src/frontend/screens/Settings/Config.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    if (result.canceled) {
+      return;
+    }
+
+    const asset = result.assets[0];
+
+    // Shouldn't happen based on how the library works
+    if (!asset) return;
+
```
cimigree: Would it make sense to add file validation here? Since there is no ticket to look at and no testing instructions, I am just guessing here, but I noticed that when a config file is first imported, it is supposed to end in .mapeoconfig

ex. 
```
if (!asset.name.endsWith('.mapeoconfig')) {
    setImportError(new Error('File name should end with .mapeoconfig'));
    await FileSystem.deleteAsync(asset.uri).catch(err => console.log(err));
    return;
  }


```

src/frontend/screens/Settings/Config.tsx
```diff
@@ -0,0 +1,177 @@
...
+      event.preventDefault();
+    });
+
+    return () => {
+      unsubscribe();
+    };
+  }, [navigation, importProjectConfigMutation.isPending]);
+
+  async function importConfigFile() {
```
cimigree, line 58: Since we use some of the same logic when creating a project, do you think it makes any sense to extract the document picking and path conversion (and maybe file validation) into a reusable function? Your call entirely, of course. Just a thought.

cimigree:
@ErikSin Would you mind updating from develop and cleaning conflicts? 
Also, is there a ticket? I have one config file settings-shipibo-v3.0.0.mapeo-settings
Is that a good one to test with? Any testing instructions or tips?

src/frontend/screens/Settings/Config.tsx
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const asset = result.assets[0];
+
+    // Shouldn't happen based on how the library works
+    if (!asset) return;
+
+    importProjectConfigMutation.mutate(convertFileUriToPosixPath(asset.uri), {
+      onSuccess: async () => {
+        await FileSystem.deleteAsync(asset.uri).catch((err: unknown) => {
```
cimigree: This doesn't trigger any UI update or state refresh other than invalidating the query. Should it refetch the project settings data to update the component's state? Or maybe I am missing something

cimigree (changes requested):
I tried importing the config file I have (settings-shipibo-v3.0.0.mapeo-settings) and I get very little UI interaction (is that how it is supposed to be -- like no success or anything) and the config ui on the screen still says the same thing as before I imported that. Is that the expected behavior? Here is what the screen looks like before and after the import...
<image src="https://github.com/user-attachments/assets/2e397975-5e00-49df-b87e-4c5827c7f02c" width="250" />

And, if I try to import a file that is not a config file, a pdf for example, the same thing happens.

So, in sum, Should the screen update with new project settings or metadata immediately after the import? 
And should there be file type validation with user messaging?

src/frontend/screens/Settings/Config.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/608/commits/4a88f58d1af7b6b3887d7a585eaba18e625ae5ba

src/frontend/screens/Settings/Config.tsx
ErikSin: adressed here:https://github.com/digidem/comapeo-mobile/pull/608/commits/4a88f58d1af7b6b3887d7a585eaba18e625ae5ba and https://github.com/digidem/comapeo-mobile/pull/608/commits/652be3f35d2bafbed4af742ffad805e7d4e9d961

src/frontend/screens/Settings/Config.tsx
ErikSin: created a custom hook that invalidates the project settings: https://github.com/digidem/comapeo-mobile/pull/608/commits/652be3f35d2bafbed4af742ffad805e7d4e9d961

ErikSin:
> I tried importing the config file I have (settings-shipibo-v3.0.0.mapeo-settings) and I get very little UI interaction (is that how it is supposed to be -- like no success or anything) and the config ui on the screen still says the same thing as before I imported that. 
> 

So this is actually an error with the backend. The backend is not throwing an error on faulty configs, so a faulty config is not returning any data, and the cache is then just returning the previous data...I will create an issue for this

ErikSin:
It should also now throw an error if the user tries to import the wrong type of file!

cimigree (commented):
My only issues with this now are around UI and messaging.
I got a new config file from Tomas which seems to work better. The one I was using was invalid, apparently.
May still be an issue or two with error handling. 
What do you think... if the config file is invalid (like the one I had been using), should we show something? Currently what happens is nothing. The loader spins and then the name doesn't change. It was certainly confusing for me and might also be confusing for a user. But maybe their chance of having an invalid file is less prone to happen? I don't know.
If I try to import a file with a different ending, I get this screen
<image src="https://github.com/user-attachments/assets/2794b643-282a-4c07-a16a-926433517d51" width="250" />
I thought we might see the message you are throwing (Invalid file type), but perhaps that is a much bigger issue because it involves retooling the ErrorBottomSheet and should be outside of this PR. Making note though, that seeing "Invalid File Type" would be helpful for a user, I think.

cimigree (approved):

PR #613, opened 2024-08-19
fix: only show connected devices that appeared during lifetime of device invite screen

### LINKED ISSUE #589: Remove disconnected devices from device list on initial navigation to device list screen
By design, [useLocalPeers](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/hooks/useLocalPeers.ts) produces a list of of disconnected and connected peers. It was designed to keep the disconnected peers on the list, so that the list of peers would not be constantly changing and jumping around. However, when the user navigates away from the screen, and back to the screen, we should clear the list of disconnected devices, and only show connected devices.

## To Do
- [ ] Clear disconnected device when navigating to device list screen

### Notion Issue:
https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=c58a7b022e084c2bae2c4379bf0c67e7&p=4a9201eab678469b9395d9ab14d7abee&pm=s

achou11:
Fixes #589 

Avoids updating `useLocalPeers()` because I'd rather that remains as close to the server state as possible, given that it's global state and can eventually be used in other areas of the app (with potentially different UX needs). The UX needs from the issue seem specific to this screen, so this opted for a wrapper hook that provides a more opinionated subset of the global state.

Testing steps:

1. Set up 2 devices, A and B. Create a new project for A.
2. On A, go through device invite flow. Stop at device list screen. If B has the app opened, it should be listed on the screen. If not, open the app on B and wait for it to show up in A.
3. Close the app on B. On A, you should see that B is still listed but marked as disconnected.
4. On A, navigate to the previous screen (or any other screen, really) and then navigate back to this device list screen. You should see that B is no longer listed. If you have other devices that are connected, those should still be listed.
5. If you open the app on B again, it should be reappear in the list on A.
---

https://github.com/user-attachments/assets/b43bfeaf-d9fb-4dbb-bed9-6f6cea7ac76f

### DIFF
--- a/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
@@ -82,7 +82,7 @@ function InvitableDeviceList() {
         // TODO: Use `device.deviceType`
         const deviceType = 'mobile';
 
-        // TODO: Update DeviceCard component to better handle potentially undefined fields
+        // TODO: Update DeviceCard component to better handle potentially undefined fields, deviceType enum
         return (
           <DeviceCard
             key={deviceId}
--- a/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
@@ -54,7 +54,7 @@ export const SelectDevice: NativeNavigationComponent<'SelectDevice'> = () => {
 
 function InvitableDeviceList() {
   const navigation = useNavigationFromRoot();
-  const devices = usePeersConnectedDuringSession();
+  const devices = useInitiallyConnectedPeers();
   const projectMembersQuery = useProjectMembers();
 
   if (projectMembersQuery.status === 'pending') {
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -91,4 +111,45 @@ const styles = StyleSheet.create({
...
+  },
 });
+
+/**
+ * Applies specialized, context-specific behavior on top of `useLocalPeers()` in the following ways:
+ *
+ * - State is bound to the lifetime of the consuming component i.e. gets re-initialized during remounts
+ * - Only connected peers are ever *added* to the state. However, subsequent updates to those peers (e.g. disconnecting) are still reflected in the state.
+ */
+function useInitiallyConnectedPeers() {
```
achou11, line 125: naming is a little confusing but hopefully these comments help clarify what this hook is doing. Basically it's keeping its own cache/state that's tied to the consuming components lifecycle, as opposed to the global state (which `useLocalPeers()` is used for)

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -91,4 +111,45 @@ const styles = StyleSheet.create({
...
+  },
 });
+
+/**
+ * Applies specialized, context-specific behavior on top of `useLocalPeers()`
+ * by only including peers that were *initially* connected from the point of view of the consuming component.
+ * If these included peers receive subsequent updates (e.g. disconnecting), they are still returned.
+ * This hook only prevents *initially* returning peers that are disconnected.
+ */
+function useInitiallyConnectedPeers() {
```
achou11, line 125: naming is a little confusing but hopefully these comments help clarify what this hook is doing. Basically it's keeping a slightly opinionated list of relevant device ids during the lifetime of the consuming component and using that to determine which of the local peers (in global state) should be returned

ErikSin:
Can you update this from develop? I need the most recent patch to test on develop!

achou11:
> Can you update this from develop? I need the most recent patch to test on develop!

done!

ErikSin (approved):
Looks good!

PR #617, opened 2024-08-21
fix: media sharing

### LINKED ISSUE #591: Possible resource mismanagement & caching issues when sharing observations with media attachments more than once
**Describe the bug**
I have noticed that some issues come up when sharing an observation that has a media attachment more than one time. The first time I hit "Share", the button shows a loading state for only a short while before the modal to share to a specific output comes up. The second time I hit "Share", the loading state lasts much longer; and even longer for a third time. Then, when I try to open the image, I am shown a "not found" cue via the exclamation mark image. I have also observed that the thumbnail itself stops loading after a while. 

**To Reproduce**
1. Go to an Observation with a media (image) attachment.
2. Press on Share. Wait for the system modal to choose your application (e.g. WhatsApp, email) to pop up.
3. Either continue to share or close the modal to return to the CoMapeo observation screen.
4. Next, press "Share" again and observe the longer wait time.
5. After the modal finally pops up, close it again, and now, press on the thumbnail image to see the "not found" image.
6. A third time pressing "Share" has me waiting at least 5 minutes (ongoing while I am writing this issue); not sure if it will culminate or not.

**Screenshots**
WhatsApp Image 2024-08-10 at 15 47 58
WhatsApp Image 2024-08-10 at 15 43 06
WhatsApp Image 2024-08-10 at 15 47 59

 - Device: Samsung Galaxy (SM-A716U)
 - Android version: 13
 - Version: 1.0.0 (build for DWeb Camp demo `com.comapeo.test`)

**Additional context**
I tried to see if internet connectivity mattered or not, but it seems not to make a difference.

I also spotted that at some point, after I saw the "not found" image for the thumbnail, that the points on the main map also disappeared, until I went to the Observations screen which seems to have reset the map state. But I could not reproduce this again so I am logging it as an anecdotal observation in case it is salient for debugging this, or a different issue.

cimigree:
closes #591 

### Description
- In the handleShare action for the button in an observation, only does the refetch for the query if needed (it is needed for the first time the photo is shared)
- It takes a very specific and very repeated set of actions to cause an error in the PhotoPreviewModal which I don't think anyone would ever do
- Moved the logic for photo query the PhotoPreviewModal.tsx from the PhotoUnpreparedView as we discussed @achou11 . Not sure if that is necessary or beneficial...
- I could move the photo querying logic up to the Observation/index if you think that would help. I will wait to hear your thoughts.

### To test
Create an observation
Click into it
Share it as many times as you want
You should not experience a delay
Thumbnail will not error out


src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -104,11 +104,16 @@ export const ButtonFields = ({
...
+    const needsRefetch = attachmentUrlQueries.some(query => !query.data);
+    let urls;
+    if (needsRefetch) {
+      const urlsQueries = await Promise.all(
+        attachmentUrlQueries.map(q => q.refetch()),
+      );
+      urls = urlsQueries.map(query => query.data!.url);
+    } else {
+      urls = attachmentUrlQueries.map(query => query.data!.url);
+    }
```
achou11: starting to question the need of using react-query (specifically `useAttachmentUrlQueries()`) for this data as opposed to just calling the project api directly. Using the query would make sense if the query state needed to show up or affect what's rendered in the UI, but in this case, we're mostly interested in getting the urls on demand based on a specific user interaction. Thoughts on doing the following instead of using that rather complex query?

```ts
// Make sure this is set up in the component
const {projectApi} = useActiveProject()

// Then handlePressShare()
async function handlePressShare() {
  const {lon, lat} = observation;
  setShareButtonLoading(true);

  // Using `Promise.allSettled()` but depending on UX, may want to use `Promise.all()`
  // If using the latter, should make sure to handle the case if one of the calls throws an error
  const urlResults = await Promise.allSettled(
    observation.attachments.map(a => {
      if (
        a.type === 'UNRECOGNIZED' ||
        a.type === 'attachment_type_unspecified'
      ) {
        return null;
      } else
        return projectApi.$blobs.getUrl({
          variant: 'original',
          type: a.type,
          name: a.name,
          driveId: a.driveDiscoveryId,
        });
    }),
  );

  const validUrls = urlResults
    .filter(
      (result): result is PromiseFulfilledResult<string> =>
        result.status === 'fulfilled' && result.status !== null,
    )
    .map(result => result.value);

  const base64Urls = await Promise.all(
    validUrls.map(url => convertUrlToBase64(url)),
  );
```

achou11: hm actually on second thought, this results in a different UX because now we're always fetching whenever the button is pressed, as opposed to using a cache after the first time. Not the worst thing but it's usually not necessary to keep refetching the urls. Could adapt this to hold some state in this component, but then we're probably just doing a lazy version of react query in that case.

Maybe my suggestion isn't worth considering after all 😅

achou11: i would be careful about using the `!` operator. that's essentially similar to telling TypeScript that we know that the value is guaranteed to not be null/undefined, which is something that we can't guarantee based on the code here.

achou11: Do you think the refetch is useful to keep here? I'm not sure it's necessary to keep, as it seemed to be what was causing the original issue before.

Potential suggestion:

- Don't refetch. Live with the current state of the attachement url queries whenever the button is pressed
- if any of the queries are missing `data`, either:
  - surface an error in the UI that prevents sharing from happening as a whole
  - filter out the attachments where `data` is not present and continue to process the ones that we have a url for
  
Not sure which of these are preferred from a UX perspective, but this additional handling should be considered even if the refetch is kept because query-refetching doesn't guarantee that a valid url will be retrieved

achou11 (changes requested):

src/frontend/screens/Observation/Buttons.tsx
cimigree: I eliminated the ! operator.
I have to refetch. Without that refetch there is never any data in the attachments when you first go to share. That is why I kept it in and made the if else. If there is another way to get that data in there the first time, I could not find it and would love to hear your ideas. Sorry. I think that is why software mansion had it in there.
I applied the filter as you suggested where the data was not present and throw an alert error if there is no data present. Not sure if that is preferred or if we should just share without the photos?

achou11: > I have to refetch. Without that refetch there is never any data in the attachments when you first go to share. That is why I kept it in and made the if else. If there is another way to get that data in there the first time, I could not find it and would love to hear your ideas. Sorry. I think that is why software mansion had it in there.

ah i see. yeah not sure what the issue is without investigating a bit. hard to tell at a glance.

> I applied the filter as you suggested where the data was not present and throw an alert error if there is no data present. Not sure if that is preferred or if we should just share without the photos?

yeah i'm not sure what's preferred. think what you have makes sense and is easy to change later if needed.

achou11: > I have to refetch. Without that refetch there is never any data in the attachments when you first go to share.

figured out the issue: 

https://github.com/digidem/comapeo-mobile/blob/0514c3b77647d493beba05c31ca39394ccb707f3/src/frontend/screens/Observation/Buttons.tsx#L84

should be omitted or `true` instead in order for the query to run. otherwise it's only running the queries when you press the button, similar to my other less-than-good suggestion in my other comment

achou11: > I applied the filter as you suggested where the data was not present and throw an alert error if there is no data present. Not sure if that is preferred or if we should just share without the photos?

going back to this, i guess the safest approach (for now) would be to do whatever mapeo mobile does or something similar? based on:

https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/screens/Observation/ObservationView.js#L152

seems like it leans towards the all-or-nothing approach. what you currently have seems most aligned with that, so i think that's fine

src/frontend/screens/Observation/Buttons.tsx
achou11, line 1: related to https://github.com/digidem/comapeo-mobile/pull/617#discussion_r1731526872, i _think_ the attachments queries call should remove the `enabledByDefault` param i.e.

```ts
  const attachmentUrlQueries = useAttachmentUrlQueries(
    observations.attachments,
    'original'
  );
```

not sure if there's any major advantage to having it not done eagerly, at least for now. i guess it'd be a similar reason for my suggestion in https://github.com/digidem/comapeo-mobile/pull/617#discussion_r1727578254, but using react-query, which is slightly better than my earlier suggestion from a UX perspective. not sure how the `refetch()` method works (if it does) if `enabled` is `false` (which it currently is as implemented) though

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -102,29 +102,56 @@ export const ButtonFields = ({
...
+
+      await Share.open({
+        title:
+          base64Urls.length > 0 ? t(m.shareMediaTitle) : t(m.shareTextTitle),
+        urls: base64Urls,
+        message: t(m.shareMessage, {
+          category_name: preset.name,
+          date: Date.now(),
+          time: Date.now(),
+          coordinates: lon && lat ? formatCoords({lon, lat, format}) : '',
```
achou11: might as well fix while you're here:

```suggestion
          coordinates: typeof lon === 'number' && typeof lat === 'number' ? formatCoords({lon, lat, format}) : '',
```

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -102,29 +102,56 @@ export const ButtonFields = ({
...
+        }),
+      });
+    } catch (error) {
+      console.error(error);
```
achou11: maybe log via Sentry here instead?

achou11 (approved):
don't think any of the new comments are blocking. as discussed, think there's a lot of potential cleanup/improvements to the pre-existing code, which i'd consider out of scope for the issue.

achou11:
can you update the PR title before merging?

cimigree:
@achou11 
I added the eager loading to the query and removed the refetch... They seem to achieve the same outcome (eager loading vs refetch) and both ways, the sharing works well no matter how many times you try to do it.
Fixed the typing/ check of long and lat as requested.
Now Logging share errors to Sentry.

Does that sounds like what you were looking for? There were so many comments I wasn't sure! ;0)

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -102,29 +102,52 @@ export const ButtonFields = ({
...
+      if (urls.length === 0) {
+        setShareButtonLoading(false);
+        Alert.alert('Error', 'Unable to share this observation.');
+        return;
+      }
+    }
+
+    try {
+      const base64Urls = await Promise.all(
+        urls.map(url => convertUrlToBase64(url as string)),
```
achou11: No need for the type coercion

```suggestion
        urls.map(url => convertUrlToBase64(url)),
```

achou11 (approved):

PR #619, opened 2024-08-22
feat: introduce improved sync screen experience

achou11:
Closes #550 
Closes #602
Closes #603
Closes #604
Towards #605
Closes #610

Important notes:

- There is a bug where sync can occur without opt-in by connected peers. This is backend-related (see https://github.com/digidem/mapeo-core-next/issues/762).

- Upgrades core to latest alpha version, which introduces a new sync state api and also the ability to set/configure sync autostopping

- This PR does not change the behavior related to sync being turned off when leaving the screen. I think that can/should be a follow-up PR that also introduces auto-stopping sync when leaving the screen after being fully synced.

---

Not an exhaustive showcase of all of the states (hard to do with just 2 devices)

- No connected devices

https://github.com/user-attachments/assets/814fce9c-dbd0-4603-b723-06a13e819672

- Connected devices, but no data to sync

https://github.com/user-attachments/assets/96882510-8a09-4ecd-876f-d6c8f104b59b

- Connected devices, data to sync

https://github.com/user-attachments/assets/f8d3aa8d-aeb4-40ae-b0cd-2e56a212d9af

PR #620, opened 2024-08-26
fix: set sync autostop when leaving screen after fully complete sync

### LINKED ISSUE #605: Sync update / Complete
**Description**
 Device has sync all its data with the found devices. 2 possible states of complete

Details : https://www.notion.so/digidem/QA-Sync-bug-3-devices-no-synced-progress-3de5f1760e934b12b673bb49b62572ec

Relates to https://github.com/digidem/comapeo-mobile/issues/603 https://github.com/digidem/comapeo-mobile/issues/604 https://github.com/digidem/comapeo-mobile/issues/602

**Todo**
## Screen 1 - Fully complete sync / device still in sync party
If device navigates to a different screen after this state, after 30 secs sync party closes. This is true ONLY when you reach this state. This is NOT true for when you actively syncing and navigate away.
Screen Shot 2024-08-13 at 8 22 30 PM

- [ ] Display _6 Devices found_ since 6 devices, other than yours, are on the same project/network
- [ ] Display _Complete!  You’re up to date. All data synced_ if all devices have synced
- [ ] Display progress bar as designed (green)
- [ ] Add _6 out of 6 devices_ on top of progress bar, if 6 devices synced
- [ ] Add 100% percentage of syncing under the progress bar
- [ ] If clicked on 'Stop', user goes to Screen 2 - see below

## Screen 2 - Fully complete sync / device out of sync party
Screen Shot 2024-08-13 at 8 22 48 PM

- [ ] Display _6 Devices found_ since 6 devices, other than yours, are on the same project/network
- [ ] Display _Complete!  You’re up to date. All data synced_ if all devices have synced
- [ ] Display progress bar as designed (green)
- [ ] Add _6 out of 6 devices_ on top of progress bar, if 6 devices synced
- [ ] Add 100% percentage of syncing under the progress bar
- [ ] Update 'Stop' button to _You’re all caught up!_

**Design**
https://www.figma.com/design/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?node-id=2596-796

achou11:
Closes #605 

~Stacked on #619~ 

- Important: before this change, leaving the sync screen disabled sync immediately for the device. Now, sync is only disabled automatically 30 seconds after leaving the screen while in the "fully complete" sync state. otherwise, the user must explicitly press the stop sync button on the sync screen in order to disable sync.

  - there may be some UX improvements that should be introduced as a result of this change e.g. showing elsewhere in the app that sync is still enabled. wondering if that's something we can address in a follow-up though
---

- Autostop after leaving screen (_For the sake of brevity, this example stops the sync 5 seconds after leaving the screen, instead of 30_)

  https://github.com/user-attachments/assets/3897a65f-8679-4162-ab8b-c6b4dad9d0b6

- Syncing in background as a result of sync staying enabled after leaving screen

  https://github.com/user-attachments/assets/e24eb131-07a9-4782-a3c5-4ada7f6af83c

### DIFF
--- a/src/frontend/screens/Sync/ProjectSyncDisplay.tsx
+++ b/src/frontend/screens/Sync/ProjectSyncDisplay.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import {useFocusEffect} from '@react-navigation/native';
 import {useQueryClient} from '@tanstack/react-query';
 import {defineMessages, useIntl} from 'react-intl';
 import {StyleSheet, View} from 'react-native';
@@ -140,27 +141,38 @@ export const ProjectSyncDisplay = ({
     dataSyncEnabled: syncState.data.isSyncEnabled,
   });
 
-  // stops sync when user leaves sync screen.
-  // The api allows us to continue syncing even if the user is not on the sync screen, but for simplicity we are only allowing sync while on the sync screen.
-  // In the future we can easily enable background sync, there are just some UI questions that need to answered before we do that.
-  React.useEffect(() => {
-    const unsubscribe = navigation.addListener('beforeRemove', () => {
-      projectApi.$sync.stop();
-      queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY, projectId]});
-    });
+  // Unset sync autostop when screen mounts
+  useFocusEffect(
+    React.useCallback(() => {
+      projectApi.$sync.setAutostopDataSyncTimeout(null);
+    }, [projectApi]),
+  );
 
-    return unsubscribe;
-  }, [navigation, projectApi, queryClient, projectId]);
+  const shouldAutostopSyncWhenLeavingScreen =
+    syncState.data.isSyncEnabled && syncStage.name === 'complete-full';
 
-  // TODO: Autostop data sync upon leaving screen while in specific state
-  // const shouldAutostopSyncWhenLeavingScreen = syncState.data.isSyncEnabled && syncStage.name === 'complete-full';
-  // React.useEffect(() => {
-  //   const unsubscribe = navigation.addListener('beforeRemove', () => {
-  //     projectApi.$sync.setAutostopDataSyncTimeout(30_000);
-  //   });
+  // Set up listener for autostopping sync (if applicable) and invalidating queries when navigating away from screen
+  useFocusEffect(
+    React.useCallback(() => {
+      const unsubscribe = navigation.addListener('beforeRemove', () => {
+        if (shouldAutostopSyncWhenLeavingScreen) {
+          projectApi.$sync.setAutostopDataSyncTimeout(30_000);
+        }
+        // TODO: All queries associated with project should be invalidated
+        queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY, projectId]});
+      });
 
-  //   return unsubscribe;
-  // }, [navigation, projectApi, shouldAutostopSyncWhenLeavingScreen]);
+      return () => {
+        unsubscribe();
+      };
+    }, [
+      navigation,
+      projectApi,
+      queryClient,
+      projectId,
+      shouldAutostopSyncWhenLeavingScreen,
+    ]),
+  );
 
   let dockContent: React.ReactNode;
   let syncInfoContent: React.ReactNode;


src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -140,27 +141,38 @@ export const ProjectSyncDisplay = ({
...
+        }
+        // TODO: All queries associated with project should be invalidated
+        queryClient.invalidateQueries({queryKey: [OBSERVATION_KEY, projectId]});
+      });
 
-  //   return unsubscribe;
-  // }, [navigation, projectApi, shouldAutostopSyncWhenLeavingScreen]);
+      return () => {
+        unsubscribe();
+      };
```
EvanHahn, line 167: Nit: why not `return unsubscribe`?

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -140,27 +141,38 @@ export const ProjectSyncDisplay = ({
...
-  //   const unsubscribe = navigation.addListener('beforeRemove', () => {
-  //     projectApi.$sync.setAutostopDataSyncTimeout(30_000);
-  //   });
+  // Set up listener for autostopping sync (if applicable) and invalidating queries when navigating away from screen
+  useFocusEffect(
+    React.useCallback(() => {
+      const unsubscribe = navigation.addListener('beforeRemove', () => {
+        if (shouldAutostopSyncWhenLeavingScreen) {
+          projectApi.$sync.setAutostopDataSyncTimeout(30_000);
+        }
```
EvanHahn, line 160: If this never runs, how will sync get stopped after you leave the screen? (I think this is mostly a question about `useFocusEffect`, not your code.)

EvanHahn (approved):
LGTM assuming my comments don't uncover hidden issues.

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
achou11: what's "this" in this case? unless i'm misunderstanding, maybe this part from the PR description answers your question?

> otherwise, the user must explicitly press the stop sync button on the sync screen in order to disable sync.

so basically you'd have to return to the sync screen and press the stop button in order to stop sync. that stop button is present in the `complete-full` sync stage if sync is enabled (a fix I introduced via https://github.com/digidem/comapeo-mobile/pull/619/commits/a320f72a86e8e8dd32c12225bdcc336c453637aa because i had initially misunderstood that screen state in the designs).

if it's asking about how `useFocusEffect()` works - it's similar to useEffect except it's only run when the screen containing this component is in focus (in this case, the sync screen). there are cases with react navigation where screens can still be mounted but not focused, which I don't think affects us in this specific case, but i think using it here makes more sense anyways.

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
achou11: mostly just a personal style preference, but it's mostly informed by https://jakearchibald.com/2021/function-callback-risks/

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
EvanHahn: Ah, okay. I _think_ the intended behavior is:

- If the screen is active, there is _no_ auto-stop timeout. The user must explicitly start and stop sync.
- If the screen is inactive, the auto-stop timeout should be 30 seconds. If the user had previously enabled sync, it will stop after 30 seconds of inactivity.

This seems like it doesn't do the second part if sync is enabled but incomplete. In that case, leaving the screen will never stop sync, even if it completes.

EvanHahn: I was wrong about the intended behavior. All good.

EvanHahn (approved):

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
achou11: for posterity since we talked about this elsewhere:

Evan's point about the expected behavior when sync finishes in the "background" (i.e. while you're not on the sync screen) is a valid one. Ideally in that case, we can set the autostop when that happens, but that logic would have to live closer to the SyncStore implementation rather than the sync screen specifically. also think that this issue could be alleviated by providing UI that exists in other parts of the app to indicate the progress of sync while you're on different screens, which I think is something that is being considered but not fully thought through yet.

PR #632, opened 2024-08-28
chore: save track use editor component

### LINKED ISSUE #609: Convert SaveTrack Screen to use Editor
The [SaveTrack Screen](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/SaveTrack/SaveTrackScreen.tsx) should use the [Editor Component](https://github.com/digidem/comapeo-mobile/tree/develop/src/frontend/sharedComponents/Editor).

## To Do
- [x] Save Track Screen Uses Editor Component
- [x] Create a `HeaderLeft` Component that tracks if the user is trying to go back. Should follow the same pattern as [ObservationEdit](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/ObservationEdit/HeaderLeft.tsx)

cimigree:
closes #609 

### Description
- Converts save track screen to use editor
- Passes a boolean to the editor for some different behavior (see below)
- Takes out the Action Row as they are not needed for a track (the camera and details)
- Uses the notes component instead of the description field from the editor (which uses passed notes and update notes) since the formatting and logic are slightly different than for an Observation
- Extracts the header left stuff to a separate component
- Removes the ability to change the Category if the category is a track

Just FYI, I could not get the .env variable to work... It never seemed to load. 
Also, there was some preexisting behavior which I wasn't sure if it was part of the feature or is a bug. Just noting it here. 
- If I delete a track (when saving or from the list) it deletes the track but not the observations associated with it
- If I create a track after another track (on the same day? Dunno. Haven't extensively tested) then the observations done in other tracks get bundled into it. So if I do a track and have one observation and save it and then later do another track and add two observations, when I click into the second track, it says there are three observations and it includes the one from the first track.


src/frontend/sharedComponents/BottomSheet.tsx
achou11, line 1: finally! was always so confused about this being here 😄

src/frontend/sharedComponents/Editor/index.tsx
achou11, line 1: seeing the changes introduced here is making me realize that I'm not entirely satisfied with how this component is implemented prior to this PR 😄 what i'm suggesting below is probably out of the scope of this PR, but feel free to use it if you think it makes sense to. also open to doing this work myself as a follow-up.

---

in an ideal situation, this component should be refactored a bit so that it's a bit more flexible about how each of the relevant parts of the editor are rendered. Instead of passing a myriad of props that are mostly just forwarded to nested components or require slightly more complex logic to determine what to render, you could specify each of the rendered sections as element prop types e.g.

```tsx
type EditorProps = {
  actionsRow?: React.ReactNode;
  // Ideally there's a way to constrain the locationView, notesView, and presetView so that they require the correct component to be used
  // e.g. presetView={<PresetView ... />} is valid, but presetView={<LocationView ... />} is not
  locationView?: React.ReactElement;
  notesView: React.ReactElement;
  photos?: Photo[];
  presetView: React.ReactElement;
};

export const Editor = ({
  actionsRow,
  locationView,
  notesView,
  photos,
  presetView,
}: EditorProps) => {
  return (
    <ScreenContentWithDock
      dockContainerStyle={{padding: 0}}
      dockContent={actionsRow}>
      <View style={styles.container}>
        {presetView}
        {locationView && <Divider />}
        {locationView}
      </View>
      {notesView}
      {photos !== undefined && <MediaScrollView photos={photos} />}
    </ScreenContentWithDock>
  );
};
```

then the usage provides more flexibility for the consuming component. for example, the Editor in the SaveTrackScreen would look like this:

```tsx
      <Editor
        presetView={
          <PresetView
            isTrack
            PresetIcon={<TrackIcon style={styles.icon} />}
            presetName={t(m.newTitle)}
          />
        }
        notesView={<TrackDescriptionField />}
      />
```

the downside of this approach is that there doesn't seem to be a straightforward way of enforcing that the right component is being used for the prop type (see the comment in the first snippet above). personally would consider it a relatively minor issue, but could argue that it leaves room for accidental misuse if not documented well enough.

Here's a commit to show the changes, if helpful: https://github.com/digidem/comapeo-mobile/commit/e14e75a7d5f7528839de69da96a60f28ec7c1447

src/frontend/screens/SaveTrack/SaveTrackScreen.tsx
```diff
@@ -1,160 +1,43 @@
...
-          icon={<ErrorIcon width={60} height={60} />}
-        />
-      </BottomSheetModal>
+      <Editor
+        photos={[]}
+        presetName={t(m.newTitle)}
+        notesComponent={<TrackDescriptionField />}
+        PresetIcon={<TrackIcon style={styles.icon} />}
+        isTrack={true}
+      />
```
achou11: nit: can remove the fragment wrapping this component

src/frontend/sharedComponents/Editor/PresetView.tsx
```diff
@@ -31,7 +33,9 @@ export const PresetView = ({
         {PresetIcon}
         <Text style={styles.categoryName}>{presetName}</Text>
       </View>
-      <Text style={styles.changeButtonText}>{formatMessage(m.change)}</Text>
+      {!isTrack && (
+        <Text style={styles.changeButtonText}>{formatMessage(m.change)}</Text>
+      )}
```
achou11: nit: don't think this component should have any knowledge of what a track is. couple of alternative suggestions:

1. specify a more explicit `disabled` prop and use that (in addition to using it for the other props in this component)
2. use the presence of `onPressPreset` in this check instead. not sure if there's a time where that's not specified but we want to display this text

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -13,39 +13,47 @@ type EditorProps = {
...
   actionsRow?: React.ReactNode;
+  notesComponent?: React.ReactNode;
+  isTrack?: boolean;
 };
 
 export const Editor = ({
-  notes,
-  updateNotes,
+  notes = '',
+  updateNotes = () => {},
```
achou11: thoughts on removing the default props here and instead updating the prop types of `DescriptionField` to be optional instead?

achou11 (approved):
Couple of non-blocking suggestions, mostly related to previously existing shortcomings of the Editor component implementation.

> Just FYI, I could not get the .env variable to work... It never seemed to load.

strange. if you haven't tried already, do any of the following help?

- start the app with `npm start -- -c` (clears the expo cache)
- uninstall the app before running the android command

regarding the bugs that you've listed, both of them sound like bugs to me (although the first one may be more debatable?). would create issues and ask relevant people about them for addressing as follow-ups.

cimigree:
> * npm start -- -c

Yes!! That worked

src/frontend/sharedComponents/Editor/index.tsx
cimigree: @achou11 I think I am going to leave this refactor to you so you can set it up just how you imagine it in all of the various components. It looks like you have gotten a lot of the work done already.

src/frontend/sharedComponents/Editor/PresetView.tsx
cimigree: Ok I passed a disabled prop and used it. I didn't use it for the other props? Not sure what you meant by that since you said we want to display the Change text.

achou11: was referring to having a single prop to determine the functionality of the touchable here. so instead of having both `onPressPreset` and `disabled` determining how the it behaves and looks. my assumption is that we should only show the "Change" text if the button can actually be pressed, so the end result i was imagining was something along the lines of:

```tsx
    <TouchableOpacity 
      disabled={disabled} // basically just updating this line
      onPress={onPressPreset} // could technically update this line to be like `disabled ? onPressPreset : undefined`, but not strictly necessary in this case
      style={styles.preset}>
      <View style={{flexDirection: 'row', alignItems: 'center'}}>
        {PresetIcon}
        <Text style={styles.categoryName}>{presetName}</Text>
      </View>
      {!disabled && <Text style={styles.changeButtonText}>{formatMessage(m.change)}</Text>}
    </TouchableOpacity>
```

PR #633, opened 2024-08-29
feature: edit a track

### LINKED ISSUE #466: Edit Track
Using the [Editor Component](https://github.com/digidem/comapeo-mobile/tree/develop/src/frontend/sharedComponents/Editor), create the screen for editing a track that has already been created.

## To Do
- [x] In Track, check if user is creator of track, show edit button on the top right header if they are. Should follow the same pattern as [ObservationHeaderRight](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/Observation/ObservationHeaderRight.tsx)
- [x] On click of edit icon, navigate to `EditTrack` Screen
- [x] In Edit Screen, user can edit `description`
- [ ] In Edit Screen, user can add more photos (but is not able to delete photos)
- [x] When user tries to go back from Track Screen, the `ConfirmDiscardBottomSheetModal` should appear, should follow the same patterns as [ObservationEdit](https://github.com/digidem/comapeo-mobile/blob/ebaf02b9e8ea8e4806a87ed364e5ed77d416300d/src/frontend/screens/ObservationEdit/HeaderLeft.tsx#L79)

cimigree:
closes #466 

### Description
- Shows an edit button on the track screen if the user created the track which is created in a separate component as requested
- Adjusts the Track screen to accommodate that separate component
- Creates a TrackEdit folder with an edit screen and header left as requested (called TrackEdit so it would sit alphabetically near the other Track folders)
- Creates a Track Edit screen that allows for the editing of a description only since Tracks will not have photos attached to them (despite what the issue says)
- Adds a function in the hooks for tracks to edit a track
- Adjusts the use of presetDisabled in the PresetScreen to make sure the change button isn't shown and the onPress does nothing if presetDisabled is passed

### Screen Recording

https://github.com/user-attachments/assets/289b192e-6796-4c0a-986e-fdcce51daeef


src/frontend/screens/Track/TrackHeaderRight.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import {View, StyleSheet} from 'react-native';
+
+import {IconButton} from '../../sharedComponents/IconButton';
+import {useTrackQuery} from '../../hooks/server/track';
+import {useDeviceInfo} from '../../hooks/server/deviceInfo';
+import {UIActivityIndicator} from 'react-native-indicators';
+import EditIcon from '../../images/Edit.svg';
```
achou11: not clear at all, but i think to be consistent with what the [observation screen header](https://github.com/digidem/comapeo-mobile/blob/dfc7d9b1d0d48b203fb3d2bf68897932c59c17f1/src/frontend/screens/Observation/ObservationHeaderRight.tsx#L7) is doing, this should be using the component from the `sharedComponents/icon`. 

I noticed the default size of them being very different when navigating between the screens 😅

src/frontend/screens/Track/index.tsx
achou11, line 1: suggestion: the `useLayoutEffect()` here isn't necessary since you've removed a variable dependency that lived inside the screen component. implemented an example of removing it in https://github.com/digidem/comapeo-mobile/commit/8bbf3af11c5d9666458b81a45ef9ac47df3f8bc0

src/frontend/sharedTypes/navigation.ts
```diff
@@ -96,6 +96,7 @@ export type RootStackParamsList = {
   SaveTrack: undefined;
   Sync: undefined;
   Track: {trackId: string};
+  TrackEdit: {trackId: string} | undefined;
```
achou11: when would `undefined` happen?

src/frontend/screens/TrackEdit/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  navigation,
+  route,
+}) => {
+  const {formatMessage} = useIntl();
+  useEffect(() => {
+    if (!route.params || !route.params.trackId) {
+      navigation.goBack();
+    }
+  }, [route, navigation]);
+  const {trackId} = route.params as {trackId: string};
```
achou11: related to other comment about the type for this screen's route param: wondering what situation this is accounting for?

achou11 (changes requested):

achou11:
Forgot to add to the review but should probably update this with the relevant tracks screens too:

https://github.com/digidem/comapeo-mobile/blob/44d5d96e3d3c62c5bcc2458eef64643a990b52bf/src/frontend/constants.ts#L6

prevents the invite bottom sheet from appearing while on the listed screens

src/frontend/sharedTypes/navigation.ts
cimigree: I was following the example of ObservationEdit.
I was thinking it could be possible for someone to delete the track and maybe then quickly try to edit it before the deletion has gone through? It was just a guess but I can't actually replicate that. Happy to remove if you think it is best.

src/frontend/screens/TrackEdit/index.tsx
cimigree: See above.

cimigree:
> Forgot to add to the review but should probably update this with the relevant tracks screens too:
> 
> https://github.com/digidem/comapeo-mobile/blob/44d5d96e3d3c62c5bcc2458eef64643a990b52bf/src/frontend/constants.ts#L6
> 
> prevents the invite bottom sheet from appearing while on the listed screens

I did the SaveTrackScreen and TrackEdit. Is all that is needed?

src/frontend/sharedTypes/navigation.ts
achou11: i _think_ in the case of the observation flow, it's accounting for when you close the app and re-open it. i believe we maintain the navigation state while you're in that flow so that you can return to it immediately after re-opening the app. i don't think we're doing the same thing for the tracks flow at the moment although that's probably something we'd like to do in the future.

if we're not currently saving the nav state for the tracks-specific flow, I would lean towards removing the related code and re-introducing it later if we end up deciding on copying that approach.

cimigree: Ok. I removed the undefined.

src/frontend/screens/TrackEdit/index.tsx
cimigree: I removed the possibility it could be undefined.

src/frontend/screens/Track/index.tsx
```diff
@@ -41,35 +35,19 @@ const m = defineMessages({
...
-                left: HIT_SLOP_PADDING,
-              }}
-              onPress={() => {
-                // TODO: navigate to edit track screen
-              }}>
-              <EditIcon />
-            </TouchableOpacity>
-          )
-        : undefined,
+      headerRight: () => <TrackHeaderRight trackId={trackId} />,
```
achou11: can remove the `useLayoutEffect()` entirely here

achou11 (approved):

src/frontend/screens/Track/index.tsx
cimigree: Oops! Thanks for the good catch yet again!

PR #636, opened 2024-08-29
fix: add date to project

### LINKED ISSUE #588: Date Added to Project Missing
The date a user was added to a project is missing from the [Device Card](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/sharedComponents/DeviceCard.tsx)

## To Do
- [ ] Add date added to `DeviceCard`

### Notion Issue:
https://www.notion.so/digidem/Project-Creation-Your-Team-Date-added-missing-3b1f246c718d44928b212051daec0f4d

cimigree:
closes #588 

### Description
- Since we now have access through the member API to a joinedAt property, this PR uses it in the Device Card in the Your Team section of Project Information to indicate when someone joined a project
- Uses react-intl to make sure it is formatted correctly in various languages
- I found the UX [here](https://www.figma.com/proto/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?node-id=1075-5095&node-type=FRAME&t=EQD08hgLH4dKDmM4-0&scaling=min-zoom&content-scaling=fixed&page-id=1061%3A3746&starting-point-node-id=1070%3A4687)
- The styles will need to be adjusted when leave project is added (I assume it hasn't been added yet)
- I had trouble styling for various font and display sizes and longer device names, so that is why the date is not right aligned for all sizes

### Screenshots
---
large text normal name length
---

<image src="https://github.com/user-attachments/assets/5977ac91-c0eb-4d45-9cb5-f32ba908c62b" width="250" />

---
default text size with normal name length
---

<image src="https://github.com/user-attachments/assets/8aebb0fe-00e4-4963-91b1-3234e4f079d8" width="250" />

---
really long name with default text size
---

<image src="https://github.com/user-attachments/assets/3194c0a6-eb82-4f8a-9321-c58354b5dcbe" width="250" />

---
really long name with larger text size
---

<image src="https://github.com/user-attachments/assets/9488e4a6-2315-4c9d-8542-a74d7d2b832a" width="250" />

### DIFF
--- a/src/frontend/sharedComponents/DeviceNameWithIcon.tsx
+++ b/src/frontend/sharedComponents/DeviceNameWithIcon.tsx
@@ -7,7 +7,7 @@ import type {
   DeviceConnectionStatus,
   DeviceType,
 } from '../sharedTypes';
-import {defineMessages, useIntl} from 'react-intl';
+import {defineMessages, useIntl, FormattedDate} from 'react-intl';
 import {Text} from './Text';
 import {MEDIUM_GREY} from '../lib/styles';
 import {ExhaustivenessError} from '../lib/ExhaustivenessError';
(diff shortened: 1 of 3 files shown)


src/frontend/sharedComponents/DeviceNameWithIcon.tsx
ErikSin, line 1: The purpose of this component was for just the name and Icon (because it is used in other screens/components). The `joinedAt` date should live outside of this in the device card

ErikSin (changes requested):
I know you probably spent alot of time tinkering with the flex, but is it possible to use space-between and proportional flex sizing (eg flex basis or flex grow/shring). I assume you probably have tried those things and flex is not as simple as it should be, but just wanted to confirm

achou11:
@cimigree I think @ErikSin addressed this issue via https://github.com/digidem/comapeo-mobile/pull/674. is it okay to close this PR?

PR #630, opened 2024-08-27
Update `@mapeo/core` to 9.0.0-alpha.23

EvanHahn:

PR #641, opened 2024-09-02
chore: update postion from expo

ErikSin:
previously when creating an observation, the `mocked` property was manually set. This property now comes from the expo-location object (ensuring data integrity as the value now comes from the phone as opposed to manually being set). I also simplified how the position object turns "null" into "undefined" and allowed the function to do that conversion.


cimigree (approved):
I tested this out as best I could using my Android device and the maestro tests with  locations provided there and it behaved the same as with develop, which is fine.
I was not able to manually set a location (when I turned off location permission), but I think that is as expected?
I also used the location mocking while watching the performance monitor and that all looked good.
It's great that we can use expo-location and you were able to simplify debounceLocation

ErikSin:
im actually going to close this as it is addressed by #630 and the updated schema chages. Thanks @EvanHahn

PR #642, opened 2024-09-02
fix: presets forever loading

### LINKED ISSUE #611: Preset icons are sometimes stuck loading
Sept 19 Update : Xime : Since the CoMapeo1.0.0-RC.3 build, it has been happening more consistently. See videos here : https://www.notion.so/digidem/Inconsistent-icons-share-behavior-1041b08162d580ae9714e9b406dfe905

Hopefully fixing it will also fix : https://www.notion.so/digidem/Share-difficulty-1041b08162d5801a9cc9eb9d2d396a82
***

**Describe the bug**

Preset icons don't seem to be appearing in multiple places. They are stuck in the loading state.

**Screenshots**

| Screenshot 1 | Screenshot 2 |
|---|---|

**To Reproduce**

I don't know how to reproduce this, but it seems to happen when joining a project for the first time.

It seems to go away when resetting the app.

**Additional context**

@achou11 saw this on one of his devices, @rudokemper on another, and @EvanHahn on a third.

I assume the bug comes from [this code](https://github.com/digidem/comapeo-mobile/blob/0da8ea773ad0d7d399abfe11b7c3eaf12cf2926f/src/frontend/hooks/server/presets.ts#L39-L60), but I'm not sure.

cimigree:
related to #611 

### Description
- I did a lot of work to make sure an image is present for the category because I imagine that our users really rely on visual images as opposed to words

_This is the major change_

- **Changes the way preset icons are fetched. Previously, on every single screen or component where there was a single icon, all of the presets were fetched and then mapped through to find the one with the correct name and then that one's image url was gotten. Instead, we now use the actual preset, which we had anyway, to just fetch the url for the image.**
  - In this new hook, I am using retry, in case there might be a temporary network problem, and I do it twice to increase chances of success
  - 2 second delay between retry attempts
  - I set a stale time of 1 minute to ensure that the icon data is considered fresh for at least a minute after it’s fetched. That way, React Query won’t refetch the icon image url, even if the component remounts, so there aren't extra requests.
  - I set a gcTime (garbage collection time) to 5 minutes so that it will stay in memory for 5 minutes after a user navigates away from the component before it is considered stale. It will be retrieved quickly if the user returns within that time and won't need to fetch data from the server
  - Keeping previous data so that while the query is fetching new data, we use the data from the previous execution. This avoids flashing or loading states.
- So, now in PresetIcon, we use the preset instead of the name and just fetch the imageurl from the preset.
  - I handle when the preset might be undefined by using the fallback (I think this mainly occurs when actively developing the app and won't happen much for a user, but it makes development smoother)
  - Now managed I manage the iconUrl with state. This way, the component can display the previously fetched icon while new data is loading
  - More robust error handling by using the query's error
  - The loading state is still checked, but now it checks whether there is already a valid iconUrl, so the component will continue showing the old icon rather than the loading spinner.
  - The fallback icon is displayed if there’s an error or if the iconUrl is missing or doesn't load. 
- In all of the places that use the PresetIcon, I pass the preset and not the name
  
  In truth, I was only able to replicate the error once accidentally. In that case, the loading spinners displayed then eventually became the fallback icon.
  It was right after creating a new project.
  Once that happened, through Maestro I ran tests on this scenario (starting app from scratch, creating a new project, going straight to the PresetChooser screen) at least 20 times and I was never able to replicate it again! So might be fixed? But more importantly, if the loaders all display, we go to the fallback icon.


src/frontend/hooks/server/presets.ts
```diff
@@ -36,25 +37,29 @@ export function usePresetsMutation() {
   });
 }
 
-export function useGetPresetIcon(size: IconSize, name?: string) {
-  const {projectId, projectApi} = useActiveProject();
+export function useGetPresetIconFromPreset(
```
ErikSin: nit: this name change seems unnecessary.

src/frontend/hooks/server/presets.ts
```diff
@@ -36,25 +37,29 @@ export function usePresetsMutation() {
...
+  preset: PresetValue,
+  size: IconSize,
+) {
+  const {projectApi} = useActiveProject();
 
-  return useQuery({
-    queryKey: ['presetIcon', projectId, size, name],
-    enabled: !!name,
+  return useQuery<string, Error>({
+    queryKey: ['presetIconFromPreset', preset.iconRef?.docId, size],
```
ErikSin: projectId needs to be part of this query key

src/frontend/screens/PresetChooser.tsx
```diff
@@ -89,16 +88,18 @@ export const PresetChooser: NativeNavigationComponent<'PresetChooser'> = ({
...
-        .sort((a, b) => {
-          return a.name.toLowerCase().localeCompare(b.name.toLowerCase());
-        });
+        .sort((a, b) =>
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+        );
 
   const handleSelectPreset = (selectedPreset: Preset) => {
+    if (!selectedPreset) return;
+
```
ErikSin: is this necessary? selectedPreset is not an optional parameter

src/frontend/sharedComponents/icons/PresetIcon.tsx
```diff
@@ -26,36 +26,44 @@ const radii = {
...
+      return <MaterialIcon name="place" size={iconSize} />;
+    }
+    const [iconUrl, setIconUrl] = useState<string | null>(null);
+    const {data, isLoading, error} = useGetPresetIconFromPreset(preset!, size);
+    useEffect(() => {
+      if (data && !iconUrl) {
+        setIconUrl(data);
+      }
+    }, [data]);
+    if (isLoading && !iconUrl) return <UIActivityIndicator size={30} />;
```
ErikSin: this state management seems a little complicated. One thing to note is that the architecture of tanstack-query is `stale while revalidate`. So `isLoading` is only true when there is no data and it is fetching. If there is data, it will serve the data and isLoading will be false EVEN WHEN it is fetching new data in the background.

src/frontend/hooks/server/presets.ts
```diff
@@ -36,25 +37,29 @@ export function usePresetsMutation() {
   });
 }
 
-export function useGetPresetIcon(size: IconSize, name?: string) {
-  const {projectId, projectApi} = useActiveProject();
+export function useGetPresetIconFromPreset(
+  preset: PresetValue,
```
ErikSin: Were just using the preset `docId` in this function. Should we just pass that directly, instead of the entire object?

src/frontend/screens/PresetChooser.tsx
```diff
@@ -89,16 +88,18 @@ export const PresetChooser: NativeNavigationComponent<'PresetChooser'> = ({
...
     : Array.from(presets)
         // Only show presets where the geometry property includes "point"
         .filter(p => p.geometry.includes('point'))
         // Sort presets by sort property and then by name, then filter only point presets
-        .sort((a, b) => {
-          return a.name.toLowerCase().localeCompare(b.name.toLowerCase());
-        });
+        .sort((a, b) =>
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+        );
```
ErikSin: the ternary is not actually necessary here (I think i originally added it...)

presets will always be an array (sometimes it can be empty, but still an array regardless)

you can just do 

```ts
const presetsList = Array.from(presets)
        // Only show presets where the geometry property includes "point"
        .filter(p => p.geometry.includes('point'))
        // Sort presets by sort property and then by name, then filter only point presets
        .sort((a, b) =>
          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),

```

ErikSin (changes requested):
So tbh i don't know what the actual issue was. I am unable to reproduce the issue. I think we need to explore why it was happening in the first place...

That being said, i think adding a retry time and delay is still useful so I think we can still merge it once a couple issues have been resolved. Namely, the state management of the `presetCircleIcon` is a little convoluded and I think should be cleared up a bit and rely more heavily on tanstack's state management.

src/frontend/screens/PresetChooser.tsx
cimigree: I think I ran into some weirdness during development, but you are right. This would not happen with a real user. removed.

src/frontend/sharedComponents/icons/PresetIcon.tsx
cimigree: OK changed.

src/frontend/hooks/server/presets.ts
cimigree: Ok I did that.

src/frontend/screens/PresetChooser.tsx
cimigree: Ok fixed. Thanks for the tip.

cimigree:
@ErikSin I made the changed you requested and rerequested a review from you. I also had to make some unrelated changes because of the PR from Evan related to how we figure out if someone created a track. Could you just peek at that (in TrackHeaderRight) to make sure it looks right? I tested it locally and it seemed to work.

src/frontend/hooks/server/presets.ts
```diff
@@ -36,25 +37,26 @@ export function usePresetsMutation() {
...
+      return await projectApi.$icons.getIconUrl(docId, {
+        mimeType: 'image/png',
+        size: size,
+        pixelDensity: 3,
+      });
     },
+    retry: 2,
+    retryDelay: 2000,
+    staleTime: 60000,
+    gcTime: 300000,
```
ErikSin: this is the default `gctime` so it does not need to be set

src/frontend/sharedComponents/icons/PresetIcon.tsx
```diff
@@ -26,36 +25,41 @@ const radii = {
...
-    if (isLoading) return <UIActivityIndicator size={30} />;
+    if (!presetDocId) {
+      return <MaterialIcon name="place" size={iconSize} />;
+    }
+    const {
+      data: iconUrl,
+      isLoading,
+      error,
+    } = useGetPresetIcon(presetDocId, size);
+    if (isLoading && !iconUrl) return <UIActivityIndicator size={30} />;
```
ErikSin: you can replace `if(isLoading && !iconUrl)` with `if(isPending)`. See https://github.com/TanStack/query/discussions/6297

ErikSin (approved):
two minor changes. Also i don't think this actually closes 611. There is a bigger problem, and this is a bandaid solution. I think there are improvements regardless. So lets not close 611 on merge (aka, can you change the description so it does not say "closes")

PR #645, opened 2024-09-03
fix: use proper timestamp for location metadata

### LINKED ISSUE #644: Observation.metadata.position.timestamp should be an ISO timestamp
**Describe the bug**
https://github.com/digidem/comapeo-mobile/blob/b423f24ff79167e190ae11eaaad48ee77b94df37/src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts#L55 converts the timestamp as milliseconds since UNIX epoch to a string representation of that number. The timestamp property should be an ISO datetime string, like all our other timestamps.

Should be `new Date(timestamp).toISOString()`

EvanHahn:
We should be using an ISO timestamp instead of a string number here.

Because `position` can be undefined and we were checking that earlier, I moved things around a tiny bit to only check that once.

Closes #644.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,6 +61,7 @@
         "geojson-geometries-lookup": "^0.5.0",
         "lodash.isequal": "^4.5.0",
         "luxon": "^3.4.4",
+        "map-obj": "^5.0.2",
         "nanoid": "^5.0.1",
         "nodejs-mobile-react-native": "^18.17.8",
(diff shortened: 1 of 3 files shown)


achou11 (approved):
wonder if this issue exists elsewhere 🤔 there's kind of a confusing assortment of location-related types where in some cases a `timestamp` field is a number and not a string

PR #526, opened 2024-07-18
Feat leave project

ErikSin:
Adds leave project screens.

Added as a separate modal since it needed to be full screen.

We couldn't add the leave project screens as a nav screen because the invite modal is declarative and opens when there is an active invite and closes when there are no more active invites. Due to the architecture of modal, they always show up on top of the navigator. This means that the modal would show up on top of the leave project screens if they were part of the nav.

https://github.com/user-attachments/assets/f55ab30d-3fce-4d21-bbc6-90d5cbfb88b0

closes #162

PR #646, opened 2024-09-03
fix: leave current project when joining another project

### LINKED ISSUE #525: Leave Project when accepting a new invite
Currently the user accepts an invite and does not leave the current project. According to the designs we should be leaving the project as well. This is a limitation of the API, which does not allow a user who is the only person on a project to leave a project (which is not desired behaviour). Once this gets fixed, we should be able to properly leave project.

Related to issue [717](https://github.com/digidem/mapeo-core-next/issues/717)

achou11:
Fixes #525 

In case someone runs into this and is initially as confused as I was: after leaving a project, the returned data from `api.listProjects()` still includes the project - it just won't have much information tied to it (except its project id). Related to  https://github.com/digidem/mapeo-core-next/issues/433

### DIFF
--- a/src/frontend/sharedComponents/LeaveProjectModalContent/LeaveProject.tsx
+++ b/src/frontend/sharedComponents/LeaveProjectModalContent/LeaveProject.tsx
@@ -14,6 +14,7 @@ import {UIActivityIndicator} from 'react-native-indicators';
 import {BottomSheetModalContent} from '../BottomSheetModal';
 import {useAcceptInvite} from '../../hooks/server/invites';
 import {ErrorBottomSheet} from '../ErrorBottomSheet';
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
 
 const m = defineMessages({
   leaveProj: {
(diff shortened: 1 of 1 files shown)


src/frontend/sharedComponents/LeaveProjectModalContent/LeaveProject.tsx
```diff
@@ -70,46 +71,38 @@ export const LeaveProject = ({
...
-          setCombinedLoading(false);
         },
       },
     );
   }
 
   return (
     <>
-      {combinedLoading ? (
+      {accept.status === 'pending' || leaveProject.status === 'pending' ? (
```
ErikSin: have you tested this out? I think this actually will still cause it to flicker becasue there is a breif time when `accept.status === success` but the `leaveProject` mutaion hasnt been started yet, so `leaveProject.status === idle`. When you made [this comment ](https://github.com/digidem/comapeo-mobile/pull/526#discussion_r1683485870) i think we both were thinking of `query`, where `pending===true` when the query hasn't been made yet. But in a mutation `pending===false` when it hasnt been called yet

achou11: hm i see what you mean. haven't tested it enough to confirm but what you describe makes sense. i was considering doing something like the following so that we could just rely on the `accept.status`:

```ts
accept.mutate({
  onSuccess: async () => {
    await leaveProject.mutateAsync({ ... })
  }
})
```

in that case, i believe the accept mutation will stay pending until `leaveProject.mutateAsync()` resolves. was a little worried that it might result in some edge cases around error-handling though

achou11: although part of me thinks that what i currently have here works as intended. would have to do some logging to confirm, but my basic understanding is that the following sequence happens:

1. Start accept mutation, so `accept.status === 'pending'`

2. The accept mutation succeeds, so `onSuccess` is called, which calls the leave project mutation so `leave.status === 'pending'`

3. After `accept.onSuccess()` finishes, `accept.status === 'idle'`, but since leave project mutation was kicked off in (2), we have `accept.status === 'idle'` and `leave.status === 'pending`.

achou11: yeah this is what i get in the logs (although replace `idle` with `success` since they're mutations i guess):

```sh
{"accept": "pending", "leave": "idle"}
{"accept": "pending", "leave": "pending"}
{"accept": "success", "leave": "pending"}
{"accept": "success", "leave": "pending"}
{"accept": "success", "leave": "success"}
{"accept": "success", "leave": "success"}
{"accept": "success", "leave": "success"}
```

So yeah i think the flicker will be a result of both mutations being `success` in this case. so same problem but slightly different reasoning than what you described. Will update this line to handle that

ErikSin: I must have done something slightly different because i originally was just using the loading state of the 2 mutations in the conditional rendering statement, and it was causing a flicker, which is why i created that `combinedLoadingState`But it seems like its not a problem. Maybe ill just test it out with a device really quick to make sure

achou11: With current check:

https://github.com/user-attachments/assets/9c067d86-8d70-445f-830c-98281e844808

---

After adding condition for when they're both in  `success` status:

https://github.com/user-attachments/assets/721902ae-c392-4f90-8e1c-a3ab70727405


I think there's a slight improvement so probably worth adding the additional check. There's additional flickering going on in the map but that's due to lots of query invalidation that's outside the scope of this PR

achou11: see ff1982ff9d1e0b6d60890eb06e37930acc64dad7 for relevant commit

ErikSin: ahhh I see now. So the 'combinedLoading' reduced the flickering not because of a race condition, but because it was set AFTER the modal was already being set to close

ErikSin (approved):
lets cherry pick this into the release!

PR #648, opened 2024-09-03
feature: sync data selector

### LINKED ISSUE #411: Project Settings / Sync data selector
## Project Settings / Sync Settings
User has access to the Sync Settings via the Project Settings
Via the Sync Settings, user be will able to decide whether their device will only sync photos (previews) or become an archive device and sync everything. 

## Settings
- [x] Update the *Project Settings* screen by adding the *Sync Settings* button

## Sync Settings
Image

User clicks on *Sync Settings* and is taken to this screen 

- [x] Display **Sync Previews (Photos Only)** and  **Photos will sync at a reduced smaller size.** 
- [x] Display a radio button next to it
- [x] Display **Sync Everything** and **Your device will sync all content at full size, including photos, audio, and videos
- [x] Display a radio button next to it
- [x] User can click one or the other 

## Sync previews
Image

- [x] User clicks on **Sync Previews (Photos Only)** , bottom sheet opens 
- [x] Display content as designed
- [x] Display warning icon 
- [ ] Disolay copy **Sync Previews? Your device will keep all existing data but new observations will sync in a smaller, preview size.** **You will no longer sync Audio or Video.**
- [x] Create **Cancel** button
- [x] If clicked, back to  *Sync Settings*
- [x] Create **Sync Previews** button
- [ ] If clicked, display ...? @ErikSin i dont see a confirmation state on figma, check w sabella
- [ ] If clicked, the user will be syncing a preview and thumbnails of their data. it will not sync audio or video

## Sync Everything
Image

- [ ] User clicks on **Sync Everything** , bottom sheet opens 
- [x] Display content as designed
- [ ] Display warning icon 
- [x] Disolay copy **Sync Everything? You are about to Sync everything. This may increase the disk space used on your device.**
- [x] Create **Cancel** button
- [x] If clicked, back to  *Sync Settings*
- [x] Create **Sync Everything**** button
- [ ] If clicked, display ...? same as sync previews
- [ ] If clicked, the user will be syncing all data and be consiered an archive device

## Media variants
See notes from Gregor on media variants https://www.notion.so/digidem/Media-Disk-Space-Management-b8801ef280d8422b801d15f3f7c78787?pvs=4#8afd5a4b51da444f9c63124c83767998

## Design
https://www.figma.com/file/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?type=design&node-id=590-1313&mode=design

cimigree:
closes #411 
### Description
- Please add an ENV variable of `EXPO_PUBLIC_FEATURE_MEDIA_MANAGER=true` as a feature flag
- Builds out a screen accessed from Project Settings to determine what type of media the user wants to have
- Sets the default to Sync Everything because that is what the current behavior is
- Makes a component that uses the bottom sheet modal content component to use with the selections
- Because of the way Select One works, if you click on a button and it is already selected nothing happens. Should I change that? Wasn't sure...
- There is nothing to be done yet with the actions (no backend) so this is just creating the screens and a persisted setting for now.
- Not sure what to use here since it won't be active for a while. Chore? Feature? Fix?

### Screen Shots
<image src="https://github.com/user-attachments/assets/3ecf65da-539b-4f87-b0e0-f9c42ddc9a66" width="250" />
<image src="https://github.com/user-attachments/assets/2bd60123-f185-4808-893c-5cbb24afaa32" width="250" />
<image src="https://github.com/user-attachments/assets/02fa410a-178d-4781-865c-0e7d6bd02f4a" width="250" />


src/frontend/screens/Settings/ProjectSettings/SyncActionSheet.tsx
```diff
@@ -0,0 +1,58 @@
...
+              text: confirmActionText,
+              variation: 'filled',
+              onPress: confirmAction,
+            },
+          ]}
+        />
+      </BottomSheetModal>
+    );
+  },
+);
```
achou11: suggestion: it's not clearly documented but it's usually unnecessary to create components that wrap the `BottomSheetModal` component as it's designed to be co-located with the usage of the `useBottomSheetModal()` hook. Instead, define what's needed for the `BottomSheetModalContent`:

```suggestion
type SyncActionSheetProps = {
  title: string;
  description: string;
  confirmActionText: string;
  confirmAction: () => void;
  onDismiss: () => void;
};

export const SyncActionSheet = ({
  title,
  description,
  confirmActionText,
  confirmAction,
  onDismiss,
}: SyncActionSheetProps) => {
  const {formatMessage: t} = useIntl();

  return (
    <BottomSheetModalContent
      title={title}
      description={description}
      icon={<Warning width={60} height={60} />}
      buttonConfigs={[
        {
          text: t(m.cancel),
          onPress: onDismiss,
          variation: 'outlined',
        },
        {
          text: confirmActionText,
          variation: 'filled',
          onPress: confirmAction,
        },
      ]}
    />
  );
};
```

Might also be worth renaming the component to make it clearer that it's just the content of the bottom sheet.

Then in `SyncSettings`, update the usage to something like the following:

```tsx
<BottomSheetModal ref={sheetRef} isOpen={modalState.isOpen}>
  <SyncActionSheet
    title={
      modalState.type === 'previews'
        ? t(m.syncPreviewsBottomSheet)
        : t(m.syncEverythingBottomSheet)
    }
    description={
      modalState.type === 'previews'
        ? t(m.syncPreviewsDescriptionBottomSheet)
        : t(m.syncEverythingDescriptionBottomSheet)
    }
    confirmActionText={
      modalState.type === 'previews'
        ? t(m.syncPreviewsBottomSheetConfirm)
        : t(m.syncEverything)
    }
    confirmAction={handleConfirm}
    onDismiss={() => {
      setModalState({type: null, isOpen: false});
      closeSheet();
    }}
  />
</BottomSheetModal>
```

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const {formatMessage: t} = useIntl();
+  const syncSetting = usePersistedSettings(store => store.syncSetting);
+  const {setSyncSetting} = usePersistedSettingsAction();
+
+  const [modalState, setModalState] = React.useState<{
+    type: 'previews' | 'everything' | null;
+    isOpen: boolean;
+  }>({
+    type: null,
+    isOpen: false,
```
achou11: is `isOpen` necessary given that there's an `isOpen` field when using the `useBottomSheetModal()` hook?

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,151 @@
...
+  },
+});
+
+export const SyncSettings = () => {
+  const {formatMessage: t} = useIntl();
+  const syncSetting = usePersistedSettings(store => store.syncSetting);
+  const {setSyncSetting} = usePersistedSettingsAction();
+
+  const [modalState, setModalState] = React.useState<{
+    type: 'previews' | 'everything' | null;
```
achou11: ideally `null` doesn't need to be included for this, since at least one of the settings will always be selected. would suggest instantiating this state based on the result from `usePersistedSettings()`. For example:

```ts
const [modalState, setModalState] = React.useState(() =>
  return {type: syncSetting, isOpen: false};
})
```

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,151 @@
...
+    setModalState({type: value, isOpen: true});
+    openSheet();
+  };
+
+  const handleConfirm = () => {
+    if (modalState.type === 'previews') {
+      setSyncSetting('previews');
+    } else if (modalState.type === 'everything') {
+      setSyncSetting('everything');
+    }
```
achou11: removing the union of `null` as suggested in a separate comment can simplify this a bit:

```ts
    setSyncSetting(modalState.type);
```

src/frontend/sharedComponents/SelectOne.tsx
```diff
@@ -26,16 +31,35 @@ export const SelectOne = <T,>({
             : index
         }
         testID={`${item.value}LanguageButton`}
-        onPress={() => value !== item.value && onChange(item.value)}>
+        onPress={() => value !== item.value && onChange(item.value)}
+        style={[radioButtonPosition === 'right' && styles.rowReverse]}>
```
achou11: nit: to avoid an unnecessary array allocation

```suggestion
        style={radioButtonPosition === 'right' ? styles.rowReverse : undefined}>
```

src/frontend/sharedComponents/SelectOne.tsx
```diff
@@ -26,16 +31,35 @@ export const SelectOne = <T,>({
...
           iconName={
             item.value === value
               ? 'radio-button-checked'
               : 'radio-button-unchecked'
           }
+          style={[
+            radioButtonPosition === 'right'
+              ? styles.alignRight
+              : styles.alignLeft,
+          ]}
```
achou11: nit: no need for this to be within an array

```suggestion
          style={
            radioButtonPosition === 'right'
              ? styles.alignRight
              : styles.alignLeft,
          }
```

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,151 @@
...
+    defaultMessage: 'Sync Previews?',
+  },
+  syncEverythingBottomSheet: {
+    id: 'screens.SyncSettings.syncEverythingButtonBottomSheet',
+    defaultMessage: 'Sync Everything?',
+  },
+  syncPreviewsDescriptionBottomSheet: {
+    id: 'screens.SyncSettings.syncPreviewsDescriptionBottomSheet',
+    defaultMessage:
+      'Your device will keep all existing data but new observations will sync in a smaller, preview size.\n\nYou will no longer sync Audio or Video.',
```
achou11: does `\n` work as expected? based on my testing, it doesn't seem to:

image

image

seems like the proper solution is to either create a separate message for this or add another rich format tag (similar to what we do for bold)

src/frontend/hooks/persistedState/usePersistedSettings.ts
```diff
@@ -5,21 +5,25 @@ import {CoordinateFormat} from '../../sharedTypes';
 type SettingsSlice = {
   coordinateFormat: CoordinateFormat;
   manualCoordinateEntryFormat: CoordinateFormat;
+  syncSetting: 'previews' | 'everything';
```
achou11: suggestion: maybe define a shared type for this to reduce some keystrokes when working with this elsewhere?

achou11 (changes requested):
overall looking good! couple of issues I noticed with the UI when compared to the designs, plus a couple of suggestions

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: That is so weird! It does work on my device as you can see from the screen shots. Good thing we test on different devices! Fixed (I hope)

src/frontend/screens/Settings/ProjectSettings/SyncActionSheet.tsx
cimigree: Ok done.

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: Removed.

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: Removed.

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: Simplified.

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
achou11: hm yeah that's odd. fwiw, I do remember it working so I was surprised to see it not working for me. maybe I need to try a couple things on my end, or at least test on a different device to confirm

src/frontend/sharedComponents/SelectOne.tsx
cimigree: Yeah I don't know why I put that in an array. Sorry. Good catch!

src/frontend/hooks/persistedState/usePersistedSettings.ts
cimigree: Ok. Done.

achou11:
not seeing any changes despite your responses to comments - did you forget to push them?

cimigree:
> not seeing any changes despite your responses to comments - did you forget to push them?

I guess I did forget. Sorry @achou11 . Just pushed now

src/frontend/screens/Settings/ProjectSettings/SyncActionSheetContent.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {defineMessages, useIntl} from 'react-intl';
+import Warning from '../../../images/Warning.svg';
+import {
+  BottomSheetModal,
+  BottomSheetModalContent,
+} from '../../../sharedComponents/BottomSheetModal';
+
+const m = defineMessages({
+  cancel: {
+    id: 'screens.SyncActionSheet.cancel',
```
achou11: nit: update to somewhat match the path of this file

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ];
+
+  return (
+    <ScrollView style={styles.container}>
+      <SelectOne
+        value={syncSetting}
+        onChange={handleOptionChange}
+        options={options}
+        radioButtonPosition="right"
+        color={SYNC_BACKGROUND}
```
achou11: wondering if this should be `MEDIUM_BLUE` instead? just a guess based on an existing shared component related to radio buttons

https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/sharedComponents/AnimatedRadio.tsx

Tangentially, seems like that component isn't used anymore. Wondering if it would've made sense to use that for this UI? Probably not but just curious... (otherwise should probably just remove that component)

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
```diff
@@ -0,0 +1,146 @@
...
+    defaultMessage: 'Sync Previews?',
+  },
+  syncEverythingBottomSheet: {
+    id: 'screens.SyncSettings.syncEverythingButtonBottomSheet',
+    defaultMessage: 'Sync Everything?',
+  },
+  syncPreviewsDescriptionBottomSheet: {
+    id: 'screens.SyncSettings.syncPreviewsDescriptionBottomSheet',
+    defaultMessage:
+      'Your device will keep all existing data but new observations will sync in a smaller, preview size.<newline/><newline/>You will no longer sync Audio or Video.',
```
achou11: needs a couple of other changes elsewhere for this to work. would need to add a new custom format e.g.

```ts
newline: () => '\n'
```

to https://github.com/digidem/comapeo-mobile/blob/ee00a488265fefd17f53d6ae59f931467d141332/src/frontend/contexts/IntlContext.tsx#L22-L26



and then every usage of that format has to be in the form of a proper xml-like tag pair e.g. `<newline></newline>` (self-closing tag format does not work, see https://github.com/formatjs/formatjs/issues/1616). 

in order to properly see this working, make sure you do the following:

1. stop the metro server
2. run `npm run extract-messages`
3. run `npm start` (which will - importantly - run `npm run build:translations`)

I suspect the disparity we were seeing before in terms of the raw newline characters is related to not having done (2). Extracting the messages via formatjs strips out newline characters (see https://github.com/formatjs/formatjs/issues/3286) which means that the corresponding message in the `translations/` folder doesn't contain the newline characters, hence why i was seeing the issue and you weren't. without the message existing in that directory, the UI will fallback to the message defined in `defaultMessage` in the app code (which contains the newlines).

src/frontend/hooks/persistedState/usePersistedSettings.ts
```diff
@@ -1,25 +1,29 @@
 import {type StateCreator} from 'zustand';
 import {createPersistedState} from './createPersistedState';
-import {CoordinateFormat} from '../../sharedTypes';
+import {CoordinateFormat, SyncSetting} from '../../sharedTypes';
 
 type SettingsSlice = {
   coordinateFormat: CoordinateFormat;
   manualCoordinateEntryFormat: CoordinateFormat;
+  syncSetting: SyncSetting;
```
achou11: the naming of this setting feels kind of vague. this is specifically related to media sync, so maybe rename that (plus any relevant types and variables) to be clearer about that? e.g. `mediaSyncMode`

achou11 (changes requested):
> Because of the way Select One works, if you click on a button and it is already selected nothing happens. Should I change that? Wasn't sure...

I would keep it as described here (i.e. keep it idempotent)

> Not sure what to use here since it won't be active for a while. Chore? Feature? Fix?

would categorize as a feature

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
achou11: think i figured out why we're seeing different things - see https://github.com/digidem/comapeo-mobile/pull/648#discussion_r1745970150

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: I used the color that was in the UI designs. Is that not the proper procedure?
Screenshot 2024-09-05 at 2 52 25 PM

Should we keep the component in case it is needed in the future?

achou11: > I used the color that was in the UI designs. Is that not the proper procedure?

nah i think that's the correct approach. especially since it's a color that already exists in the codebase (albeit with an unnecessarily restrictive name).

> Should we keep the component in case it is needed in the future?

I don't think we should, but would leave that for a separate PR

src/frontend/hooks/persistedState/usePersistedSettings.ts
cimigree: Changed to mediaSyncSetting... Hope that is ok

src/frontend/screens/Settings/ProjectSettings/SyncSettings.tsx
cimigree: Ok I made them separate text components instead of altering the IntlContext

achou11: extracted the above into a separate issue: https://github.com/digidem/comapeo-mobile/issues/669

achou11 (approved):

PR #653, opened 2024-09-04
fix: updates hook for track editing

cimigree:
This is making CI fail for everything on develop.
This fix changes the hook from what it used to be to what it is now to be able to tell if a track was created by the device that is trying to edit it.


ErikSin (approved):
Thanks for catching!

PR #656, opened 2024-09-04
fix: manage "obscured" config variables via babel env variable transform

gmaclennan:
**NOTE:** This PR will require developers to add these environment variables to their local `.env` file:

- `MAPBOX_ACCESS_TOKEN`
- `COMAPEO_METRICS_URL`
- `COMAPEO_METRICS_API_KEY`

The values for these variables are in our 1Password Technology vault. Attempts to build the app will fail with an error if these are not set (to avoid runtime errors)

This PR changes how we add runtime semi-sensitive configuration values to the app.

Currently we use Expo's [environment variables](https://docs.expo.dev/guides/environment-variables/). The value of any environment variable that starts with `EXPO_PUBLIC_` is used to replace any runtime code in the app JS bundle, as long as the environment variable is accessed via `process.env.EXPO_PUBLIC_MY_VAR`.

When building locally, these environment variables can be set in `.env` files. However when building on EAS, these variables must be set in `eas.json`, which must be checked into source control. Expo is doing this to really reinforce that these variables are public (because they can be read by someone who scans the APK), however it is limiting for our needs.

We have some configuration values that are not "secrets" as such, but we would not like to make them too visible to the public to avoid abuse. These include:

- Mapbox public access token, which can be "stolen" by someone to use on their own map, which would run up a bill for our mapbox account if used a lot.
- Metrics URL and API Key, which someone could use to post spurious metrics data.

Keeping these in source code makes it easy for a bot or a casual observer to discover them. Embedding them in the code at runtime makes it much harder for someone to find. However we should still treat these values as "public" and ensure that no damage can be done with them.

This PR switches away from using expo environment variables for these "semi-sensitive" configuration values, and instead relies on the babel plugin `transform-inline-environment-variables`. The plugin is configured with an "allow-list" of environment variables to include, to avoid the small risk of leaking other sensitive environment variables in the app.

I think continuing to use expo environment variables (e.g. `EXPO_PUBLIC_`) for other public configuration values, such as feature flags, makes sense, since it is more explicit.


cimigree:
@gmaclennan One of the CI failures was fixed by this PR which might not have made it into this one since we merged it into develop: https://github.com/digidem/comapeo-mobile/pull/653

gmaclennan:
I've switched the base branch for this to `develop`, so we can merge there and then cherry-pick this onto `release/1.0.0`

gmaclennan:
I added secrets to this repo and modified the workflow to pass these as env variables

achou11 (approved):
Added a commit to update the docs. As a follow-up, would like to create a `.env.template` file to make this easier for onboarding and also provide in-file context for what some of these variables are actually used for

PR #661, opened 2024-09-05
feature: sync preview label

### LINKED ISSUE #413: Sync Preview Label

**Additional context**

cimigree:
closes #413 
### Description
- Alters the sync setting to include a default, which seems to be what is required by the issue 
- No label if the user has not selected a sync setting
- Creates a media label component that uses the sync setting and then renders the required text depending on props passed so that it can match the requirements of the component
- Renders the media label from the Observation and from the PhotoPreviewModal
- I did not add the feature flag because the default is to show nothing, but let me know if that is incorrect

### Screen shots
<image src="https://github.com/user-attachments/assets/cbb58822-8b7e-435e-a22e-1d92fd8e2a21" width="250" />
<image src="https://github.com/user-attachments/assets/0dc11ae2-75f4-45ff-b8be-e539858da12a" width="250" />
<image src="https://github.com/user-attachments/assets/d798d594-0663-4118-baa8-1feddef675b0" width="250" />
<image src="https://github.com/user-attachments/assets/6393ad86-e6bd-49e5-a254-cea2d2c924e8" width="250" />


achou11:
For whenever you revisit this PR:

I think the way the original issue that this addresses was described/framed gave it the perception of an unnecessarily large scope. This PR is introducing the settings screen related to media sync but I think that's not desirable for now because the backend doesn't support selective media sync (see https://github.com/digidem/comapeo-core/issues/434). If the design presented this screen as something that's currently in an unchangeable state, then it would be more okay.

I would suggest refocusing this PR such that it strictly implements just the feature about the media labels in the observation screen, as that can be implemented safely, even though for now you'll usually always see `Full size and previews available` because the backend syncs all media

PR #631, opened 2024-08-27
fix: address various issues related to invites

achou11:
Fixes #587 

Introduces a notable refactor of how invites are handled in the app in order to fix a few issues related to invites. Some UI requirements need us to have a stable reference to an invite (or its ID) after it's updated (either in the background or due to user intervention e.g. cancelled, rejected, error, accepted, etc).

Not overly excited about certain parts of the implementation (especially some of the effects-related code). Think there could be some clever react-query option to help with some of this, but it's not entirely clear to me. Open to exploring if there's an obvious approach.

Other issues that are fixed:

- subsequent relevant invites open the bottom sheet. before, the following sequence did not work as expected:

  1. A invites B
  2. A cancels invite or B declines invite
  3. A invites B again
  
  After (3), the invite bottom sheet did not open for B when it should.

- While device is going through leave project flow via an invite, if the invite is cancelled, update the UI to reflect that.

- fixes font size inheritance when using bold formatter via react-intl.
  - can extract this to a separate PR

Also introduces the following changes:

- upgrades `@gorhom/bottom-sheet` to latest
  - this was mostly as an attempt to fix some weird behavior with bottom sheet modals as I was working on this. Upgrading didn't help so it's not strictly necessary to include it in this PR, so happy to omit it or make a separate PR if preferred.

- fixes some of the copy for the different invite states (i used the designs [here](https://www.figma.com/design/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?node-id=1196-1914&t=zoLa1eFxP7AA5qjL-0) as basis)

- more thorough handling of other invite states (e.g. error)

A list of scenarios that I've tested with these changes (and generally work as expected or can be further polished with follow-ups):

### Inviting a device that does not have to leave an existing project

1. A creates project. B does not have existing project.
2. A invites B -> B sees invite bottom sheet appear
3. A cancels invite -> B sees invite bottom sheet update to show that invite was cancelled.
4. B dismisses invite sheet
5. A sends another invite -> B sees invite bottom sheet appear
6. B rejects invite -> B sees invite bottom sheet disappear -> A gets updated UI showing that invite was rejected
7. A sends another invite -> B sees invite bottom sheet appear
8. B accepts invite -> A sees confirmation that B accepted invite -> B sees invite bottom sheet in-progress acceptance UI -> B eventually sees success UI in bottom sheet

### Inviting a device that is already on an existing project

1. A creates project. B does not have existing project.
[...]

PR #665, opened 2024-09-05
fix: properly inherit style for translated bold text

achou11:
Extracted from #631 

Uses the built-in `Text` component from React Native for wrapping rich-formatted translation strings instead of our custom component. This prevents issues related to style specificity due to our custom component defining its own opinionated styling, when in practice we just want to make sure that these formatted translations are only applying the minimum styling (in this case, bold). The base `Text` component supports nesting and applies styling based on specificity related to nesting. For example, if we have a translation string like `'Hello <bold>World</bold>'` and use it as follows:

```tsx
<CustomText>{t(...)}</CustomText>
```

it will output:

```tsx
<CustomText>Hello <Text style={{fontWeight: 'bold'}}>World</Text></CustomText>
```

The styles from `CustomText` will be applied to `World`, but then any styles specified by the nested `Text` (in this case, `fontWeight`) will take precedence and be applied.

To test, try changing the style of the text containing the formatted translation. The style of the formatted translation should match what's specified, with the exception of it being bold.

Only example as of now is in the passcode settings: 

https://github.com/digidem/comapeo-mobile/blob/c5ee20c1a0af69c8c35f67359676038e4f66c704/src/frontend/screens/AppPasscode/PasscodeIntro.tsx#L60

### DIFF
--- a/src/frontend/contexts/IntlContext.tsx
+++ b/src/frontend/contexts/IntlContext.tsx
@@ -1,11 +1,10 @@
 import * as React from 'react';
 import {IntlProvider as ReactIntlProvider, CustomFormats} from 'react-intl';
-import {StyleSheet} from 'react-native';
+import {StyleSheet, Text} from 'react-native';
 
 import messages from '../../../translations/messages.json';
 import {usePersistedLocale} from '../hooks/persistedState/usePersistedLocale';
 import {TranslatedLocale} from '../lib/intl';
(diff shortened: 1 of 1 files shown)


cimigree (approved):
I was able to play around and apply new styles to the formatted text as you said. Thanks for the thorough explanation and testing instructions.

PR #676, opened 2024-09-09
fix: update message content when sharing observation

### LINKED ISSUE #516: Add Observation details to Observation Share
When sharing an observation through an external service (eg. Whatsapp, email...etc), the observation details should also be shared

## To do
- [ ] Add Observations details to share message

### Notion Issue
https://www.notion.so/digidem/f551401ee13f4652b3b3f23e52db9783?v=c58a7b022e084c2bae2c4379bf0c67e7&p=1d4e4229705d49b0a7822047c2752a46&pm=s

achou11:
Fixes #516 

Haven't extensively tested this on various kinds of categories (yet) but tried my best to port over what Mapeo Mobile does in terms of the final result

---

image

<br/>

image

<br/>

image

### DIFF
--- a/src/frontend/screens/Observation/Buttons.tsx
+++ b/src/frontend/screens/Observation/Buttons.tsx
@@ -69,12 +69,6 @@ const m = defineMessages({
     id: 'screens.Observation.shareMessageFooter',
     defaultMessage: 'Sent from CoMapeo',
   },
-  shareMessageFieldNoAnswer: {
-    id: 'screens.Observation.shareMessageFieldNoAnswer',
-    defaultMessage: 'No answer',
-    description:
-      'Placeholder text for fields on an observation which are not answered',
-  },
   fallbackCategoryName: {
     id: 'screens.Observation.fallbackCategoryName',
     defaultMessage: 'Observation',
@@ -160,12 +154,11 @@ export const ButtonFields = ({
           continue;
         }
 
-        const displayedValue =
-          (Array.isArray(value) ? value : [value])
-            .map(v => {
-              return getValueLabel(v, field).trim();
-            })
-            .join(', ') || t(m.shareMessageFieldNoAnswer);
+        const displayedValue = (Array.isArray(value) ? value : [value])
+          .map(v => {
+            return getValueLabel(v, field).trim();
+          })
+          .join(', ');
 
         completedFields.push({label: field.label, value: displayedValue});
       }
--- a/src/frontend/screens/Observation/Buttons.tsx
+++ b/src/frontend/screens/Observation/Buttons.tsx
@@ -178,7 +178,7 @@ export const ButtonFields = ({
         message: createObservationShareMessage({
           categoryName: preset ? preset.name : t(m.fallbackCategoryName),
           coordinateFormat: format,
-          fields: completedFields,
+          completedFields,
           footerText: t(m.shareMessageFooter),
           observation,
           timestamp: formatDate(observation.createdAt, {format: 'long'}),
@@ -256,7 +256,7 @@ const styles = StyleSheet.create({
 function createObservationShareMessage({
   categoryName,
   coordinateFormat,
-  fields,
+  completedFields,
   footerText,
   observation,
   timestamp,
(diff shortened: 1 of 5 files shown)


src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -139,19 +151,38 @@ export const ButtonFields = ({
...
+
+        const displayedValue =
+          (Array.isArray(value) ? value : [value])
+            .map(v => {
+              return getValueLabel(v, field).trim();
+            })
+            .join(', ') || t(m.shareMessageFieldNoAnswer);
+
+        completedFields.push({label: field.label, value: displayedValue});
+      }
```
achou11: logic inspired by Mapeo Mobile implementation

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/Observation/ObservationShare.js#L60-L65

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -139,19 +151,38 @@ export const ButtonFields = ({
...
+        if (value === undefined || value === null || value === '') {
+          continue;
+        }
+
+        const displayedValue =
+          (Array.isArray(value) ? value : [value])
+            .map(v => {
+              return getValueLabel(v, field).trim();
+            })
+            .join(', ') || t(m.shareMessageFieldNoAnswer);
```
ErikSin: should it be saying `{question} no answer` because thats not what i am getting:
image
image

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -221,3 +252,59 @@ const styles = StyleSheet.create({
...
+  coordinateFormat,
+  fields,
+  footerText,
+  observation,
+  timestamp,
+  titleText,
+}: {
+  categoryName?: string;
+  coordinateFormat: CoordinateFormat;
+  fields: Array<{label: string; value: string}>;
```
ErikSin: nit: dont call this fields, so we dont confuse it with actual fields. Maybe like fieldsShareText or something along those line

src/frontend/screens/Observation/Buttons.tsx
ErikSin: I see that your checking if the value is undefined twice, so i assume the second check is for typescript

ErikSin (approved):
I see that your checking twice for value being undefined. Im assuming the second time is just to satisfy typescript? Its also not sharing the fields if they are unanswered which is the same as mapeo

src/frontend/screens/Observation/Buttons.tsx
achou11: Based on what Mapeo does, unanswered fields are completely omitted from the share message body. That's as much as I know in terms of contextual knowledge though 😄

ErikSin: Sorry my comment was a little confusing. What you have is the desired behaviour. its just more of a clarification questions as to what `t(m.shareMessageFieldNoAnswer)` is for if we never actually share a field with no answer

achou11: ah i understand what your question is asking after a second read. I think the `|| t(...)` fallback is not necessary given the check a few lines up (which I think you're referring to?). think i'll remove that fallback if there aren't strong objections

src/frontend/screens/Observation/Buttons.tsx
achou11: renamed to `completedFields` via b28d696ca8a3c2a65b17161cbe01bb298708e418

src/frontend/screens/Observation/Buttons.tsx
achou11: removed the fallback string (and its translation) via 0d1c59793077bfb1f77ed267f3e0d9862beb2705

ErikSin: I think that makes sense!

PR #684, opened 2024-09-10
fix: update copy to replace Mapeo with CoMapeo

### LINKED ISSUE #683: Copy/ Update Mapeo to say CoMapeo
There's a few places in the app where Mapeo is still being mentioned 
Todo - update copy to dislay the name CoMapeo
details - https://www.notion.so/digidem/Update-Mapeo-to-say-CoMapeo-eaa3576c064743628210d6b4620659a2

achou11:
Fixes #683 

Updates all user-facing copy to use `CoMapeo` instead of `Mapeo`, as well as any translation-related variables in the app code.

~I added a separate commit that updates the files in the `messages/` directory. Didn't know if this causes issues with CrowdIn integration but figured it would avoid doing an extra step as a follow-up. Can easily revert if not desired.~ EDIT: reverted

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,3 +69,7 @@ jobs:
         run: npm run lint:types
       - name: Run unit tests
         run: npm test
+        env:
+          MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
+          COMAPEO_METRICS_URL: ${{ secrets.COMAPEO_METRICS_URL }}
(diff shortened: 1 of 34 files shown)


ErikSin (approved):
Lets revert the messages commit. Crowdin should be able to this and i think it would be less error prone if we didn't touch those files.

achou11:
> Lets revert the messages commit. Crowdin should be able to this and i think it would be less error prone if we didn't touch those files.

Done via https://github.com/digidem/comapeo-mobile/pull/684/commits/b5b285a9f252d7e9bf8a2711680d3583453fa9f5! Will wait for CI to pass before merging

PR #687, opened 2024-09-11
fix: handle category-related changes after importing a config more elegantly

### LINKED ISSUE #686: Custom Config imported / App Crash
Have not manage to reproduce by Jen. Ximena reproduced on 2 different devices but can no longer reproduce

Details and videos - https://www.notion.so/digidem/Custom-Config-imported-App-Crash-148f4de78c1040e1aaf00f868313e749

achou11:
Fixes #686 
Towards #611

More closely emulates the behavior of Mapeo Mobile when it comes to displaying an observation that has an unknown/unmatched preset/category.

Also properly invalidates query caches for other kinds of data that are affected by the introduction of a new project config.

Note that this doesn't fully fix the issue described in #611 , as I believe the cause is out of the scope of this PR. See https://github.com/digidem/comapeo-core/issues/821#issuecomment-2344231495. It does however make some progress towards fixing it. Before, I don't think those icons would ever load. Now, they will eventually load (helps to navigate in and out of screens for now).

---

https://github.com/user-attachments/assets/8c3487e9-fe90-4943-ba51-b9726ff5304b

### DIFF
--- a/src/frontend/hooks/server/presets.ts
+++ b/src/frontend/hooks/server/presets.ts
@@ -2,11 +2,8 @@ import {
   useSuspenseQuery,
   useMutation,
   useQueryClient,
-  useQuery,
-  keepPreviousData,
 } from '@tanstack/react-query';
 import {PresetValue} from '@mapeo/schema';
-import {IconSize} from '../../sharedTypes';
 
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 
@@ -34,28 +31,5 @@ export function usePresetsMutation() {
     onSuccess: () => {
       queryClient.invalidateQueries({queryKey: [PRESETS_KEY]});
     },
-  });
-}
-
-export function useGetPresetIcon(docId: string | undefined, size: IconSize) {
-  const {projectId, projectApi} = useActiveProject();
-
-  return useQuery<string, Error>({
-    queryKey: ['presetIconFromDocId', projectId, docId, size],
-    enabled: !!docId,
-    queryFn: async () => {
-      if (!docId) {
-        throw new Error('Preset icon reference not found');
-      }
-      return await projectApi.$icons.getIconUrl(docId, {
-        mimeType: 'image/png',
-        size: size,
-        pixelDensity: 3,
-      });
-    },
-    retry: 2,
-    retryDelay: 2000,
-    staleTime: 60000,
-    placeholderData: keepPreviousData,
   });
 }
--- a/src/frontend/hooks/useObservationWithPreset.ts
+++ b/src/frontend/hooks/useObservationWithPreset.ts
@@ -7,13 +7,5 @@ export const useObservationWithPreset = (observationId: string) => {
   const {data: presets} = usePresetsQuery();
   const preset = matchPreset(observation.tags, presets);
 
-  if (!observation) {
-    throw new Error('Observation does not exist');
-  }
-
-  if (!preset) {
-    throw new Error('Preset does not exist');
-  }
-
   return {observation, preset};
 };
--- a/src/frontend/hooks/useSelectFileAndImportConfig.ts
+++ b/src/frontend/hooks/useSelectFileAndImportConfig.ts
@@ -3,6 +3,8 @@ import {selectFile} from '../lib/selectFile';
 import * as FileSystem from 'expo-file-system';
 import {PROJECT_SETTINGS_KEY, useImportProjectConfig} from './server/projects';
 import {convertFileUriToPosixPath} from '../lib/file-system';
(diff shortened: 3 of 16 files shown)


src/frontend/hooks/useObservationWithPreset.ts
```diff
@@ -7,13 +7,5 @@ export const useObservationWithPreset = (observationId: string) => {
   const {data: presets} = usePresetsQuery();
   const preset = matchPreset(observation.tags, presets);
 
-  if (!observation) {
-    throw new Error('Observation does not exist');
-  }
-
-  if (!preset) {
-    throw new Error('Preset does not exist');
-  }
```
achou11, line 16: for posterity, this is what caused the error seen in the issue

src/frontend/hooks/server/presets.ts
```diff
@@ -36,26 +33,3 @@ export function usePresetsMutation() {
...
-      return await projectApi.$icons.getIconUrl(docId, {
-        mimeType: 'image/png',
-        size: size,
-        pixelDensity: 3,
-      });
-    },
-    retry: 2,
-    retryDelay: 2000,
-    staleTime: 60000,
-    placeholderData: keepPreviousData,
```
achou11, line 59: pretty sure specifying these is not helpful. opted to remove them entirely unless there's a notable issue that can be fixed by one of them being specified (query uses defaults for some of these if not specified)

src/frontend/hooks/useSelectFileAndImportConfig.ts
```diff
@@ -24,7 +27,12 @@ export function useSelectFileAndImportConfig() {
       }
     },
     onSuccess: () => {
-      return queryClient.invalidateQueries({queryKey: [PROJECT_SETTINGS_KEY]});
+      return Promise.all([
+        queryClient.invalidateQueries({queryKey: [FIELDS_KEY]}),
+        queryClient.invalidateQueries({queryKey: [ICONS_KEY]}),
+        queryClient.invalidateQueries({queryKey: [PROJECT_SETTINGS_KEY]}),
+        queryClient.invalidateQueries({queryKey: [PRESETS_KEY]}),
+      ]);
```
achou11, line 35: I _think_ it makes sense to have the mutation wait for all of these to resolve. the tradeoff is a slightly longer mutation time in the `active` state when importing a config

src/frontend/hooks/server/presets.ts
ErikSin: I think we should keep the retry to 2 or even 1 (as a stop gap to the stale fastify port problem)

ErikSin (approved):

src/frontend/hooks/server/presets.ts
achou11: React Query defaults to 3 based on the docs: https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

image

Can make it explicit if that's preferred though

ErikSin: I think we should make it less than 3 because we know what the problem is, and so failing faster would be better for the UI

achou11: unless I'm misunderstanding and you're suggesting that we should reduce the retry count (which I don't think is what you mean)

achou11: made it 3 explicitly via https://github.com/digidem/comapeo-mobile/pull/687/commits/0c36158e7df21c5a557143e53d53b9c95fe11079

ErikSin: That is what I am suggesting. Have tanstack query retry less times so that it throws an error faster. On each retry it remain in the pending state. That way the user doesn't see the loading spinner for so long and just shows the default icon when the fastify server is using a stale port

achou11: > I think we should make it less than 3 because we know what the problem is, and so failing faster would be better for the UI

Ah i see. I'll reduce to 2 in that case. i think it's still useful to have a couple of retries because the issue doesn't always happen (but that probably largely depends on the device and other factors). I'd be curious to see how prevalent it is in a release build vs development mode

achou11: > That way the user doesn't see the loading spinner for so long and just shows the default icon when the fastify server is using a stale port

Just to clarify, the issue not the stale port (at least in the context of this PR) but the server not being ready when resuming the app. Point understood though

achou11: reduced to 2 via 898be8912e40b14fdc9171231b30742d50f5f75a

PR #688, opened 2024-09-11
fix: always include timestamp & coords on observation position

EvanHahn:
We should treat these values as required.

### DIFF
--- a/src/frontend/screens/ManualGpsScreen/index.tsx
+++ b/src/frontend/screens/ManualGpsScreen/index.tsx
@@ -24,7 +24,7 @@ import {IconButton} from '../../sharedComponents/IconButton';
 import SaveCheck from '../../images/CheckMark.svg';
 import {Select} from '../../sharedComponents/Select';
 import {Text} from '../../sharedComponents/Text';
-import type {Position, CoordinateFormat} from '../../sharedTypes';
+import type {CoordinateFormat} from '../../sharedTypes';
 
 import {
   latitudeIsValid,
--- a/src/frontend/screens/ManualGpsScreen/index.tsx
+++ b/src/frontend/screens/ManualGpsScreen/index.tsx
@@ -104,28 +104,26 @@ export const ManualGpsScreen = ({
         );
       }
 
-      const position: Position = {mocked: false};
-
-      if (convertedData.coords) {
-        const {lon: longitude, lat: latitude} = convertedData.coords;
-
-        const lonIsValid =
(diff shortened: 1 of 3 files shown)


src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts
```diff
@@ -45,17 +45,14 @@ export function useMostAccurateLocationForObservation() {
         },
         debounceLocation()(location => {
           if (ignore) return;
-
-          const position: Position = {mocked: false};
```
achou11, line 49: the `Position` type is now an unused import in this file. mind removing it?

src/frontend/screens/Settings/CreateTestData.tsx
```diff
@@ -258,7 +258,7 @@ function useCreateFakeObservationsMutation() {
       const promises = [];
 
       for (let i = 0; i < count; i++) {
-        const fakeCoordinates = randomPosition({
+        const [lon = 0, lat = 0] = randomPosition({
```
achou11: nit: would probably prefer to just throw if `randomPosition()` doesn't return a tuple of 2 elements here. 0 is unfortunately a [significant value](https://en.wikipedia.org/wiki/Null_Island) in the context of geospatial coordinates.

the TS typing they have (`number[]`) is inaccurate, as the implementation always returns a tuple:

https://github.com/Turfjs/turf/blob/6889237200a088fa5539f92f361fc96135e09ac1/packages/turf-random/index.ts#L30-L44

achou11 (dismissed):
lgtm!

src/frontend/screens/ManualGpsScreen/index.tsx
achou11, line 1: same comment about removing the `Position` type import in this file

src/frontend/screens/Settings/CreateTestData.tsx
EvanHahn: Fixed in 9142abf01fbfad057de4c22a46e03fc49b4cb586. Also opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70521 to fix the upstream types.

src/frontend/screens/ManualGpsScreen/index.tsx
EvanHahn: Removed in 2128be9777a12e986f749e0993e3dcc74db8096f.

EvanHahn:
Closing, because I think we plan to do this in a pretty different way.

PR #702, opened 2024-09-12
Fix/manual gps when no location service

### LINKED ISSUE #685: Location not shared/ Cant save manual GPS
**Problem:**

- Device not sharing their location with CoMapeo
- User tries to create an observation, CM reminds user to share location
- User declines
- User tries to create an observation
    - in GPS - display*s Searching*
- User can save the observation without the prompt of saving location manually

https://www.notion.so/digidem/Location-not-shared-Cant-save-manual-GPS-962afb0da23441ab9998ba8b2fab0ca7

ErikSin:
fixes: #685 

image

### DIFF
--- a/src/frontend/screens/ObservationCreate/index.tsx
+++ b/src/frontend/screens/ObservationCreate/index.tsx
@@ -16,7 +16,6 @@ import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 import {HeaderLeft} from './HeaderLeft';
 import {ActionsRow} from '../../sharedComponents/ActionRow';
 import {Alert, type AlertButton} from 'react-native';
-import noop from '../../lib/noop';
 
 const m = defineMessages({
   observation: {
@@ -78,7 +77,7 @@ const m = defineMessages({
   },
 });
 
-const MINIMUM_ACCURACY = 10;
+const MAXIMUM_ACCURACY = 10;
 
 export const ObservationCreate = ({
   navigation,
@@ -230,13 +229,13 @@ export const ObservationCreate = ({
     ];
 
     if (!value) {
-      noop();
       return;
     }
 
     // If the user has already inputted a manual location, do not check if location is accurate
     if (value.metadata?.manualLocation) {
       createObservation();
+      return;
     }
 
     const accuracy = value.metadata?.position?.coords?.accuracy;
@@ -254,7 +253,7 @@ export const ObservationCreate = ({
     if (
       accuracy &&
       typeof accuracy === 'number' &&
-      accuracy < MINIMUM_ACCURACY
+      accuracy >= MAXIMUM_ACCURACY
     ) {
       Alert.alert(
         formatMessage(m.weakGpsTitle),
--- a/src/frontend/screens/ObservationCreate/index.tsx
+++ b/src/frontend/screens/ObservationCreate/index.tsx
@@ -15,6 +15,7 @@ import {ErrorBottomSheet} from '../../sharedComponents/ErrorBottomSheet';
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 import {HeaderLeft} from './HeaderLeft';
 import {ActionsRow} from '../../sharedComponents/ActionRow';
+import {Alert, AlertButton} from 'react-native';
 
 const m = defineMessages({
   observation: {
@@ -36,7 +37,47 @@ const m = defineMessages({
     defaultMessage: 'What is happening here?',
     description: 'Placeholder for description/notes field',
   },
+  noGpsTitle: {
+    id: 'screens.ObservationCreate.noGpsTitle',
+    defaultMessage: 'No GPS signal',
+    description: 'Title of dialog when trying to save with no GPS coords',
+  },
+  noGpsDesc: {
+    id: 'screens.ObservationCreate.noGpsDesc',
+    defaultMessage:
+      'This observation does not have a location. You can continue waiting for a GPS signal, save the observation without a location, or enter coordinates manually',
+    description: 'Description in dialog when trying to save with no GPS coords',
+  },
+  weakGpsTitle: {
+    id: 'screens.ObservationCreate.weakGpsTitle',
+    defaultMessage: 'Weak GPS signal',
+    description: 'Title of dialog when trying to save with low GPS accuracy.',
+  },
+  weakGpsDesc: {
+    id: 'screens.ObservationCreate.weakGpsDesc',
+    defaultMessage:
+      'GPS accuracy is low. You can continue waiting for better accuracy, save the observation with low accuracy, or enter coordinates manually',
+    description:
+      'Description in dialog when trying to save with low GPS accuracy.',
+  },
+  saveAnyway: {
+    id: 'screens.ObservationCreate.saveAnyway',
+    defaultMessage: 'Save',
+    description: 'Button to save regardless of GPS state',
+  },
+  manualEntry: {
+    id: 'screens.ObservationCreate.manualEntry',
+    defaultMessage: 'Manual Coords',
+    description: 'Button to manually enter GPS coordinates',
+  },
+  keepWaiting: {
+    id: 'screens.ObservationCreate.keepWaiting',
+    defaultMessage: 'Continue waiting',
+    description: 'Button to cancel save and continue waiting for GPS',
+  },
 });
+
+const MINIMUM_ACCURACY = 10;
 
 export const ObservationCreate = ({
   navigation,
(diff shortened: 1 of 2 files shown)


src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -169,12 +210,56 @@ export const ObservationCreate = ({
...
+        onPress: () => navigation.navigate('ManualGpsScreen'),
+        style: 'cancel',
+      },
+      {
+        text: formatMessage(m.keepWaiting),
+        onPress: () => {},
+      },
+    ];
+
+    if (!value) throw new Error('no observation saved in persisted state ');
```
achou11: Probably shouldn't throw an error here in the case that this does actually happen, as it will cause the app to crash. Would suggest a no-op via early return or gracefully surfacing an error of some sort for now.

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -224,6 +309,10 @@ export const ObservationCreate = ({
   );
 };
 
+function isGpsAccurate(accuracy?: number): boolean {
+  return typeof accuracy === 'number' && accuracy < MINIMUM_ACCURACY;
+}
```
achou11: nit: don't think it's necessary to have a separate function for this right now

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -169,12 +210,56 @@ export const ObservationCreate = ({
...
+
+    // If we don't have accuracy, allow save anyway (this is a remnant from mapeo: https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L125C3-L125C50)
+    if (accuracy && !isGpsAccurate(accuracy)) {
+      Alert.alert(
+        formatMessage(m.weakGpsTitle),
+        formatMessage(m.weakGpsDesc),
+        confirmationOptions,
+      );
+      return;
+    }
```
achou11: if the observation location is set manually, we don't want to surface these. should add a check for `value.metadata?.manualLocation` being not true in both of these.

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -15,6 +15,7 @@ import {ErrorBottomSheet} from '../../sharedComponents/ErrorBottomSheet';
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 import {HeaderLeft} from './HeaderLeft';
 import {ActionsRow} from '../../sharedComponents/ActionRow';
+import {Alert, AlertButton} from 'react-native';
```
achou11: nit:

```suggestion
import {Alert, type AlertButton} from 'react-native';
```

achou11 (changes requested):

src/frontend/screens/ObservationCreate/index.tsx
achou11: ~Ah with the fix applied in 5aab158e01d7a7c9f1fad97c5c22da0b37c80e78 i think that the situation I wrote above doesn't happen anymore~.  EDIT: nvm, i wasn't able to reproduce the above situation even without that commit, so maybe I'm just misreading code at this point 😅  

Still think it might be worth having that check anyways for clarity though. The `liveLocation` variable is a little confusing to me at a glance otherwise

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/d3519214756fc2fbe99980e4ad1e4ca8fb6024f9

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/d3519214756fc2fbe99980e4ad1e4ca8fb6024f9

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/d3519214756fc2fbe99980e4ad1e4ca8fb6024f9

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/d3519214756fc2fbe99980e4ad1e4ca8fb6024f9

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -169,12 +211,68 @@ export const ObservationCreate = ({
...
+        style: 'cancel',
+      },
+      {
+        text: formatMessage(m.keepWaiting),
+        onPress: () => {},
+      },
+    ];
+
+    if (!value) {
+      noop();
```
achou11: This is not necessary

```suggestion
      noop();
```

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -169,12 +211,68 @@ export const ObservationCreate = ({
...
+      return;
+    }
+
+    // If the user has already inputted a manual location, do not check if location is accurate
+    if (value.metadata?.manualLocation) {
+      createObservation();
+    }
```
achou11: Need to add a return if you're doing this

```suggestion
    if (value.metadata?.manualLocation) {
      createObservation();
      return;
    }
```

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -15,6 +15,8 @@ import {ErrorBottomSheet} from '../../sharedComponents/ErrorBottomSheet';
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 import {HeaderLeft} from './HeaderLeft';
 import {ActionsRow} from '../../sharedComponents/ActionRow';
+import {Alert, type AlertButton} from 'react-native';
+import noop from '../../lib/noop';
```
achou11: See other comment where this is used.

```suggestion
```

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -169,12 +211,68 @@ export const ObservationCreate = ({
...
+        confirmationOptions,
+      );
+      return;
+    }
+
+    // If we don't have accuracy, allow save anyway (this is a remnant from mapeo: https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L125C3-L125C50)
+    if (
+      accuracy &&
+      typeof accuracy === 'number' &&
+      accuracy < MINIMUM_ACCURACY
```
achou11: This should be checking if the accuracy is too large (since accuracy is the numeric range of distance in this case)

```suggestion
      accuracy >= MINIMUM_ACCURACY
```

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -36,8 +38,48 @@ const m = defineMessages({
...
+    description: 'Button to manually enter GPS coordinates',
+  },
+  keepWaiting: {
+    id: 'screens.ObservationCreate.keepWaiting',
+    defaultMessage: 'Continue waiting',
+    description: 'Button to cancel save and continue waiting for GPS',
+  },
 });
 
+const MINIMUM_ACCURACY = 10;
```
achou11: think this should be renamed to `MAXIMUM_ACCURACY` or something similar? See other comment where this is used, but this should be representing the maximum value we allow without asking the user about potentially inserting a manual one.

achou11 (changes requested):

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: Im a little confused here. I think you are right, but historically we have used `accuracy < MINIMUM_ACCURACY` in comapeo mobile. Why has it worked before?

ErikSin: see: https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L126

achou11: Mapeo Mobile performs a different check. It's checking to see that the accuracy is valid and good enough (i.e. `accuracy < 10`) first and if it is, goes straight ahead to saving the draft.

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L85-L95

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ObservationEdit/SaveButton.js#L116-L127

Your PR is doing checking for cases where it's invalid or insufficient first, in order to potentially open the dialogs about manually setting a coordinate.

ErikSin: RIght....I should probably just read the code 😬

achou11: Yeah i think the `MINIMUM_ACCURACY` naming is a bit confusing since accuracy is one of those inverted quantities (in this case at least). Ideally it would be more descriptive in this context 😅

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/9ed6b36b9c35fd46ccb358820f0d0980cfc3aab7

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/9ed6b36b9c35fd46ccb358820f0d0980cfc3aab7

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/9ed6b36b9c35fd46ccb358820f0d0980cfc3aab7

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/9ed6b36b9c35fd46ccb358820f0d0980cfc3aab7

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/702/commits/9ed6b36b9c35fd46ccb358820f0d0980cfc3aab7

ErikSin:
Ok! i have tested it out on my device and it is doing what it should be doing: Asking for GPS when accuracy is greater than 10 and saving otherwise

achou11 (approved):
Might want to update this PR with develop first since there have been some changes introduced (although no merge conflicts it seems)

achou11:
Also should update the PR title before merging!

PR #705, opened 2024-09-16
fix: do not set observation metadata when using manual gps coordinate

### LINKED ISSUE #689: Saving an observation with a manually entered coordinate should not update the observation's `metadata` field
There's a (presumably unintentional) difference in how observations are persisted between Mapeo Mobile and CoMapeo Mobile in the case of manually entering a coordinate for the observation.

In Mapeo Mobile, we update the `Observation.lat` and `Observation.lon` fields of the draft observation:

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/ManualGpsScreen/index.tsx#L84-L91

https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/hooks/useDraftObservation.ts#L198-L209

In CoMapeo Mobile, we update `Observation.lat` and `Observation.lon`, but we also update `Observation.metadata.position`:

https://github.com/digidem/comapeo-mobile/blob/b59ac86d8a383246e24afd52b3b19482cb600b09/src/frontend/screens/ManualGpsScreen/index.tsx#L127-L130

https://github.com/digidem/comapeo-mobile/blob/b59ac86d8a383246e24afd52b3b19482cb600b09/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts#L68-L87

Updating `Observation.metadata.position` is not desired in this case because that is ideally reserved for information provided by something like a device sensor (e.g. information provided by a native library like `expo-location`).

I'm not immediately sure of user-facing issues that this could cause.

achou11:
Fixes #689 

Paired on this with @EvanHahn

### DIFF
--- a/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
+++ b/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
@@ -35,10 +35,22 @@ export type DraftObservationSlice = {
     newDraft: () => void;
     deletePhoto: (uri: string) => void;
     existingObservationToDraft: (observation: Observation) => void;
-    updateObservationPosition: (props: {
-      position: Position | undefined;
-      manualLocation: boolean;
-    }) => void;
+    updateObservationPosition: (
+      props:
+        | {
+            manualLocation: false;
+            position: Position | undefined;
+          }
+        | {
+            manualLocation: true;
+            position: {
+              coords: {
+                latitude: number;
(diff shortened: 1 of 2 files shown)


EvanHahn (approved):
LGTM!

PR #716, opened 2024-09-16
chore: update dependencies

EvanHahn:
This updates all of our (Co)Mapeo dependencies to their latest versions.  This should be the last time we do this before the 1.0 releases of those modules.

_I did not test this manually._


ErikSin:
@cimigree can you e2e this?

cimigree:
@ErikSin, 

Everything passed except the help text for entering details on an observation were not visible. I don't know if that is a requirement change or a change in the default config or what...
<image src="https://github.com/user-attachments/assets/c3747494-83cb-430f-b21b-46bc876d635e" width="250" />
It used to say "Wood" in the blank space...

cimigree (approved):
@ErikSin, 

Everything passed except the help text for entering details on an observation were not visible. I don't know if that is a requirement change or a change in the default config or what...
<image src="https://github.com/user-attachments/assets/c3747494-83cb-430f-b21b-46bc876d635e" width="250" />
It used to say "Wood" in the blank space...

PR #732, opened 2024-09-17
chore: update CoMapeo dependencies to final release versions

EvanHahn:
This updates CoMapeo dependencies to their final release versions.  Notably, several names changed. For example, `@mapeo/core` is now `@comapeo/core`.

There are no other changes in this repo other than renames and version bumps.

I did a quick manual test on a real Android phone and I was successfully able to make an observation.

PR #738, opened 2024-09-17
chore: update CoMapeo dependencies to final release versions (#732)

ErikSin:
This updates CoMapeo dependencies to their final release versions. Notably, several names changed. For example, `@mapeo/core` is now `@comapeo/core`.

There are no other changes other than renames and version bumps.


ErikSin:
ci is failing but the next cherry pick will fix this. Merging as is as i am trying to keep the order of commits as similar as possible with develop

PR #742, opened 2024-09-17
test: load env variables so `npm test` works locally

EvanHahn:
We [set environment variables in CI][0] so that `npm test` works, but it doesn't work locally because Jest isn't loading them.

This loads them when running Jest using `dotenv-cli`.

[0]: https://github.com/digidem/comapeo-mobile/blob/9b6cf8f71a0d960174c7e8cd5ce1b6bb8f1dd5a7/.github/workflows/ci.yml#L72-L75


gmaclennan (commented):
I feel like this doesn't belong in here - it should be part of the jest test runner or the runner command to load env vars from a file. I feel like this could come back and bite us in the future, because expo cli is loading .env files too and we could get a surprise.

EvanHahn:
The problem, I think, is that Jest loads this config and immediately crashes because the environment variables are required. As an alternative, we could disable these checks in test. Would that work?

gmaclennan:
That works. An alternative approach would be to use [`dotenv-cli`](https://www.npmjs.com/package/dotenv-cli) and change the `test` script in `package.json` to `dotenv jest`. This works for me locally.

Pros: This allows for Jest tests to be added which depend on env variables e.g. test renders of the map
Cons: Anyone running the tests will need to define the required variables in their local `.env`.

Given how we use the environment variables currently, and our current Jest tests, your solution is fine for now, we just need to remember it when we maybe add additional env variables and add Jest tests that try to use them.

EvanHahn:
Changed to use `dotenv-cli`. Merging.

PR #744, opened 2024-09-19
fix: update to `@comapeo/core@1.0.1`

### LINKED ISSUE #671: Project Invite/ Invitee accepts invite but doesnt see its in the project
Update: As of Sept 17, using CoMapeo1.0.0-RC.3, the issue is consistently happening whenever a device is being invited.

 ***

See video and details : https://www.notion.so/digidem/Project-invite-invited-device-didn-t-complete-process-after-accepting-a-k-a-ghost-peasy-2e35a3cb71e24cf2b4a55cca50d34135

Device A invites Device B to project

Device B accepts

Device A see the confirmation that B has accepted

Device B invitation is still open, and the button to “Accept Invite” can be clicked but nothing happens

Device B can click “Decline” and sees they are not on a project

Device A has a ghost device on their project

Important to note : 
This has happened to Jen once
Ximena has not seen this happened
We cannot reproduce

EvanHahn:
_I haven't yet tested this manually._

Fixes #671.


gmaclennan:
The hypercore transitive dep is still wrong unfortunately. Just away from my desk but I can fix when I'm back. I think you need to uninstall Mapeo core then re-install to force the package-lock to dedupe to the hypercore 10.17 version.

gmaclennan (approved):
My mistake, I was looking at the wrong `package-lock.json` (the one for the frontend, rather than the backend). The file `src/backend/package-lock-json` only has the correct version of hypercore, `v10.17.0`. 

image

We should at some point address how to use types from the backend in the front-end without bringing all the dependencies with them.

Side-note: this is a case-in-point for https://github.com/digidem/comapeo-core/issues/841. Updating `@mapeo/core` in the frontend resulted in the pinned version of `corestore`, but still kept two versions of `hypercore`. In the backend we were ok for some reason, but could have had this issue. Npm lockfiles (and dependencies in any programming language) are a headache!

image

(frontend `npm list hypercore` showing above, can be ignored because I don't think we bundle it anyway)

gmaclennan:
Let's get this tested manually before merging. I can't test until about 10am EDT.

EvanHahn:
Tested this manually and it looks good.

Merging.

PR #747, opened 2024-09-19
fix: unpublish dns-sd service

gmaclennan:
We were not unpublishing the zeroconf service when the app went into the background, but we were re-publishing the service when the app came into the foreground (the same with wifi coming off and on).

When zeroconf attempts to publish a service with the same name, it appends ` (${n})` to the published service name.

The result of this was that the app would leave multiple published dns-sd services e.g.

```
8130fb3b21ac48b6
8130fb3b21ac48b6 (1)
8130fb3b21ac48b6 (2)
```

Only the most recent would have the correct port - the older services would have the port from previous app opens.

Leaving these services published does not break anything per se, but the consequences are:

1. Increased mdns network traffic, because other mdns devices share the service information.
2. Slower connections between CoMapeo devices and possible performance overhead, because CoMapeo will attempt to connect to _every_ service published.

This PR correctly unpublishes the service when the app goes into the background or the wifi turns off. Unfortunately the react-native-zeroconf library has not implemented an event for publishing or unpublishing failing, so we implement a timeout, but don't throw an error, because just trying again is a reasonable solution. We do report the error to Sentry though so that we can track these failures.

I have tested this by opening and closing the app while viewing [this DNS-SD Browser](https://apps.apple.com/gb/app/discovery-dns-sd-browser/id1381004916?mt=12) for MacOS, but you could also test via `avahi-browse -a -v --ignore-local --resolve`.

I do not have two devices at the moment to test on real devices, this would benefit from that.

I don't think this is necessary for MVP, but could be worth releasing in a hotfix if there are discovery issues with large numbers of devices on the same network.

### DIFF
--- a/src/frontend/contexts/LocalDiscoveryContext.tsx
+++ b/src/frontend/contexts/LocalDiscoveryContext.tsx
@@ -340,7 +340,7 @@ function publishZeroconf(
     };
 
     zeroconf.on(
-      // @ts-expect-error - the types are wrong, this is the correct event name
+      // @ts-expect-error We can remove this when <https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70693> is merged.
       'published',
       onPublish,
     );
--- a/src/frontend/contexts/LocalDiscoveryContext.tsx
+++ b/src/frontend/contexts/LocalDiscoveryContext.tsx
@@ -373,9 +373,9 @@ function stopZeroconf(zeroconf: Zeroconf): Promise<void> {
 
 function unpublishZeroconf(
   zeroconf: Zeroconf,
-  publishedNames: Set<string>,
+  publishedNamesToBeMutated: Set<string>,
 ): Promise<void> {
-  if (publishedNames.size === 0) return Promise.resolve();
+  if (publishedNamesToBeMutated.size === 0) return Promise.resolve();
   return new Promise((resolve, reject) => {
     const timeoutId = setTimeout(() => {
       cleanup();
@@ -387,14 +387,14 @@ function unpublishZeroconf(
       zeroconf.off('remove', onRemove);
     };
     const onRemove = (name: string) => {
-      publishedNames.delete(name);
-      if (publishedNames.size === 0) {
+      publishedNamesToBeMutated.delete(name);
+      if (publishedNamesToBeMutated.size === 0) {
         cleanup();
         resolve();
       }
     };
     zeroconf.on('remove', onRemove);
-    for (const name of publishedNames) {
+    for (const name of publishedNamesToBeMutated) {
       zeroconf.unpublishService(name);
     }
   });
--- a/src/frontend/contexts/LocalDiscoveryContext.tsx
+++ b/src/frontend/contexts/LocalDiscoveryContext.tsx
@@ -111,17 +111,17 @@ export function createLocalDiscoveryController(mapeoApi: MapeoClientApi) {
       // publishedName could be different from the name we requested, if there
       // was a conflict on the network (the conflict could come from the same
       // name still being registered on the network and not yet cleaned up)
-      const publishedName = await publishZeroconf(zeroconf, {name, port}).catch(
-        e => {
-          // Publishing could fail (timeout), but we don't want to throw the
-          // state machine start(), because that would leave the state machine
-          // in an "error" state and stop other things from working. By silently
-          // failing (with the report to Sentry), we are able to try again next
-          // time.
-          Sentry.captureException(e);
-        },
-      );
-      if (publishedName) publishedNames.add(publishedName);
+      try {
+        const publishedName = await publishZeroconf(zeroconf, {name, port});
+        publishedNames.add(publishedName);
+      } catch (e) {
+        // Publishing could fail (timeout), but we don't want to throw the
+        // state machine start(), because that would leave the state machine
+        // in an "error" state and stop other things from working. By silently
+        // failing (with the report to Sentry), we are able to try again next
+        // time.
+        Sentry.captureException(e);
+      }
       await startZeroconfPromise;
     },
     async stop() {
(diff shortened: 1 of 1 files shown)


gmaclennan:
This is possibly creating too much noise with Sentry because of reporting timeouts when stopping service adverstisement. Could investigate when that happens (probably when wifi is turned off) and be more selective in what we report to sentry.

src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -91,33 +98,48 @@ export function createLocalDiscoveryController(mapeoApi: MapeoClientApi) {
...
+        e => {
+          // Publishing could fail (timeout), but we don't want to throw the
+          // state machine start(), because that would leave the state machine
+          // in an "error" state and stop other things from working. By silently
+          // failing (with the report to Sentry), we are able to try again next
+          // time.
+          Sentry.captureException(e);
+        },
       );
+      if (publishedName) publishedNames.add(publishedName);
```
EvanHahn: I found this last conditional a little confusing, because it implies that `publishZeroconf` can return a falsy value. I think using a normal `try`/`catch` feels a bit clearer:

```suggestion
      try {
        const publishedName = await publishZeroconf(zeroconf, {name, port});
        publishedNames.add(publishedName);
      } catch (e) {
        // Publishing could fail (timeout), but we don't want to throw the
        // state machine start(), because that would leave the state machine
        // in an "error" state and stop other things from working. By silently
        // failing (with the report to Sentry), we are able to try again next
        // time.
        Sentry.captureException(e);
      }
```

src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -319,6 +375,35 @@ function stopZeroconf(zeroconf: Zeroconf): Promise<void> {
...
+      cleanup();
+      reject(new Error('Timed out unpublishing zeroconf service'));
+    }, ZEROCONF_UNPUBLISH_TIMEOUT_MS);
+
+    const cleanup = () => {
+      clearTimeout(timeoutId);
+      zeroconf.off('remove', onRemove);
+    };
+    const onRemove = (name: string) => {
+      publishedNames.delete(name);
```
EvanHahn: I was pretty surprised to see this mutate `publishedNames`. We can avoid that by returning a result type:

```javascript
function unpublishZeroconf(
  zeroconf: Zeroconf,
  publishedNames: Readonly<Set<string>>,
): Promise<{
  publishedNames: Set<string>;
  err?: Error;
}> {
  if (publishedNames.size === 0) return Promise.resolve({publishedNames});
  publishedNames = new Set(publishedNames);
  return new Promise(resolve => {
    const timeoutId = setTimeout(() => {
      cleanup();
      resolve({
        publishedNames,
        err: new Error('Timed out unpublishing zeroconf service'),
      });
    }, ZEROCONF_UNPUBLISH_TIMEOUT_MS);

    const cleanup = () => {
      clearTimeout(timeoutId);
      zeroconf.off('remove', onRemove);
    };
    const onRemove = (name: string) => {
      publishedNames.delete(name);
      if (publishedNames.size === 0) {
        cleanup();
        resolve({publishedNames});
      }
    };
    zeroconf.on('remove', onRemove);
    for (const name of publishedNames) {
      zeroconf.unpublishService(name);
    }
  });
}
```

We can then use it above like this:

```javascript
// ...
(async () => {
  const unpublishResult = await unpublishZeroconf(
    zeroconf,
    publishedNames,
  );
  ({publishedNames} = unpublishResult);
  if (unpublishResult.err) {
    // See above for why we silently fail here
    Sentry.captureException(unpublishResult.err);
  }
})(),
// ...
```

More verbose, but I think it avoids a pretty unusual mutation of the argument.

(You could argue that we're mutating `zeroconf` too, but I feel like that's a bit different; that's passing a handle to some external interface, which the set is not.)

src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -298,6 +320,40 @@ function startZeroconf(zeroconf: Zeroconf): Promise<void> {
...
+      clearTimeout(timeoutId);
+      zeroconf.off('published', onPublish);
+    };
+    const onPublish = ({name: publishedName}: ZeroconfService) => {
+      cleanup();
+      resolve(publishedName);
+    };
+
+    zeroconf.on(
+      // @ts-expect-error - the types are wrong, this is the correct event name
```
EvanHahn:
```suggestion
      // @ts-expect-error We can remove this when <https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70693> is merged.
```

EvanHahn (changes requested):
LGTM overall, but a few nitpicks.

src/frontend/contexts/LocalDiscoveryContext.tsx
gmaclennan: Agree to this suggestion. I've used this pattern (adding catch() and testing return value) a couple of times and it's probably not a great pattern!

src/frontend/contexts/LocalDiscoveryContext.tsx
gmaclennan: Not such a fan of this suggestion. `publishedNames` is really tracking state that should be internal to zeroconf. It feels weird that it gets "out of sync" if it's not mutated - unpublish is async, so changes to this code to read `publishedNames` before unpublish resolves would not get the correct answer. We don't currently have code that would be affected by this, and we might not, but it seems like we're opening ourselves to a race condition. I think because `publishedNames` is the state of "which names are published", then it should be mutated as names are published / unpublished. If there is an error during the unpublishing of one of the names, we don't want the unpublishign of the other names to be ignored / missing from state.
TL;DR I think this change will introduce bugs, and IIFEs can also be hard to read.

EvanHahn: Renamed the arg in 5cb2e2fc808da2d28e4258967fd1d583f8f38f3a.

EvanHahn (approved):

EvanHahn:
> This is possibly creating too much noise with Sentry because of reporting timeouts when stopping service adverstisement. Could investigate when that happens (probably when wifi is turned off) and be more selective in what we report to sentry.

I didn't address this comment because this PR was blocking a release, but I still think it's worth considering.

I'll bring it up with the team soon.

PR #749, opened 2024-09-23
chore: move query invalidation of config-related queries to useImportProjectConfig()

achou11:
Moves the query invalidations that are relevant to a change in the config (e.g. importing a new one) to `useImportProjectConfig()`, which is the more relevant mutation and is used internally by `useSelectFileAndImportConfig()`. This should not have any notable user-facing impact, as it's only callsite was `useSelectFileAndImportConfig()`.

This will avoid issues where we use `useImportProjectConfig()` but forget to do important query invalidations.


cimigree (approved):
Makes sense and worked fine when I tested it out.

PR #756, opened 2024-09-25
fix: Only allow coordinator to update config

### LINKED ISSUE #755: Only Coordinator can import Config
Coordinators, not participants, should be able to import a config

## To Do
- [ ] Disable participants from importing a config

ErikSin:
fixes #755 

### Participant Screen:
image

### Coordinator Screen:
image

### DIFF
--- a/src/frontend/hooks/server/projects.ts
+++ b/src/frontend/hooks/server/projects.ts
@@ -13,6 +13,7 @@ export const PROJECT_KEY = 'project';
 export const PROJECT_MEMBERS_KEY = 'project_members';
 export const ORIGINAL_VERSION_ID_TO_DEVICE_ID_KEY =
   'originalVersionIdToDeviceId';
+export const OWN_ROLE = 'my_role';
 
 export function useProject(projectId?: string) {
   const api = useApi();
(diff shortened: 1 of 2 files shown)


achou11 (dismissed):
I suggest creating a query that wraps `projectApi.$getOwnRole()`, which should make this PR more succinct.

also, would rename the PR to classify this as a fix instead of a chore

(also sorry for jumping in, didn't see the assigned reviewer until after reading through the code)

cimigree:
I will wait until Andrew's requests are addressed before reviewing if that is ok, @ErikSin

ErikSin:
@achou11 i have added your suggestions! Thanks

src/frontend/hooks/server/projects.ts
```diff
@@ -147,3 +148,14 @@ export function useImportProjectConfig() {
     },
   });
 }
+
+export function useGetOwnRole() {
+  const {projectId, projectApi} = useActiveProject();
+
+  return useQuery({
+    queryKey: [OWN_ROLE, projectId],
```
ErikSin: `OWN_ROLE` is not fully necessary at this point, but in the future there could be the possibility of your role changing, hence why I created this query key

cimigree: Could you use a more descriptive key besides OWN_ROLE? USER_ROLE_KEY or something?

cimigree (approved):
One minor suggestion. I think it's good to have this query in place. I think we will use it again soon for determining editing capabilities on observations.

I tested it out on two devices running in development and it worked great.

achou11:
@ErikSin another reminder to rename the PR to classify this as a fix, not a chore

src/frontend/hooks/server/projects.ts
ErikSin: yeah, i think it would be good to distinguish between a user role query vs this devices role query. Maybe `THIS_USERS_ROLE_KEY`?

PR #758, opened 2024-09-25
feat: implement custom map file support

### LINKED ISSUE #622: Menu Path / Settings
**Description**
Goal : Improved workflow for importing a background map into CoMapeo, by allowing the user to pick the background map file from a file-picker in the app, rather than requiring the user to place the background map file in a specific location
An interim solution until the Map Manager feature is brought to CoMapeo
 
Relates to https://github.com/digidem/comapeo-mobile/issues/624 https://github.com/digidem/comapeo-mobile/issues/623

**Todo**
Screen Shot 2024-08-26 at 12 33 57 PM

- [ ] Add **Map Management** button in **App Setings**
- [ ] Add **Background Maps** button  in **Map Management**

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2179-1778&t=pHGODPSppOLjQUPF-0

### LINKED ISSUE #623: Map Upload / Success
**Description**
Goal : Improved workflow for importing a background map into CoMapeo, by allowing the user to pick the background map file from a file-picker in the app, rather than requiring the user to place the background map file in a specific location
An interim solution until the Map Manager feature is brought to CoMapeo
To note : 
- Coordinator and Particpants can uploap or remove a map
- Accepted file types are .mbttiles, .mappackage

Relates to https://github.com/digidem/comapeo-mobile/issues/622 https://github.com/digidem/comapeo-mobile/issues/624
**Todo**

## Screen 1
Screen Shot 2024-08-26 at 12 42 49 PM

- [ ] Display screen as designed
- [ ] Use copy _About Offline Maps_
- [ ] _Adding an offline map will enable you to see a map when you are offline. You can only add one (1) offline map to your CoMapeo project._ 
- [ ] _Only your device will have access to the offline map._
- [ ] Add a  **Choose file**  button
- [ ] Use copy under button _Accepted file types are .mbttiles, .mappackage_

## Screen 2
Screen Shot 2024-08-26 at 12 43 02 PM

- [ ] If Choose file button clicked, open the devices file manager
- [ ] It will display the accepted file types

## Screen 3
Screen Shot 2024-08-27 at 6 55 33 PM

- [ ] Display screen as designed
- [ ] Use copy _Offline Map Added_
- [ ] _You will now the available offline areas on the map._
- [ ] Add a  **Close**  button

## Screen 4
Screen Shot 2024-08-26 at 12 43 42 PM

- [ ] Update Screen 1
- [ ] Display _Map Name_ an _Date Added_
- [ ] Add box and display Map Name and Server Address, under server name
- [ ] Add date added, under date added
- [ ] Add **Remove Map** button
- [ ] Add **Replace Offline Map** button
- [ ] When **Replace Offline Map** button is clicked, go to Screen 1

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2179-1778&t=pHGODPSppOLjQUPF-0

### LINKED ISSUE #624: Replace or Remove Map
**Description**
Goal : Improved workflow for importing a background map into CoMapeo, by allowing the user to pick the background map file from a file-picker in the app, rather than requiring the user to place the background map file in a specific location
An interim solution until the Map Manager feature is brought to CoMapeo
To note : 
- Coordinator and Particpants can upload, replace or remove a map
- Accepted file types are .mbttiles, .mappackage

Relates to https://github.com/digidem/comapeo-mobile/issues/623 https://github.com/digidem/comapeo-mobile/issues/622

**Todo**
Screen Shot 2024-08-26 at 12 45 35 PM

- [ ] Display bottom sheet as designed
- [ ] Use copy _Delete Offline Map?_
- [ ] _This will delete the map and its offline areas. No collected observation data will be deleted.  This cannot be undone._ 
- [ ] Add **Delete Map** button
- [ ] If clicked, go to Screen 1 https://github.com/digidem/comapeo-mobile/issues/623
- [ ] Add **Close** button
- [ ] If clicked, go to Screen 4 https://github.com/digidem/comapeo-mobile/issues/623

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2179-1778&t=pHGODPSppOLjQUPF-0

achou11:
~Stacked on #796~

Closes #622
Closes #623
Closes #624 

Refer to the designs in Figma instead of the screenshots from the issues, since they're a bit outdated.

Implementation details:

- ~the error state in the case of validly named file that is actually invalid as a custom map was not detailed in the design, so I had to implement that using my own judgment.~ EDIT: Mostly eliminated by actually verifying the file (see 5284dc3557ac867add9fb17a00040e942d82c022), although I kept the UI pieces so that we handle the super edge case better in case it really does happen for whatever reason.

- for trying it out, you can use fixtures from https://github.com/digidem/styled-map-package/tree/main/test/fixtures

Known issues:

- ~Removing the custom map may not properly update the rendered map (should eventually be fixed in core via https://github.com/digidem/comapeo-core/pull/906) - will be resolved before merging. The demo video is using the changes that fix this issue but they're not in the PR yet, so if you're trying to test it yourself locally it probably won't work.~ EDIT: no longer an issue with b2c015a95e5fc9747be2a8461f52eada402ae82c

- Potential loading issue with download icon, causing a visual flicker when initially loading. I'm not sure if this is a dev-only issue but wondering how to fix 🤔 

---

https://github.com/user-attachments/assets/b6bb5e29-71ef-4408-83d6-02b2330dcf5c

image

image

image

image

image

image

image

image


src/frontend/sharedComponents/icons/index.tsx
```diff
@@ -213,3 +215,7 @@ export const LocationFollowingIcon = ({
 export const StopIcon = ({color = WHITE, size = 30}: FontIconProps) => (
   <FontAwesomeIcon color={color} name="stop" size={size} />
 );
+
+export function DownloadIcon(props: FontIconProps) {
+  return <AntDesignIcon {...props} name="download" />;
+}
```
achou11, line 220: The Ant variant was the closest to what the designs had - ideally we would stick with the same icon sets across all icons though...

src/frontend/lib/file-system.ts
```diff
@@ -1,3 +1,45 @@
...
+import * as DocumentPicker from 'expo-document-picker';
+
+// @ts-expect-error Only null when on web https://github.com/expo/expo/issues/5558
+export const DOCUMENT_DIRECTORY: string = FileSystem.documentDirectory;
+
 export function convertFileUriToPosixPath(fileUri: string) {
   return fileUri.replace(/^file:\/\//, '');
 }
+
+// TODO: Some overlap with selectFile() from lib/utils but fixes some usage limitations. Ideally use this for everything
```
achou11, line 11: just noting this but if there's an interest in unifying the implementations, I'd rather address in a follow up

src/frontend/screens/Settings/MapManagement/CustomMapDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function LoadingOverlay() {
+  return (
+    <View pointerEvents="none" style={styles.overlay}>
+      <View style={[styles.overlay, {opacity: 0.1, backgroundColor: BLACK}]} />
+      <Loading size={10} />
+    </View>
+  );
+}
```
achou11, line 99: could be useful to extract as a shared component, but would prefer to avoid the extra work if not necessary for now

src/backend/src/app.js
```diff
@@ -2,19 +2,23 @@ import debug from 'debug'
...
+const require = createRequire(import.meta.url)
+
+/** @type {import('../types/rn-bridge.js')} */
+const rnBridge = require('rn-bridge')
+
 // Do not touch these!
 const DB_DIR_NAME = 'sqlite-dbs'
 const CORE_STORAGE_DIR_NAME = 'core-storage'
+const CUSTOM_MAPS_DIR_NAME = 'maps'
+const DEFAULT_CUSTOM_MAP_FILE_NAME = 'default.smp'
```
ErikSin, line 21: Just confirming that this requires the developer to know the names of the directories? If so, that seems quite fragile, maybe something to consider refactoring at the UK retreat.

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,345 @@
...
+                      uri: asset.uri,
+                    },
+                    {
+                      onError: () => {
+                        FileSystem.deleteAsync(asset.uri, {
+                          idempotent: true,
+                        }).catch(noop);
+                      },
+                      onSuccess: () => {
+                        mapAddedBottomSheet.openSheet();
```
ErikSin: This <ChooseMapFile /> is identical to the one below. We should create one instance instead of having identical code

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,364 @@
...
+    defaultMessage: 'Delete Map',
+  },
+
+  importErrorTitle: {
+    id: 'screens.Settings.MapManagement.BackgroundMaps.importErrorTitle',
+    defaultMessage: 'Import Error',
+  },
+  importErrorDesciption: {
+    id: 'screens.Settings.MapManagement.BackgroundMaps.importErrorDescription',
+    defaultMessage: 'Unable to import the file. Please go back and try again.',
```
achou11, line 101: ideally this should include the name of the file or the map name, but given how the error modal works and the context in which the error occurs, it's a bit tricky to get that to work seamlessly. for now, intentionally settling for just a generalization

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,364 @@
...
+        title={
+          selectCustomMapMutation.error || importCustomMapMutation.error
+            ? m.importErrorTitle
+            : undefined
+        }
+        description={
+          selectCustomMapMutation.error || importCustomMapMutation.error
+            ? m.importErrorDesciption
+            : undefined
+        }
```
achou11: we only surface a custom title + description in the case of an import error. otherwise just fallback to the default messaging from the component

src/frontend/sharedTypes/navigation.ts
```diff
@@ -61,14 +61,6 @@ export type RootStackParamsList = {
...
   DirectionalArrow: undefined;
   P2pUpgrade: undefined;
-  MapSettings: undefined;
-  BackgroundMaps: undefined;
-  BackgroundMapInfo: {
-    bytesStored: number;
-    id: string;
-    styleUrl: string;
-    name: string;
-  };
```
achou11, line 71: artifact of Mapeo Mobile

src/frontend/hooks/server/maps.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export function useSelectCustomMapFile() {
+  return useMutation({
+    mutationFn: () => {
+      return selectFile({
+        extensionFilters: ['smp'],
+      });
+    },
+  });
+}
```
ErikSin: I think based on how simple this function is, we should just have a `useSelectCustomFile(extenstionfilters?:[string])`, and the consuming component can set the extensions. What do you think?

src/frontend/hooks/server/maps.ts
```diff
@@ -0,0 +1,134 @@
...
+        }).catch(noop);
+
+        throw new Error('Invalid map file');
+      }
+    },
+    onSuccess: () => {
+      refresh();
+      queryClient.invalidateQueries({
+        queryKey: [MAPS_QUERY_KEY],
+      });
```
ErikSin, line 71: I don't fully understand the need for the refresh token? From a surface perspective it looks like it is doing the same thing as invalidating the query

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,364 @@
...
+            {
+              dangerous: true,
+              variation: 'filled',
+              text: t(m.deleteMapButtonText),
+              icon: <MaterialIcon size={30} name="delete" color={WHITE} />,
+              onPress: () => {
+                removeCustomMapMutation.mutate();
+                removeMapBottomSheet.closeSheet();
+              },
+            },
```
ErikSin: would it not be simpler to have 1 bottom sheet modal, where the contents are conditionally rendered?

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,364 @@
...
+            name={customMapInfoQuery.data.name}
+            dateAdded={customMapInfoQuery.data.created}
+            size={customMapInfoQuery.data.size}
+            onRemove={() => {
+              removeMapBottomSheet.openSheet();
+            }}
+          />
+        ) : (
+          <Loading size={10} />
+        );
```
ErikSin: if its fetching, shouldnt we always be showing a loading state? Under the assumption that there can only be 1 background map, swr doesn't apply here

ErikSin (changes requested):
upon looking at this logic for way too long 😬 ... I think i figured out a possible way to seperate the logic to be a little clearer

```ts
// BackgroundMaps.tsx

return (
      <ScrollView contentContainerStyle={styles.container}>
        <Text style={styles.aboutText}>{t(m.about)}</Text>
        <View style={styles.descriptionContainer}>
          <Text>{t(m.description1)}</Text>
          <Text>{t(m.description2)}</Text>
        </View>
        { customMapInfoQuery.isFetching ?
             <Loader> : !customMapInfoQuery.data ?
             <ChooseMapFile />:
             <CustomMapDetails/>
        }
        { customMapInfoQuery.isError &&
             <>
                  <Text style={styles.infoLoadErrorText}>
                   {t(m.customMapInfoLoadError)}
                  </Text>
                <Button
                  fullWidth
                  variant="outlined"
                 onPress={() => {
                  removeCustomMapMutation.mutate();
               }}>
                  <Text style={styles.removeMapFileButtonText}>
                     {t(m.removeMapFile)}
                  </Text>
                </Button>
             </>
        }
      </ScrollView>
)

```

`ChooseMapFile` could then just deal with its own loading state. But this would require ChooseMapFile to have its own ErrorModal.
And anytime the maps are fetching, the loader is shown. Since the use can only have 1 custom map we don't need to worry about showing the previous map. This would mean that `CustomMapDetails` would not have to have a loading state, because the parent component is dealing with it

I think this would simplifiy the switch statement with if statements inside of it

src/backend/src/app.js
achou11: yeah there's some duplication between frontend and backend in terms of knowing where this directory is. However, there's no good way to share code between the two unless you involve an unnecessary IPC roundtrip, which I'd rather avoid in this case.

src/frontend/hooks/server/maps.ts
achou11: yeah the tricky part here is that the invalidation we need is controlled by the consuming map view, not react query. for example:

1. We have an initial value of: `http://localhost:8000/maps/style.json`. This is used by the map view, which makes the http request internally to fetch a response. let's say `{ "name": "my-map" }` for the sake of the example.
2. We do something that should change what style.json we get back in 1 (such as add or remove a map). 
3. We invalidate the query, but that isn't enough in some cases because you can still get the same value of `http://localhost:8000/maps/style.json`, in which case the mapview won't actually get a new response from that URL because of HTTP caching.

In order to actually force the mapview to fully fetch the response of the same url and ignore the cache, we introduce the refresh token in the form of a url query string, which counts as a new url, even though the query string goes unused by the server that we're calling. we did something similar in Mapeo Mobile for other http resources that would have the same issue (e.g. [here](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/api.ts))

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
achou11: in this case i would argue against it for the following reasons:

1. One is a full-sized sheet and the other isn't. If i remember correctly, we had issues with dynamically switching between full-size and non-full-size (some issue with the component implementation)

2. It provides a more explicit separation of states that can occur. Part of this is due to a limitation of the how the bottom sheet component is designed, but i'd rather avoid code that does something like `isOpen={mapAddedOpen || mapRemovedOpen}`, because then there's a greater risk of introducing or seeing the wrong sheet content, unless you're careful about the checks you make in it.

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
achou11: To clarify: this is handling the case where a map currently exists and we're doing something that would trigger another fetch (e.g. removing or updating).

It is showing a loading state (see the `loading` prop in the `CustomMapDetails`) but it's doing a more visually graceful loading state. I found it very jarring for the information to just suddenly disappear, turn into the three dots, and then into some other UI, so I implemented a simple overlay that prevents less layout shift in this specific case.

It looks like this, which may look weird when seeing as a screenshot, but I argue that it's preferable to just showing the dots:

image

achou11:
> ChooseMapFile could then just deal with its own loading state. 

Haven't tried out your suggestion just yet, but I'm a little hesitant about introducing an additional query "boundary" when it may not be necessary to. I usually prefer to keep all of the query logic at the same level and just have the rendered components be pretty dumb.

What I currently like about what I have implemented is that it clearly handles all of the states that this UI can end up in:

1. Initial load -> show loading
2. Initial load succeeds, which has the following outcomes:
  a. no custom map info -> show upload file UI
  b. custom map info exists -> show custom map info
3. Refetching the custom map info (due to removing the map, but in the future, could be due to updating or adding a map). This has outcomes identical to (2)
4. An error occurs due to:
  a. a bad file being chosen
  b. some server error (which is relatively unlikely but could happen)

While the implementation doesn't look pretty, I don't find it overly difficult to follow. Just happens to be a consequence of how to work with react + react query.

Even if I move a lot of the logic to another nested component, I'd still probably end up with the switch statement approach, even though it may be slightly less complicated.

Overall I find it easier to approach building the UI based on react query's `status` value, even if it results in some duplicated UI code. The `data` oriented approach feels messy to me because it doesn't guarantee what state the query is in, so you'd have to include specific one-off checks if you go about it that way.

achou11:
here's a first attempt at what you suggested that takes into the various states that I mentioned (no additional component separation just yet):

```tsx
 <ScrollView contentContainerStyle={styles.container}>
        <Text style={styles.aboutText}>{t(m.about)}</Text>
        <View style={styles.descriptionContainer}>
          <Text>{t(m.description1)}</Text>
          <Text>{t(m.description2)}</Text>
        </View>

        {/* Think your suggestion was to make this a separate component */}
        {customMapInfoQuery.status === 'pending' ? (
          <Loading size={10} />
        ) : customMapInfoQuery.data ? (
          <CustomMapDetails
            loading={customMapInfoQuery.isFetching}
            name={customMapInfoQuery.data.name}
            dateAdded={customMapInfoQuery.data.created}
            size={customMapInfoQuery.data.size}
            onRemove={() => {
              removeMapBottomSheet.openSheet();
            }}
          />
        ) : customMapInfoQuery.isFetching ? (
          <Loading size={10} />
        ) : (
          <ChooseMapFile
            onChooseFile={() => {
              selectCustomMapMutation.mutate(undefined, {
                onSuccess: asset => {
                  if (!asset) return;

                  importCustomMapMutation.mutate(
                    {
                      uri: asset.uri,
                    },
                    {
                      onError: () => {
                        FileSystem.deleteAsync(asset.uri, {
                          idempotent: true,
                        }).catch(noop);
                      },
                      onSuccess: () => {
                        mapAddedBottomSheet.openSheet();
                      },
                    },
                  );
                },
              });
            }}
          />
        )}

        {customMapInfoQuery.status === 'error' && (
          <>
            <Text style={styles.infoLoadErrorText}>
              {t(m.customMapInfoLoadError)}
            </Text>
            <Button
              fullWidth
              variant="outlined"
              onPress={() => {
                removeCustomMapMutation.mutate();
              }}>
              <Text style={styles.removeMapFileButtonText}>
                {t(m.removeMapFile)}
              </Text>
            </Button>
          </>
        )}
      </ScrollView>
```


Aside from the error stuff, which I agree can just be inlined as you suggested, I don't know how I feel about this.

EDIT: The more I look at it, the more okay I am with it. Definitely appreciate that it's less code in general. just find it a little harder to understand that ternary, which I could break up into a switch statement or variable in the render body, but that would be closer to the initial criticisim 😄

achou11:
Alright did a little cleanup that moves towards what you suggested and I'm feeling much better about it:

```tsx
 <ScrollView contentContainerStyle={styles.container}>
        <Text style={styles.aboutText}>{t(m.about)}</Text>
        <View style={styles.descriptionContainer}>
          <Text>{t(m.description1)}</Text>
          <Text>{t(m.description2)}</Text>
        </View>

        <CustomMapInfoSection
          onChooseFile={() => {
            selectCustomMapMutation.mutate(undefined, {
              onSuccess: asset => {
                if (!asset) return;

                importCustomMapMutation.mutate(
                  {
                    uri: asset.uri,
                  },
                  {
                    onError: () => {
                      FileSystem.deleteAsync(asset.uri, {
                        idempotent: true,
                      }).catch(noop);
                    },
                    onSuccess: () => {
                      mapAddedBottomSheet.openSheet();
                    },
                  },
                );
              },
            });
          }}
          onRemoveMap={() => {
            removeMapBottomSheet.openSheet();
          }}
        />

        {customMapInfoQuery.status === 'error' && (
          <>
            <Text style={styles.infoLoadErrorText}>
              {t(m.customMapInfoLoadError)}
            </Text>
            <Button
              fullWidth
              variant="outlined"
              onPress={() => {
                removeCustomMapMutation.mutate();
              }}>
              <Text style={styles.removeMapFileButtonText}>
                {t(m.removeMapFile)}
              </Text>
            </Button>
          </>
        )}
      </ScrollView>
```

where `CustomMapSection` is:

```tsx
function CustomMapInfoSection({
  onChooseFile,
  onRemoveMap,
}: {
  onChooseFile: () => void;
  onRemoveMap: () => void;
}) {
  const customMapInfoQuery = useGetCustomMapInfo();

  if (customMapInfoQuery.status === 'pending') {
    return <Loading size={10} />;
  }

  if (customMapInfoQuery.data) {
    return (
      <CustomMapDetails
        loading={customMapInfoQuery.isFetching}
        name={customMapInfoQuery.data.name}
        dateAdded={customMapInfoQuery.data.created}
        size={customMapInfoQuery.data.size}
        onRemove={onRemoveMap}
      />
    );
  }

  return customMapInfoQuery.isFetching ? (
    <Loading size={10} />
  ) : (
    <ChooseMapFile onChooseFile={onChooseFile} />
  );
}
```

Thanks for bearing with me! Now the question is: is this an improvement over what I currently have implemented? Would say that it's maybe enough of an improvement to commit to - thanks for the suggestion and making me think harder about this!

EDIT: Part of me wants to include the error rendering in that `CustomMapInfoSection` component, but then i think i'll run into some implementation messiness

src/frontend/hooks/server/maps.ts
achou11: addressed via dcac9e0472ef7ee072605e3e3be74883ed5668c7

Think ideally this stuff in `useSelectFileAndImportConfig.ts` also lives in the file I created here, but would consider that more of a follow-up item

ErikSin:
> Alright did a little cleanup that moves towards what you suggested and I'm feeling much better about it:
> 
> ```tsx
>  <ScrollView contentContainerStyle={styles.container}>
>         <Text style={styles.aboutText}>{t(m.about)}</Text>
>         <View style={styles.descriptionContainer}>
>           <Text>{t(m.description1)}</Text>
>           <Text>{t(m.description2)}</Text>
>         </View>
> 
>         <CustomMapInfoSection
>           onChooseFile={() => {
>             selectCustomMapMutation.mutate(undefined, {
>               onSuccess: asset => {
>                 if (!asset) return;
> 
>                 importCustomMapMutation.mutate(
>                   {
>                     uri: asset.uri,
>                   },
>                   {
>                     onError: () => {
>                       FileSystem.deleteAsync(asset.uri, {
>                         idempotent: true,
>                       }).catch(noop);
>                     },
>                     onSuccess: () => {
>                       mapAddedBottomSheet.openSheet();
>                     },
>                   },
>                 );
>               },
>             });
>           }}
>           onRemoveMap={() => {
>             removeMapBottomSheet.openSheet();
>           }}
>         />
> 
>         {customMapInfoQuery.status === 'error' && (
>           <>
>             <Text style={styles.infoLoadErrorText}>
>               {t(m.customMapInfoLoadError)}
>             </Text>
>             <Button
>               fullWidth
>               variant="outlined"
>               onPress={() => {
>                 removeCustomMapMutation.mutate();
>               }}>
>               <Text style={styles.removeMapFileButtonText}>
>                 {t(m.removeMapFile)}
>               </Text>
>             </Button>
>           </>
>         )}
>       </ScrollView>
> ```
> 
> where `CustomMapSection` is:
> 
> ```tsx
> function CustomMapInfoSection({
>   onChooseFile,
>   onRemoveMap,
> }: {
>   onChooseFile: () => void;
>   onRemoveMap: () => void;
> }) {
>   const customMapInfoQuery = useGetCustomMapInfo();
> 
>   if (customMapInfoQuery.status === 'pending') {
>     return <Loading size={10} />;
>   }
> 
>   if (customMapInfoQuery.data) {
>     return (
>       <CustomMapDetails
>         loading={customMapInfoQuery.isFetching}
>         name={customMapInfoQuery.data.name}
>         dateAdded={customMapInfoQuery.data.created}
>         size={customMapInfoQuery.data.size}
>         onRemove={onRemoveMap}
>       />
>     );
>   }
> 
>   return customMapInfoQuery.isFetching ? (
>     <Loading size={10} />
>   ) : (
>     <ChooseMapFile onChooseFile={onChooseFile} />
>   );
> }
> ```
> 
> Thanks for bearing with me! Now the question is: is this an improvement over what I currently have implemented? Would say that it's maybe enough of an improvement to commit to - thanks for the suggestion and making me think harder about this!
> 
> EDIT: Part of me wants to include the error rendering in that `CustomMapInfoSection` component, but then i think i'll run into some implementation messiness

This implementation on a surface level looks much cleaner and is defintely easier to follow.

achou11:
Heads up: I updated core and IPC deps with the fix to the issue mentioned in the PR description via b2c015a95e5fc9747be2a8461f52eada402ae82c

src/backend/package.json
```diff
@@ -13,8 +13,8 @@
   "author": "Digital Democracy",
   "license": "MIT",
   "dependencies": {
-    "@comapeo/core": "2.0.0",
-    "@comapeo/ipc": "2.0.0",
+    "@comapeo/core": "2.0.1",
+    "@comapeo/ipc": "2.0.2",
```
achou11, line 17: just highlighting that we need this update to core to fix an issue that happens when removing a custom map.

package.json
```diff
@@ -108,14 +108,15 @@
     "tiny-typed-emitter": "^2.1.0",
     "uint8array-extras": "^0.5.0",
     "utm": "^1.1.1",
+    "valibot": "^0.42.1",
```
achou11, line 111: I'm introducing this dep for response body validation, which is used when interfacing with the map server

src/backend/src/app.js
achou11: for posterity: the way this custom map feature works is the following:

1. The frontend is responsible for moving a file to and from an app-specific directory on the device. In this case, we the android "document directory" which is something like `file:///data/user/0/com.comapeo.dev/files/` and make a `maps/default.smp` there.
2. The backend is configured to use the path in (1) as the path to watch for determining if there's a custom map file that can be used for the map server.

src/frontend/hooks/server/maps.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+}
+
+export function useSelectFile() {
+  return useMutation({
+    mutationFn: (opts: Parameters<typeof selectFile>[0]) => {
+      return selectFile(opts);
+    },
+  });
+}
```
ErikSin: Do we need this twice?

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,336 @@
...
+          }
+          buttonConfigs={[
+            {
+              dangerous: true,
+              variation: 'filled',
+              text: t(m.deleteMapButtonText),
+              icon: <MaterialIcon size={30} name="delete" color={WHITE} />,
+              onPress: () => {
+                removeCustomMapMutation.mutate();
+                removeMapBottomSheet.closeSheet();
```
ErikSin, line 210: Any concerns about not having a loading state

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -0,0 +1,336 @@
...
+          ]}
+        />
+      </BottomSheetModal>
+
+      <BottomSheetModal
+        fullScreen
+        ref={mapAddedBottomSheet.sheetRef}
+        isOpen={mapAddedBottomSheet.isOpen}>
+        <BottomSheetModalContent
+          icon={<GreenCheckSvg />}
```
ErikSin: this should have a [marginTop of 80](https://www.notion.so/digidem/Spacing-74d09a0bdefa4bc5b9508336f7b2142a?pvs=4#9dfdb9bddf6a4f25a627e6b27b5dde0e)

src/frontend/screens/Settings/MapManagement/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    <ScrollView>
+      <List>
+        <ListItem
+          onPress={() => {
+            navigation.navigate('BackgroundMaps');
+          }}>
+          <ListItemText primary={t(m.backgroundMaps)} />
+        </ListItem>
+      </List>
+    </ScrollView>
```
ErikSin, line 34: [This should have a top padding of 40](https://www.notion.so/digidem/Spacing-74d09a0bdefa4bc5b9508336f7b2142a?pvs=4#515a2f0e66d944348c7dd350df02cf6c)

ErikSin (approved):
Nice job! seems to be working. A few small comments, but non-blocking

src/frontend/screens/Settings/MapManagement/index.tsx
achou11: seems somewhat questionable for that design spec to apply to this page, especially when compared to other settings-like pages.

App settings:

image

Screen in question (with your suggestion):

image

My guess is that either:

- the app settings page needs to be updated too
- the design spec does not apply to pages like these

achou11: > the app settings page needs to be updated too

ah yeah it this seems likely when looking at the screenshot in https://github.com/digidem/comapeo-mobile/issues/622

I think for the sake of consistency, I'm going to avoid making this change so that it matches the app settings layout. we can address both of those in a follow up (since there are a good number of design changes needed anyways)

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
achou11: yeah this ideally is fixed in the `BottomSheetModalContent` component but addressed here as a one-off for now (since we probably do the same in other places). See e65020cab5368512d268a2fe20755ba79e8b5aa1

image

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
achou11: not super concerned because this happens so quickly and also we have a different modal for an error that may occur. Added a small change anyways just in case though, via b79e78d5f8d165a156f4fa8cd265f8e708329875

src/frontend/hooks/server/maps.ts
achou11: oh no we don't - i copied this to a different file but forgot to actually use that one. Removed this one via 5a702b4a90eef5685a1b7419ca27aff8b8f572ba

Issue #360, opened 2024-05-15
New audio recording


ximenabb:
## New audio recording
Tapping on either “While using the app” or “Only this time” takes the user this interface.
User can record up to 5 minutes of audio.

## Todo

**New Audio Recording - Start**
<img width="193" alt="Screen Shot 2024-05-15 at 3 53 16 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/7d7d3f49-59df-4eb3-97d8-89f4994a1b37">
- [x] Display screen as designed
- [x] Display counter at 00:00
- [x] Display text Record up to 5 minutes
- [x] Display red circle record button, when tapped start recording as _New Audio Recording - In Progress_

**New Audio Recording - In Progress**
<img width="200" alt="Screen Shot 2024-05-15 at 3 57 17 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/c3e36011-b1ac-4576-a300-de78ee13777d">
- [x] Display countdown of minutes left to the recording, starting at 5mins
- [x] Update the red circle record button to the recording button, as designed
- [x] To display the progress in color - the record screen “fills up” with blue, e.g. a horizontal block of blue starts at the bottom of the screen and slowly expands to fill the screen until the entire screen is blue at 5 minutes.  !!! SWM team - can you let me know if that logic makes sense or need more details? 

**New Audio Recording - 60-second Alert**
<img width="190" alt="Screen Shot 2024-05-15 at 4 07 21 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/2e95c0b3-4389-4cbd-b0d3-3e5070f1d7be">
- [x] When Audio Recording reaches 60-seconds left, brighten the blue to Mapeo Blue

**Stop recording button**
- [x] Tapping the “Stop” button, ends the Audio Recording and  displays the Playback interface

**Playback interface**
<img width="197" alt="Screen Shot 2024-05-15 at 4 11 13 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/ce973467-d019-46a9-b2e4-b588dc145ef4">
- [x] Tapping the “X” button, attaches the audio to Observation Draft and takes user to “Success” confirmation.
- [x] Display audio progress bar
- [x] Display _Total length_ and the length in minutes
- [x] Display Play button, to replay audio
- [x] Display Trash Can button, when clicks takes to the Delete screen

**Success Confirmation**
[...]

PR #760, opened 2024-09-26
feat: adds audio permissions

cimigree:
Adds functionality for asking for permissions for Audio. Adds Audio navigation and dummy screen. Starts to address this issue: [#360 ](https://github.com/digidem/comapeo-mobile/issues/360)
Here are the [designs](https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/%22Ready-for-Dev%22-Hand-off-Assets?node-id=0-1&node-type=canvas&t=DhW0UI0lwgUo0EC5-0)
I found this command helpful to reset the permissions to be able to test different scenarios:
` adb shell pm reset-permissions`
Make sure to add this to your env file: EXPO_PUBLIC_FEATURE_AUDIO=true

### DIFF
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -326,11 +326,13 @@ export const createDefaultScreenGroup = ({
       component={HowToLeaveProject}
       options={{headerShown: false}}
     />
-    <RootStack.Screen
-      name="Audio"
-      options={audioNavigationOptions}
-      component={Audio}
-    />
+    {process.env.EXPO_PUBLIC_FEATURE_AUDIO && (
+      <RootStack.Screen
+        name="Audio"
+        options={audioNavigationOptions}
+        component={Audio}
+      />
+    )}
     {process.env.EXPO_PUBLIC_FEATURE_TEST_DATA_UI && (
       <RootStack.Screen
         name="CreateTestData"
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -76,6 +76,10 @@ import {SettingsPrivacyPolicy} from '../../screens/Settings/DataAndPrivacy/Setti
 import {TrackEdit} from '../../screens/TrackEdit/index.tsx';
 import {Config} from '../../screens/Settings/Config';
 import {HowToLeaveProject} from '../../screens/HowToLeaveProject.tsx';
+import {
+  Audio,
+  navigationOptions as audioNavigationOptions,
+} from '../../screens/Audio/index.tsx';
 
 export const TAB_BAR_HEIGHT = 70;
 
@@ -322,7 +326,11 @@ export const createDefaultScreenGroup = ({
       component={HowToLeaveProject}
       options={{headerShown: false}}
     />
-
+    <RootStack.Screen
+      name="Audio"
+      options={audioNavigationOptions}
+      component={Audio}
+    />
     {process.env.EXPO_PUBLIC_FEATURE_TEST_DATA_UI && (
       <RootStack.Screen
         name="CreateTestData"
--- a/src/frontend/sharedComponents/ActionRow.tsx
+++ b/src/frontend/sharedComponents/ActionRow.tsx
@@ -6,7 +6,7 @@ import AudioIcon from '../images/observationEdit/Audio.svg';
 import DetailsIcon from '../images/observationEdit/Details.svg';
 import {useNavigation} from '@react-navigation/native';
 import {Preset} from '@comapeo/schema';
-import {PermissionAudioBottomSheetContent} from './PermissionAudioBottomSheetContent';
+import {PermissionAudioBottomSheetContent} from '../screens/Audio/PermissionAudioBottomSheetContent';
 import {Audio} from 'expo-av';
 import {useBottomSheetModal} from '../sharedComponents/BottomSheetModal';
 import {BottomSheetModal} from './BottomSheetModal';
--- a/src/frontend/sharedComponents/ActionRow.tsx
+++ b/src/frontend/sharedComponents/ActionRow.tsx
@@ -4,11 +4,14 @@ import {ActionTab} from './ActionTab';
 import PhotoIcon from '../images/observationEdit/Photo.svg';
 import AudioIcon from '../images/observationEdit/Audio.svg';
 import DetailsIcon from '../images/observationEdit/Details.svg';
-import {useNavigationFromRoot} from '../hooks/useNavigationWithTypes';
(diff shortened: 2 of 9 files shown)


src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -322,7 +326,11 @@ export const createDefaultScreenGroup = ({
       component={HowToLeaveProject}
       options={{headerShown: false}}
     />
-
+    <RootStack.Screen
+      name="Audio"
+      options={audioNavigationOptions}
+      component={Audio}
+    />
```
ErikSin: Should this be under the feature flag?

src/frontend/sharedComponents/ActionRow.tsx
```diff
@@ -59,5 +84,14 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
...
 
-  return <ActionTab items={bottomSheetItems} />;
+  return (
+    <>
+      <ActionTab items={bottomSheetItems} />
+      <PermissionAudio
+        closeSheet={closeAudioPermissionSheet}
+        isOpen={isAudioPermissionSheetOpen}
+        sheetRef={audioPermissionSheetRef}
+      />
```
ErikSin: I think it would be better to have the `<BottomSheetModal / >` component live on this page, so you don't need to pass the props down (that's how it is designed to be used)

And then I think you could rename the `PermissionAudio` to be more descriptive such as `PermissionAudioBottomSheetContent`

So it would look like
```ts
< BottomSheetModal closeSheet={closeAudioPermissionSheet}
    isOpen={isAudioPermissionSheetOpen}
    sheetRef={audioPermissionSheetRef} >
        < PermissionAudioBottomSheetContent />
</ BottomSheetModal>
```

src/frontend/sharedComponents/PermissionAudio.tsx
```diff
@@ -1,72 +1,100 @@
...
+    const response = await Audio.requestPermissionsAsync();
+    setPermissionResponse(response);
+    if (response.status === 'granted') {
+      closeSheet();
+      navigation.navigate('Audio');
+    } else if (response.status === 'denied' && response.canAskAgain) {
+      closeSheet();
+    } else if (response.status === 'denied' && !response.canAskAgain) {
+      handleOpenSettings();
+    }
```
ErikSin: is this logic backwards?

src/frontend/sharedComponents/PermissionAudio.tsx
```diff
@@ -1,72 +1,100 @@
-import React, {FC, useCallback, useEffect} from 'react';
+import React, {FC, useState, useEffect} from 'react';
```
ErikSin: can remove unused `useEffect`

src/frontend/sharedComponents/PermissionAudio.tsx
```diff
@@ -1,72 +1,100 @@
...
+    };
+    actionButtonText = t(m.allowButtonText);
+  } else if (permissionResponse.status === 'denied') {
+    onPressActionButton = handleOpenSettings;
+    actionButtonText = t(m.goToSettingsButtonText);
+  } else {
+    onPressActionButton = async () => {
+      await handleRequestPermission();
+    };
+    actionButtonText = t(m.allowButtonText);
```
ErikSin: this may be a preference thing, so feel free to ignore, but this feels cleaner as 2 seperate ternaries:

```ts
const onPressActionButton = !permissionResponse
  ? async () => {
      await handleRequestPermission();
    }
  : permissionResponse.status === 'denied'
  ? handleOpenSettings
  : async () => {
      await handleRequestPermission();
    };

const actionButtonText= !permissionResponse
  ? t(m.allowButtonText)
  : permissionResponse.status === 'denied'
  ? t(m.goToSettingsButtonText)
  : t(m.allowButtonText);
```

src/frontend/sharedComponents/PermissionAudio.tsx
```diff
@@ -1,72 +1,100 @@
...
     <BottomSheetModal
-      fullScreen
       ref={sheetRef}
+      isOpen={isOpen}
       onDismiss={closeSheet}
-      isOpen={isOpen}>
+      fullScreen>
       <BottomSheetModalContent
         icon={<AudioPermission />}
         title={t(m.title)}
```
ErikSin: this should have a [paddingTop of 80](https://www.notion.so/digidem/Spacing-74d09a0bdefa4bc5b9508336f7b2142a?pvs=4#059606ae2b9840018976a1d1400f1836)

ErikSin (changes requested):
minor changes, mostly related to formatting.

But the functionality seems to be doing what we want it to do!

src/frontend/Navigation/Stack/AppScreens.tsx
cimigree: I am not sure. You can only get to the Audio file from the bottom sheet item in the action row which is under the feature flag. What do you think?

src/frontend/sharedComponents/PermissionAudio.tsx
cimigree: It actually isn't. I am not sure of the logic behind why "canAskAgain" is what it is, I just observed how it worked. If your response is canAskAgain, it is immediately after your very first response to the device query asking for Permission -- the first time you answer it saying, no, I don't give permission, and so immediately after that, in that case you don't need to go to settings and can just have the sheet close. 

If it's a response of !canAskAgain, you have already said you don't give permission but are asking for using audio again, so now you need to go to settings to understand what it is we are asking you (which is I assume what codesign wanted).

I am not sure if I explained that well! I can walk you through it over screen share if that's easier.

src/frontend/sharedComponents/ActionRow.tsx
cimigree: Ok.

src/frontend/sharedComponents/PermissionAudio.tsx
cimigree: I see... each button on its own. Sure.

src/frontend/Navigation/Stack/AppScreens.tsx
ErikSin: I think it wouldn't hurt to put this under a feature flag

src/frontend/sharedComponents/PermissionAudioBottomSheetContent.tsx
ErikSin, line 1: one other small things, I don't think this is a shared component. Its only being used by the audio screen, so I think it should just go in `screens/Audio`

ErikSin (approved):
Looks good, Just one comment about where the file belongs. Just run `npx tsc` after you move the file to makes sure all the file imports were properly changed before merging in.

PR #381, opened 2024-05-24
feat: audio recording

CDFN:
This PR adds most of recording feature which is subject of #360 and #361.

PR #762, opened 2024-09-26
feat: adds creation, saving and playback of audio

### LINKED ISSUE #360: New audio recording

cimigree:
closes #360 

### Description
Adds the ability to create an audio file from an observation
Adds the necessary files and logic so that when you hit "stop" on an audio recording, you are given the Playback option to be able to hear what you just recorded. 
Continues to address issue https://github.com/digidem/comapeo-mobile/issues/360
Used this PR that [Andrew created](https://github.com/digidem/comapeo-mobile/compare/develop...360/create-audio-recording)
Here are the [designs](https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/%22Ready-for-Dev%22-Hand-off-Assets?node-id=0-1&node-type=canvas&t=DhW0UI0lwgUo0EC5-0)
Uses some code from [SWM](https://github.com/digidem/comapeo-mobile/pull/381)

Instructions
Make sure to add this to your env file: EXPO_PUBLIC_FEATURE_AUDIO=true
Go to create an observation
Hit the microphone
Grant permission to record audio
You will be taken to the create audio recording screen
Watch the screen change color and the time pass and the countdown change.
It will stop automatically at 5 minutes
Hit the stop button to stop recording.
You can play, stop, play, stop, and then when you get to the end it resets to the beginning. Hope that is right. Designs weren't clear.
At that point, the audio is added to state. 
You can delete it with the trashcan.
If you hit the X in the top left, the audio is added to the draft observation, the success modal opens, and you are given the option to return to the editor or record another audio, which I wired up to work.
Once you save the Observation, the audio file or files is added to the observation.
You can't see these yet... That will be another PR!


src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -0,0 +1,149 @@
...
+              uri,
+            };
+            addAudio(audioRecording);
+            openSuccessSheet();
+          }}
+          backImage={props => <CloseIcon color={props.tintColor} />}
+        />
+      ),
+    });
+  }, [navigation, openSuccessSheet]);
```
ErikSin: this is missing `addAudio`, `createdAt`, `duration`, and `uri` from the dependency array

src/frontend/screens/Audio/CreateRecording/index.tsx
ErikSin, line 1: We should [prevent the user from going back](https://reactnavigation.org/docs/preventing-going-back) on this page (except for when `recording.staus === idle`)

src/frontend/screens/Audio/CreateRecording/RecordingIdle.tsx
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    navigation.setOptions({
+      headerLeft: props => (
+        <CustomHeaderLeft
+          tintColor={props.tintColor}
+          headerBackButtonProps={props}
+        />
+      ),
+    });
+  }, [navigation]);
```
ErikSin, line 30: this block of code is unnecessary. By default it has a `headerLeft`

src/frontend/screens/Audio/ContentWithControls.tsx
ErikSin, line 1: Everyrtime you use `ContentWithControl`, the prop `controls` uses `<Controls.Row >`. Would it make sense to just integrate `Controls.Row` directly into this component so the developer does not need to import and wrap it everytime?

src/frontend/screens/Audio/CreateRecording/RecordingActive.tsx
```diff
@@ -0,0 +1,64 @@
...
+    (MAX_RECORDING_DURATION_MS - duration) / 60_000,
+  );
+
+  const elapsedTimeValue = useDerivedValue(() => {
+    return withTiming(duration, {duration: 500});
+  }, [duration]);
+
+  useEffect(() => {
+    navigation.setOptions({headerLeft: () => null});
+  }, [navigation]);
```
ErikSin: I think `headerShown:false` would be clearer here.

src/frontend/screens/Audio/CreateRecording/useAudioRecording.ts
```diff
@@ -0,0 +1,117 @@
...
+  if (!state) {
+    return {
+      status: 'idle',
+      startRecording: async () => {
+        const createdAt = Date.now();
+        const {recording, status} = await Audio.Recording.createAsync(
+          Audio.RecordingOptionsPresets.HIGH_QUALITY,
+          status => {
+            setState(prev => {
+              if (!prev) return prev;
```
ErikSin: what is the logic here? If Im understanding it, it the state is null, don't set the state. But the state is null by default. So we want to ignore the state being set until the recording is complete (which makes sense). But then why the return statement below this line? would we not want to just set the state to null if there is a previous state?

cimigree:
@ErikSin Ready for continuation of review. I changed the multi ternary to an if statement. I find it a little more readable. I moved the createdAt to happen when the recording is saved. I had to keep the header in RecordingIdle because it is needed when you go to record another audio recording after you have recorded the first one. Got rid of the unlink function because resetting seems to do the trick upon deletion as you noticed.

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -0,0 +1,147 @@
...
+        />
+      </BottomSheetModal>
+      <RecordingSuccessModal
+        sheetRef={successSheetRef}
+        isOpen={isSuccessSheetOpen}
+        onReturnToEditor={handleReturnToEditor}
+        onRecordAnother={handleRecordAnother}
+      />
+    </>
+  );
```
ErikSin: I think it would be better to have one bottom sheet modal with 2 different `< BottomSheetContents/ >` .That way you wouldn't have to import useBottomSheetModalTwice

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -273,7 +287,9 @@ export const ObservationCreate = ({
         <SaveButton
           onPress={checkAccuracyAndLocation}
           isLoading={
-            createObservationMutation.isPending || createBlobMutation.isPending
+            createObservationMutation.isPending ||
+            createBlobMutation.isPending ||
+            createAudioBlobMutation.isPending
```
ErikSin, line 292: `createAudioBlobMutation.isPending` needs to be added to the useEffect dependencies

src/frontend/screens/Audio/CreateRecording/useAudioRecording.ts
```diff
@@ -0,0 +1,35 @@
...
+    setRecording(null);
+    setStatus(null);
+    setUri(null);
+  };
+
+  async function stopRecording() {
+    if (!recording) return;
+    await recording.stopAndUnloadAsync();
+    setUri(recording.getURI());
+  }
```
ErikSin: We should probably make all 3 of these function `useCallback` functions.

src/frontend/hooks/useDraftObservation.ts
```diff
@@ -92,6 +94,13 @@ export const useDraftObservation = () => {
     [deletePersistedPhoto, deletePhotoPromise],
   );
 
+  const addAudio = useCallback(
+    (audioRecording: AudioRecording) => {
+      addAudioRecording(audioRecording);
+    },
+    [addAudioRecording],
+  );
+
```
ErikSin: creating this function is unneccessary, you can just export `AddAudioRecording` directly (i dont think renaming it is necessary)

src/frontend/screens/Audio/Controls.tsx
```diff
@@ -0,0 +1,119 @@
...
+      {...pressableProps}
+      style={({pressed}) => [
+        styles.basePressable,
+        typeof pressableProps.style === 'function'
+          ? pressableProps.style({pressed})
+          : pressableProps.style,
+        pressed && styles.pressablePressed,
+      ]}>
+      {children}
+    </Pressable>
```
ErikSin: it would probably be good to turn this into a touchable opacity to remain consistent with the rest of the app (unless it drastically changes how it works/looks)

src/frontend/screens/Audio/CreateRecording/useAutoStopRecording.ts
```diff
@@ -0,0 +1,12 @@
+import {useEffect} from 'react';
+
+export function useAutoStopRecording(
+  minutesRemaining: number,
+  onPressStop: () => void,
+) {
+  useEffect(() => {
+    if (minutesRemaining === 0) {
```
ErikSin: just to be safe, i think we should do `minutesRemaining <= 0`

src/frontend/screens/Audio/Playback.tsx
ErikSin, line 1: this should probably be a shared component because it is going to be used by other screens in the future

src/frontend/screens/Audio/constants.ts
ErikSin, line 1: To remain consistent with other constants, i don't think it needs it own file and can just be stored in `Audio.index.tsx`

src/frontend/screens/Audio/useAudioPlayback.ts
```diff
@@ -0,0 +1,85 @@
+import {Audio, AVPlaybackStatus, AVPlaybackStatusSuccess} from 'expo-av';
+import {useCallback, useEffect, useState, useRef} from 'react';
+import {Sound} from 'expo-av/build/Audio/Sound';
+
+export const useAudioPlayback = (recordingUri: string) => {
+  const recordedSoundRef = useRef<Sound | null>(null);
+  const [isPlaying, setPlaying] = useState(false);
+  const [duration, setDuration] = useState<number>(0);
+  const [currentPosition, setCurrentPosition] = useState(0);
+  const [isReady, setReady] = useState(false);
```
ErikSin: I think we can take out isReady completely (or don't make it state). Its not really being used in a meaningful way, and we don't export it

src/frontend/screens/Audio/useAudioPlayback.ts
```diff
@@ -0,0 +1,85 @@
...
+  }, []);
+
+  useEffect(() => {
+    let soundInstance: Sound | null = null;
+    Audio.Sound.createAsync({uri: recordingUri})
+      .then(({sound, status}) => {
+        if ('error' in status && status.error) {
+          console.error('Error while creating audio playback', status.error);
+          return;
+        }
```
ErikSin, line 31: as a follow up pr, i think we should handle audio errors (because we do not handle them in the UI at all)

ErikSin (changes requested):
Some adjustment but the overall architecture seems sound. I am a little worried though about the fact that there is no error handling, so I think we should address that in a follow up PR.

src/frontend/screens/Audio/CreateRecording/RecordingIdle.tsx
cimigree: As I said above, without it, when I go to record another audio recording, there is no header and I am stuck on the screen. Is there another way to make sure there is a header on the page in that situation?

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
cimigree: I did it but I personally I don't like it, tbh... Much more messy and seems like a lot more code.

src/frontend/screens/Audio/useAudioPlayback.ts
cimigree: Ok

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -1,40 +1,64 @@
...
   const navigation = useNavigationFromRoot();
+  const {startRecording, stopRecording, reset, status, uri} =
+    useAudioRecording();
+
+  let currentState = 'idle';
+  if (status && status.isRecording) {
+    currentState = 'active';
+  } else if (uri) {
+    currentState = 'done';
+  }
```
ErikSin: slight optimization would be: const currentState = status?.isRecording ? 'active' : uri ? 'done' : 'idle';

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -1,40 +1,64 @@
...
+        onDelete={() => {
+          navigation.goBack();
+          reset();
+        }}
+        onRecordAnother={reset}
+      />
+    );
+  }
+
+  return null;
```
ErikSin: this can be removed

src/frontend/screens/Audio/useAudioPlayback.ts
```diff
@@ -0,0 +1,82 @@
...
+  const stopPlayback = async () => {
+    if (!recordedSoundRef.current || !isPlaying) return;
+
+    try {
+      await recordedSoundRef.current!.pauseAsync();
+      setPlaying(false);
+    } catch (error) {
+      console.error('Failed to pause sound:', error);
+    }
+  };
```
ErikSin: Sorry, I did not catch this in my last review, but these should also be memoized with useCallback. Generally I think it is good practice to memoize any function or values being exported from a hook

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -0,0 +1,241 @@
...
+              onPress={handleReturnToEditor}
+              variant="outlined"
+              color="ComapeoBlue">
+              {t(m.returnToEditorButtonText)}
+            </Button>
+            <Button fullWidth onPress={handleRecordAnother}>
+              {t(m.recordAnotherButtonText)}
+            </Button>
+          </View>
+        </View>
```
ErikSin: Is there a reason that `<BottomSheetModalContent />` is not being used here?

ErikSin (changes requested):
when i delete a recording, the bottom sheet is not closing

cimigree:
> when i delete a recording, the bottom sheet is not closing

Hm. I can't figure out why that would be, can't replicate.

Here is what it looks like on my Pixel. Is this what you are talking about?

https://github.com/user-attachments/assets/9de7f5d2-9041-40e4-9be7-551f5615db4f

ErikSin:
> > when i delete a recording, the bottom sheet is not closing
> 
> Hm. I can't figure out why that would be, can't replicate.
> 
> Here is what it looks like on my Pixel. Is this what you are talking about?
> deletionvideo.mp4

yeah, i can recreate it sometimes....im just trying to figure out why its only happening sometimes

ErikSin:
> > > when i delete a recording, the bottom sheet is not closing
> > 
> > 
> > Hm. I can't figure out why that would be, can't replicate.
> > Here is what it looks like on my Pixel. Is this what you are talking about?
> > deletionvideo.mp4
> 
> yeah, i can recreate it sometimes....im just trying to figure out why its only happening sometimes

hmm, i think i figured how when (but not why) it is happening. If you close the app while taking an observation, and then open the app again, it will open on the `ObservationCreate` screen - which is expected behaviour. But from here if you take an audio clip and try to delete it the bottom sheet will stay open

ErikSin:
https://github.com/user-attachments/assets/7a4e3e3e-e46e-4a0e-8901-6b98482c660c

cimigree:
> > > > when i delete a recording, the bottom sheet is not closing
> > > 
> > > 
> > > Hm. I can't figure out why that would be, can't replicate.
> > > Here is what it looks like on my Pixel. Is this what you are talking about?
> > > deletionvideo.mp4
> > 
> > 
> > yeah, i can recreate it sometimes....im just trying to figure out why its only happening sometimes
> 
> hmm, i think i figured how when (but not why) it is happening. If you close the app while taking an observation, and then open the app again, it will open on the `ObservationCreate` screen - which is expected behaviour. But from here if you take an audio clip and try to delete it the bottom sheet will stay open

@ErikSin For me, that is not happening on either of my devices even when I follow the steps you do exactly, unless while in an observation, I actually force close the app and then **restart expo** and then reopen the app. And in that case, that lack of ability to close the sheet doesn't even always happen... I noticed you have the expo development screen when you restart, which I do not. I am thinking this might not happen in the real world because expo won't be restarting mid go?
And because of this I sort of timeboxed this. I worked for 1.5 hours and this is what I came up with and I don't think it really works for the particular situation where expo reloads ... If you know something else to do, I would love to hear it. Or I can completely redo the bottomsheetmodal but I have a feeling that is not a good idea.

ErikSin:
Could you build an APK and confirm your theory? :)

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -1,40 +1,56 @@
...
+
+  if (currentState === 'done') {
+    return (
+      <RecordingDone
+        uri={uri || ''}
+        duration={status?.durationMillis || 0}
+        onDelete={() => {
+          reset();
+        }}
+        onRecordAnother={reset}
```
ErikSin: onDelete and onReset are now the same thing. It would just be clearer is reset was just a prop of RecordingDone and just pass reset once

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -0,0 +1,235 @@
...
+                bold: message => (
+                  <Text style={{fontWeight: 'bold'}}>{message}</Text>
+                ),
+              }}
+            />
+          </Text>
+        </View>
+      );
+      return (
+        <View style={{marginTop: 80}}>
```
ErikSin: Also instead of wrapping this, you can just add a `marginTop` of 80 directly to the `SuccessIcon`

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -1,40 +1,51 @@
...
+import {RecordingDone} from './RecordingDone';
+import {RecordingIdle} from './RecordingIdle';
+import {useAudioRecording} from './useAudioRecording';
 
 export function CreateRecording() {
-  const navigation = useNavigationFromRoot();
+  const {startRecording, stopRecording, reset, status, uri} =
+    useAudioRecording();
+
+  const currentState = status?.isRecording ? 'active' : uri ? 'done' : 'idle';
```
ErikSin: nit: can you give this a more meaningful name, perhaps `recordingState`?

src/frontend/screens/Audio/PermissionAudioBottomSheetContent.tsx
```diff
@@ -37,48 +36,38 @@ const m = defineMessages({
   },
 });
 
-type ObservationCreateNavigationProp =
-  NativeRootNavigationProps<'ObservationCreate'>['navigation'];
-
 interface PermissionAudioBottomSheetContentProps {
   closeSheet: () => void;
-  navigation: ObservationCreateNavigationProp;
+  setPendingAction: (action: 'navigateToAudio' | null) => void;
```
ErikSin: i think it would be better to pass a function that handles the entire action (since there is only one action that we want), instead of passing the entire setter. Also see comment below about changing the setState to handle a boolean instead of the type `'navigateToAudo' | null`

eg.
instead of `setPendingAction: (action: 'navigateToAudio' | null) => void` do `setNavigateToAudioTrue: ()=>void`

and then in the parent component
```ts
<PermissionBottomSheetContent setNavigateToAudioTrue={()=>{setShouldNavigateToAudio(true)}}  {...rest} />
```

src/frontend/screens/Audio/index.tsx
```diff
@@ -6,6 +6,8 @@ import {CustomHeaderLeft} from '../../sharedComponents/CustomHeaderLeft';
 import {StatusBar} from 'expo-status-bar';
```
ErikSin: this `StatusBar` is causing a flicker in the screen when you navigate back. Notice in the video how the screen jumps up and down when navigating between the screen. I think we can just get rid of it completely. I also think that `Audio/CreateRecording/index` should become the new `Audio/index`

https://github.com/user-attachments/assets/7135013d-a4d6-4975-a015-2793dcaf092a

src/frontend/sharedComponents/ActionRow.tsx
```diff
@@ -51,6 +51,10 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
     openOnMount: false,
   });
 
+  const [pendingAction, setPendingAction] = useState<'navigateToAudio' | null>(
+    null,
+  );
```
ErikSin: I think it is clearer to use a boolean instead of creating 2 opposing states. Eg `const [shouldNavigateToAudio, setShouldNavigateToAudio] = useState(false)`

ErikSin (changes requested):
Sorry, i just noticed the use of `statusBar`. the SWM team was using them, and it was causing a flickering, so I think we should just remove them entirely

src/frontend/screens/Audio/index.tsx
cimigree: @ErikSin, There is going to be something else in Audio/index in the next PR related to Playback from the Observation. Is it ok if I leave it as is? Not the Status Bar... Having the two files

ErikSin (approved):
As a follow up, lets add the metadata to the blob

PR #767, opened 2024-10-02
chore: updated copy

### LINKED ISSUE #766: Sync Screens/ Update Copy
Referencing this issue : https://www.notion.so/digidem/Sync-Device-discovery-inconsistent-10c1b08162d580099a26fdeb94138c5c

When syncing with more than 3 devices, we notice the number of “discovered devices” is not the same across all devices, even if on the same project and connected. Ultimately the end result of syncing is successful.

## To do
Revisit the sync screens by updating the copy 

- [ ] Label Devices found (under name of project) : If no  “No devices found” 
- [ ] Label Devices found (under name of project) : if devices found “Devices found”
- [ ] Before syncing : “Devices available"
- [ ] While syncing : “You are syncing with your team”
- [ ] Complete: “Complete! Waiting for other devices to join”
- [ ] Progress bar - remove "X out of X devices"

https://www.figma.com/design/uzynHXR0xgJu19lsCoJyXN/%F0%9F%90%9A-Shell-App?node-id=2596-796

ErikSin:
closes #766 

Updates the copy so user does not see the amount of devices available as there is a bug in the count. 

image

image

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -1313,16 +1313,13 @@
     "message": "All data synced"
   },
   "screens.Sync.ProjectSyncDisplay.devicesAvailableToSync": {
-    "message": "{count} {count, plural, one {device} other {devices}} available"
+    "message": "Devices available"
   },
   "screens.Sync.ProjectSyncDisplay.devicesFound": {
-    "message": "{count} {count, plural, one {Device} other {Devices}} found"
+    "message": "Devices found"
   },
   "screens.Sync.ProjectSyncDisplay.noDevicesAvailableToSync": {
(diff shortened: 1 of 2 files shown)


cimigree (approved):
Copy looks good and I tested it out and it made sense and seemed to work. It doesn't say Remove x out of x devices?
Just wondering, Are you leaving all of the variables so that they can be dealt with later? I didn't have any trouble with speed but I have only two fast devices...

ErikSin:
I have take out the unused variables! Thanks for the reminder

Issue #625, opened 2024-08-26
Coordinator / Adding Server errors


ximenabb:
**Description**
Coordinator has clicked on **Add Server** and see below possible errors if the server is not responding or in use.
Use this template for other possible errors

**Todo**

## Screen 1 - Adding url - error

<img width="281" alt="Screen Shot 2024-09-18 at 12 34 46 PM" src="https://github.com/user-attachments/assets/2d7c169d-2b5f-46aa-939f-0eb59fc81d16">

- [ ] Use usual design patterns when URL incorrect

## Screen 2 - Adding a url - Device offline
<img width="277" alt="Screen Shot 2024-09-18 at 12 35 59 PM" src="https://github.com/user-attachments/assets/d871d7d2-5ef8-4f38-aa85-60e07dc9eb39">

- [ ] Display screen as designed
- [ ] Use copy _No Internet Access_
- [ ] Use copy _Your device is offline. Connect to the Internet to add a server_
- [ ] Add a  **Go Back**  button

## Screen 3 - Unable to connect

<img width="293" alt="Screen Shot 2024-09-18 at 12 38 28 PM" src="https://github.com/user-attachments/assets/2acb84f2-3c68-4b43-9b51-475e259e7277">

- [ ] Display screen as designed
- [ ] Use copy _Unable to connect  to Server_
- [ ] _The server may not be configured correctly or there might be another issue._ 
- [ ] _Go back and try again._ 
- [ ] Add a  **Go Back**  button

## Screen 4 - Server already in use
<img width="283" alt="Screen Shot 2024-09-18 at 12 39 25 PM" src="https://github.com/user-attachments/assets/a4d62ec6-2086-4f95-bcd8-1b1f57dbdb27">

- [ ] Display screen as designed
- [ ] Use copy _Server already in use_
- [ ] _Go back to add a different server. _ 
- [ ] Add a  **Go Back**  button

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=2909-1931&t=3MJwxcx0iGMHqAbR-1

ximenabb:
Updated figma link

PR #771, opened 2024-10-02
chore: update icon on error bottom sheet

ErikSin:
After discussion with design team, it was determined that the error icon should be switched to just the red exclamation point icon.
image

### DIFF
--- a/src/frontend/sharedComponents/ErrorBottomSheet.tsx
+++ b/src/frontend/sharedComponents/ErrorBottomSheet.tsx
@@ -8,7 +8,7 @@ import {
   BottomSheetModal,
   useBottomSheetModal,
 } from './BottomSheetModal';
-import {LogoWithErrorIcon} from './LogoWithErrorIcon';
+import ErrorIcon from '../images/Error.svg';
 
 const m = defineMessages({
   somethingWrong: {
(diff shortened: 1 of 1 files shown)


ErikSin:
I also added the ability to add custom messages to the error bottom sheet which is needed for #625

src/frontend/sharedComponents/ErrorBottomSheet.tsx
```diff
@@ -74,8 +76,9 @@ export const ErrorBottomSheet = (props: ErrorModalProps) => {
       onDismiss={closeSheet}
       isOpen={isOpen}>
       <BottomSheetModalContent
-        icon={<LogoWithErrorIcon />}
-        title={formatMessage(m.somethingWrong)}
+        icon={<ErrorIcon style={{marginTop: 80}} />}
+        title={formatMessage(title || m.somethingWrong)}
+        description={description ? formatMessage(description) : undefined}
```
cimigree, line 81: It looks like from the other Issue the description is supposed to be 16 pixels. Might need to set up a ternary for the styles here as well.

cimigree (changes requested):
Looks good.

Just adding a note that it may make sense down the line to get rid of the LogoWithErrorIcon and FatalError because I am guessing the ServerLoading error will be replaced by #625 (not a change request...)

Change request is about that the description should be smaller font to be in line with designs.

src/frontend/sharedComponents/ErrorBottomSheet.tsx
ErikSin: I updated the font size (as well as the icon size to match the design system):
image

https://github.com/digidem/comapeo-mobile/pull/771/commits/077e8ab254ae86f503796c235e2cee57d48a957a

cimigree (approved):

PR #778, opened 2024-10-03
fix: clears observation refs when clearing tracks

### LINKED ISSUE #634: Observations from previous track included in new track
**Describe the bug**
Observations from a previously completed track are incorrectly included in a newly created track.

**To Reproduce**
1. Start a Track
2. Add an observation
3. End the Track
4. Start another Track
5. Add an observation
6. End the Track
7. Click into the observation list
8. Click on the 2nd Track
9. You will see that the 2nd track contains the observation from the first track and from the 2nd track too

**Expected behavior**
Only the observations created during the track should be displayed with it

**Screenshots**
---
**Observation List with both Tracks and all observations placed correctly**
<image src="https://github.com/user-attachments/assets/34ead4b6-66c9-4ba8-9c38-73791ae34590" width="250" />

---
**First Track with correct Observation list**
<image src="https://github.com/user-attachments/assets/ddd0a38c-c118-4fbc-af8d-1c97a9f7bffe" width="250" />

---
**Second Track with all observations**
<image src="https://github.com/user-attachments/assets/03e1f7d5-6a75-48e6-8ae7-3ea2839112dd" width="250" />

---
 - Device: Samsung Galaxy A25 5G
 - OS: 
 - Browser N/A
 - Version: develop branch
 
 **Additional Info**
 The observations are coming from this function:
```
   const trackObservations = observations.filter(observation =>
    track.observationRefs.some(ref => ref.docId === observation.docId),
  );
```
From this file:
`src/frontend/screens/Track/index.tsx`

ErikSin:
fixes #634 
Properly clears the observation refs when clearing a track. Previously was resetting `observation` to and empty array but `observationRefs` needs to be reset instead. This was due to a schema change and the front end did not update properly.

### DIFF
--- a/src/frontend/hooks/persistedState/usePersistedTrack.ts
+++ b/src/frontend/hooks/persistedState/usePersistedTrack.ts
@@ -72,7 +72,7 @@ export const tracksStore = createPersistedStore<TracksStoreState>(
           trackingSince: null,
           distance: 0,
           isTracking: false,
-          observations: [],
+          observationRefs: [],
           description: '',
         };
       }),
--- a/src/frontend/hooks/persistedState/usePersistedTrack.ts
+++ b/src/frontend/hooks/persistedState/usePersistedTrack.ts
@@ -66,15 +66,13 @@ export const tracksStore = createPersistedStore<TracksStoreState>(
         };
       }),
     clearCurrentTrack: () =>
-      set(() => {
-        return {
-          locationHistory: [],
-          trackingSince: null,
-          distance: 0,
-          isTracking: false,
-          observationRefs: [],
-          description: '',
-        };
+      set({
+        locationHistory: [],
+        trackingSince: null,
+        distance: 0,
+        isTracking: false,
+        observationRefs: [],
+        description: '',
       }),
     setTracking: (val: boolean) =>
       set(() =>


achou11 (approved):
A little concerned about why a type error was not surfacing before this change 🤔 good catch though!

achou11:
> A little concerned about why a type error was not surfacing before this change 🤔 good catch though!

ah, it's because of how the `set()` call was being used in there. if you convert it from

```ts
set(() => {
  return {
    locationHistory: [],
    trackingSince: null,
    distance: 0,
    isTracking: false,
    observations: [], // should error but doesn't!
    description: '',
  };
})
```

to 

```ts
set({
  locationHistory: [],
  trackingSince: null,
  distance: 0,
  isTracking: false,
  observations: [], // properly errors here
  description: '',
})
```

a type error appears as expected. might want to fix that in this PR to prevent future mistakes

ErikSin:
> A little concerned about why a type error was not surfacing before this change 🤔 good catch though!

~~yeah...i have a feeling its all the typescript generic fandaggaling that is done when [creating the function createPersistedState](https://github.com/digidem/comapeo-mobile/blob/6dbdd9d5676737be12138dea134503cf1adf0bf9/src/frontend/hooks/persistedState/createPersistedState.ts#L37)....~~

oh, jk i just saw your other comment

PR #765, opened 2024-10-02
feat: Adding Remote Archive

ErikSin:
closes #621 
closes #625
closes #627 

Adds the ability for user to add a Remote Archive (formerly known as Cloud Server)

 ** To Note. this wont be merged into develop as is. We are creating this branch to create a build for COP 16
 
 Also 13 of the "files changed" are just dependencies and tests, so this PR is not as big as it seems

image

image

image

image

image

image

PR #779, opened 2024-10-07
chore: add util to normalize remote archive URLs

EvanHahn:
_This is meant to be merged into #765._

This adds a new function, `normalizeRemoteArchiveUrl`. When the user types a URL, we should run it through this normalizer before passing it to the backend.

Some quick examples for how it works:

```javascript
normalizeRemoteArchiveUrl("example.com")
// => "https://example.com/"

normalizeRemoteArchiveUrl("100.64.0.42:1234")
// => "https://100.64.0.42:1234/"

normalizeRemoteArchiveUrl("https://example.com/path")
// => "https://example.com/path/"

normalizeRemoteArchiveUrl("invalid URL")
// Error

normalizeRemoteArchiveUrl("http://incorrect-protocol.example")
// Error
```

This isn't integrated into the frontend yet, but should be once we start adding remote archives.


src/frontend/lib/ipAddress.test.ts
EvanHahn, line 1: I don't think it's worth reviewing `ipAddress.ts` or `ipAddress.test.ts` too closely, because they're mostly copy-pasted from Node's source code.

EvanHahn:
CI failures seem unrelated to this change.

PR #782, opened 2024-10-08
feat: audio error handling

cimigree:
### Description
Adds error handling for audio
Uses the Error Bottom Sheet as it is perfect for the needs here

### Instruction
Do some audio stuff and make sure everything works properly
Throw some errors...
For example, in useAudioRecording.ts in startRecording do 
```
    try {
      throw new Error('Simulated error in startRecording');

      const {recording: audioRecording} = await Audio.Recording.createAsync(
        Audio.RecordingOptionsPresets.HIGH_QUALITY,
        stat => setStatus(stat),
      );
      setRecording(audioRecording);
    } catch (err) {
      const error = normalizeError(err);
      setError(error);
    }
  }, [setRecording, setStatus, setError]);
```
Then go to make an audio recording and you will see the error.
Try it with and without a message to see each option
Another one where you can see an actual error...
Make a recording. Go to play it back...
In Playback.tsx
modify to this:
export function Playback({
  uri,
  leftControl,
  rightControl,
}: {
  uri: string;
  isModalOpen?: boolean;
  leftControl?: ReactNode;
  rightControl?: ReactNode;
}) {
  uri = 'https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3';
  const {
    duration,
    currentPosition,
    isPlaying,
    stopPlayback,
    startPlayback,
    error,
    clearError,
  } = useAudioPlayback(uri);
  
  You will see an error that looks like this:
<image src="https://github.com/user-attachments/assets/db8fefcd-c889-47ae-be30-125961493019" width="250" />


src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -26,26 +28,43 @@ export function CreateRecording() {
...
+      <ErrorBottomSheet
+        error={error}
+        description={
+          error?.message
+            ? ({
+                id: 'error.dynamic',
+                defaultMessage: error.message,
+              } as MessageDescriptor)
+            : undefined
+        }
```
ErikSin: I don't think we should add an error message. This is not translatable and any messages will need to be approved by codesign first

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -26,26 +28,43 @@ export function CreateRecording() {
...
+      )}
+      {recordingState === 'done' && (
+        <RecordingDone
+          uri={uri || ''}
+          duration={status?.durationMillis || 0}
+          reset={reset}
+        />
+      )}
+      <ErrorBottomSheet
+        error={error}
```
ErikSin: I think we should throw an error if the `recordingState` is done and there is no uri:

```ts
if(recordingState === 'done' && !uri){
    setError(new Erorr('no uri'))
}

return (...rest)
```

ErikSin (changes requested):
Small changes, but otherwise looks good

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -26,26 +35,41 @@ export function CreateRecording() {
...
-        duration={status?.durationMillis || 0}
-        onPressStop={stopRecording}
-      />
-    );
-  }
+  useEffect(() => {
+    if (recordingState === 'done' && !uri) {
+      setHasError(true);
+    }
+  }, [recordingState, uri, setHasError]);
```
ErikSin: i think we should take out this `useEffect`. Better alternative:

```ts
if(!error && recordingState === 'done' && !uri){
     setHasError(true)
}

```

src/frontend/screens/Audio/CreateRecording/useAudioRecording.ts
```diff
@@ -5,31 +5,56 @@ export function useAudioRecording() {
...
-    );
-    setRecording(audioRecording);
-  }, [setRecording, setStatus]);
+    try {
+      const {recording: audioRecording} = await Audio.Recording.createAsync(
+        Audio.RecordingOptionsPresets.HIGH_QUALITY,
+        stat => setStatus(stat),
+      );
+      setRecording(audioRecording);
+      setHasError(false);
```
ErikSin: why do we need to setHasError to false here?

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -26,26 +35,41 @@ export function CreateRecording() {
...
+        />
+      )}
+      <ErrorBottomSheet
+        error={hasError ? new Error('An error occurred') : null}
+        clearError={clearError}
+        tryAgain={
+          recordingState === 'idle'
+            ? startRecording
+            : recordingState === 'active'
+              ? stopRecording
```
ErikSin: If it throws an error while recording, to reset, the user will call stop recording, which will set the recoding status to done which i think is undesired behaviour.

I think regardless of the state, the "try again" should reset the the recording, and have the user start again.

ErikSin (changes requested):
as i think about this i think it would make sense NOT to pass a clearError, and have the user just use the reset. If there is an error at any time during the recording, we just want to reset and have the user retry. That requires us to reset(), which encapsulates setting the error to false.

I think it also means we shouldnt pass anything to `tryAgain` in the ErrorBottom Sheet, because the goBack button will call clear error and the clear error wil call the reset, so the try again and go back will be the same.

Also, sorry this was not said before, but I think passing the Error as an Error, and not a boolean, is still the best practice. Right now we dont have a customized error message, but in the future we might want to. So passing the error to the bottom sheet is something we still want to do, but we just dont want to expose the error message right now.

cimigree:
@ErikSin I didn't do the reset thing in Playback. Wasn't sure if that was warranted there as the error might just be a matter of loading timing for example and resetting would involve clearing out the uri.

ErikSin (approved):
Looks good!

PR #783, opened 2024-10-08
feat: set preset language to match app language

ErikSin:
Sets the language of presets based on app language.

### DIFF
--- a/src/frontend/hooks/server/fields.ts
+++ b/src/frontend/hooks/server/fields.ts
@@ -1,15 +1,17 @@
 import {useQuery} from '@tanstack/react-query';
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
+import {usePersistedLocale} from '../persistedState/usePersistedLocale';
 
 export const FIELDS_KEY = 'fields';
 
 export const useFieldsQuery = () => {
   const {projectId, projectApi} = useActiveProject();
+  const lang = usePersistedLocale(store => store.locale);
 
   return useQuery({
-    queryKey: [FIELDS_KEY, projectId],
+    queryKey: [FIELDS_KEY, projectId, lang],
     queryFn: async () => {
-      return projectApi.field.getMany();
+      return projectApi.field.getMany({lang});
     },
   });
 };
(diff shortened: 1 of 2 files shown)


achou11 (commented):
From my understanding, the `lang` option is available for _all_ datatype method calls e.g. `getByDocId()`, `getByVersionId()`, etc. Shouldn't we update all of the relevant queries to specify this? Otherwise we would get inconsistencies in the UI based on which queries are being used.

ErikSin:
> From my understanding, the `lang` option is available for _all_ datatype method calls e.g. `getByDocId()`, `getByVersionId()`, etc. Shouldn't we update all of the relevant queries to specify this? Otherwise we would get inconsistencies in the UI based on which queries are being used.

I believe that this is the only query. we don't ever call `preset.getByDocId()` or `preset.getByVersionId()` anywhere in front end.

achou11:
> I believe that this is the only query. we don't ever call `preset.getByDocId()` or `preset.getByVersionId()` anywhere in front end.

Ah ok. Is it okay that only the presets are returning the translated versions (e.g. we don't need to do something similar for fields)?

ErikSin:
> > I believe that this is the only query. we don't ever call `preset.getByDocId()` or `preset.getByVersionId()` anywhere in front end.
> 
> Ah ok. Is it okay that only the presets are returning the translated versions (e.g. we don't need to do something similar for fields)?

hmmm good question...ill look into that

ErikSin:
ok! yeah, it looks like the fields can be translated!

achou11:
> ok! yeah, it looks like the fields can be translated!

just to be clear, _all_ datatypes (e.g.  observation, track, field, and preset) can technically be translated i.e. any call to `getByDocId()`, `getByVersionId()`, and `getMany()` accept that `lang` option, hence the first question I had. I think for the purposes of the config-related UI, getting the translated fields and presets is sufficient, but probably need a follow up to update other relevant datatype queries to do the same thing.

achou11 (approved):
Should create a follow-up issue to pass in that lang option to other relevant queries

PR #785, opened 2024-10-10
fix: Fix changing preset & store `presetRef` on observation

gmaclennan:
This fixes two issues related to editing a preset for an observation. I consider them blocking bugs because:

1. The bug when changing a preset results in the loss of user data (e.g. loss of data that a user has entered under observation details).
2. The bug of not storing a `presetRef` means that if a user changes presets, there is no way of linking existing observations back to their original preset. This will result in loss of information if presets are updated in the future.

The fix required storing the selected preset on the draft observation state. There are more elegant ways to fix this, but they would require more refactoring of code. My goal here was to minimize changes while fixing the bug.

### Changing a preset

Changing a preset on an observation should:

1. Remove any tags which were applied by the previous preset.
2. Keep all tags that were not applied by the previous preset (e.g. tags that were edited by the user by filling in fields, but also any other tags)
3. Remote any tags defined in the `removeTags` property of the new preset.

This was the [logic in Mapeo Mobile (legacy)](https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/screens/CategoryChooser.tsx#L91-L104) (except we did not respect the `addTags` and `removeTags` preset properties).

Prior to this change, the [updatePreset logic](https://github.com/digidem/comapeo-mobile/blob/bcca1fb716d75a765c8f05c4dcf878edad6475e0/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts#L157-L163) tried to only keep tags that were defined by fields in the new preset, and discarded all other tags, which include tags from users entering data in fields from the previous preset. In addition, the logic was incorrect because it was matching fieldIds to tag keys. This results in loss of user data when editing a category, which is not what we want.

### Storing a preset

We should be storing the preset that was used to edit an observation on `observation.presetRef`. We normally try to match a preset based on tags, so that if a preset is updated, we match and existing observation to the new preset. However, the `presetRef` acts as a fallback and also a way of auditing an observation's history, so that you can know which preset was originally used (including the exact field definitions).

This presetRef information was not saved in the current code. This PR adds this functionality.

### DIFF
--- a/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
+++ b/src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
@@ -26,6 +26,7 @@ export type DraftObservationSlice = {
   audioRecordings: [];
   value: Observation | null | ClientGeneratedObservation;
   observationId?: string;
+  preset?: Preset;
   actions: {
     addPhotoPlaceholder: (draftPhotoId: string) => void;
     replacePhotoPlaceholderWithPhoto: (draftPhoto: DraftPhoto) => void;
@@ -75,6 +76,7 @@ const draftObservationSlice: StateCreator<DraftObservationSlice> = (
         audioRecordings: [],
         value: null,
         observationId: undefined,
+        preset: undefined,
       });
     },
     updateObservationPosition: props => {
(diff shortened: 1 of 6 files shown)


gmaclennan:
I have not yet tested this on a device.

src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
```diff
@@ -184,7 +206,8 @@ export const usePersistedDraftObservation = createPersistedState(
 export const usePreset = () => {
   const {data: presets} = usePresetsQuery();
   const tags = usePersistedDraftObservation(store => store.value?.tags);
-  return !tags ? undefined : matchPreset(tags, presets);
+  const preset = usePersistedDraftObservation(store => store.preset);
+  return preset || !tags ? undefined : matchPreset(tags, presets);
```
achou11: don't remember operator precedent rules off the top of my head so just checking: is this the intention?

`preset || (!tags ? undefined : matchPreset(...))`

achou11 (approved):
Noting that I haven't tested on a device yet either, but code changes seem good to me

src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
gmaclennan: Good point, I re-wrote this so that the intention is clearer.

gmaclennan:
I have tested this on a device and identified some issues. Update pending.

gmaclennan:
Unfortunately with the current way the code is written I can't think a way of writing more tests for this. Instead I have tested this by writing console.log messages and verifying the changes were as-expected:

1. New observations have a `presetRef` defined.
2. When you change the preset on an existing observation, it removes tags from the previous preset and applies tags from the new one.

The code would become cleaner and less prone to bugs if the creation of a draft from an existing observation was moved from `useEffect` on the edit screen to the handler for the user pressing the edit button. That can be addresses in follow-up cleanup.

achou11 (approved):
Thanks for the fixes despite the tangle!

PR #795, opened 2024-10-14
chore: Changes necessary to run end to end tests.

cimigree:
Some small but tedious changes necessary to run the end to end tests and have them reflect some of the changes in the app.


ErikSin (approved):
none of us can really test this, but it doesn't effect the actual code so I think you can just yolo merge it

PR #796, opened 2024-10-15
chore: update core and ipc to v2.0.0

achou11:
Towards #622, #623, and #624

- Updates `@comapeo/core` and `@comapeo/ipc` to `2.0.0`.
- App changes:
  - **Removes support for embedded static maps**. This is related to inserting an asar-formatted file into a highly specific directory on the android device, which is the "legacy" approach for custom maps inherited from Mapeo. This was more of a stop-gap feature that wasn't expected to be heavily used anyways, and we made the decision to drop support for this entirely in core. #318 will introduce an improvement to that feature that is much more reliable and easier to use.
  - Removes some map server boilerplate code since that is now handled within the `MapeoManager` itself.

---

Online default style:

image

Offline fallback:

image


src/backend/package-lock.json
achou11, line 1: confirming that the `hypercore` version does not unexpectedly change here (`cd src/backend && npm list hypercore`):

```sh
mapeo-mobile-node-next@1.0.0 /Users/andrewchou/GitHub/digidem/comapeo-mobile/src/backend
└─┬ @comapeo/core@2.0.0
  ├─┬ corestore@6.8.4
  │ └── hypercore@10.17.0 deduped
  └── hypercore@10.17.0
```

src/backend/patches/@comapeo+core+1.0.1+001+fastify-controller-force-close.patch
achou11, line 1: No longer needed since updated core includes this (see https://github.com/digidem/comapeo-core/commit/d4b9904e22ea1ef445d2df0fbfe6b79f051abe2b)

src/backend/patches/@electron+asar+3.2.13+001+remove-original-fs-require.patch
achou11, line 1: no longer needed since updated core removes support for the static maps approach, which relied on using asar files (see https://github.com/digidem/comapeo-core/commit/f21a675e992bf1bb908f5a938adcd4b2f4c2efac#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L157)

src/backend/src/app.js
```diff
@@ -58,47 +52,27 @@ process.on('exit', (code) => {
...
  * @param {string} options.defaultConfigPath
- * @param {string} options.fallbackMapPath Path to app-specific external file storage folder
- *
  */
 export async function init({
   version,
   rootKey,
   migrationsFolderPath,
-  sharedStoragePath,
+  // sharedStoragePath,
```
achou11: opted to just comment this out as opposed to remove since we'll need to use it for supporting custom maps in a follow-up that's coming soon

src/backend/src/app.js
```diff
@@ -58,47 +52,27 @@ process.on('exit', (code) => {
...
   defaultConfigPath,
-  fallbackMapPath,
 }) {
   log('Starting app...')
   log(`Device version is ${version}`)
 
   const privateStorageDir = rnBridge.app.datadir()
   const dbDir = join(privateStorageDir, DB_DIR_NAME)
   const indexDir = join(privateStorageDir, CORE_STORAGE_DIR_NAME)
-  const staticStylesDir = join(sharedStoragePath, 'styles')
```
achou11, line 78: the removal of this means that devices that have installed the app previously will have a lingering directory, which shouldn't be a problem but might be something worth documenting somehow?

package.json
```diff
@@ -28,9 +28,8 @@
   },
   "dependencies": {
     "@bam.tech/react-native-image-resizer": "^3.0.7",
-    "@comapeo/ipc": "1.0.1",
+    "@comapeo/ipc": "2.0.0",
     "@dev-plugins/react-navigation": "^0.0.6",
-    "@dr.pogodin/react-native-fs": "^2.24.1",
```
achou11, line 33: no longer used at all given the removal of the external shared directory path

src/backend/src/app.js
```diff
@@ -56,49 +50,27 @@ process.on('exit', (code) => {
  * @param {string} [options.version] Device Version
  * @param {Buffer} options.rootKey
  * @param {string} options.migrationsFolderPath
- * @param {string} options.sharedStoragePath Path to app-specific external file storage folder
```
achou11, line 59: we no longer need this option because we're strictly going to use the app's "internal" storage - AKA the document directory - to store the custom maps. The path to this directory is already available via nodejs-mobile e.g. `const privateStorageDir = rnBridge.app.datadir()`

The setup for this will be done in the other PR

achou11 (commented):
While working on #758 to build on top of this PR, I realized that we can get rid of some more boilerplate + complexity. See 48a919397e63ffc081c65139ff908b36247b1c40 for relevant changes.

cimigree (approved):
Passes end to end tests but that is all I can speak to.

PR #798, opened 2024-10-15
feat: audio playback, edit, delete

### LINKED ISSUE #361: Playback / Delete / Share
## Playback / Delete / Share
User has successfully saved an audio recording on a observation. User clicks on the recording from the observation. User can play it back, delete it or share it.

## Todo
Screen Shot 2024-05-15 at 4 25 21 PM

**Playback**
- [ ] Task is already here https://github.com/digidem/comapeo-mobile/issues/360
- [x] Only difference is, tapping the “X” button takes user to the Observation

**Delete Audio Recording**
- [x] Display bottom sheet as designed
- [x] Display red icon and _Delete?_
- [x] Display text _Your Audio Recording will be permanently deleted. This cannot be undone._ 
- [x] Display `Delete` button, when clicked redirects to the Observation
- [x] Display `Cancel` button, when clicked close the modal

**Share**
- [ ] **No** creating a custom share sheet (we need to update figma)
- [ ] When tap share, use android share sheet

## Design
https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/Tracks-and-Audio-(Share)?node-id=529-657&t=xhcTH3fFF4S8gJjD-0

cimigree:
### Description
closes #361 
- Adds the ability to see audio thumbnail(s) on an existing observation or when editing an observation
- Thumbnail is pretty basic right now because we don't have access to the duration and created at from the backend yet but I left some placeholders to guide future develop
- Can click on the thumbnail and listen to it from editing an observation or just when viewing an observation
- Adds the ability to Edit an existing observation in regard to audio. Can add another audio or delete an audio file when you are editing an observation (but only when editing -- not when viewing)
- Adds the ability to Share an audio file through Android sharing (as specified in the ticket, different from the design docs)
- I made some choices how best to handle the different shape of the audio gotten from the backend and how and where to transform it. I did not want to add multiple files and much more complexity to handle processing audio in components as is done with photos. Audio files are much more simple and it seemed unnecessary to add so much complexity.
- I also decided to pass a few parameters instead of adding context. Mostly just to know if someone is editing a file which will determine how some navigation is done and whether or not to show the delete button.


src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -75,6 +103,12 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> = ({
 
     async function createDraftFromExistingObservation(docId: string) {
       const observation = await projectApi.observation.getByDocId(docId);
+      const audioAttachments =
+        observation.attachments.filter(
+          attachment => attachment.type === 'audio',
+        ) || [];
+      setAudioAttachments(audioAttachments);
+
```
ErikSin, line 111: Thanks for the work you put into. I am a little worried that the logic is a bit challenging to follow. If I understand correctly,

1. you are pulling attachment of type `audio` and setting it to state as `AudioAttachment` on line 106
2. on every render, all audio attachments are processed, with `useAudioProcessing` to return `audioWithUri` on line 83
3. On every change of `audioWithUri` you run a useEffect on line 85
4. The useEffect stringifies the `audioWithUrl`and the state object `processedAudioRecordings` and comparest them. if they are not the same, you are setting the`processedAudioRecordings` to the Audio with URL
5. The you set the `processedAudioRecording` as a prop to `existingObservationToDraft` on line 124

I think we could simplify it and make it more maintainable for everyone. Would you be open to pairing up on this, or I can provide async feedback if you would prefer?

cimigree: @ErikSin Yes it is confusing. Sorry. I think I have a more clear solution and am working on a little refactor. Still not quite fixed but I will push up another try soon. If that is no good, I think pairing might be more efficient than async if we can find an overlapping time.

ErikSin (changes requested):
Thanks for clearing up that logic. It makes a lot more sense now.

However, I now see that the media thumbnail has 2 separate architectures to do similar thing. 
1. The photos thumbnails take an `attachment`, and loads each url separately inside the component, while
2.  The audio is loading ALL the urls beforehand and then passing it to the thumbnails. 

While your way is not wrong, I think for consistency we should follow the same pattern as the photos.

If we follow the same pattern as the photos we also wouldn't have to create the attachments separately and pass them to the `existingObservationToDraft()`. We could just derive it from the observation. Therefore were not dealing with 2 separate states that represent the same thing

Issue #768, opened 2024-10-02
Accessibility Improvements


ximenabb:
## 1 Update input across the app
Inputs throughout are not aligning to our design guidelines which have been reviewed for visual accessibility, specifically in line weight and color. Marking this as a correction. (SF, Sept 24). Broader fix for all inputs (including text fields), everywhere.
Notion issue : https://www.notion.so/digidem/Correction-Update-input-across-the-app-867d8fd78cba406cac423a8c2a0719c8

## 2  Correct colors for accessibility on orange buttons
Notion issue : https://www.notion.so/digidem/Correction-Correct-colors-for-accessibility-on-orange-buttons-d94e351785dd4188bc110a7c761b002c

PR #803, opened 2024-10-21
chore: update orange icons to accessible color

ErikSin:
Updates the add button and sync button to use an accessbile color (also updates both of them to use SVGs).

There also was a duplicate export of the `SyncIcon` so I deleted the file, and updated the import to come from icons/index

fixes (partially) #768


src/frontend/sharedComponents/AddButton.tsx
```diff
@@ -27,16 +26,9 @@ const AddButtonNoMemo = ({
...
     <TouchableOpacity disabled={disabled || isLoading} onPress={onPress}>
-      {!isLoading ? (
-        <Image
-          source={require('../images/add-button.png')}
-          style={styles.button}
-        />
-      ) : (
-        <Loading size={15} />
-      )}
+      {isLoading ? <UIActivityIndicator size={50} /> : <AddButtonSVG />}
```
ErikSin, line 31: I also updated this loader. This was a legacy thing before we were using UiActivityIndicator. Updating it to be consistent with the rest of the app

cimigree, line 31: Can you remove the button styles as they are no longer being used?

src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -25,7 +25,7 @@ export const HomeHeader: FC<
         onPress={() => {
           navigation.navigate('Sync');
         }}>
-        <SyncIconCircle />
+        <SyncIconCircle testID="MAIN.sync-icon" />
```
cimigree, line 28: Thanks for adding the test id. It is working!

cimigree (approved):
Wow that is a bright orange! I like it!
Just one bit of code to remove please and then you can merge. All looked good and seemed to be working.

PR #806, opened 2024-10-23
feat: audio playback from observation, edit, delete

### LINKED ISSUE #361: Playback / Delete / Share

cimigree:
### Description
closes #361 

- Adds the ability to see audio thumbnail(s) on an existing observation or when editing an observation
- Thumbnail is pretty basic right now because we don't have access to the duration and created at from the backend yet but I left some placeholders to guide future develop
- Can click on the thumbnail and listen to it from editing an observation or just when viewing an observation
- Adds the ability to Edit an existing observation in regard to audio. Can add another audio or delete an audio file when you are editing an observation (but only when editing -- not when viewing)
- Adds the ability to Share an audio file through Android sharing (as specified in the ticket, different from the design docs)
- Passes the attachment to the Audio Thumbnail and the Thumbnail does the logic (as is true with Photo Thumbnail) to decide what to do as was requested.
- I also decided to pass a few parameters instead of adding context. Mostly just to know if someone is editing a file which will determine how some navigation is done and whether or not to show the delete button.


src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
```diff
@@ -1,71 +1,104 @@
...
-      <Text style={styles.timeSince}>
-        <FormattedRelativeTime
-          value={(Date.now() - record.createdAt.getTime()) / 1000}
-          numeric="auto"
-          updateIntervalInSeconds={1}
-        />
-      </Text>
+      onPress={handlePress}
+      disabled={isPending || !data?.url || isError}>
+      <PlayArrow width={48} height={48} />
```
ErikSin: The thumbnail should follow the photo thumbnail model that uses an `UIActivictyIndicator `while loading and and `AlertIcon` on Error

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -168,18 +164,17 @@ export const ObservationCreate = ({
...
-      return createBlobMutation.mutateAsync(
-        // @ts-expect-error Due to TS array filtering limitations. Fixed in TS 5.5
-        photo,
-      );
-    });
 
-    const audioPromises = unsavedAudioRecordings.map(audio => {
-      return createAudioBlobMutation.mutateAsync(audio);
-    });
+    const attachmentPromises = [savablePhotos, unsavedAudioRecordings].map(
```
ErikSin: i think this needs to be spread this otherwise you have 1 array with only 2 items in them (those items being arrays)

src/frontend/screens/Audio/ExistingRecording.tsx
```diff
@@ -0,0 +1,148 @@
...
+    const localUri = `${FileSystem.cacheDirectory}${tempFileName}`;
+
+    try {
+      const downloadResult = await FileSystem.downloadAsync(uri, localUri);
+      await Share.open({url: downloadResult.uri});
+      await FileSystem.deleteAsync(downloadResult.uri, {idempotent: true});
+    } catch (err) {
+      // just keep while developing
+      console.error('Error sharing file:', err);
+    }
```
ErikSin: As discussed in the call, this should trigger a loading state

ErikSin (commented):
Just 3 more comments outside of the stuff we discussed today

src/frontend/sharedComponents/MediaScrollView/index.tsx
```diff
@@ -4,38 +4,80 @@ import {Dimensions, ScrollView, StyleSheet} from 'react-native';
...
+  return 'uri' in attachment && !('driveDiscoveryId' in attachment);
+};
+
+export const MediaScrollView: FC<MediaScrollView> = ({
+  attachments,
+  isEditing = false,
+}) => {
   const scrollViewRef = React.useRef<ScrollView>(null);
-  const length = photos?.length ?? 0;
+  const length = attachments.length;
```
ErikSin: this is not needed anymore!

src/frontend/sharedComponents/MediaScrollView/index.tsx
```diff
@@ -4,38 +4,80 @@ import {Dimensions, ScrollView, StyleSheet} from 'react-native';
...
   const navigation = useNavigationFromRoot();
   React.useLayoutEffect(() => {
     scrollViewRef.current && scrollViewRef.current.scrollToEnd();
-  }, [photos?.length]);
+  }, [attachments.length]);
 
-  if (photos?.length === 0) return null;
+  if (length === 0) return null;
```
ErikSin: this can be removed as well

src/frontend/sharedComponents/MediaScrollView/index.tsx
```diff
@@ -4,38 +4,80 @@ import {Dimensions, ScrollView, StyleSheet} from 'react-native';
...
+  attachment: Audio | Photo,
+): attachment is AudioAttachment => {
+  return 'type' in attachment && attachment.type === 'audio';
+};
+
+const isUnsavedAudio = (
+  attachment: Audio | Photo,
+): attachment is UnsavedAudio => {
+  return 'uri' in attachment && !('driveDiscoveryId' in attachment);
+};
```
ErikSin: i think we should seperate this into a file in `lib` and then use there everywhere you do a check.

src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
```diff
@@ -1,71 +1,150 @@
...
+  onPress?: () => void;
+  disabled?: boolean;
 }
 
-export const AudioThumbnail: FC<AudioThumbnail> = ({
+type AudioThumbnailImageProps = {
+  isLoading: boolean;
+  error?: Error | null;
+  uri?: string;
+  onPress?: () => void;
```
ErikSin: is there a reason `onPress` is optional?

src/frontend/sharedTypes/navigation.ts
```diff
@@ -102,7 +102,7 @@ export type RootStackParamsList = {
   DataAndPrivacy: undefined;
   SettingsPrivacyPolicy: undefined;
   HowToLeaveProject: undefined;
-  Audio: undefined;
+  Audio: {existingUri?: string; isEditing: boolean} | undefined;
```
ErikSin: i think "existingUri" implies the uri of saved audio, which is not the case. So perhaps just "uri"

src/frontend/screens/Audio/ExistingRecording.tsx
```diff
@@ -0,0 +1,191 @@
...
+        }
+      }
+    };
+
+    downloadAudio();
+
+    return () => {
+      isCancelled = true;
+      if (localUri) {
+        FileSystem.deleteAsync(localUri, {idempotent: true}).catch(() => {});
```
ErikSin: I understand that the mentality is to have the uri available as soon as they press share, and in order to do that we want to start loading it into memory as soon as this screen opens. But this will cause it to load into memory twice. On first render, local uri is not available, so `useAudioPlayback` will take the uri, and load it into memory with `Audio.Sound.createAsync()`. Then the useeffect you have written here will also start to load into memory. then when that is loaded into memory, a uri will be created, which will replace the original uri in `useAudioPlayback`. BUT useAudioPlayback will still call `Audio.Sound.createAsync()`on this locauri and start loading into memory again.

I think the original approach of only downloading it onClick of the share button avoids unnecessarily having it twice in memory (and yes it will still be in memory twice, but only if the user chooses to share it, not every single time)

src/frontend/hooks/persistedState/usePersistedDraftObservation/photosMethods.ts
```diff
@@ -6,23 +6,23 @@ type Setter = StoreApi<DraftObservationSlice>['setState'];
...
+  set({attachments: newAttachments});
 }
 
 export function replaceDraftPhotos(
   set: Setter,
   get: Getter,
   draftPhoto: DraftPhoto,
 ) {
-  const updatedPhotosState = get().photos.map(p => {
+  const updatedPhotosState = get().attachments.map(p => {
```
ErikSin: rename to "updatedAttachments"

ErikSin (changes requested):
Its breaking. The Playback screen is throwing an error. The audio is still playing, but the error bottom sheet is popping up - and i am unable to close the bottom sheet and I have to close the app entirely.

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
```diff
@@ -243,19 +243,19 @@ const SharedBackButton = ({
...
+  // attachments are created right before an observation is made, so we need to check # attachments that are about to be saved
+  const {attachments: originalAtts, ...originalNoAttachments} = original;
 
-  if (originalAtts.length !== numberOfPhotos) return false;
+  if (originalAtts.length !== numberOfAttachments) return false;
 
-  const {attachments, ...edittedNoPhotos} = editted;
-  return isEqual(orignalNoPhotos, edittedNoPhotos);
+  const {attachments, ...editedNoAttachments} = edited;
+  return isEqual(originalNoAttachments, editedNoAttachments);
```
ErikSin: Sorry, this logic doesnt work anymore. originally you could not delete photos, so i was just checking the # of photos. If a user added a photo, the original number of photos would be different from the editted number of photos.

But now a user can delete an audio file. So in theory a user could create an new audio file and delete an old audio file. That would mean the # of attachments would be the same so `isEqual` would return to true even though it is false. So we could actually just use `isEqual` on the `original`  and `edittted`. I didnt just do that originally because I was worried that the `attachments` array order would effect `isEqual` but its better for the user to get a false negative then a false positive.

src/frontend/screens/Audio/ExistingRecording.tsx
cimigree: So I think I have fixed the issue and the crash. Sorry about that. I made the change because the Share was not working after the first Share... so I had to cache the uri (localURI -- now just used for sharing). I did it wrong in a rush at the end of the day. So,  I made a change to cache the uri when someone presses Share so that a user can share multiple times and it won't load twice or crash the playback! I am only deleting it on unmount ( I think I did that right but could use your eyes on it).

src/frontend/sharedComponents/MediaScrollView/index.tsx
cimigree: Isn't it needed for the scroll view? I just subbed attachments.length but the length is still needed I think.

src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
cimigree: No! Fixed.

src/frontend/sharedComponents/MediaScrollView/index.tsx
cimigree: Ok done and used.

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
cimigree: I made the fix you suggested and tested it out and it seems to work.

src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
```diff
@@ -204,10 +206,29 @@ const draftObservationSlice: StateCreator<DraftObservationSlice> = (
...
-      }),
+    addAudio: (audio: UnsavedAudio) => {
+      set({attachments: [...get().attachments, audio]});
+    },
+    deleteAudio: (uri: string) => {
+      const extractedName = uri.split('/').pop();
+      const updatedAttachments = get()
+        .attachments.map(attachment => {
+          if ('uri' in attachment && attachment.uri === uri) {
+            return null;
```
ErikSin: why are we returning null?

src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
```diff
@@ -1,71 +1,150 @@
...
+  style?: StyleProp<ViewStyle>;
+  size?: number;
+  isEditing: boolean;
+};
+
+interface SharedThumbnailProps {
   style?: StyleProp<ViewStyle>;
+  size: number;
+  isEditing: boolean;
+  onPress?: () => void;
```
ErikSin: this `onPress` is completely unused. Should we just get rid of it?

src/frontend/screens/Audio/ExistingRecording.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const localFilePath = `${FileSystem.cacheDirectory}${tempFileName}`;
+        const downloadResult = await FileSystem.downloadAsync(
+          uri,
+          localFilePath,
+        );
+        setLocalUri(downloadResult.uri);
+      }
+      await Share.open({url: localUri!});
+    } catch (err) {
+      console.error('Error sharing file:', err);
```
ErikSin: should remove

src/frontend/screens/Audio/ExistingRecording.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const handleShare = async () => {
+    setLoading(true);
+    try {
+      if (!localUri) {
+        const tempFileName = `audio_${Date.now()}.m4a`;
+        const localFilePath = `${FileSystem.cacheDirectory}${tempFileName}`;
+        const downloadResult = await FileSystem.downloadAsync(
+          uri,
+          localFilePath,
+        );
```
ErikSin, line 104: This is breaking when I try and share an audio file before it has been written to the db (aka when I am creating and observation and have not saved the observation yet.

This is the error I am getting:
`Error sharing file: [Error: Call to function 'ExponentFileSystem.downloadAsync' has been rejected.
→ Caused by: java.lang.IllegalArgumentException: Expected URL scheme 'http' or 'https' but was 'file']`

Idk if we necessarily want to be sharing an audio file before it is saved? Unless design explicitly wanted that, I think we should follow the pattern of the rest of the observations and only allow them to share after it isaved

ErikSin (changes requested):
If you open an audio file, the share it (I technically dont even have to share it, i can just press the share button then cancel), go back, then reopen the audio file it will throw an error when you try to play it.

When I explicitly consoled the error, this is what i got:
`[Error: com.google.android.exoplayer2.upstream.HttpDataSource$HttpDataSourceException: java.io.IOException: java.util.concurrent.ExecutionException: java.net.ConnectException: Failed to connect to /127.0.0.1:43517]`...such a helpful error....

src/frontend/hooks/persistedState/usePersistedDraftObservation/index.ts
cimigree: It was how I figured to delete unsaved audios.

src/frontend/screens/Audio/ExistingRecording.tsx
cimigree: Sounds good. I will remove file sharing if file hasn't been saved yet

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -216,8 +167,91 @@ export function RecordingDone({duration, uri, reset}: RecordingDoneProps) {
...
+const RenderModalContent = ({
+  modalContentType,
+  t,
+  handleDelete,
+  closeSheet,
+  handleReturnToEditor,
+  handleRecordAnother,
+}: {
+  modalContentType: ModalContentType;
+  t: ReturnType<typeof useIntl>['formatMessage'];
```
ErikSin: instead of passing `t` as a prop, can you just use the `useIntl()` hook directly in this component

src/frontend/screens/Audio/CreateRecording/RecordingDone.tsx
```diff
@@ -60,15 +60,21 @@ interface RecordingDoneProps {
   duration: number;
   uri: string;
   reset: () => void;
+  isEditing?: boolean;
```
ErikSin: `isEditing` is a mandatory prop for the parent component, so I think we should making this a mandatory prop here as well, and then you can get rid of ` isEditing = false,` on line 73

src/frontend/screens/Audio/CreateRecording/index.tsx
```diff
@@ -6,7 +6,7 @@ import {RecordingIdle} from './RecordingIdle';
 import {useAudioRecording} from './useAudioRecording';
 import {ErrorBottomSheet} from '../../../sharedComponents/ErrorBottomSheet';
 
-export function CreateRecording() {
+export function CreateRecording({isEditing = false}: {isEditing: boolean}) {
```
ErikSin: if `isEditing` is a mandatory prop, why are we setting a default value?

src/frontend/screens/Audio/index.tsx
```diff
@@ -1,16 +1,31 @@
...
 import {CreateRecording} from './CreateRecording';
+import {ExistingRecording} from './ExistingRecording';
+import {useDraftObservation} from '../../hooks/useDraftObservation';
 
 export const MAX_RECORDING_DURATION_MS = 5 * 60_000;
 
-export function Audio() {
+export function Audio({route}: NativeRootNavigationProps<'Audio'>) {
+  const {deleteAudio} = useDraftObservation();
+  const {isEditing = false, uri, isSavedUri = false} = route.params ?? {};
```
ErikSin: same comment as above, if `isEditing` is a required prop, do we need to set a default value?

ErikSin: Also, we do not need the last part of line 14, can get rid of `?? {}`

src/frontend/screens/Audio/useAudioPlayback.ts
```diff
@@ -31,27 +31,31 @@ export const useAudioPlayback = (recordingUri: string) => {
...
+        }
       });
 
     return () => {
-      if (soundInstance) {
-        soundInstance.unloadAsync().catch(err => {
+      isCancelled = true;
+      if (recordedSoundRef.current) {
+        recordedSoundRef.current.unloadAsync().catch(err => {
           handleError(err);
```
ErikSin: why are we handling the error here but not on line 29. I think in both cases we should skip the error as the app will still function if the sound isn't unloaded

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -129,10 +127,11 @@ export const ObservationCreate = ({
   const createObservation = React.useCallback(() => {
     if (!value) throw new Error('no observation saved in persisted state ');
 
-    const savablePhotos = photos.filter(photo => photo.type === 'processed');
-    const savableAudioRecordings = audioRecordings;
+    const savablePhotos = attachments.filter(isProcessedDraftPhoto);
```
ErikSin: we should be consistent with the language here and call this `unsavedPhotos`

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
```diff
@@ -243,19 +243,18 @@ const SharedBackButton = ({
...
 ) {
-  if (!editted || !('docId' in editted)) return true;
-  // attachments are created right before an observation is made, so we need to check # photos that are about to be saved
-  const {attachments: originalAtts, ...orignalNoPhotos} = original;
+  if (!edited || !('docId' in edited)) return true;
+  // attachments are created right before an observation is made, so we need to check # attachments that are about to be saved
+  const {attachments: originalAtts} = original;
 
-  if (originalAtts.length !== numberOfPhotos) return false;
+  if (originalAtts.length !== numberOfAttachments) return false;
```
ErikSin, line 257: this is check is wrong now. number of attachments is no longer a good incdicator since audio files can be created and deleted

ErikSin (changes requested):
Some minor changes.

I also refactored the `isEditing prop`. As much as possible, the Editor should just be a UI layer, and not hold any logic. I just derived `isEditing` from the route name where it was needed

src/frontend/sharedComponents/CustomHeaderLeftClose.tsx
cimigree: I thought it was a good way of doing a quick check before we do the more involved check... If they aren't the same length then something definitely changed so we can return false right away before doing the isEqual check next.

ErikSin (approved):
Thanks Cindy!

PR #808, opened 2024-10-29
chore: Add Detox e2e tests & BrowserStack CI

gmaclennan:
This adds e2e tests using Detox, plus CI integration with BrowserStack for running tests on real devices.

Unfortunately, right now it's not possible to run the app with any permissions (camera and location) enabled in BrowserStack testing. When testing locally, Detox installs the APK via `adb install -g`. The `-g` argument causes the app to be installed with all the permissions already enabled. When running Detox tests on BrowserStack, BrowserStack uses their own mechanism to install the APK, and does not expose an ADB connection.

Using Detox tests, there is no way to respond to a permission dialog in tests, because Detox uses gray box testing.

As a temporary fix, this PR introduces a `EXPO_PUBLIC_IS_E2E` environment variable, which is set when building for detox. If this variable is set, then the app does not ask for permissions.

This means that all components in the app will be in the "no permissions granted" state.


cimigree:
Instructions:
1. Need to add some env variables ... talk to Cindy or Gregor about that.
2. in terminal type npm run test:e2e
3. You will be given the prompt ?Rebuild app -- choose yes the first time by using the cursor and hit return
4. You will be given the prompt ? Rebuild beackend -- choose yes the first time by using the cursor and hit return
5. ? Run tests locally or in the cloud? -- choose cloud with the down cursor and hit enter
6. You will be given a list of devices. Use the cursor to scroll down, the space key to make a selection (your choice) and then hit enter
7. The cli will do its thing. It will take quite a while. Your terminal should look like this afterwards
8. At this point I cannot add you to Browserstack because of our plan, but I am sure you will be able to be added soon!
Screenshot 2024-11-13 at 5 37 57 PM

ErikSin:
Im going to close this since we are probably not going to be using detox!

gmaclennan:
@cimigree did this get to a stage where tests were running and passing on browserstack? I ask because if this is actually working, this is better than not having tests, so even if we plan to not use detox in the future, it seems worth using this as an interim measure. If it's not working and still requires additional work then maybe best to just close and discard.

PR #810, opened 2024-10-30
Tracks permission

### LINKED ISSUE #809: Location Permission
Location permission 
- [ ]  Integrate location request as [designed](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=3487-617&t=zh8yydymqjChphjd-4)
- [ ] "CoMapeo collects location data to track your route on the map, even when the app is closed. Your location data is only stored on your device by default, and is not stored or sent anywhere."

Todo
- [x] confirming app should be tracking when the app is fully closed

ErikSin:
Explicitly asks user for background location. closes #809 

image

image

### DIFF
--- a/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
+++ b/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
@@ -44,7 +44,7 @@ export const GPSBackgroundPermissionDisabled = ({
       return;
     }
 
-    if (backgroundPermission.canAskAgain) {
+    if (!backgroundPermission.canAskAgain) {
       handleOpenSettings();
     }
   }
--- a/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
+++ b/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
@@ -1,0 +1,83 @@
+import * as React from 'react';
+import {defineMessages, useIntl} from 'react-intl';
+import {StyleSheet, View, Linking} from 'react-native';
+import {Text} from '../../../sharedComponents/Text';
+import {MapPinErrorIconSmall} from '../../../sharedComponents/MapPinErrorIcon/MapPinErrorIconSmall';
+import {Button} from '../../../sharedComponents/Button';
+import * as Location from 'expo-location';
+
+const handleOpenSettings = () => {
+  Linking.sendIntent('android.settings.LOCATION_SOURCE_SETTINGS');
+};
+
+const m = defineMessages({
+  useLocation: {
+    id: 'screens.MapScreen.GPSBackgroundPermissionDisabled.useLocation',
+    defaultMessage: 'Use Your Location',
+  },
+  collectsLocation: {
+    id: 'screens.MapScreen.GPSBackgroundPermissionDisabled.collectsLocation',
+    defaultMessage:
+      'CoMapeo collects location data to track your route on the map, even when the app is closed. Your location data is only stored on your device by default, and is not stored or sent anywhere.',
+  },
+  turnOn: {
+    id: 'screens.MapScreen.GPSBackgroundPermissionDisabled.turnOn',
+    defaultMessage: 'Turn On',
+  },
+});
+
+type GPSBackgroundPermissionDisabledProp = {
+  setBackgroundStatusGranted: (status: boolean) => void;
+};
+
+export const GPSBackgroundPermissionDisabled = ({
+  setBackgroundStatusGranted,
+}: GPSBackgroundPermissionDisabledProp) => {
+  const {formatMessage} = useIntl();
+
+  async function askBackgroundLocationPermission() {
+    const backgroundPermission =
+      await Location.requestBackgroundPermissionsAsync();
+
+    if (backgroundPermission.granted) {
+      setBackgroundStatusGranted(true);
+      return;
+    }
+
+    if (backgroundPermission.canAskAgain) {
+      handleOpenSettings();
+    }
+  }
+  return (
+    <View style={styles.container}>
+      <MapPinErrorIconSmall style={{marginBottom: 20}} />
+      <Text style={styles.title}>{formatMessage(m.useLocation)}</Text>
+      <Text style={styles.description}>
+        {formatMessage(m.collectsLocation)}
+      </Text>
+      <Button fullWidth onPress={askBackgroundLocationPermission}>
+        {formatMessage(m.turnOn)}
+      </Button>
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    alignItems: 'center',
+    display: 'flex',
+    justifyContent: 'center',
+    padding: 20,
+    paddingTop: 30,
+  },
+  title: {
+    fontSize: 24,
+    fontWeight: '500',
+    marginBottom: 10,
+    textAlign: 'center',
+  },
+  description: {
+    textAlign: 'center',
+    marginBottom: 30,
+  },
+});
--- a/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
+++ b/src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
@@ -1,14 +1,9 @@
 import * as React from 'react';
 import {defineMessages, useIntl} from 'react-intl';
-import {StyleSheet, View, Linking} from 'react-native';
+import {StyleSheet, View} from 'react-native';
 import {Text} from '../../../sharedComponents/Text';
 import {MapPinErrorIconSmall} from '../../../sharedComponents/MapPinErrorIcon/MapPinErrorIconSmall';
 import {Button} from '../../../sharedComponents/Button';
-import * as Location from 'expo-location';
-
-const handleOpenSettings = () => {
-  Linking.sendIntent('android.settings.LOCATION_SOURCE_SETTINGS');
-};
 
 const m = defineMessages({
   useLocation: {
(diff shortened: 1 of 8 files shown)


src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
```diff
@@ -0,0 +1,83 @@
...
+  container: {
+    alignItems: 'center',
+    display: 'flex',
+    justifyContent: 'center',
+    padding: 20,
+    paddingTop: 30,
+  },
+  title: {
+    fontSize: 24,
+    fontWeight: '500',
```
achou11, line 57: looks a bit light compared to the design (and the foreground implementation).

```suggestion
    fontWeight: 'bold',
```

src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
```diff
@@ -0,0 +1,83 @@
...
+  async function askBackgroundLocationPermission() {
+    const backgroundPermission =
+      await Location.requestBackgroundPermissionsAsync();
+
+    if (backgroundPermission.granted) {
+      setBackgroundStatusGranted(true);
+      return;
+    }
+
+    if (backgroundPermission.canAskAgain) {
```
achou11: based on the [docs](https://docs.expo.dev/versions/v50.0.0/sdk/location/#permissionresponse), i think this should be flipped. otherwise i commonly run into situations where pressing the button does nothing.

> Indicates if user can be asked again for specific permission. If not, one should be directed to the Settings app in order to enable/disable the permission.

```suggestion
    if (!backgroundPermission.canAskAgain) {
```

src/frontend/screens/MapScreen/GPSPermissions/GPSForegroundPermissionDisabled.tsx
```diff
@@ -26,26 +26,28 @@ const m = defineMessages({
...
+  async function askForegroundLocationPermission() {
+    const foregroundPermission =
+      await Location.requestForegroundPermissionsAsync();
+
+    if (foregroundPermission.granted) {
+      setForegroundStatusGranted(true);
+      return;
+    }
+
+    if (foregroundPermission.canAskAgain) {
```
achou11: same comment about this being flipped

```suggestion
    if (!foregroundPermission.canAskAgain) {
```

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
```diff
@@ -1,34 +1,42 @@
...
   const [backgroundStatus] = Location.useBackgroundPermissions();
   const [foregroundStatus] = Location.useForegroundPermissions();
 
-  const [isGranted, setIsGranted] = useState<boolean | null>(null);
+  const [foregroundStatusGranted, setForegroundStatusGranted] = useState<
+    boolean | null
+  >(null);
+  const [backgroundStatusGranted, setBackgroundStatusGranted] = useState<
+    boolean | null
+  >(null);
```
achou11: probably missing some context, but can these not be derived from the `backgroundStatus` and `foregroundStatus`?

achou11 (changes requested):
Some small requests but overall looking good.

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
ErikSin: yeah, you would think. But its actually a [problem with the hooks](https://github.com/expo/expo/issues/30351). The hooks are not actually responsive, so hence why I am forcing the state here. I actually think that we should refactor the entire permissions landscape because this exposed some really annoying edge cases

src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
ErikSin: I recently learned that we should be following exactly what is on Figma (when it comes to font sizing and weight). The real issue here is that it should be using the font Rubik (as it is a header). We actually are improperly importing Rubik into the app. I have a follow up PR that will fix all the fonts. So I think we should keep the font weight to 500 and when it is converted to rubik it will be consistent with the figma

src/frontend/screens/MapScreen/GPSPermissions/GPSForegroundPermissionDisabled.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/810/commits/290ca88b575d19f0103839b117e82d6bb254d3c8

src/frontend/screens/MapScreen/GPSPermissions/GPSBackgroundPermissionDisabled.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/810/commits/290ca88b575d19f0103839b117e82d6bb254d3c8

achou11: seems to be pointing to the other file, not this one

ErikSin: whoops! I had it on my vscode unsaved

ErikSin: https://github.com/digidem/comapeo-mobile/pull/810/commits/11a34c8341d672175c178fe0eb8ea3af9e20e31e

achou11 (approved):
Wouldn't consider this blocking, but ran into this slightly annoying issue where there's a lot of unprompted moving around in this sequence:

1. Press tracks tab, which opens up the background location permissions request sheet
2. Press "Turn On" button to get directed to app's location settings
3. Do not grant the desired permission. Press hardware back button (or gesture back) to go back to the app.

After 3, I _usually_ get immediately redirected to the settings app (see the video). Personally, I would expect to only get directed to the settings app when explicitly pressing that button in the bottom sheet (i.e. after step 2). my guess is that fixing this is maybe outside the scope of the PR, but in case there's an easy fix for it you're aware of.

https://github.com/user-attachments/assets/b6e014c9-06e3-4ac2-a1ed-d43624d38cd4

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
```diff
@@ -1,34 +1,86 @@
...
+      handleOpenSettings();
+    }
+  }
+
+  const renderContent = () => {
+    if (!foregroundPermission || !backgroundPermission) {
+      return (
+        <View style={{display: 'flex', minHeight: 200}}>
+          <Loading />
+        </View>
```
ErikSin, line 45: there is a small (but unlikely) chance that the user can open the track modal BEFORE the permission have been loaded. I needed to set a height to the view in order for this to not just be a sliver of a bottom sheet. So I arbitrarly set 200, but I am open to suggestions if we don't want to use an arbitrary number

achou11: have unfortunately run into this before 🥲

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
```diff
@@ -1,34 +1,84 @@
...
+  }
+
+  async function askBackgroundLocationPermission() {
+    if (backgroundPermission!.canAskAgain) {
+      const permission = await Location.requestBackgroundPermissionsAsync();
+      setBackgroundPermission(permission);
+    } else {
+      handleOpenSettings();
+    }
+  }
```
achou11: nit: would inline these to where they're used so you can avoid the non-null operator usage e.g. `backgroundPermission!.canAskAgain` becomes `backgroundPermission.canAskAgain`

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
```diff
@@ -38,13 +88,7 @@ export const GPSPermissionsModal = React.memo(() => {
...
       onDismiss={onBottomSheetDismiss}
       handleComponent={() => null}>
-      <BottomSheetView>
-        {isGranted ? (
-          <GPSPermissionsEnabled />
-        ) : (
-          <GPSPermissionsDisabled setIsGranted={setIsGranted} />
-        )}
-      </BottomSheetView>
+      <BottomSheetView>{renderContent()}</BottomSheetView>
```
achou11, line 89: nit: `renderContent()` being a function instead of a variable has a small overhead. probably insignificant in this case but can be somewhat prone to issues if it was more complex

src/frontend/screens/MapScreen/GPSPermissions/GPSPermissionsModal.tsx
```diff
@@ -1,34 +1,84 @@
...
 import {BottomSheetModal, BottomSheetView} from '@gorhom/bottom-sheet';
 import {TAB_BAR_HEIGHT} from '../../../Navigation/Stack/AppScreens';
-import {StyleSheet} from 'react-native';
-import {useFocusEffect} from '@react-navigation/native';
+import {StyleSheet, Linking, View} from 'react-native';
+import {GPSBackgroundPermissionDisabled} from './GPSBackgroundPermissionDisabled';
+import {Loading} from '../../../sharedComponents/Loading';
+
+const handleOpenSettings = () => {
+  Linking.sendIntent('android.settings.LOCATION_SOURCE_SETTINGS');
```
achou11, line 14: if the user ends up going down this path, there's an issue where the app doesn't realize that the permission has been granted when returning to it. in order for the app to recognize that the permission is granted, they have to restart the app.

this is related to the issue you referred to in expo location, so i don't think there's a trivial way of properly handling this case. the "easy" short-term solution would be to inform the user that a restart would be needed.

don't think this is a blocker, but something to keep an eye on for later.

achou11 (approved):
definitely an improvement! although noticed another relatively minor issue (existed before newest commits)

PR #813, opened 2024-10-31
chore: Removes feature flag for tracks.

cimigree:
Also edits the env template to remove feature flags for what are now features!


ErikSin (approved):
Thanks!

PR #816, opened 2024-10-31
chore: update `@comapeo/core` and `@comapeo/schema`

EvanHahn:
This updates `@comapeo/core` and `@comapeo/schema` to their latest versions. This should have no user impact but should make it easier to add new features, such as Remote Archive or Alerts.

I tested this by installing the app from scratch and making sure I could get to the map screen.


ErikSin (approved):
Thanks Evan!

PR #818, opened 2024-11-04
chore: updated buttons and some headers and labels to use rubik

### LINKED ISSUE #769: Implement type ramp, typefaces, positioning and truncation
See Notion issue : https://www.notion.so/digidem/Correction-Implement-type-ramp-typefaces-positioning-and-truncation-10b1b08162d580e0891ef1d9c6a93055

- [ ] to investigate why the issue has occurred in the first place.
- [ ] Discuss with sabella if theres a need to discuss design
- [ ] estimate time to fix
- [ ] Implement fix

ErikSin:
closes #769 

Added rubik font via expo.
Updated the text component to take a `variant` prop. Setting `variant=primary` will use font rubik
Updated the `button` component to use variant "primary"
Updated selected header and labels to use variant "primary"
Update the nav header to use Rubik

Updated these headers and labels:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=3468-2389&node-type=frame&t=SFrkpMNpu1M2cDRi-0

# To Do:
- we should incrementally update all headers and labels to use rubik.


src/frontend/sharedComponents/Text.tsx
```diff
@@ -1,15 +1,82 @@
...
+  const computedStyle = React.useMemo(
+    () => [
+      style,
+      {fontFamily: getFontFamily({styleObj: style, variant: variant})}, // Set the correct fontFamily based on fontWeight
+    ],
+    [style, variant],
+  );
+
+  return (
+    <RNText
```
cimigree: Using these hardcoded font sizes seems like it will make it harder to adjust the font sizes globally.
Not using a theme or stylesheet seems messy.

src/frontend/sharedComponents/Text.tsx
```diff
@@ -1,15 +1,82 @@
...
+ *
+ * Font, font sizing and font weight, should follow designs in figma.
+ * Generally all headers, labels, and button us Rubik (set `variant` to "primary").
+ * Paragraph and body text should use the device's System Font.
+ * Refer to this [design sytem](https://www.notion.so/digidem/Typography-a1600603854445ad82b8d67d4067be1a) for more info.
+ */
 export const Text = ({
   children,
   style,
+  variant = 'body',
```
cimigree: Requiring developers to manually set variant="primary" seems error-prone. Maintaining this across so many different components seems unmanageable.

src/frontend/sharedComponents/Text.tsx
```diff
@@ -1,15 +1,82 @@
...
+        {color: BLACK, fontSize: variant === 'primary' ? 16 : 14},
+        computedStyle,
+      ]}
+      {...otherTextProps}>
+      {children}
+    </RNText>
+  );
+};
+
+function getFontFamily({
```
cimigree: I have some concerns about this method.
The mapping assumes that fontWeight directly corresponds to specific font files, which might not always be accurate. Can't styles come from various sources (inline styles, stylesheets, theme providers)?
Also, merging styles and inspecting them on each render seems like a lot of overhead. Plus, I am just having trouble following the logic.

cimigree (changes requested):
Although most of the screens shown in the Figma are addressed by this PR, I am pretty confused about how this PR resolves all of the issues in #769 . I have included some screen shots to show what seems on first check to not be fixed. 
One thing for sure is the truncation is not addressed at all, is it? 
The issues says "Make sure all titles and labels are using Rubrik"
It also says: "Correct font sizes throughout, ensure type ramp is applied" and 
"Apply truncation rules to areas that risk overrunning"

Also, it seems like there are a number of ways to do the styling in a way that is more easy to maintain in the future and for future developers.
One way would be to use a centralized theme or style sheet (a styles.ts file where you could have a heading, label, body, etc and call them rubikheader or whatever)
Or another idea would be to make specialized text components, eg. heading, label, bodytext etc
In addition to being confused about and not loving the changes to Text.tsx I am just concerned about future implementation and how much of it there will be and how the issue does not seem resolved.

No truncation
---
image

Alignment is off
---
image

This title is supposed to be bold according to the Figma
---
image

Figma:

Screenshot 2024-11-05 at 10 51 21 AM

ErikSin:
> Although most of the screens shown in the Figma are addressed by this PR, I am pretty confused about how this PR resolves all of the issues in #769 . I have included some screen shots to show what seems on first check to not be fixed. One thing for sure is the truncation is not addressed at all, is it? The issues says "Make sure all titles and labels are using Rubrik" It also says: "Correct font sizes throughout, ensure type ramp is applied" and "Apply truncation rules to areas that risk overrunning"

Sorry, I shouldn't have tagged the original issue. This fixes only certain screens that were addressed by Sabella in call. WE identified that doing all the screens at once would be too much work so she wanted me to to these screens first.

> Also, it seems like there are a number of ways to do the styling in a way that is more easy to maintain in the future and for future developers. One way would be to use a centralized theme or style sheet (a styles.ts file where you could have a heading, label, body, etc and call them rubikheader or whatever) Or another idea would be to make specialized text components, eg. heading, label, bodytext etc In addition to being confused about and not loving the changes to Text.tsx I am just concerned about future implementation and how much of it there will be and how the issue does not seem resolved.

So I am open to other ways of doing this. 

For the first alternative of having a theme, are you suggesting we have a styles.ts that looks like:

```ts
const styles = Stylesheet.create({
     rubikHeader:{fontFamily: rubik}
})

export rubikHeader = styles.rubikHeader
```
and then in the component we would do

```ts
<Text style={rubikHeader}>...someText</Text>
```

If this is what you are suggesting, could you just elaborate on why you think that is a better alternative?

For the second alternative, im also curious what the advantages of having a `< Header >` vs a `< Text variant='header' >` is? To me this seems like the same approach, with a slightly different implementation.

Im open to either of these, just curious as to why you think they are better alternatives :)

src/frontend/sharedComponents/Text.tsx
ErikSin: So in react native there is only one available font, the system font (Roboto for Android). So if we were to set the font to anything else, typescript won't throw an error, BUT the phone would still use the system font.

So in order to use a font, we need to include the font files (or import it through expo - which technically speaking just imports the file for us).

So a each font has a different file for each font weight and italization. So we cannot just import one rubik file. if we had the one rubik file and did `{fontFamily: 'rubik', fontWeight:"bold"}` the font weight won't show up. it would look like normal rubik. What we have to do is do` {fontFamily: 'Rubik_700Bold'}`. The `fontWeight` property technically only works for the system font.

So what this function does is allows developers to use fontWeight normally. And based on the fontWeight, the correct `fontFamily` will be chosen. So if the developer sets the `fontWeight` to 700, then this `getFontFamily `would return `{fontFamily: 'Rubik_700Bold'}`. 

Unfortunately with custom fonts we can't just use `fontWeight` and expect the system to know which file to use

cimigree:
Yes, your example is what I was suggesting. If we have a styles.ts file with predefined styles, it might make it easier to apply the styles across the app.
Like this maybe?
```
export const typography = StyleSheet.create({
  header: {
    fontFamily: 'Rubik_500Medium',
    fontSize: 24,
    // other header styles
  },
  label: {
    fontFamily: 'Rubik_400Regular',
    fontSize: 16,
    // other label styles
  },
  bodyText: {
    fontFamily: 'System',
    fontSize: 14,
    // other body text styles
  },
});

```
Then use it like this:
```
import { typography } from './styles';

<Text style={typography.header}>Some Header Text</Text>
```

Are we using variant "header"?

That could be ok. But, if it's its own component, then there can be additional logic in there if needed...

Something like:
```
export const Header = (props: TextProps) => (
  <Text style={typography.header} {...props} />
);

export const Label = (props: TextProps) => (
  <Text style={typography.label} {...props} />
);

```
And then use it like so:
```
<Header>Welcome</Header>
<Label>Username</Label>
```

Doing this would make the code more semantic and then developers don't have to remember about additional logic (the word primary is that your choice?). Also, everything would be centralized and if we change it at the top level it would be applied throughout. 
Having to pass variant="primary" or the exact fontweight seems much more fallible and to me -- a temporary solution only. I thought we try to develop with the future in mind even if you can't do everything requested in the ticket.
Wouldn't it have been nice if we had these set up already and then if you pulled Rubik in it would have been applied in all of the screens and components?
You are the expert, obviously, but even if you don't want to refactor, could you fix the one screen that does not have the bold text as it is in Figma for this?

src/frontend/sharedComponents/Text.tsx
cimigree: Thanks for this thorough explanation. I still think the logic is a little complex and having either separate components or a centralized typography style sheet would be more maintainable in the future.

ErikSin:
> Yes, your example is what I was suggesting. If we have a styles.ts file with predefined styles, it might make it easier to apply the styles across the app. Like this maybe? export const typography = StyleSheet.create({ header: { fontFamily: 'Rubik_500Medium', fontSize: 24, // other header styles }, label: { fontFamily: 'Rubik_400Regular', fontSize: 16, // other label styles }, bodyText: { fontFamily: 'System', fontSize: 14, // other body text styles }, });
> 
> Then use it like this: import { typography } from './styles';
> 
> Some Header Text
> 
> Are we using variant "header"?
> 
> That could be ok. But, if it's its own component, then there can be additional logic in there if needed...
> 
> Something like: export const Header = (props: TextProps) => ( <Text style={typography.header} {...props} /> );
> 
> export const Label = (props: TextProps) => ( <Text style={typography.label} {...props} /> );
> 
> And then use it like so:
> Welcome Username
> 
> Doing this would make the code more semantic and then developers don't have to remember about additional logic (the word primary is that your choice?).

I agree that having a text component, which is declarative, would be ideal (whether its  `< Header />` or `< Text variant="header" />`). The only issue is the delineation between what is a header vs label vs body is not as simple. Something can be a label, but can be bolded differently from another label, for example. Or another common example is something can be a header, but sized differently from another header (page header vs section header).

We could break up all of these categories into different component, ie < Label1 />, <Label 2 />, < PageHeader / >, < SectionHeader /> < BodyLarge /> < BodyRegular />....etc

But i am hesitant to do that because it would require a lot of mental overhead of knowing which styling belongs to which component. According to the design system there are 10 different variants AND none of those include how its bolded. 

So my mentality (which is not say I am right) is that the developer cares about the font, the weight, and the size. There is already properties in `style` that deal with weight and size, so I wanted to keep that the same (hence my getFontFamily() function => allowing the developer to just use fontWeight normally). And then all the developer need to do is determine if the font is rubik (aka "primary") or the default font.

That being said, I see that manually setting the font weight and font inside each component is not the best for scalability and maintenance. Im open to alternatives, but I dont think having 10 different components is helpful, especially if those components really only deal with the size and font, but not font weight. Open to suggestions if you see a middle ground

cimigree:
Sure what you say makes sense. Is the term "primary" a choice you made? I am not crazy about it as providing clear meaning of what it is/ does. Can that at least be something else? 
Also, still missing that bolded text see screen shot above.

ErikSin:
So I spoke to Sabella Today. I misunderstood the requirements. All headers, labels, and buttons, use the same font (rubik) and the same font weight (500). All body text, uses the system font and the regular font weight. The only difference is between all the variant are the font sizes.

So based on that new information I did what you suggested and created semantic components. I opted to use` < Header varinant="{someVariant}" />` as opposed to `< Header1 />, <Header2 /> etc`. The reason I did that was to take advantage of JS doc. So when a developer uses the `HeaderText` or `BodyText` component, they can put there mouse over the component and VS code will tell them which variant is which size. This will avoid having to open the file each time to determine what variant is which size.

I also added the @deperecated tag to our `Text` component. I think it will be too much in one PR to replace all the Text components. So we should do it for every file we touch. The deprecated tag will notify us that it should be replaced when we open a file.

src/frontend/sharedComponents/Text/BodyText.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
+import {Text as RNText, StyleProp, TextProps, TextStyle} from 'react-native';
+import {BLACK} from '../../lib/styles';
+
+type Variant = 'large' | 'regular' | 'smallMeta' | 'tinyMeta';
```
cimigree, line 5: Would you consider more semantic names? Or more conventional names? eg. semantic: caption, subtitle, body, etc, or conventional: small, medium, large, xsmall, xlarge?

src/frontend/sharedComponents/Text/BodyText.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+export const BodyText = ({
+  children,
+  style,
+  variant,
+  ...otherTextProps
+}: React.PropsWithChildren<BodyProps>) => {
+  let computedStyle: StyleProp<TextStyle>;
+
+  let fontSize: number;
```
cimigree: could you do something like

```
const fontSizeMap = {
  large: 20,
  regular: 16,
  small: 14,
  xsmall: 12,
};

const fontSize = fontSizeMap[variant || 'regular'];
```

since computerstyle is just fontSize?
Then 
`return (
  <RNText style={[{color: BLACK, fontSize}, style]} {...otherTextProps}>
    {children}
  </RNText>
);`

src/frontend/sharedComponents/Text/HeaderText.tsx
```diff
@@ -0,0 +1,73 @@
...
+ */
+export const HeaderText = ({
+  children,
+  style,
+  variant,
+  ...otherTextProps
+}: React.PropsWithChildren<HeaderProps>) => {
+  let fontSize: number;
+
+  switch (variant) {
```
cimigree: Could you use an object/ map here too perhaps?

cimigree (commented):
Yay! I am glad we can use more simple text components! I looked at most of the screens and they looked good. Just a few suggestions -- take them or leave them.

src/frontend/sharedComponents/Text/BodyText.tsx
ErikSin: I also don't love these names, but I am trying to stay consistent with the [design system](https://www.notion.so/digidem/Typography-a1600603854445ad82b8d67d4067be1a?pvs=4)

cimigree (commented):
Hmm. This doesn't look right with the size of my device name.

cimigree (approved):

PR #837, opened 2024-11-11
fix: observations disappearing when tracking starts

ErikSin:
closes #835

PR #838, opened 2024-11-12
fix: observations disappearing when tracking starts (#837)

ErikSin:


ErikSin:
I tested this on my device and it seems like it was working (although I had to clear the storage in order for it to work again)

cimigree (approved):

PR #839, opened 2024-11-12
feat: show linked track in observation

### LINKED ISSUE #820: Display a Track in an Obvservation
https://www.notion.so/digidem/Tracks-Cant-see-Track-in-the-Obvs-detail-view-1341b08162d58055ad9bda590caa7a69

ErikSin:
Closes #820.

Also updated some styling to be more consistent with design specs, and to allow the parent to control the padding styles.

image

### DIFF
--- a/src/frontend/screens/Observation/FieldDetails.tsx
+++ b/src/frontend/screens/Observation/FieldDetails.tsx
@@ -1,27 +1,28 @@
 import * as React from 'react';
 import {View, Text, StyleSheet} from 'react-native';
-import {MEDIUM_GREY, DARK_GREY, BLACK, LIGHT_GREY} from '../../lib/styles';
+import {MEDIUM_GREY, DARK_GREY, BLACK} from '../../lib/styles';
 import {
   FormattedFieldProp,
   FormattedFieldValue,
 } from '../../sharedComponents/FormattedData';
 import {Field, Observation} from '@comapeo/schema';
+import {ViewStyleProp} from '../../sharedTypes';
 
 export const FieldDetails = ({
   fields,
   observation,
+  style,
 }: {
   fields: Field[];
   observation: Observation;
+  style?: ViewStyleProp;
 }) => {
   return (
     <View>
       {fields.map(field => {
         const value = observation.tags[field.tagKey];
         return (
-          <View
-            key={field.docId}
-            style={[styles.section, styles.optionalSection]}>
+          <View key={field.docId} style={[styles.section, style]}>
             <Text style={styles.fieldTitle}>
               <FormattedFieldProp field={field} propName="label" />
             </Text>
@@ -52,11 +53,6 @@ const styles = StyleSheet.create({
   },
   section: {
     flex: 1,
-    marginHorizontal: 15,
     paddingVertical: 15,
-  },
-  optionalSection: {
-    borderTopColor: LIGHT_GREY,
-    borderTopWidth: 1,
   },
 });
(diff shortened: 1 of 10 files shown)


src/frontend/screens/Observation/index.tsx
```diff
@@ -96,33 +114,52 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
+        {track === 'loading' ? (
+          <Loading />
         ) : (
-          <ButtonFields
-            isMine={isMine}
-            observationId={observationId}
-            fields={fields}
-          />
+          <>
+            <View style={[styles.section, {flex: 1}]}>
```
cimigree: I'm curious about the necessity of applying flex: 1 in so many places. Is there a specific reason for this repetition? This View has flex: 1 here and styles.section also has flex: 1. And then below, the root also has flex: 1.

src/frontend/screens/Observation/TrackList.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const m = defineMessages({
+  track: {
+    id: 'screens.Observation.TrackList.track',
+    defaultMessage: 'Track',
+  },
+});
+
+export function TrackList({track}: TrackListProps) {
```
cimigree: nit: This isn't really a list even though I get why it contains a tracklistitem. Can it be called something more semantic? Like TrackLink?

src/frontend/screens/Observation/FieldDetails.tsx
```diff
@@ -1,27 +1,28 @@
...
   return (
     <View>
       {fields.map(field => {
         const value = observation.tags[field.tagKey];
         return (
-          <View
-            key={field.docId}
-            style={[styles.section, styles.optionalSection]}>
+          <View key={field.docId} style={[styles.section, style]}>
             <Text style={styles.fieldTitle}>
```
cimigree: Would it make sense to use the new Text styles here? BodyText? Or HeaderText? The Details (and the regular text) is very light on my screen. It looks different than I think what is asked for in the Figma.
<image src="https://github.com/user-attachments/assets/fdab91d6-588e-4283-b9ee-a41b2a2bc1b7" width="250" />

cimigree (approved):
I really like moving the accordion to a component! 
I have a few small comments in lined. The bit about the text being light is really just since you are in here. I think it is like that in production, right? Up to you if you want to deal with it now. Same with the other small comments.
It is very cool how I could go back and forth from the track to the observation to the track again etc through the links. I think this little glitch is probably fine, and it is hard to explain...
but if you go to the track from the observation, the back button takes you back to the observation, which is as I would expect. If you go to the track in the observation list, and then go to the observation and hit back, you go back to the track, which is as I would expect.
If you go to the observation from the observation list, to the track, to the observation and hit back, you go back to the observation list. Which is not what I would expect. Might be as expected for others though or too complex and minor to adjust.

ErikSin:
> but if you go to the track from the observation, the back button takes you back to the observation, which is as I would expect. If you go to the track in the observation list, and then go to the observation and hit back, you go back to the track, which is as I would expect. If you go to the observation from the observation list, to the track, to the observation and hit back, you go back to the observation list. Which is not what I would expect. 

Yeah, its kind of a weird quirk of react-navigation. If the navigation stack is `[observationList, observation, track]`, and then from `track` the user navigates to `observation`, react navigation will check the stack first and see if observation is in the stack, it is is, it will just go back to that point in the stack instead of adding observation on top of the stack (ie, the stack will become `[observationList, observation]` instead of `[observationList, observation, track, observation]`. That explains why the back button then takes the user back to the observationList instead of the track screen. Its actually quite a nice feature to make sure that the user does not get stuck in really deep navigation holes that would require them to see the same screen several time when they are trying to go back.

Funnily enough React Nav actually took out this behaviour in their latest release...

Anyways I circumvented this behaviour by using `navigation.push` instead of `navigation.navigate`. Now the user can endlessly go between tracks and observations and the navigation will be maintained.

cimigree (approved):
Wow it's much darker and easier to see now!

PR #840, opened 2024-11-12
fix: bottom sheet modal works when animations are disabled in android settings

cimigree:
There is an issue on Android 14 with the setting remove animations that can be accessed through accessibility in settings.
If a user has that enabled like this:  Accessibility -> color and motion -> enable Remove animations, then the bottom sheet modal does not work, which makes our app kind of unusable.
This PR upgrades the bottom sheet modal package and its dependencies (react-native-reanimated and react-native-gesture-handler) so that the modal works on Android 14 and forward even if a user has the accessibility feature of remove animations enabled.

Here is the issue as stated in react-native-bottom-sheet:
https://github.com/gorhom/react-native-bottom-sheet/issues/1674
Here is the changelog for react-native-bottom-sheet: https://github.com/gorhom/react-native-bottom-sheet/blob/master/CHANGELOG.md

It doesn't look like there are any breaking changes in the dependency react-native-gesture-handler:
https://github.com/software-mansion/react-native-gesture-handler/releases?page=2

And here is react-native-reanimated:
https://github.com/software-mansion/react-native-reanimated/releases?page=2
The animation on audio seems to be working fine, as before.
Passes maestro tests.

All that being said, I am not really sure how necessary this upgrade is and there could be unforeseen changes.


ErikSin (approved):
looks good to me. I think we should Cherry Pick this on the on the rc/v1.1.0 branch as this does fix a potential problem

PR #842, opened 2024-11-13
fix: drawer not closing when animations disabled

ErikSin:
Fixes drawer not closing when animations disabled. Caused by outdated dependencies.


cimigree (approved):
I ran the tests and all worked fine! The only thing is that the full sheet modal is not opening all the way if animations are disabled. If they are enabled, then it does open all the way. 

<image src="https://github.com/user-attachments/assets/46af9bbc-1755-4ff9-8389-82352edd0f8c" width="250" />
<image src="https://github.com/user-attachments/assets/3289661c-565f-4bf4-b819-0173f65dcf2e" width="250" />

ErikSin:
> The only thing is that the full sheet modal is not opening all the way if animations are disabled. If they are enabled, then it does open all the way.

I think that fine given that its not worse behaviour than before.... (aka it just wouldn't open)

Issue #359, opened 2024-05-15
Microphone Permissions


ximenabb:
## Microphone Permissions
As we introduce audio, comapeo will ask users to give it permission to the microphone. 
Important to note - we are not creating a custom "Permission Bottom Sheet - Microphone".

## Todo

**User clicks on mic icon for first time : Request Bottom Sheet - Microphone**
<img width="468" alt="Screen Shot 2024-05-15 at 7 03 13 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/e37a934c-688d-4427-911c-e95824a03b7a">

- [ ] Display screen as designed
- [ ] Add microphone icon with _Recording Audio with CoMapeo_ as title
- [ ] Display text _To record audio while using the app  and in the background CoMapeo needs to access your microphone._
- [ ] Display `Not now` button, when clicked redirect to to the starting state (New / Edit Obvservation)
- [ ] Display `Allow` button, when clicked open the android system permission dialog

**User wh
<img width="201" alt="Screen Shot 2024-05-15 at 7 03 31 PM" src="https://github.com/digidem/comapeo-mobile/assets/78557542/956b51a3-afb6-41ab-9168-b7b60c7978cd">
o have previously denied permission**

- [ ] Display screen like the _Request Bottom Sheet - Microphone_ screen
- [ ] Display `Not now` button, when clicked redirect to to the starting state (New / Edit Obvservation)
- [ ] Display `Go to Settings` button, when clicked open the android settings

**Tapping on either “While using the app” or “Only this time”**
- [ ] Take user to https://github.com/digidem/comapeo-mobile/issues/360

**Tapping "Don't allow"**
- [ ] First time - Take user to starting state
- [ ] Second time - take to settings

## Design
https://www.figma.com/design/1Knqv1NrijUuqMo1pvYyF7/Tracks-and-Audio-(Share)?node-id=529-812&t=V0vmc8SZM3TxMGuM-0

PR #843, opened 2024-11-13
fix: microphone permission modal has correct text

### LINKED ISSUE #831: Microphone permission denial is not as designed
**Describe the bug**
When someone clicks on the microphone the first time, the Bottom sheet is supposed to say Allow or Not Now the first time. But then the second time, it is supposed to say Go To Settings (see issue #359 )

Image
That is not happening. It keeps saying Allow, Not Now

**To Reproduce**
1. Go to make a new observation
2. Click on the microphone
3. Click Allow and then deny permission (don't allow)
4. Click on the Microphone again
5. It says Not Now, Allow

**Expected behavior**
For step 5, it should say Not Now/ Go To Settings

**Screenshots**
See ticket above with what it should be

**Additional context**
Clicking on Allow the second time (after an original denial) takes someone to settings, which is the correct behavior. But the button text is wrong.

cimigree:
closes #831 

### Description
This PR removes some old console.logs.
Also! 
1. it makes sure that the text on the permission modal is correct. It should say allow, not now if someone has not denied permissions.
2. Once someone has denied permissions, it should say Go to Settings, which now it does.
I had to add the useEffect so that when someone returns from settings, the behavior is correct. There were many ways to achieve this! I hope you like the way I landed on. I considered at least 4 other ways that also worked.

As a reminder, `adb shell pm reset-permissions` will reset the permissions (from "granted") but if they are denied, they will remain "denied"... Not sure why that is.

### To test
1. Start an observation. 
2. Press microphone. Modal should open and say "Allow/ Not Now". Press Not Now. Modal should close.
3. Press microphone. Modal should say "Allow/ Now Now". Press Allow. Grant permission. You should navigate to the the audio recording screen.
4. Run `adb shell pm reset-permissions`
5. Reopen the app, grant other permissions
6. Press microphone. Modal should now say Not Now, Go to Settings. Hit Not Now and modal should close.
7. Press microphone. Modal should now say Not Now, Go to Settings. Hit Go to Settings and you should Go to settings. Don't enable microphone permissions. Modal should still be open. You can hit Not Now and it will close.
8. Press microphone. Modal should now say Not Now, Go to Settings. Hit Go to Settings and you should Go to settings. Enable Microphone permissions. Use back button to go back to the app. You should navigate to audio now (sorry there is some flashing as the sheet closes).

### DIFF
--- a/src/frontend/sharedComponents/ActionRow.tsx
+++ b/src/frontend/sharedComponents/ActionRow.tsx
@@ -55,6 +55,8 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
   });
 
   const [shouldNavigateToAudio, setShouldNavigateToAudio] = useState(false);
+  const [permissionStatus, setPermissionStatus] =
+    useState<Audio.PermissionStatus | null>(null);
 
   const handleCameraPress = () => {
     navigation.navigate('AddPhoto');
@@ -65,6 +67,7 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
 
   const handleAudioPress = useCallback(async () => {
     const {status} = await Audio.getPermissionsAsync();
+    setPermissionStatus(status);
     if (status === 'granted') {
       navigation.navigate('Audio', {
         isEditing: currentRoute?.name === 'ObservationEdit',
@@ -118,6 +121,8 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
         <PermissionAudioBottomSheetContent
           closeSheet={closeAudioPermissionSheet}
           setShouldNavigateToAudioTrue={() => setShouldNavigateToAudio(true)}
+          permissionStatus={permissionStatus}
+          isOpen={isAudioPermissionSheetOpen}
         />
       </BottomSheetModal>
     </>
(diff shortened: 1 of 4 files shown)


ErikSin:
Just to clarify. What was happening before was the user was setting the permissions in setting, but the app was improperly detecting when that permission was being set?

cimigree:
@ErikSin 
> Just to clarify. What was happening before was the user was setting the permissions in setting, but the app was improperly detecting when that permission was being set?

There were two things wrong.
Problem number one and the one that Ximena noticed, was just that the text in the modal was wrong. If you deny permissions, the next time you press the microphone the modal is supposed to say Go To Settings (see issue https://github.com/digidem/comapeo-mobile/issues/359 ) and it was saying Allow/ Not now instead. So I fixed that so it now says Go To Settings if you had previously denied permission.

Problem number two, once the user navigated back to the app from settings giving permission for the microphone, the behavior wasn't right  because the modal was still open saying Allow/ Not now and if you clicked allow, you would go to settings again... You had to press Not Now, the modal would close and then when you pressed the microphone again then you would correctly be navigated to Audio. So I fixed that.

src/frontend/sharedComponents/ActionRow.tsx
```diff
@@ -55,6 +55,8 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
   });
 
   const [shouldNavigateToAudio, setShouldNavigateToAudio] = useState(false);
+  const [permissionStatus, setPermissionStatus] =
```
ErikSin: perhaps call this `AudioPermission`, or something that includes "Audio"

ErikSin (approved):
One small suggestion, but otherwise it looks good!

PR #844, opened 2024-11-13
feat: Shows Remote Detection Alerts

### LINKED ISSUE #811: Alerts Detail
**What we shared with CMI -**
CMI or partner submits new alerts to a Mapeo Cloud API endpoint (different endpoint for each project) in GeoJSON format, with a strictly specified schema for GeoJSON feature properties.
Mapeo user syncs with Mapeo Cloud when online.
Other Mapeo users sync with each other, sharing the alerts offline.
All users who have synced see an overlay on their map showing alerts.

**Limitations:**
- Alerts displayed on the map will not be interactive - there will be no info popups or ways for the user to click the alerts to see more information.
- Any info about the alerts will need to be displayed as part of the map style, e.g. by displaying a text layer as part of the map style that shows labels based on alert metadata.
- There will be no linking (at this stage) of observations with alerts.

**In order to implement these features we will need the following from CMI:**
- Details of the format of alerts data and what metadata it includes, in order to reach agreement on a schema for storage in Mapeo of alert data.
- A Style JSON ([Mapbox Style Specification](https://docs.mapbox.com/style-spec/guides/)) with layer definitions for each geometry type of alerts that you wish to display. This will be hard-coded in the app for display of alerts when they are available. Style spec needs to be compatible with Maps SDK for Android v9.3.0+.

**What CMI replied**

https://docs.google.com/document/d/1hTvhtltAswEehajb0TxE9E6XyK6xVVKteP-YuD_Tu80/edit?tab=t.0
Stylesheet for alerts
Here, we are rendering Polygons (with stroke) and LineStrings, and adding a label if the following fields are found: alert_type, month_detec, and year_detec.

It’s tricky to provide a hard-coded style for alerts that will work with many different background maps (imagery, vector) **but we have found that a bright, vivid red works best;** we are using this same style for alerts on our front end tooling.

CMI thinks that it could be good in the future to have these style properties be set as a fallback, to rename “alerts” to something more generalized like “overlay”, and to allow the Mapeo config designer to provide a style in their configuration (or elsewhere). But for now, we have named things as “alerts” in the layers and sources. Awana should feel free to adopt whatever approach you deem best for now.

ErikSin:
Closes #811

image


gmaclennan (commented):
Not a thorough review, but read it through on my phone and no red flags from me.

PR #845, opened 2024-11-14
fix: delete map shows confirmation twice

### LINKED ISSUE #833: Map / Delete maps shows confirmation x2
https://www.notion.so/digidem/Map-Delete-maps-shows-confirmation-x2-1351b08162d58044a695ccd534c7f057

cimigree:
closes #833 

### Description
The problem: We were closing the modal immediately after initiating the mutation, but the mutation itself triggers a state change that might cause the modal to re-open. Specifically, the onSuccess handler calls refresh(), which updates the refreshToken and may cause components depending on it to re-render, such as the BackgroundMaps which would then rerender the BottomSheetModal (would still be isOpen true)

Solution: Move the close sheet into the onSuccess handler of removeCustomMapMutation, so the modal closes only after the mutation completes and any state changes settle.

I built an APK to test this because it wasn't happening for me while debugging

### To test
You can download map files from here: https://drive.google.com/drive/folders/19iZUWrlTQIBpqr7MnP0u5E5dmFudJYKt?ths=true if you need to
App Settings -> Map Management -> Background Maps -> Choose File
Hit Close once the modal appears
UI Box with Map and Date should be on the screen
Hit Remove Map -> Delete Map
Modal should close and Box with Map should no longer be there
If you go to the map screen it should be back to the default map

### DIFF
--- a/src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
+++ b/src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
@@ -14,7 +14,7 @@ import {
 import ErrorSvg from '../../../images/Error.svg';
 import GreenCheckSvg from '../../../images/GreenCheck.svg';
 import noop from '../../../lib/noop';
-import {DARK_GREY, RED, WHITE} from '../../../lib/styles';
+import {RED, WHITE} from '../../../lib/styles';
 import {
   BottomSheetModal,
   BottomSheetModalContent,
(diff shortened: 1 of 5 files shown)


src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
ErikSin, line 1: Can you replace the < Text/> components in the file?

cimigree:
@ErikSin I hope it is ok but while I was doing the text stuff I just thought I would fix it in all of the background map files... Some of what was here didn't match the Figma, so I did everything to match the Figma (if there was a Figma)

ErikSin (approved):
I just noticed one more small thing. The 'Settings/MapManagement/index` uses a < ScrollView > < List /> </ ScrollView > . Can you update that to use the  < FullScreenMenuList /> component?

PR #847, opened 2024-11-14
Fix: share button sometimes not working after first share

### LINKED ISSUE #834: Share button /Inconsistent
https://www.notion.so/digidem/Share-difficulty-again-11a1b08162d5806e94e9d39378870e2f

cimigree:
closes #834 

### Description

- I came across this when developing Audio
- The urls for the attachments only work once for converting into readable files.
- So, in the Share button, after the first Share was pressed, the network call for converting the url to a base64 image was failing and no amount of time waiting would fix that.
- This PR uses the getUrl query to fetch a new url for each attachment every time the share button is pressed because the hook can't be used in regular functions or conditionally.
- We can't set the base64 in state because technically opening the share menu is putting the app in the background and state would clear out when share is pressed (something else I came across in the Audio development)
- This is working well, so hoping it works for you too!
- I also added failOnCancel: false so there is no error thrown if a user decides not to share the observation

### To test

1. Create an observation or observations with attachments and some without
2. Share them as many times as you want without leaving the observation screen.
3. I can't test on What's App so please test on What's App for me

### DIFF
--- a/src/frontend/screens/Observation/Buttons.tsx
+++ b/src/frontend/screens/Observation/Buttons.tsx
@@ -6,7 +6,6 @@ import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
 import {defineMessages, useIntl} from 'react-intl';
 import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
 import {useDeleteObservation} from '../../hooks/server/observations';
-import {Text} from '../../sharedComponents/Text';
 import Share from 'react-native-share';
 import {useObservationWithPreset} from '../../hooks/useObservationWithPreset.ts';
 import {formatCoords} from '../../lib/utils.ts';
@@ -17,6 +16,8 @@ import * as Sentry from '@sentry/react-native';
 import {CoordinateFormat} from '../../sharedTypes/index.ts';
 import {getValueLabel} from '../../sharedComponents/FormattedData.tsx';
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
+import {BodyText} from '../../sharedComponents/Text/BodyText.tsx';
+import {isPhotoOrAudio} from '../../lib/attachmentTypeChecks.ts';
 
 const m = defineMessages({
   delete: {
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -8,16 +8,15 @@ import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
 import {useDeleteObservation} from '../../hooks/server/observations';
 import {Text} from '../../sharedComponents/Text';
```
ErikSin: can you replace this Text Element?

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -117,36 +113,47 @@ export const ButtonFields = ({
     ]);
   }
 
-  async function handlePressShare() {
+  async function fetchFreshUrls() {
```
ErikSin, line 117: I think this can be simplified a bit. 2 main things. if you await`Promise.all`, it resolves (or rejects) all the promises inside the array. So it is redundant (and in certain edgecases lead to unwanted behaviour) to await all the objects inside the array. 

Secondly you can prefilter the array using the functions you created to determine if it is type audio or photo before you put them in the `promise.all`. That way you dont have to do a bunch of checks afterwards for null/undefined as well as type photo and audio

Here is what I would suggest

```ts

//defined outside the react component:
function isPhotoOrAudio(
  attachment: Attachment,
): attachment is Attachment & {type: 'photo' | 'audio'} {
  return isSavedPhoto(attachment) || isAudioAttachment(attachment);
}

// inside the react component:
async function fetchFreshUrls() {
    const {attachments} = observation;

    if (attachments.length === 0) {
      return [];
    }

    return await Promise.all(
      attachments.filter(isPhotoOrAudio).map(attachment => {
        return projectApi.$blobs.getUrl({
          driveId: attachment.driveDiscoveryId,
          name: attachment.name,
          type: attachment.type,
          variant: 'original',
        });
      }),
    );
  }

ErikSin (changes requested):
just left a comment about simplifying the code

src/frontend/screens/Observation/Buttons.tsx
cimigree: [9af300f](https://github.com/digidem/comapeo-mobile/pull/847/commits/9af300f79d54f28a85d7487d12fb983f71421c2f)

src/frontend/screens/Observation/Buttons.tsx
cimigree: [9af300f](https://github.com/digidem/comapeo-mobile/pull/847/commits/9af300f79d54f28a85d7487d12fb983f71421c2f)

ErikSin (approved):
Seems to be working well! Thanks for doing that

PR #849, opened 2024-11-18
feat: add number as field

### LINKED ISSUE #836: Number field needed for configuration questions
https://www.notion.so/digidem/Number-field-needed-for-configuration-questions-57cc870f07c84e09a07b34eb219532b1

Erik to investigate whats possible

ErikSin:
closes #836

### DIFF
--- a/src/frontend/screens/ObservationFields/Number.tsx
+++ b/src/frontend/screens/ObservationFields/Number.tsx
@@ -20,7 +20,8 @@ export const Number = React.memo<{field: Field}>(({field}) => {
             field.tagKey,
             newVal
               .replace(/[^0-9.-]/g, '') // Allow digits, decimal, and negative sign
-              .replace(/(?!^)-/g, ''), // Remove any minus sign that is not at the start
+              .replace(/(?!^)-/g, '') // Remove any minus sign that is not at the start
+              .replace(/(\..*?)\./g, '$1'), // Remove additional decimal points
           )
         }
         keyboardType="numeric"
--- a/src/frontend/screens/ObservationFields/Number.tsx
+++ b/src/frontend/screens/ObservationFields/Number.tsx
@@ -16,7 +16,12 @@ export const Number = React.memo<{field: Field}>(({field}) => {
         testID="OBS.details-inp"
         value={typeof value === 'string' ? value : ''}
         onChangeText={newVal =>
-          updateTags(field.tagKey, newVal.replace(/[^0-9]/g, ''))
+          updateTags(
+            field.tagKey,
+            newVal
+              .replace(/[^0-9.-]/g, '') // Allow digits, decimal, and negative sign
+              .replace(/(?!^)-/g, ''), // Remove any minus sign that is not at the start
+          )
         }
         keyboardType="numeric"
         style={styles.textInput}
--- a/src/frontend/screens/ObservationFields/Number.tsx
+++ b/src/frontend/screens/ObservationFields/Number.tsx
@@ -1,0 +1,45 @@
+import * as React from 'react';
+import {StyleSheet, TextInput} from 'react-native';
+import {QuestionLabel} from './QuestionLabel';
+import {Field} from '@comapeo/schema';
+import {usePersistedDraftObservation} from '../../hooks/persistedState/usePersistedDraftObservation';
+import {useDraftObservation} from '../../hooks/useDraftObservation';
+
+export const Number = React.memo<{field: Field}>(({field}) => {
+  const tags = usePersistedDraftObservation(store => store.value?.tags);
+  const {updateTags} = useDraftObservation();
+  const value = tags ? tags[field.tagKey] : '';
+  return (
+    <React.Fragment>
+      <QuestionLabel field={field} />
+      <TextInput
+        testID="OBS.details-inp"
+        value={typeof value === 'string' ? value : ''}
+        onChangeText={newVal =>
+          updateTags(field.tagKey, newVal.replace(/[^0-9]/g, ''))
+        }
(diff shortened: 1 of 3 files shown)


src/frontend/screens/ObservationFields/QuestionLabel.tsx
```diff
@@ -1,21 +1,24 @@
...
+import {HeaderText} from '../../sharedComponents/Text/HeaderText';
+import {Header} from 'react-native/Libraries/NewAppScreen';
 
 interface Props {
   field: Field;
 }
 
 export const QuestionLabel = ({field}: Props) => {
-  const hint = <FormattedFieldProp field={field} propName="placeholder" />;
+  // const hint = <FormattedFieldProp field={field} propName="placeholder" />;
```
cimigree: Does the commented code and console log have a purpose here? If not, can you please remove or adjust?

src/frontend/screens/ObservationFields/QuestionLabel.tsx
```diff
@@ -1,21 +1,24 @@
 import * as React from 'react';
 import {View, StyleSheet} from 'react-native';
 import {FormattedFieldProp} from '../../sharedComponents/FormattedData';
-import {Text} from '../../sharedComponents/Text';
 import {Field} from '@comapeo/schema';
+import {HeaderText} from '../../sharedComponents/Text/HeaderText';
+import {Header} from 'react-native/Libraries/NewAppScreen';
```
cimigree: The import for Header is not being used/ is a mistake probably

src/frontend/screens/ObservationFields/Number.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {Field} from '@comapeo/schema';
+import {usePersistedDraftObservation} from '../../hooks/persistedState/usePersistedDraftObservation';
+import {useDraftObservation} from '../../hooks/useDraftObservation';
+
+export const Number = React.memo<{field: Field}>(({field}) => {
+  const tags = usePersistedDraftObservation(store => store.value?.tags);
+  const {updateTags} = useDraftObservation();
+  const value = tags ? tags[field.tagKey] : '';
+  return (
+    <React.Fragment>
```
cimigree, line 13: should there be/ can there be a minimum and/ or maximum? And is it intended that negative numbers and decimals aren't allowed?

cimigree (changes requested):
I am sure this will be a great addition! I have a few minor questions and comments.

src/frontend/screens/ObservationFields/Question.tsx
```diff
@@ -22,5 +23,9 @@ export const Question = ({field}: QuestionProps) => {
     return <SelectMultiple field={field as SelectMultipleField} />;
   }
 
+  if (field.type === 'number') {
```
cimigree, line 26: Could this be a map to be a little more scalable/ elegant? Something like 
```
const FIELD_COMPONENTS: Record<string, React.ComponentType<{ field: Field }>> = {
  selectOne: SelectOne,
  selectMultiple: SelectMultiple,
  text: TextArea,
  number: Number,
};

export const Question = ({ field }: QuestionProps) => {
  const FieldComponent = FIELD_COMPONENTS[field.type] || TextArea;
  return <FieldComponent field={field} />;
};

```

src/frontend/screens/ObservationFields/QuestionLabel.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/849/commits/a5dd8f98482e4cddb5678275bdd2b98201ec362e

src/frontend/screens/ObservationFields/QuestionLabel.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/849/commits/a5dd8f98482e4cddb5678275bdd2b98201ec362e

src/frontend/screens/ObservationFields/Question.tsx
ErikSin: I tried this but there are some annoying typing issues that arise.  the `field` prop for each of these components are actually slightly modified subsets of `Field` in schema. This is because the React Component property has certain properties that are necessary for the UI but not necessary for the actual schema.

I tried widening the definition of `React.ComponentType<{ field: Field }>>` to include the other types of props, but that lead into a TS battle of narrowing props that ended up making the mapping still complicated

cimigree (commented):
Do you think it's ok that numbers like this are allowed? Should we just trust the observers?

<image src="https://github.com/user-attachments/assets/0f18d6f2-1863-40e6-90fa-7071a0a50fa2" width="250" />

ErikSin:
> Do you think it's ok that numbers like this are allowed? Should we just trust the observers?

Its fixed now!

src/frontend/screens/ObservationFields/Number.tsx
```diff
@@ -20,7 +20,8 @@ export const Number = React.memo<{field: Field}>(({field}) => {
             field.tagKey,
             newVal
               .replace(/[^0-9.-]/g, '') // Allow digits, decimal, and negative sign
-              .replace(/(?!^)-/g, ''), // Remove any minus sign that is not at the start
+              .replace(/(?!^)-/g, '') // Remove any minus sign that is not at the start
```
cimigree, line 23: This works, but you might consider `.replace(/^(-?)(\d*)(\.?\d*).*/, '$1$2$3') `instead to have one line for the regex. It does the same thing but in one line instead of two...

cimigree (approved):

PR #850, opened 2024-11-18
feat: show count of other devices on sync screen

EvanHahn:
This reverts commit 78c5be40fbc36c4bf0e4d417353c03e5964123d9, where we removed these counts.

I *think* this problem is fixed.

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -1527,13 +1527,16 @@
     "message": "All data synced"
   },
   "screens.Sync.ProjectSyncDisplay.devicesAvailableToSync": {
-    "message": "Devices available"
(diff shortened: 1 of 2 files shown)


ErikSin:
~~This was done on purpose. The count was often inaccurate and it was requested that no count was shown as that was less confusing.~~

Sorry, I realize now you are saying that this problem has been fixed in the backend, so the count should be accurate now.

EvanHahn:
This still seems to be a bug (see https://github.com/digidem/comapeo-core/issues/872#issuecomment-2537298587). Closing, because we don't want to merge this.

PR #851, opened 2024-11-18
Fixes crash when permission is taken away.

### LINKED ISSUE #848: Tracks/ Give permission then remove via device settings - Crash
https://www.notion.so/digidem/Tracks-Give-permission-then-remove-via-device-settings-Crash-13e1b08162d580db9f75c08250ba66ba

cimigree:
closes #848 

### Description

- The error that was showing when permission was removed while a track is in progress was `startDate.getTime() is undefined or is not a function.` (or something like that)
- The error is fixed by the change in this PR
- Now you can start a track, revoke permission, go back to the app, and when you click on the Track dude in the bottom tabs, the "Use Your Location" modal pops up.
- The timer is still timing though...
- It looks like the track is continuous but it is hard to tell because the space I have wifi is limited
- This doesn't solve the problem of what to include in the track but it does fix the crash as far as I can tell
- I built it as a release candidate and the behavior was the same -- no crash

### DIFF
--- a/src/frontend/hooks/useFormattedTimeSince.ts
+++ b/src/frontend/hooks/useFormattedTimeSince.ts
@@ -3,7 +3,7 @@ import {Duration} from 'luxon';
 
 export const useFormattedTimeSince = (start: Date | null, interval: number) => {
   const [currentTime, setCurrentTime] = useState(new Date());
-  let startDate = start ? start : new Date();
+  let startDate = start ? new Date(start) : new Date();
 
   useEffect(() => {
     setCurrentTime(new Date());


ErikSin:
This fixes the crash but compromises the data integrity. If it is still able to take a track while the location service is off, the track will not be accurate. Obviously the app crashing was not ideal, but we need a way of catching the crash and doing something instead of just ignoring it and continuing on

cimigree:
@ErikSin I don't think there is a crash though... And that error is a real error. The startDate is not being read as a date. So I am not clear what you mean or what to do next.

ErikSin:
So previously if the user turned off their permission the app would crash. This behaviour meant that tracks would stop tracking (as well as blow up the entire app). With your current fix, the app doesn't crash anymore BUT tracking still remains. If the user was turn it back on, it would create a straight line from the last point to the new point once it has been turned back on. This is not ideal behaviour as now we are creating a false straight line and effecting the integrity of the data.

So we need to find an inbetween where the app does not crash, BUT it should turn off tracks.

ErikSin (approved):

PR #854, opened 2024-11-20
fix: removes audio from observation sharing

### LINKED ISSUE #812: Whatsapp - Sharing Obvs w Audio -doesnt share audio
https://www.notion.so/digidem/Whatsapp-Sharing-Obvs-w-Audio-doesnt-share-audio-1371b08162d5807d8883eb42714e328a?pvs=4

cimigree:
closes #812 
See issue for explanation.
From Gregor in Slack:

> This issue was discussed during co-design, and the limitation of being unable to share audio along with an observation was clearly explained, and that is why we added the ability to share audio attachments individually as a work-around / compromise. It does seem perhaps that that previous conversation has been forgotten - I tried to be very clear from the start that we would be unable to reliably support sending an observation with audio attachments via whatsapp. 

And Jen's response:

> I will need to dig into issues in coming days, but even if it is a known issue for whatsapp share, adding audio should not degrade the ability to send an observation with the photos, excluding audio files.

So, this PR removes audio attachments from being shared. At this point, the only type of attachments to share is of type photo, so the function focuses on just photos.

### DIFF
--- a/src/frontend/lib/attachmentTypeChecks.ts
+++ b/src/frontend/lib/attachmentTypeChecks.ts
@@ -81,7 +81,3 @@ export function isDraftPhoto(attachment: any): attachment is DraftPhoto {
 export function isPhoto(attachment: any): attachment is Photo {
   return isSavedPhoto(attachment) || isDraftPhoto(attachment);
 }
-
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks Cindy!

PR #855, opened 2024-11-21
fix: simplify technical implementation of importing custom map file

achou11:
The previous approach:

1. Pick a map file via the file picker, which **_copies the file to the app's cache storage_**.
2. If valid, **_move_** the file from the cache storage into the app's document storage.

Both steps have a performance overhead that is most noticeable when working with large files. The overhead in (1) is something we've run into before. The issue in (2) is more theoretical, with the suspicion that potentially expensive deletions need to happen internally when moving a file from one storage directory.

This PR changes the approach to the following:

1. Pick a map file via the file picker, which no longer copies it into the cache storage.
2. **_Copy_** the selected file into the app document storage using the SAF uri, which represents the original location of the selected file.

This should have no user-facing impact, aside from potential performance improvements due to doing less work internally.

### DIFF
--- a/src/frontend/lib/file-system.ts
+++ b/src/frontend/lib/file-system.ts
@@ -28,17 +28,22 @@ export async function selectFile(opts: {
     throw new Error();
   }
 
-  const hasValidExtension = opts.extensionFilters
-    ? opts.extensionFilters.some(extension =>
-        asset.uri.endsWith(`.${extension}`),
-      )
-    : true;
-
-  if (!hasValidExtension) {
-    FileSystem.deleteAsync(asset.uri).catch(err => {
-      console.log(err);
+  if (opts.extensionFilters) {
+    const hasValidExtension = opts.extensionFilters.some(extension => {
+      return asset.name.endsWith(`.${extension}`);
     });
-    throw new Error('Invalid extension');
+
+    if (!hasValidExtension) {
+      if (opts.copyToCache) {
+        FileSystem.deleteAsync(asset.uri, {
+          idempotent: true,
+        }).catch(err => {
+          console.log(err);
+        });
+      }
+
+      throw new Error('Invalid extension');
+    }
   }
 
   return asset;
--- a/src/frontend/lib/file-system.ts
+++ b/src/frontend/lib/file-system.ts
@@ -9,14 +9,16 @@ export function convertFileUriToPosixPath(fileUri: string) {
 }
 
 // TODO: Some overlap with selectFile() from lib/utils but fixes some usage limitations. Ideally use this for everything
-export async function selectFile(opts: {
-  copyToCache?: boolean;
+export async function selectFile({
+  mimeFilters,
+  extensionFilters,
+}: {
   mimeFilters?: Array<string>;
   extensionFilters?: Array<string>;
-}) {
+} = {}) {
   const documentResult = await DocumentPicker.getDocumentAsync({
-    type: opts.mimeFilters,
-    copyToCacheDirectory: opts.copyToCache,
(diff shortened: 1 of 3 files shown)


src/frontend/lib/file-system.ts
```diff
@@ -28,17 +28,22 @@ export async function selectFile(opts: {
...
-        asset.uri.endsWith(`.${extension}`),
-      )
-    : true;
-
-  if (!hasValidExtension) {
-    FileSystem.deleteAsync(asset.uri).catch(err => {
-      console.log(err);
+  if (opts.extensionFilters) {
+    const hasValidExtension = opts.extensionFilters.some(extension => {
+      return asset.name.endsWith(`.${extension}`);
```
achou11, line 35: just highlighting the subtle change here. using the `name` is more reliable and agnostic to the `copyToCache` option because the uri only contains the original extension when copying to the cache.

src/frontend/lib/file-system.ts
```diff
@@ -28,17 +28,22 @@ export async function selectFile(opts: {
...
-    throw new Error('Invalid extension');
+
+    if (!hasValidExtension) {
+      if (opts.copyToCache) {
+        FileSystem.deleteAsync(asset.uri, {
+          idempotent: true,
+        }).catch(err => {
+          console.log(err);
+        });
+      }
```
achou11: we only need to attempt deletion of the cached file if the `copyToCache` option is specified

src/frontend/screens/Settings/MapManagement/BackgroundMaps.tsx
```diff
@@ -154,11 +153,6 @@ export function BackgroundMapsScreen() {
                       uri: asset.uri,
                     },
                     {
-                      onError: () => {
-                        FileSystem.deleteAsync(asset.uri, {
-                          idempotent: true,
-                        }).catch(noop);
-                      },
```
achou11, line 161: the asset here used to refer to the one in the cache directory, but given that we're no longer doing that, there's no need for this attempted deletion (i.e. we don't want to attempt to delete the original file, which I think fails anyways)

gmaclennan:
I would just drop the `opts.copyToCache` option. It's not being used yet, and it's probably not a great idea in most circumstances. You can add it if needed in the future, although probably in most cases it's going to be better to copy _after_ getting a return value from the document picker, so that you can show better user feedback.

gmaclennan:
> I would just drop the `opts.copyToCache` option. It's not being used yet, and it's probably not a great idea in most circumstances. You can add it if needed in the future, although probably in most cases it's going to be better to copy _after_ getting a return value from the document picker, so that you can show better user feedback.

e.g. always set it to `false` and don't expose it as an option for useSelectFile().

achou11:
> I would just drop the `opts.copyToCache` option. It's not being used yet, and it's probably not a great idea in most circumstances. You can add it if needed in the future, although probably in most cases it's going to be better to copy _after_ getting a return value from the document picker, so that you can show better user feedback.

yeah that's valid. had kept it in there because the result of `DocumentPicker.getAsync()` provides more useful information if copying to the cache directory, at least based on my memory of how we used it in Mapeo Mobile.

achou11:
@gmaclennan I implemented your suggestion via b1db178accbe10e08e7ebc9c2c9ffa2536e8603d

cimigree (approved):
There was a lot of flashing during the import (maybe more than before? not sure) and seemed to take longer strangely, but it did work fine in the end. Your logic and changes make sense! And I see the difference in the uri and the name!

PR #878, opened 2024-12-05
fix: sizing of full screen modal

### LINKED ISSUE #875: Action buttons on full-screen variant of bottom modal sheet are not visible on the screen
**Describe the bug**

When you perform some kind of action in the app that may fail, we typically open a modal-like bottom sheet to display that an error has occurred. This is a full-screen view that has buttons for taking action at the bottom of the sheet. However, as of 718496c3ec07bf8fced4c8ef56fc3ee98bbfaf2c, these buttons are not displaying properly (either being completely off-screen or partially). As a result, there's no way to dismiss this sheet and the user has to close and re-open the app, which is disruptive and not intended.

We have two variants of this sheet:

- one that gives the option of just going back, which currently looks like this (no actionable buttons present at all): 

  

- another that gives an option to go back and an option to try again (the button to go back is pushed further down than expected and the try again button is just off the screen entirely):

  

The change in https://github.com/digidem/comapeo-mobile/commit/b87c15ec78c2ffb9b76899108d405c5d7818394e#diff-69ce228e0685c294fef05e9dff8a5e6ee8765353a4a0fd10627862d1bcc17deaR63 seems to have introduced this issue. Was able to fix it by reverting it, which probably reintroduces the issue it was meant to solve originally 🥲 (i didn't test that specifically).

**To Reproduce**

Any action that can trigger a failure to open up the error sheet will suffice. I ran into this in the custom map import flow i.e.

1. Go to map management screen
2. Press `Choose file`
3. Choose a file with a non-accepted format (e.g. `.txt`, `.pdf`, `.jpg`, etc)
4. Error sheet shows up

**Expected behavior**

- In the case of the `Go Back` variant, it should look like this:

  

- In the case of the `Go Back` + `Try Again` variant, it should look like this:

cimigree:
closes #875 

- Adds snappoints prop to bottom sheet modal to help sizing on full screen modals.
- The original issue was that the project invite sheet wasn't occupying the full screen when fullScreen was true, so a user couldn't see or check the box that is in there
- If we set snapPoints={['100%']} when fullScreen is true that makes the modal "snap to" 100% of the screen height, so now the modal always occupies the full screen when needed and doesn't affect any other modals

This PR also:

- Reverts modal sizing to 100% so the modals that already were working aren't broken anymore
- Reverts the reversal of the top margin on the audio permission sheet because that is no longer needed

Here is a screen shot of the modal that had been causing the issue that made me change the height in the Content file which caused the other modals to break -- now it is good!

<image src="https://github.com/user-attachments/assets/09868c6b-38f7-4b8e-b5f6-3faec6dc1c06" width="250" />

### DIFF
--- a/src/frontend/screens/Audio/PermissionAudioBottomSheetContent.tsx
+++ b/src/frontend/screens/Audio/PermissionAudioBottomSheetContent.tsx
@@ -103,22 +103,24 @@ export const PermissionAudioBottomSheetContent: FC<
         : t(m.allowButtonText);
 
   return (
-    <BottomSheetModalContent
-      icon={<AudioPermission />}
-      title={t(m.title)}
-      description={t(m.description)}
-      buttonConfigs={[
-        {
-          variation: 'outlined',
-          onPress: closeSheet,
-          text: t(m.notNowButtonText),
-        },
-        {
-          variation: 'filled',
-          onPress: onPressActionButton,
-          text: actionButtonText,
-        },
-      ]}
-    />
+    <View style={{paddingTop: 80}}>
+      <BottomSheetModalContent
+        icon={<AudioPermission />}
+        title={t(m.title)}
+        description={t(m.description)}
+        buttonConfigs={[
+          {
+            variation: 'outlined',
+            onPress: closeSheet,
+            text: t(m.notNowButtonText),
+          },
+          {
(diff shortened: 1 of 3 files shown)


achou11 (approved):
changes lgtm, thanks for the fixes!

PR #880, opened 2024-12-09
style: add white background to DrawerMenuIcon

luandro:
Adds a white background to the DrawerMenuIcon to improve visibility and contrast, important for when using a custom background map which makes the button invisible.

Logged issue: https://www.notion.so/digidem/Menu-not-visible-with-satellite-imagery-14a1b08162d580d1a5aef5362d809991?pvs=4

### Before
comapeo_dark_menu

### After
mapeo_background_button


ErikSin:
Im going to close this for now. Any design changes have to go through codesign/design. So we can reopen if this goes through the right channel and is approved :)

PR #881, opened 2024-12-09
style: add close icon to DrawerMenuIcon

luandro:
Adds a close icon option to the DrawerMenuIcon component, making ui clearer.

comapeo_close_button


ErikSin:
@luandro this needs to go to codesign before we can make any changes.

PR #882, opened 2024-12-09
chore: Add working helper text

luandro:
Updated the helper text in QuestionLabel to use the old but working placeholder and added hint style for a more consistent look and feel.

### DIFF
--- a/src/frontend/screens/ObservationFields/QuestionLabel.tsx
+++ b/src/frontend/screens/ObservationFields/QuestionLabel.tsx
@@ -13,7 +13,11 @@ export const QuestionLabel = ({field}: Props) => {
       <HeaderText variant="header3">
         <FormattedFieldProp field={field} propName="label" />
       </HeaderText>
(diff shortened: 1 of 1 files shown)


gmaclennan:
When this is not working, is it because the config defines a `placeholder` but not `helperText`, or is there a bug elsewhere? I'm not sure this is the right place to fix this, because it may cause us headaches in the future. If it's not a bug, but a need for a workaround to use the `placeholder` prop from config as helper text, and it's too complicated to update the configs, then maybe we can add a workaround to either preset import or preset builder e.g. `preset.placeholder = preset.placeholder || preset.helperText`

luandro:
Your totally right, there's no bug. Just built a config using `helperText` instead of `placeholder` and it works perfectly. I think there might have been a bug at the time I built the first CoMapeo cat files, where `helperText` wasn't working and only `placeholder` worked. There already are CoMapeo cats with `placeholder` in the wild, so ` preset.placeholder = preset.placeholder || preset.helperText` solution might be a good one.

PR #883, opened 2024-12-09
Chore refactor observation track with test

ErikSin:
Refactors the track list to be testable. Includes separating the logic for finding a track, writing tests for that logic, and simplifying to component to simply render views instead of processing logic

image
image


src/frontend/screens/Observation/TrackAccordian.tsx
```diff
@@ -14,29 +17,39 @@ const m = defineMessages({
...
-        <TrackListItem
-          track={track}
-          onPress={() => {
-            navigation.push('Track', {trackId: track.docId});
-          }}
-          testID={`trackListItem:${track.docId}`}
-        />
-      }
-    />
+  return !track ? null : (
```
achou11: stylistic preference but would suggest either an early return or switching the ternary here

src/frontend/screens/Observation/findAssociatedTrack.test.ts
```diff
@@ -0,0 +1,164 @@
+import {Track} from '@comapeo/schema';
+import {findAssociatedTrack} from './findAssociatedTrack';
+
+const tracks: Track[] = [
```
achou11: ideally this uses [`@mapeo/mock-data`](https://github.com/digidem/mapeo-mock-data) to avoid potentially becoming misaligned or outdated in the future.

src/frontend/screens/Observation/findAssociatedTrack.ts
```diff
@@ -0,0 +1,14 @@
+import {type Track} from '@comapeo/schema';
+
+export function findAssociatedTrack({
+  tracks,
+  observationId,
+}: {
+  tracks: Track[] | undefined;
+  observationId: string;
+}) {
+  if (!tracks) return undefined;
```
achou11: when seeing something like this I usually adjust the implementation so that it's not handling as many inputs. It's a bit easier to test and I subjectively find it a bit more readable. Also, in the case of where this function is actually being used in the app, the input for `tracks` is never undefined, so not sure if accounting for that here is necessary.

```ts
export function findAssociatedTrack({
  tracks,
  observationId,
}: {
  tracks: Track[];
  observationId: string;
}) {
  return tracks.find(trackData =>
    trackData.observationRefs.some(ref => ref.docId === observationId),
  );
}
```

If `tracks` is `undefined` for whatever reason, let the caller of the function handle that i.e.

```ts
const trackOfInterest = allTracks === undefined ? undefined : findAssociatedTrack({ tracks: allTracks, observationId })

// Wouldn't suggest this but alternatively, if you really want to avoid a ternary:
const trackOfInterest = findAssociatedTrack({ tracks: allTracks || [], observationId })
```

src/frontend/screens/Observation/findAssociatedTrack.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  it('returns the correct track with the associated observationId', () => {
+    const track = findAssociatedTrack({tracks, observationId: 'ghi789'});
+    expect(track?.docId).toBe('def456');
+  });
+
+  it('returns undefined when there is no matching observation docId', () => {
+    const track = findAssociatedTrack({tracks, observationId: 'qwerty'});
+    expect(track).toBeUndefined();
+  });
+});
```
achou11: at a glance, not sure if I find these tests very useful as is. maybe it's because of the hardcoding of a single value you're testing against? usually testing against hardcoded values is more helpful when you have a wider set of fixture data to work with.
 
I also think part of it is because I have very little inclination of what the "correct" result is when using the function. Taking this example:

```ts
    const track = findAssociatedTrack({tracks, observationId: 'ghi789'});
    expect(track?.docId).toBe('def456');
```

I read this test as saying: "If you have these specific tracks and you provide observation id ghi789, you should get the track with doc id def456."

From my perspective, this feels opaque in the sense that I have to make sure that the data being tested against actually makes sense. I have to evaluate not only the implementation of the function being tested but the **implementation of the data in the test** as well. It's a little less reassuring with a single hardcoded value, as I have to ensure that it looks like an appropriate fixture value.

I think ideally what would happen is that the test would clarify the relation between the observation and the tracks being tested, such that it becomes apparent what the actual result should be as a person reading the test. For example, each of the following situations should work with their own data to replicate the situations of interest:

- The observation has nothing to do with the tracks being tested against i.e. you have an observation document and an array of unrelated track documents.

  ```ts
  it('returns undefined when no associated track is found', () => {
    const [observation] = generate('observation');

    // Mostly for TS
    assert(observation);

    const tracks = generate('track', {count: 10});

    const foundTrack = findAssociatedTrack({
      tracks,
      observationId: observation.docId,
    });

    expect(foundTrack).toBeUndefined();
  });
  ```

- The observation is referenced by one of the tracks in the test i.e. you have an observation document and an array of track documents, one of which is updated so that it points to the observation of interest.

  ```ts
  it('returns the correct track with the associated observationId', () => {
    const [observation] = generate('observation');

    // Mostly for TS
    assert(observation);

    const tracks = generate('track', {count: 10});

    // Can pick a specific one or randomly choose, doesn't matter given how this test is written
    const targetTrack = tracks[3];

    // Mostly for TS
    assert(targetTrack);

    targetTrack.observationRefs.push({
      docId: observation.docId,
      versionId: observation.versionId,
    });

    const foundTrack = findAssociatedTrack({
      tracks,
      observationId: observation.docId,
    });

    expect(foundTrack).toStrictEqual(targetTrack);
  });
  ```

- There are no tracks to test against (trivial so no example)

The above example implementations are still a little bit flawed, but I find that they make the expected outcomes more noticeable to the reader of the test.

achou11 (commented):
Added few suggestions but nothing blocking. I think the test implementation could use some work but achieves the basic goal.

Seems like this PR is also fixing/adjusting a couple of things for the tracks UI - would update the PR title if that's the case.

achou11 (approved):
Meant to submit my last review as an approval - my b 😅

src/frontend/screens/Observation/findAssociatedTrack.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      highlights: ['beach', 'sunset'],
+    },
+  },
+];
+
+describe('Tests findAssociatedTrack', () => {
+  it('returns undefined when tracks prop is undefined', () => {
+    const track = findAssociatedTrack({tracks: undefined, observationId: '1'});
+    expect(track).toBeUndefined();
+  });
```
achou11: in tandem with the suggestion about removing the undefined tracks case, would update this to test when tracks is an empty array

ErikSin:
@mockdata is not playing well with my esm file. changing my test to an `mts` file makes the test pass, but eslint is very unhappy...any suggestions?

achou11:
ugh yeah more of an issue with jest and our environment than the mock-data module itself. there doesn't seem to be a trivial approach to getting it to work with jest. not sure what the best path forward is aside from sticking with something closer to hardcoded values I guess 😬 

---

Wouldn't recommend this, but one workaround is to run the CLI version of the mock-data module and read the data from that. Added this to the top of the test file, which works but is also pretty hacky 😅 

```ts
import {generate as mockGenerate} from '@mapeo/mock-data';
import {execSync} from 'node:child_process';

// you'd use this in the actual tests
const generate: typeof mockGenerate = (schema, opts) => {
  const result = execSync(
    `npx generate-mapeo-data --schema ${schema} --count ${opts?.count || 1}`,
    {
      encoding: 'utf8',
    },
  );

  return JSON.parse(result);
};
```

PR #884, opened 2024-12-09
chore: update `@comapeo/ipc` to 2.1.0

EvanHahn:
This adds support for `MapeoClientApi#getIsArchiveDevice` and `MapeoClientApi#setIsArchiveDevice`.

_I didn't test this manually._


cimigree (approved):
I ran the E2E tests and also tested with my new logic using the media sync settings and all looked good.

PR #885, opened 2024-12-09
chore: media management logic

### LINKED ISSUE #879: Add logic for media management
## Description
The Media Management toggle currently exists, but...
The logic needs to be added to make the toggle actually do something.
It currently exists, but is behind a feature flag.

### Todos:

- [ ] Add a hook file to house the queries to the backend related to setting and getting the media management settings
```
await mapeoApi.getIsArchiveDevice()
await mapeoApi.setIsArchiveDevice(true)
```

- [ ] Use that logic instead of the localstorage logic that is already in there `const {setMediaSyncSetting} = usePersistedSettingsAction();`
- [ ] Remove the logic related to usePersistedSettings && setMediaSyncSetting that is no longer needed
- [ ] Remove feature flag EXPO_PUBLIC_FEATURE_MEDIA_MANAGER=true

cimigree:
Closes #879 

### Description
- Adds new hook to use backend functions for getting and setting the media sync settings.
- Uses those new hooks in the media sync screen
- Takes the media sync settings out of local settings/ state
- Takes out the media management flag


src/frontend/hooks/server/mediaSync.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export function isArchiveDevice(value: MediaSyncSetting): boolean {
+  return value === 'everything';
+}
+
+export function useGetMediaSyncSetting() {
+  const api = useApi();
+
+  return useQuery({
```
ErikSin: were going to be switching all our queries to `suspenseQueries`, so if you don't mind doing that here.

src/frontend/screens/Settings/ProjectSettings/MediaSyncSettings.tsx
```diff
@@ -71,14 +71,17 @@ const m = defineMessages({
...
-    () => mediaSyncSetting,
-  );
+  const [modalType, setModalType] =
+    React.useState<MediaSyncSetting>('everything');
+
+  React.useEffect(() => {
+    if (!isLoading && mediaSyncSetting) {
+      setModalType(mediaSyncSetting);
+    }
+  }, [mediaSyncSetting]);
```
ErikSin: you're state syncing here, which should be avoided. if you switch to suspense, you wont need to worry about the loading state.

src/frontend/screens/Settings/ProjectSettings/MediaSyncSettings.tsx
```diff
@@ -120,7 +123,7 @@ export const MediaSyncSettings = () => {
   return (
     <ScrollView style={styles.container}>
       <SelectOne
-        value={mediaSyncSetting}
+        value={modalType}
```
ErikSin: this value should be `mediaSyncSetting` so it is always in line with actual data.

ErikSin: There is a slight problem as when the value is chosen, it will take a few seconds for the cache to refetch. Since it is aligned with the value of the cache, it will be a little delayed. You can avoid this by doing an [optimistic update](https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates)

ErikSin (changes requested):
good job. just some slight adjustments to account for the fact that we are going to be using suspense moving forward AND there is some state syncing that should be avoided (but the suspense will help with that)

src/frontend/hooks/server/mediaSync.ts
cimigree: Sorry! I didn't realize. Should be set now.

src/frontend/screens/Settings/ProjectSettings/MediaSyncSettings.tsx
cimigree: Ok

src/frontend/hooks/server/mediaSync.ts
```diff
@@ -0,0 +1,60 @@
...
+      return {previousSetting};
+    },
+    onError: (err, newSetting, context) => {
+      if (context?.previousSetting) {
+        queryClient.setQueryData(
+          [MEDIA_SYNC_SETTING_KEY],
+          context.previousSetting,
+        );
+      }
+    },
```
ErikSin: I think it would be simpler (and more maintainable) to [optimistic update via the ui](https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#via-the-ui)

You could remove all this code.

And in the component you could simply

```ts
const {mutate: setMediaSyncSetting, variable, isPending} = useSetMediaSyncSetting();

<SelectOne
        value={isPending ? variable: mediaSyncSetting}
        {...restProps}
      />

```

ErikSin (approved):
Thanks Cindy! Looks great

PR #891, opened 2024-12-11
chore: use Invite type directly from core

achou11:
There were a couple of places where we were using `MapBuffers<InviteInternal>` to represent an invite from core instead of `Invite`. I believe this was because the latter wasn't available/exported for some time, but [now it is](https://github.com/digidem/comapeo-core/blob/f073ca916f573ab66240f35805b4df5ecceb11e4/src/invite-api.js#L28). This PR shouldn't have any user-facing impact as it only replaces type imports.


cimigree (approved):
Makes sense. I tested out the project invite flow with a couple of different scenarios on my two devices and all worked fine.

PR #892, opened 2024-12-17
chore: add workflow for lockfile checks

achou11:
Copies over the workflow we've set up in [core](https://github.com/digidem/comapeo-core/blob/main/.github/workflows/lockfile.yml). On pull requests, it runs a job for each of the following:

1. checking that the lockfile version doesn't change. this usually happens if someone uses an incompatible package manager version to manage dependencies, which we want to prevent

2. highlighting changes in dependencies that the PR introduces, if any, by writing a PR comment with a table

Introducing this will be helpful for avoiding any surprise updates and highlighting intended changes moving forward

---

Note that this only does a check on the top-level lockfile. Ideally this would also account for the backend's lockfile, but haven't fully figured out how to get that working just yet.

### DIFF
--- a/.github/workflows/lockfile.yml
+++ b/.github/workflows/lockfile.yml
@@ -1,0 +1,34 @@
+name: Check Lockfile
+on:
+  pull_request:
+
+jobs:
+  lockfile_version:
+    name: Lockfile version check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repo
+        uses: actions/checkout@v4
+      - name: Check package-lock.json version has not been changed
+        uses: mansona/npm-lockfile-version@v1
+        with:
+          version: 3
+  lockfile_changes:
+    name: Lockfile changes check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repo
+        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


gmaclennan:
> Note that this only does a check on the top-level lockfile. Ideally this would also account for the backend's lockfile, but haven't fully figured out how to get that working just yet.

Run the action a second time and pass:

```
with:
  path: src/backend/package-lock.json
```

However it needs an update to our lockfile workflow fork to include the comment header when identifying which comment to update (the header contains the path), otherwise it will be overwritten.

gmaclennan:
https://github.com/digidem/npm-lockfile-changes/pull/2

achou11:
any ideas for how to go about the lockfile version check on the backend? doesn't accept a path input like the lockfile changes one, unfortunately

did a quick try with just cd-ing into the backend directory before running the version check action but that didn't seem to work. maybe i did something wrong there?

gmaclennan:
No not sure. There are probably some things we can do around checking out a subdirectory, or just the package-lock, but I don't think it's worth the effort at this point for this check.

achou11:
Decided to merge as is and will leave doing checks for the backend as a follow-up

PR #893, opened 2024-12-17
chore: upgrade production dependencies

ErikSin:
Upgraded all production dependencies and uses exact versioning for all dependecies.

Also adds a test that checks all dependencies use exact versioning


package.json
```diff
@@ -24,145 +24,147 @@
...
+    "expo-system-ui": "2.9.4",
+    "expo-task-manager": "11.7.3",
+    "expo-updates": "0.24.13",
+    "geojson": "0.5.0",
+    "geojson-geometries-lookup": "0.5.0",
+    "lodash.isequal": "4.5.0",
+    "luxon": "3.5.0",
+    "map-obj": "5.0.2",
+    "nanoid": "5.0.9",
+    "nodejs-mobile-react-native": "18.20.4",
```
achou11, line 86: for future reference, wondering if upgrading anything related to nodejs-mobile (or any runtime we use) should be left as a separate task just because of its importance and how it could affect other parts of our setup (e.g. native prebuilds)

src/frontend/hooks/persistedState/createPersistedState.ts
achou11, line 1: as a reminder, the changes here are related to the change in zustand's persistence middleware behavior ([link](https://zustand.docs.pmnd.rs/migrations/migrating-to-v5#persist-middleware-no-longer-stores-item-at-store-creation))

src/frontend/packageJson.test.ts
```diff
@@ -5,6 +5,7 @@ import * as semver from 'semver';
 describe('frontend package.json', () => {
   let frontendMapeoDependencies: Record<string, string>;
   let backendMapeoDependencies: Record<string, string>;
+  let allDependencies: Record<string, string>;
```
achou11: would rename this to `allFrontendDependencies` (and rename other related variables accordingly)

src/frontend/packageJson.test.ts
```diff
@@ -37,6 +40,13 @@ describe('frontend package.json', () => {
       expect(semver.satisfies(frontendVersion, backendVersion)).toBe(true);
     }
   });
+
+  it('all front end dependencies use exact version', () => {
+    for (const version of Object.values(allDependencies)) {
+      const isExact = semver.valid(version) !== null;
+      expect(isExact).toBe(true);
+    }
+  });
```
achou11: given this test, i think you can remove the test that does this for the mapeo-related frontend deps (the test with the title `"uses exact versions for @mapeo dependencies"`)

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -162,11 +162,13 @@ export const ButtonFields = ({
         completedFields.push({label: field.label, value: displayedValue});
       }
 
+      console.log({length: base64Urls.length});
```
achou11: remove

```suggestion
      console.log({length: base64Urls.length});
```

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -162,11 +162,13 @@ export const ButtonFields = ({
...
+
       await Share.open({
         subject: `${t(m.comapeoAlert)} — _*${preset ? preset.name : t(m.fallbackCategoryName)}*_ — ${formatDate(observation.createdAt, {format: 'long'})}`,
         title:
           base64Urls.length > 0 ? t(m.shareMediaTitle) : t(m.shareTextTitle),
-        urls: base64Urls,
+        urls: !base64Urls.length ? undefined : base64Urls,
```
achou11, line 169: is there any need to avoid passing in an empty array? would think what was here before should be okay

otherwise, a suggestion to switch the ternary

```suggestion
        urls: base64Urls.length ? base64Urls : undefined,
```

src/frontend/AppNavigator.tsx
```diff
@@ -33,8 +32,6 @@ export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
     };
   }, []);
 
-  useReactNavigationDevTools(rootNavigationRef);
```
achou11, line 36: just checking - removing this dep and the integration here is intentional? not against it, just making sure there's some context to this change

package.json
```diff
@@ -24,145 +24,147 @@
     "build:backend": "node ./scripts/build-backend.mjs",
     "extract-messages": "formatjs extract 'src/frontend/**/*.{ts,tsx}' --ignore='**/*.d.ts' --format crowdin --out-file ./messages/en.json",
     "eas-build-post-install": "npm run build:backend && npm run build:translations && npm run build:intl-polyfills",
-    "tests": "maestro start-device --platform android && npm run android && maestro test e2e"
+    "tests": "maestro start-device --platform android && npm run android && maestro test e2e",
+    "upgrade-dependencies:all": "npm-check --update --save-exact",
+    "upgrade-dependencies:production": "npm-check --update --save-exact --production",
+    "upgrade-dependencies:dev-deps": "npm-check --update --save-exact --dev-only"
```
achou11: just to reduce the typing a bit:

```suggestion
    "upgrade-deps:all": "npm-check --update --save-exact",
    "upgrade-deps:production": "npm-check --update --save-exact --production",
    "upgrade-deps:dev": "npm-check --update --save-exact --dev-only"
```

achou11 (changes requested):
App seems to run well after doing some basic flows. Provided a couple of smaller suggestions that I think should be addressed before merging. Also:

- update the PR title
- update the nodejs-mobile patch file names to match the new version of nodejs-mobile-react-native and their references in the patches readme

src/frontend/screens/Observation/Buttons.tsx
ErikSin: passing the empty array was breaking it in the newer version (not in the changelog).

src/frontend/AppNavigator.tsx
ErikSin: It was used for debugging in a previous pr and was not taken out aftwards

package.json
ErikSin: Im not fully sure? For a major version, I would say yes. But for a minor/patch version, if everything is running properly, i don't know if there is a need. But I also don't have as much background on how these dependencies are interacting

src/frontend/packageJson.test.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/893/commits/7993eb9ff21ad00e84957db58012f9a637f3f772

src/frontend/packageJson.test.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/893/commits/7993eb9ff21ad00e84957db58012f9a637f3f772

src/frontend/screens/Observation/Buttons.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/893/commits/b68cbb1cb333a5967a181d17407dd40b7a1ad42e

package.json
achou11: yeah i think in this case it worked out fine. would say that moving forward, we should definitely be more careful about minor+major version upgrades of nodejs-mobile

achou11 (approved):
The links in the patches readme need to be updated to use the new file names, but otherwise changes lgtm!

Issue #874, opened 2024-12-03
Alerts/ Support Multipoint


ximenabb:
The front-end code ([PR#844](https://github.com/digidem/comapeo-mobile/pull/844)) supports displaying alerts of type Point, LineString, MultiLineString, Polygon and MultiPolygon. 
It does not support `MultiPoint`

ToDo

- [ ] Add support of `MultiPoint`

EvanHahn:
Haven't tested it, but I _think_ this is all that's required in `src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx`:

```diff
@@ -31,6 +31,7 @@ export const RemoteDectionAlertsMapLayer = () => {
             'Point',
             'MultiLineString',
             'MultiPolygon',
+            'MultiPoint',
           ],
           ['has', 'metadata.alert_type'],
           ['has', 'month_detec'],
@@ -59,7 +60,7 @@ export const RemoteDectionAlertsMapLayer = () => {
       {/* Circle Layer for Points */}
       <MapboxGL.CircleLayer
         id="comapeo-alerts-point"
-        filter={['==', '$type', 'Point']}
+        filter={['in', '$type', 'Point', 'MultiPoint']}
         style={{
           circleRadius: 5,
           circleColor: '#FF0000',
```

ErikSin:
We could do it the way that Evan has recommended, the only thing is that it would be in-distinguishable from normal points. Does the CMI team want them styled differently (aka Not red dots) from the other alerts?

ximenabb:
@ErikSin they can be the same style (red dot) as other alerts

gmaclennan:
Yes, generally there should be no need to style `Multi` geometry variants differently than their non-multi versions, e.g. `Line` and `MultiLine` should be the same, `Polygon` and `MultiPolygon` should be the same, and `Point` and `MultiPoint` should be the same.

PR #895, opened 2024-12-17
Remote detection alerts

### LINKED ISSUE #811: Alerts Detail

ErikSin:
closes #811 and #874

Currently the backend does not support multipoints, so unable to test


src/frontend/hooks/server/remoteDetectionAlert.ts
```diff
@@ -0,0 +1,15 @@
+import {useSuspenseQuery} from '@tanstack/react-query';
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
+
+export const REMOTE_DETECTION_ALERTS_KEY = 'alerts';
+
+export function useRemoteDectionAlerts() {
```
achou11:
```suggestion
export function useRemoteDetectionAlerts() {
```

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import MapboxGL from '@rnmapbox/maps';
+
+import {RemoteDetectionAlert} from '@comapeo/schema';
+import {FeatureCollection} from 'geojson';
+import {useRemoteDectionAlerts} from '../../../hooks/server/remoteDetectionAlert';
+import {flatten} from 'flat';
+import {includeKeys} from 'filter-obj';
+
+export const RemoteDectionAlertsMapLayer = () => {
```
achou11: I suggest extracting any of the constant `filter` props used in this component to constants that live inside this file but outside the component

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
```diff
@@ -0,0 +1,129 @@
...
+        geometry: alert.geometry,
+        properties: {
+          ...flatten(
+            includeKeys(alert, [
+              'metadata',
+              'detectionDateStart',
+              'detectionDateEnd',
+              'sourceId',
+            ]),
+          ),
```
achou11: what's the desired outcome here? This feels prone to "overflattening" and I would suggest using explicit fields where flattening isn't needed:

```
          metadata: alert.metadata, // Maybe this one needs flattening? Not sure what the desired outcome is...
          detectionDateStart: alert.detectionDateStart,
          detectionDateEnd: alert.detectionDateEnd,
          sourceId: alert.sourceId,
```

Doing the above potentially avoids needing to install deps too (for now)

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
achou11:
```suggestion
export const RemoteDetectionAlertsMapLayer = () => {
```

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -114,6 +115,7 @@ export const MapScreen = () => {
 
         {isFinishedLoading && (
           <>
+            <RemoteDectionAlertsMapLayer />
```
achou11: just checking since I may lack some context: this layer will be rendered _underneath_ the other layers - is that what's desired?

achou11 (changes requested):

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: Im not sure I fully understand what you mean? Are you saying:

```ts
<MapboxGL.SymbolLayer
        id="mapeo-alerts-label"
       // Make the object inside "filter" a CONST?
        filter={[
          'all',
          [
            'in',
            '$type',
            'Polygon',
            'LineString',
            'Point',
            'MultiLineString',
            'MultiPolygon',
            'MultiPoint',
          ],
          ['has', 'metadata.alert_type'],
          ['has', 'month_detec'],
          ['has', 'year_detec'],
        ]}
```

src/frontend/screens/MapScreen/index.tsx
ErikSin: I think that makes sense because (as it is now) the remote detection alerts are not clickable, while the other layers are.

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
achou11: my guess is that it's related to the mapbox filter expressions, in which case I would suggest just lifting the relevant `metadata` fields explicitly such that they're in the `properties` geojson field e.g.

```ts
properties: {
  alertType: metadata.alert_type,
  ...
}
```

and update the filter expression to be `['get', 'alertType']`
  
There seems to be a way to write a mapbox style expression to query nested fields in a object `properties` field but I think it's not necessary to try it.

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
achou11: yeah - I'm suggesting defining those filters as consts that live outside the component as a minor optimization

ErikSin: So it should look like:

```ts
// outside the component:
ALERTS_LABEL_FILTER = [
          'all',
          [
            'in',
            '$type',
            'Polygon',
            'LineString',
            'Point',
            'MultiLineString',
            'MultiPolygon',
            'MultiPoint',
          ],
          ['has', 'metadata.alert_type'],
          ['has', 'month_detec'],
          ['has', 'year_detec'],
        ]

// Inside the component:
<MapboxGL.SymbolLayer
        id="mapeo-alerts-label"
        filter={ALERTS_LABEL_FILTER}/>
        
```

Is that what you mean?

achou11: yep! none of these filters (as of writing this) rely on any component props or state so there's not really a need to have them defined within a component, as they'll create a new JS value each time the component rerenders unless you explicitly memoize (which is not needed due to no deps on props or state). given that they're usually nested arrays too, in theory it's relatively more expensive for a new value to be created on each render compared to other kinds of values

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: I believe (but im not 100% sure) that there can be certain properties in metadata (including `alert_type`) that is used by `properties`. So this flattening allows us to access this.

Gregor and I discussed this, and we came to this code (using the deps)

Your way only allows for the `alert_type` to be extracted. But maybe that ok? Im not sure if there are other properties that are needed in metadata?

achou11: > Your way only allows for the alert_type to be extracted. But maybe that ok? Im not sure if there are other properties that are needed in metadata?

Understandable, just feels like a premature optimization for something that can be implemented with much less code - an example of accounting for something that isn't needed yet

achou11: seems like the only code that's using the properties are the layer filters, which so far only needs the `alert_type` from `metadata`. Would make more sense to make more of the metadata fields accessible in `properties` if - for example - the layer was pressable and some interaction callback needed access to those fields via the geojson properties

ErikSin: https://github.com/digidem/comapeo-mobile/pull/895/commits/c2e9f5bf27ba68e39198168a2c0f15229cc3154f

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/895/commits/5506b1f3d00344d92637693a41f48137e48fd627

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
```diff
@@ -0,0 +1,133 @@
...
+    'in',
+    '$type',
+    'Polygon',
+    'LineString',
+    'Point',
+    'MultiLineString',
+    'MultiPolygon',
+    'MultiPoint',
+  ],
+  ['has', 'alert_type'],
```
achou11: This has to match the key used in the corresponding geojson `properties`

```suggestion
  ['has', 'alertType'],
```

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
```diff
@@ -0,0 +1,133 @@
...
+      id="alerts-source"
+      shape={convertRemoteDetectionAlertsToFeatures(alerts)}>
+      {/* Symbol Layer for Labels */}
+      <MapboxGL.SymbolLayer
+        id="comapeo-alerts-label"
+        filter={LABEL_FILTER}
+        style={{
+          textField: [
+            'concat',
+            ['get', 'alert_type'],
```
achou11: Needs to match key used in geojson properties

```suggestion
            ['get', 'alertType'],
```

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
```diff
@@ -0,0 +1,133 @@
...
+      return {
+        type: 'Feature',
+        geometry: alert.geometry,
+        properties: {
+          alertType: alert.metadata.alert_type,
+          detectionDateStart: alert.detectionDateStart,
+          detectionDateEnd: alert.detectionDateEnd,
+          sourceId: alert.sourceId,
+          month_detec: dateStart.getMonth() + 1,
+          year_detec: dateStart.getFullYear(),
```
achou11: no need to mix camelcase and snakecase. stick to one or the other (probably lean towards camelcase). Relevant for when filtering using the mapbox style expressions syntax (see other comments)

```suggestion
          alertType: alert.metadata.alert_type,
          detectionDateStart: alert.detectionDateStart,
          detectionDateEnd: alert.detectionDateEnd,
          sourceId: alert.sourceId,
          monthDetec: dateStart.getMonth() + 1,
          yearDetec: dateStart.getFullYear(),
```

If applying the above suggestion, you'll need to update the filters that reference `month_detec` and `year_detec`

achou11 (changes requested):

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/895/commits/313a10eafaa3eddf45bb9151022851dd480ef2f1

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/895/commits/313a10eafaa3eddf45bb9151022851dd480ef2f1

src/frontend/screens/MapScreen/MapLayers/RemoteDetectionAlertsLayer.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/895/commits/313a10eafaa3eddf45bb9151022851dd480ef2f1

package-lock.json
achou11, line 1: there should no longer be changes in this file

achou11:
Mind updating with develop and undoing the lockfile changes? If tests pass after doing those, think the PR will be ready for merging

src/frontend/screens/MapScreen/MapLayers/ObservationMapLayer.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
 import MapboxGL from '@rnmapbox/maps';
 
-import {useObservations} from '../../hooks/server/observations';
-import {usePresetsQuery} from '../../hooks/server/presets';
-import {useNavigationFromHomeTabs} from '../../hooks/useNavigationWithTypes';
+import {useObservations} from '../../../hooks/server/observations.ts';
+import {usePresetsQuery} from '../../../hooks/server/presets.ts';
+import {useNavigationFromHomeTabs} from '../../../hooks/useNavigationWithTypes.ts';
```
achou11: out of curiosity - are you adding the file extensions manually or is that some external formatting feature being applied? Not opposed to it, but does add a little extra noise to the diff

package-lock.json
ErikSin: im not sure how to fix this. i have removed the package lock and ran npm i. any suggestions?

achou11: I usually do the following:

1. copy and paste the state the lockfile was in when you branched off develop
2. clear node_modules
3. Run `npm i` (there should be a diff that undoes the diff listed here)

achou11: Happy to push a fix if helpful!

ErikSin: yeah, go ahead!

src/frontend/screens/MapScreen/MapLayers/ObservationMapLayer.tsx
ErikSin: i think it is vscode doing that. I moved these files into a folder and vscode updated the imports. I will remove

package-lock.json
achou11: fixed via https://github.com/digidem/comapeo-mobile/pull/895/commits/1cec0ccd3e9e44ee46c10cd49c7427a02fc26899

src/frontend/screens/MapScreen/MapLayers/ObservationMapLayer.tsx
achou11: yeah fwiw, i do prefer explicit file extensions personally (and in some cases that's actually necessary), but more of a stylistic thing in the context of this repo

achou11 (approved):

PR #896, opened 2024-12-18
chore: update backend native deps

gmaclennan:
This updates the holepunch.to managed native deps to the latest versions, using the updated prebuilds.

This also changes native module prebuilds from hard-coded versions to actually use the prebuild versions that match the dependencies.

I have tested this on in an emulator to check the app loads and I can create an observation, but I have not tested beyond that.

The dependency updates are all within semver, so there should be no breaking changes, however it would be good to update comapeo-core to the same versions and check all the tests pass.

This solves issues with these modules on x64, but x64 is still blocked by better-sqlite3 which is still broken on x64.


scripts/build-backend.mjs
```diff
@@ -33,10 +33,10 @@ await Promise.all([
 fs.cpSync('./src/backend', './nodejs-assets/backend', {recursive: true});
 fs.mkdirSync('./nodejs-assets/nodejs-project/node_modules', {recursive: true});
 
-fs.writeFileSync('./nodejs-assets/BUILD_NATIVE_MODULES.txt', '1', {
+fs.writeFileSync('./nodejs-assets/BUILD_NATIVE_MODULES.txt', '0', {
```
achou11: This needs to stay enabled because it's how nmrn internally decides to look for prebuilds in the first place. Otherwise the prebuilds will not be copied into the final apk.

> I have tested this on in an emulator to check the app loads and I can create an observation, but I have not tested beyond that.

This change prevents the app from running at all for me. Were you testing in an emulator prior to making this specific change?

scripts/download-prebuilds.mjs
```diff
@@ -3,48 +3,56 @@
...
 import fs from 'node:fs';
 import path from 'node:path';
+import {createRequire} from 'node:module';
+import {fileURLToPath} from 'node:url';
+
+const require = createRequire(import.meta.url);
 
 // NOTE: we currently don't support building for intel arch (android-x64)
-const TARGETS = ['android-arm', 'android-arm64'];
+const TARGETS = ['android-arm', 'android-arm64', 'android-x64'];
```
achou11, line 11: should remove the note above

ErikSin:
@achou11 do you mind taking a look at this PR?

achou11:
> @achou11 do you mind taking a look at this PR?

yeah I had started looking at it yesterday. My understanding is that we probably don't want to integrate this for the upcoming release just yet so I de-prioritized. Is that accurate?

EDIT: guess this could be merged into main without being part of the next release, but just a bit tedious to remember to not include this in the release branch, which is harder to do if using the release action trigger since it just uses the current state of the main branch (i think)

EDIT 2: actually it'd make more sense to merge this prior to the release because it fixes the issue of our current version of core (which we updated on develop recently) using different versions of the relevant native dependencies than what we were using prior to this PR.

achou11 (changes requested):

scripts/download-prebuilds.mjs
achou11: addressed via 4bf1a37e66f07f97640507280b2e8efd5161f595

scripts/build-backend.mjs
achou11: addressed via 473b7ef5cf6e1ad80cec8303577f29041ddbbda8

achou11 (approved):

PR #897, opened 2024-12-18
chore: upgrade eslint configuration

achou11:
- Upgrades eslint to v9 and install other necessary eslint modules for writing the configuration
- Upgrades other eslint plugins to latest compatible versions
- Removes rules related to React Native (via `@react-native/eslint-config`) due to compat issues with the typescript eslint plugin it uses internally.
- ~applies very trivial fixes that could be automated by the eslint check (i.e. with its `--fix`) flag.~ EDIT: applies fixes to relevant and easy-to-address issues
- updates the relevant npm script
- introduces more granular application of rules based on kind of file or directory (e.g. separate rules for backend vs frontend, rules specific for tooling files)

Remaining items (can be done as follow-ups):

- [x] ~decide whether or not not having rules specific to React Native for now is a blocker. Can spend some time attempting to get it working but would be okay with doing as a follow-up.~ EDIT: addressed via https://github.com/digidem/comapeo-mobile/pull/897/commits/ad35fdc6d9bf5a9a3ce02e1d744c25f032808884

- [x] adjust the react configuration to omit non-applicable rules (e.g. no need for `dom`)


eslint.config.mjs
```diff
@@ -0,0 +1,94 @@
...
+});
+
+// TODO: Incorporate react-native
+const frontendConfig = tseslint.config({
+  name: 'frontend',
+  files: ['src/frontend/**/*.{js,jsx,ts,tsx}'],
+  extends: [
+    tseslint.configs.recommended,
+    pluginQuery.configs['flat/recommended'],
+    // TODO: incorporate 'disable-dom' and 'disable-web-apis' presets
```
achou11: filed an issue for the bug I was running into when trying to add them: https://github.com/Rel1cx/eslint-react/issues/923

package.json
```diff
@@ -164,7 +167,8 @@
     "rimraf": "5.0.5",
     "semver": "7.6.0",
     "type-fest": "4.26.0",
-    "typescript": "5.3.3"
+    "typescript": "5.3.3",
```
achou11: ideally we'd try to update TypeScript here but not strictly necessary to. Open to doing though.

src/frontend/hooks/server/maps.ts
achou11, line 1: ESLint autofix

src/frontend/contexts/PhotoPromiseContext/index.tsx
achou11, line 1: ESLint autofix

src/frontend/hooks/useFormattedTimeSince.ts
achou11, line 1: ESLint autofix

src/frontend/lib/utils.ts
achou11, line 1: ESLint autofix

eslint.config.mjs
```diff
@@ -0,0 +1,94 @@
...
+    '*.config.{js,mjs,cjs}',
+    'scripts/*.{js,mjs,cjs}',
+    'expo-config-plugins/*.{js,mjs,cjs}',
+  ],
+  languageOptions: {
+    globals: {
+      ...globals.node,
+      ...globals.nodeBuiltin,
+      ...globals.worker,
+    },
```
ErikSin, line 39: Should we include `global.jest`? In preparation for #898 ?

eslint.config.mjs
```diff
@@ -0,0 +1,94 @@
...
+
+// TODO: Incorporate react-native
+const frontendConfig = tseslint.config({
+  name: 'frontend',
+  files: ['src/frontend/**/*.{js,jsx,ts,tsx}'],
+  extends: [
+    tseslint.configs.recommended,
+    pluginQuery.configs['flat/recommended'],
+    react.configs['recommended-typescript'],
+    react.configs['disable-dom'],
```
ErikSin: should we also disable the `web-api`?

ErikSin (commented):
Im glad that we are using the new eslint architecture. There are a lot of new rules introduced by the premade/recommended configs. Running `eslint` introduced 99 problems (but a bitch aint one 😆 ) - 71 error and 28 warnings.

I think we should using these recommended configs but Im wondering if it should be the job of this PR to solve the new errors associated with it?

eslint.config.mjs
achou11: no, because `globals.jest` is relevant for test files that are run in a jest environment. this configuration is for configuration files related to dev/build tooling.

probably makes sense to add a config in `frontendConfig` that targets test files and specifies the globals there instead

eslint.config.mjs
achou11: I think the `web-api` has some useful rules that are still relevant to React Native code (e.g. https://eslint-react.xyz/docs/rules/web-api-no-leaked-interval), although maybe it's worth disabling a rule like https://eslint-react.xyz/docs/rules/web-api-no-leaked-resize-observer. i'm on the fence about https://eslint-react.xyz/docs/rules/web-api-no-leaked-event-listener because it's generally a helpful rule but does add some noise due to some apis using the `const subscription = foo.addEventListener()` pattern. If the noise is minimal compared to the utility, I'm currently in favor of adding ignore comments where the rule is complaining as opposed to disabling the rule altogether.

achou11 (commented):
> I think we should using these recommended configs but Im wondering if it should be the job of this PR to solve the new errors associated with it?

I'm in favor of addressing the issues as follow-ups, mostly because we haven't made an effort to address the issues prior to this PR anyways. I've added the automatic fixes in this PR but if the issues don't fall under that category, I think it's safer to have dedicated PRs to address them.

I'm mostly interested in keeping this PR's scope to updating the tooling and configuration. I think decisions around things like the following are better served as follow-ups:

- deciding which rules to enable/disable
- incorporating CI checks
- addressing the actual issues

src/backend/package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "mapeo-mobile-node-next",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
achou11, line 4: just noting that the massive diff in this file (aside from removing eslint) is due to the lockfile version changing. I _think_ this is desirable in our case, as v3 is associated for npm v9 and later, which aligns with the versions of node that matter to us (18 + 20)

https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#lockfileversion

not really sure why the lockfile version changed in this PR versus other PRs where we updated backend deps though 🤔

eslint.config.mjs
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/897/commits/5ad46a7e9aeb513d5da44cf216af407ad5a7be82

achou11:
as of https://github.com/digidem/comapeo-mobile/pull/897/commits/e4dbe19be91ccdd790ac9e1bbf19a8d53f7e5439, `npm run lint:eslint` reports: `210 problems (109 errors, 101 warnings)`. Great to read through the complaints, many/most of which are valid and should be addressed via follow-ups 😄

src/frontend/hooks/useFormattedTimeSince.ts
```diff
@@ -2,8 +2,8 @@ import {useEffect, useState} from 'react';
 import {Duration} from 'luxon';
 
 export const useFormattedTimeSince = (start: Date | null, interval: number) => {
-  const [currentTime, setCurrentTime] = useState(new Date());
-  let startDate = start ? new Date(start) : new Date();
+  const [currentTime, setCurrentTime] = useState(() => new Date());
```
gmaclennan, line 5: Which rule complained about this? These are the kind of rules in eslint that bug me a bit. Yes it saves an unnecessary `new Date()` on each render, but that is a pico-optimization.

achou11: https://eslint-react.xyz/docs/rules/hooks-extra-prefer-use-state-lazy-initialization

yes it's a minor optimization but it's the unfortunate reality of how react works. i'd rather enforce consistency on this rather than relying on everyone to have a deep enough knowledge about this potentially consequential limitation in React

achou11:
did an initial pass on adjusting rules that are unanimously less helpful and addressing the resulting errors and warnings in https://github.com/digidem/comapeo-mobile/pull/897/commits/6f4b60835caed4c9817d1fc5b74cb3b7b4fe2cba.

Kind of wish I did a stacked PR so that changes in the app code could be easier to see. happy to move things around if desirable.

couple of observations:

- not sure about turning off the unused variables rule entirely. while potentially annoying in terms of prototyping, i was able to catch a bug with it enabled. my original adjustment to the rule was to allow unusued variables if prefixed with underscores, which is a common practice. i'd personally be happier with keeping a rule like that in place.

- not sure about turning off the no explicit `any` rule entirely. i found a couple of bugs because of it being used too freely without any context as to why. i'd prefer that people use the eslint disable comment to ignore the rule if they really need to use it, as it communicates proper intent.

while the bugs found so far are relatively minor, nothing stops us from introducing major bugs with these rules being disabled down the line. i'd rather put in the early foundations to mitigate those cases, even if they present some inconveniences in the development experience

ErikSin:
To be discussed tomorrow.

I agree with Andrew about the `No unused vars`, and I find it useful to keep unnecessary code out. 

The only one I dont agree with is`@eslint-react/hooks-extra/no-direct-set-state-in-use-effect`. I think we as a team are good at catching when `useState` is being used unnecessary and there are more cases then now when do have to use it in a useEffect.

Otherwise the rest of the rules seem good for me.

achou11:
> The only one I dont agree with is@eslint-react/hooks-extra/no-direct-set-state-in-use-effect. I think we as a team are good at catching when useState is being used unnecessary and there are more cases then now when do have to use it in a useEffect.

Yeah agreed, i don't find this rule super helpful although I kind of understand the reasoning. Will disable it 👍

achou11:
Based on conversations, I made a couple of revisions:

- re-enabled no unused vars rule (https://github.com/digidem/comapeo-mobile/pull/897/commits/1793981e6cd1541567c5fb1a5bc944d90f42e268)

- re-enabled no unused styles rule (https://github.com/digidem/comapeo-mobile/pull/897/commits/59eb8e9260c1f39b18784dc52fafe49f2945292f)

ErikSin (approved):

PR #902, opened 2025-01-17
chore: update Expo to SDK 52

achou11:
Way more involved than I expected it to be. 

- Changes to how the splash screen is configured. There are some upstream issues with newer Android versions in terms of how the splash screen behaves in development builds, but apparently they do not affect production builds. Haven't tested that yet though. See https://expo.dev/changelog/2024/11-12-sdk-52#known-limitations-of-splash-screen-implementation

- I've opted for upgrading some dependencies related to expo that aren't aligned with `expo-doctor` i.e. installing newer versions than what they list as compatible. Some of these benefit from useful fixes in later versions. There was a case where one of them needed an upgrade to properly support React Native 0.76 (forgot which one as I'm writing this).

- We should probably move away from using `react-native-indicators`. The only two components we use from there are the `DotIndicator` (we can probably implement our own) and the `UIActivityIndicator` (we can use https://reactnative.dev/docs/activityindicator instead). For now, I patched the dependency to fix an error that was coming up during runtime

- Updates some deps that are adjacent to all of these Expo updates. Also removes some dependencies that used to be relevant but are no longer based on my evaluation.


app.json
```diff
@@ -19,6 +20,20 @@
...
+          "backgroundColor": "#050F77",
+          "imageWidth": 200,
+          "android": {
+            "resizeMode": "contain"
+          },
+          "ios": {
+            "resizeMode": "contain"
+          }
+        }
+      ],
```
achou11: recommended approach to the splash screen configuration is now via a build plugin as opposed to the top-level `expo` field in the app config

package.json
```diff
@@ -38,75 +38,74 @@
...
     "react-native-scale-bar": "1.0.6",
-    "react-native-screens": "3.29.0",
+    "react-native-screens": "4.5.0",
     "react-native-share": "12.0.3",
-    "react-native-svg": "15.10.1",
+    "react-native-svg": "15.11.1",
     "react-native-url-polyfill": "2.0.0",
     "react-native-vector-icons": "10.2.0",
     "react-native-zeroconf": "0.13.8",
-    "readable-stream": "4.5.2",
```
achou11: not entirely sure if removing this is okay but didn't seem to be necessary based on a some checks. might need to dig into the commit history a bit to know 🤔

package.json
```diff
@@ -117,25 +116,25 @@
...
+    "@babel/runtime": "7.26.0",
     "@comapeo/core": "2.2.0",
     "@comapeo/schema": "1.2.0",
     "@formatjs/cli": "6.2.0",
-    "@react-native-community/cli": "12.3.6",
-    "@react-native/babel-preset": "0.73.21",
-    "@react-native/eslint-config": "0.73.2",
-    "@react-native/metro-config": "0.73.5",
-    "@react-native/typescript-config": "0.74.0",
+    "@mapeo/mock-data": "2.1.2",
```
achou11: missing explicit dev dependency

src/frontend/App.tsx
```diff
@@ -48,7 +48,13 @@ const mapeoApi = createMapeoClient(messagePort, {timeout: Infinity});
...
 localDiscoveryController.start();
 initializeNodejs();
-SplashScreen.preventAutoHideAsync();
+
+SplashScreen.setOptions({
+  fade: true,
+});
+SplashScreen.preventAutoHideAsync().catch(err => {
+  console.log(err);
+});
```
achou11, line 57: as mentioned, these won't seem to work when you're working with a dev build on newer Android versions

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -143,8 +144,8 @@ function rotatePhoto(
     'JPEG',
     CAPTURE_QUALITY,
     getPhotoRotation(acc),
-  ).then(({uri}) => {
-    return {uri};
+  ).then(resized => {
+    return {uri: resized.uri};
```
achou11, line 148: tiny ESLint error that was bothering me and worth fixing

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -85,6 +85,7 @@ export const CameraView = ({onAddPress}: Props) => {
     ref.current
       .takePictureAsync(captureOptions)
       .then(pic => {
+        if (!pic) return;
```
achou11, line 88: not really sure when this happens in practice, but guessing it's fine to just do an early return in this case.

package.json
```diff
@@ -38,75 +38,74 @@
...
-    "@sentry/react-native": "5.20.0",
+    "@sentry/react-native": "6.5.0",
     "@tanstack/react-query": "5.62.8",
     "@turf/helpers": "7.1.0",
     "@turf/random": "7.1.0",
     "@types/luxon": "3.4.2",
     "assert": "2.1.0",
     "buffer": "6.0.3",
     "cheap-ruler": "4.0.0",
+    "debug": "4.4.0",
```
achou11, line 59: missing explicit prod dependency

app.json
achou11: definitely need to confirm that these settings are actually being applied. Can/should do so in the following ways:

1. Test via a dev build on older Android versions
2. Test via a non-dev build on new Android versions

app.json
```diff
@@ -1,5 +1,6 @@
 {
   "expo": {
+    "newArchEnabled": false,
```
achou11, line 3: just noting that we explicitly disable the new arch

README.md
```diff
@@ -20,7 +20,7 @@ The next version of Mapeo Mobile
 
       - Note that **Java 17** is the minimum version required for this project.
 
-      - When setting up the Android-specific tooling, you will also need to install the [Android NDK](https://developer.android.com/ndk/). This can be installed using Android Studio by going to the `SDK Tools` tab in the `SDK Manager`. This project uses **NDK 25.1.8937393**.
+      - When setting up the Android-specific tooling, you will also need to install the [Android NDK](https://developer.android.com/ndk/). This can be installed using Android Studio by going to the `SDK Tools` tab in the `SDK Manager`. This project uses **NDK 26.1.10909125**.
```
achou11, line 23: our ndk tooling has updated due to react native changes. doesn't seem to cause issues with nmrn

achou11 (commented):
Opening this up for review. Haven't had a chance to go through the app in detail to figure out if anything's regressed - would appreciate more eyes to help with this.

package.json
achou11: ah yeah added via https://github.com/digidem/comapeo-mobile/commit/6d9e242ea896c9f2559d9b4fd6b439e2ad6e895d to account for issues with how rpc-reflector was being bundled. Should no longer be needed at this point

app.json
cimigree: @achou11 How old for version of Android? I can test on an emulator if that is useful.

achou11: Based on https://expo.dev/changelog/2024/11-12-sdk-52#known-limitations-of-splash-screen-implementation, the autohiding prevention doesn't work on _dev_ builds in Android 13+, but should be fine with _release_ builds.

If you're able to test the release builds on Android 13, 14, and/or 15, that'd be helpful! (can also do it on my end if necessary)

achou11: one thing I'd particularly take note of is the size of the logo in the splash screen. without specifying the `imageWidth`, it becomes incredibly tiny on my Android 15 Pixel 6, although I've only tested in dev builds

achou11:
amazingly have not found any notable regressions yet within the app!

cimigree:
I spent a decent amount of time trying to get this PR to run against the e2e tests and found it really, really slow and glitchy on an emulator. I used a Pixel 6 with Android 11, a Pixel 6 with Android 14, and a Pixel 8Pro with Android 15. The last one ran the fastest but was still quite glitchy. When I cleared out the Android caches and wiped the data from my emulator and clicked through, it seemed like things were working fine though.

So, I decided to try my device and click through the tests manually. That went fine (tedious, but fine). But then I discovered I was able to run the tests on my connected device! Do you think you might be able to  as well? I had to make a few tweaks to the commands to bring everything up to date, but it worked better and was not slow like it was on the emulator.

One thing that is cool and new is that the camera now works on the emulator.
<image src="https://github.com/user-attachments/assets/2d40108a-d2c6-4785-a9a7-818b50e64cc3" width="100" />

I haven't had a chance to test the release build yet but can do that first thing tomorrow.

app.json
cimigree: Just curious about why this is...

ErikSin:
I ran the app on my device. I only have one device on me so I wasn't able to test sync or invites.

Everything seems to be running well. One noticeable difference is that the expo camera now makes a noise when a picture is captured. Not a blocker, but if its easy to turn off, I think we should

achou11:
Thanks both!

> I spent a decent amount of time trying to get this PR to run against the e2e tests and found it really, really slow and glitchy on an emulator.

Curious why it'd be so slow 🤔 Only thing I can think of if from our side is that an intel-based build is somehow being run on your arm-based emulator, although not sure why or how that would happen. Otherwise maybe a regression from Expo's side?

> One noticeable difference is that the expo camera now makes a noise when a picture is captured. Not a blocker, but if its easy to turn off, I think we should

@ErikSin I think this was already the case before? i'm able to hear a sound in the currently released version of CoMapeo. Is it that the sound is happening under unexpected circumstances? It's easy to address - just not sure what the desired behavior is/was so I preserved it

ErikSin:
> @ErikSin I think this was already the case before? i'm able to hear a sound in the currently released version of CoMapeo. Is it that the sound is happening under unexpected circumstances? It's easy to address - just not sure what the desired behavior is/was so I preserved it

Interesting, it is not happening on the released version of CoMapeo on my device. Perhaps it device specific and the upgrade changed something? I think if it is happening on some devices already lets not address it unless we are asked to by PS or QA

app.json
cimigree: Yes it is quite small! I tested on a release-candidate build
This is a Pixel 8A running Android 14.
<image src="https://github.com/user-attachments/assets/681770b5-4eec-4cb3-8bae-52c7a0b8e0f3" width="300" />

Same was true on my Samsung Galaxy A25 also on Android 14(sorry didn't get a screenshot but it was basically the same as the Pixel).

cimigree (commented):
The code looks fine.
As I said, I ran tests on the dev version yesterday and reported back.

Today I built release candidate version. Not sure if that is the same as "release"
I tested taking photos and using the camera and that went well.
Observations were fine.
I tried to do Tracks which was fine.
Permissions worked as expected.
I uploaded a new config and that went well.
I recorded Audio and that was fine.

What didn't work for me.

- Uploading a map file. I tried all of the files in here:

https://drive.google.com/drive/search?q=parent:19iZUWrlTQIBpqr7MnP0u5E5dmFudJYKt
And none of them worked. The one from 11/27 just caused the action to freeze. i.e nothing happened after I selected the file except I went back to the About Custom Map screen. Then When I press Choose File nothing happened. The others triggered "something went wrong."

- Inviting another phone to a project.

Just a simple invite appeared to work, but then the invitation modal popped up again and the app froze. When I reopened the app, I was on the project.
With inviting a phone that was already on a project, the flow worked as expected until I tried to leave the project and accept the invite, then I got the something went wrong screen which I could not close (App was frozen). When I reloaded I was on the new project.

One thing I noticed is that when I open the app after it had been frozen it seems to take a long time but is faster to reopen when it hadn't been frozen. Is that expected?

Not sure what all of this means, but it might make sense to try to get the image bigger on the splash screen and make sure your invite flow is working better than mine at least before moving ahead.

achou11:
> What didn't work for me.
> 
>     * Uploading a map file. I tried all of the files in here:
> 
> 
> https://drive.google.com/drive/search?q=parent:19iZUWrlTQIBpqr7MnP0u5E5dmFudJYKt And none of them worked. The one from 11/27 just caused the action to freeze. i.e nothing happened after I selected the file except I went back to the About Custom Map screen. Then When I press Choose File nothing happened. The others triggered "something went wrong."

I think issues with the larger one have existed prior to this. Did you try the other ones right after the attempt with the large one failed? I think the failure of the large one sometimes causes all sorts of issues with the app afterwards, so wouldn't be surprised if it somehow affected other attempts during the same "session". I was able to successfully use the smaller ones even after the attempt with the big one failed.

>     * Inviting another phone to a project.
> 
> 
> Just a simple invite appeared to work, but then the invitation modal popped up again and the app froze. When I reopened the app, I was on the project. With inviting a phone that was already on a project, the flow worked as expected until I tried to leave the project and accept the invite, then I got the something went wrong screen which I could not close (App was frozen). When I reloaded I was on the new project.

Yeah not sure if that's related to any of the changes here. Can look into it a bit, although would have to wait until tomorrow when I have more access to various devices (currently working from elsewhere today).

> One thing I noticed is that when I open the app after it had been frozen it seems to take a long time but is faster to reopen when it hadn't been frozen. Is that expected?

Have no idea 😅 "Frozen" meaning after the failed attempt to add the custom map file? Again, might be related to preexisting issues with the app after a failed attempt to use a large map file.

app.json
achou11: mostly to reduce the scope of this upgrade 😄 there are some unknowns on our end in terms of whether things like nodejs-mobile will Just Work™️  with the new architecture enabled. Based on https://reactnative.dev/blog/2024/10/23/the-new-architecture-is-here#gradual-migration, i think the idea is that things shouldn't break when it's enabled but I'd rather find out in a dedicated follow-up than here.

cimigree:
> I think issues with the larger one have existed prior to this. Did you try the other ones right after the attempt with the large one failed? I think the failure of the large one sometimes causes all sorts of issues with the app afterwards, so wouldn't be surprised if it somehow affected other attempts during the same "session". I was able to successfully use the smaller ones even after the attempt with the big one failed.
> 
Yes it was the same session. I didn't even know if those were valid mapfiles, so good to know it can work.

> Have no idea 😅 "Frozen" meaning after the failed attempt to add the custom map file? Again, might be related to preexisting issues with the app after a failed attempt to use a large map file.

What I was referring to here was frozen after the invite process which froze in both cases -- inviting a phone that was on no project (it froze after the modal showed up the second time) and when inviting a phone already on a project (it seemed to leave the project and then eventually went to the something went wrong modal which couldn't be closed) and app was then frozen.

I followed the same steps with invitations on my devices on the develop branch and everything worked smoothly.

achou11:
> What I was referring to here was frozen after the invite process which froze in both cases -- inviting a phone that was on no project (it froze after the modal showed up the second time) and when inviting a phone already on a project (it seemed to leave the project and then eventually went to the something went wrong modal which couldn't be closed) and app was then frozen.
> 
> I followed the same steps with invitations on my devices on the develop branch and everything worked smoothly.

Ah thanks for the clarification. Will definitely look into this tomorrow then.

app.json
```diff
@@ -19,6 +20,20 @@
           "RNMapboxMapsDownloadToken": "sk.eyJ1IjoiYnN0ZWZhbmN6eWsiLCJhIjoiY2x1dXV1NHNlMGU5dzJqcnh1Zno4YW4xcyJ9.P_LsVHh_HWTh25x5rRdrQg"
         }
       ],
+      [
+        "expo-splash-screen",
+        {
+          "image": "./assets/splash.png",
+          "backgroundColor": "#050F77",
+          "imageWidth": 300,
```
achou11, line 28: anything greater than this seems to result in the image being cutoff. not sure if it's worth fiddling around with this too much

app.json
achou11: increased the size to 300 (see e3e0ebf0cb1c6f90c6deae1c75eeac25a755b7fd) but anything larger seems to result in the image being cutoff

achou11:
@cimigree have you been able to reproduce the invites issue reliably? I haven't been able to get it to fail in the same way as you've described yet...

app.json
cimigree: Ok it is not as big as it used to be but bigger than it was!

cimigree:
> @cimigree have you been able to reproduce the invites issue reliably? I haven't been able to get it to fail in the same way as you've described yet...

Invite flow worked both ways -- not on a project and on a project. Phew! Very glad neither of us could get it to fail!

cimigree (approved):

PR #911, opened 2025-01-21
fix: observations hidden behind tracks

### LINKED ISSUE #869: Tracks / Obvs hidden behind track
https://www.notion.so/digidem/Tracks-Obvs-hidden-behind-track-14a1b08162d5803c994ae3576d0cf632

cimigree:
closes #869 

Changes the order of layers so that the tracks are laid down first and below the observations.
<image src="https://github.com/user-attachments/assets/49cf9929-9e08-42e3-94da-f1278bec1543" width="250" />

### DIFF
--- a/src/frontend/screens/MapScreen/index.tsx
+++ b/src/frontend/screens/MapScreen/index.tsx
@@ -114,9 +114,9 @@ export const MapScreen = () => {
 
         {isFinishedLoading && (
           <>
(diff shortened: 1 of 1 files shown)


achou11 (approved):
Good catch! Took me a second to understand why it needs to be this way but makes sense if you think about it in terms of layers being applied 😅

PR #913, opened 2025-01-21
chore: upgrade production dependencies

### LINKED ISSUE #906: Update Production Dependencies
Update production dependencies to @latest with the exact version.

Update all production dependencies excluding `expo` (and all related dependencies), `react-native`, and `react navigation`

The `npm` command `upgrade-deps:production` can be used, which uses [npm-check](https://github.com/dylang/npm-check).

cimigree:
closes #906 
**FYI: I tested using Maestro and all passed**

Updated
---
Updated all of the patch updates.
Updates all of the minor updates except react-native-gesture-handler  2.21.2   ❯  2.22.0 because that requires react-native 74
Update to @sentry/react-native            5.20.0   ❯  6.5.0

Could not completely update
---
react-native-mmkv    (was able to do a patch update 2.12.1 to 2.12.2 )     
react-native-safe-area-context  
react-native-screens       (was able to update, but not to latest)
because they all require react native 74

uint8array-extras  could not be updated at all. Kept getting Hermes errors. I suspect this has happened to others because we are on quite an old version

Types
---
One required types update to geojson types


package.json
```diff
@@ -85,36 +85,36 @@
...
-    "react-native-mmkv": "2.12.1",
+    "react-native-mmkv": "2.12.2",
     "react-native-progress": "5.0.1",
-    "react-native-reanimated": "3.16.5",
+    "react-native-reanimated": "3.16.7",
     "react-native-restart": "0.0.27",
     "react-native-safe-area-context": "4.14.0",
     "react-native-scale-bar": "1.0.6",
-    "react-native-screens": "3.29.0",
+    "react-native-screens": "4.4.0",
```
ErikSin: We use react native screen only for react navigation. This current version explicitly doesn't support react-navigation v6. So lets not upgrade it here. and lets upgrade it when you upgrade react-navigation

ErikSin (approved):
One comment, but otherwise looks good to me!

PR #914, opened 2025-01-21
chore: clean up translations build script

achou11:
Relies on built-in Node APIs to handle functionality done by our usage of `rimraf` and `glob`, which allows us to remove them as direct dev deps

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -154,7 +154,6 @@
         "prettier": "3.5.1",
         "react-native-svg-transformer": "1.5.0",
         "react-test-renderer": "18.3.1",
-        "rimraf": "5.0.5",
         "semver": "7.7.1",
         "type-fest": "4.34.1",
         "typescript": "5.7.3",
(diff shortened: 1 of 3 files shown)


ErikSin:
im not fully sure how to check this...but i think CI is passing its fine!

ErikSin (approved):

achou11:
> im not fully sure how to check this...but i think CI is passing its fine!

for something like this, i usually do the following:

1. Run the relevant script on `develop`.
2. Rename the any outputted assets for comparing to later
3. Switch to the branch with the changes
4. Run the same script
5. Compare the output to what was produced in (2)

achou11:
> for something like this, i usually do the following:
> 
>     1. Run the relevant script on `develop`.
> 
>     2. Rename the any outputted assets for comparing to later
> 
>     3. Switch to the branch with the changes
> 
>     4. Run the same script
> 
>     5. Compare the output to what was produced in (2)

and upon doing this i realized that the output is _slightly_ different 😄 . before, the `messages.json` would have the keys sorted in reverse alphabetical order whereas now they're sorted alphabetically i.e. 

before:

```json
{ 
  "way": { ... }, 
  ..., 
  "api": { ... }
}
```

after:

```json
{ 
  "api": { ... }, 
  ..., 
  "way": { ... }
}
```

From what I can tell, this doesn't have any user-facing impact. The only place where it would have is the displayed languages in the language settings, but we [sort the options in alphabetical order based on the english name](https://github.com/digidem/comapeo-mobile/blob/750fa5148b71befb34632fec99fe2f78f676e656/src/frontend/lib/intl.ts#L67-L69) before displaying them so the order in the actual json file doesn't matter.

PR #918, opened 2025-01-23
chore(deps): update @comapeo/core to 2.3.1

achou11:
Also:

- updates explicit `@comapeo/schema` dep in root to 1.4.1
- updates `@mapeo/mock-data` to ~`2.1.3`~ ~`2.1.4`~ `2.1.5` and moves it to a dev dependency (not used in app code)

Manually verified the backend lockfile changes to make sure that important transitive deps like Hypercore are not unintentionally updated.


ErikSin (approved):
tested on my device and everything seems to be working

PR #919, opened 2025-01-23
chore: update dependency

ErikSin:
Updgrade comapeo/core to `@latest`


achou11:
@ErikSin didn't realize you were also doing this work. Mind if we defer to #918 since it handles all of the necessary dep updates?

ErikSin:
haha yes, i just saw thats. Whoops!

PR #921, opened 2025-01-29
fix: fix issues building backend when working with fresh installation

achou11:
Follow-up to #896 , which I didn't properly test on a "fresh" installation of the repo. The problem was that the script for downloading prebuilds was deriving information about the relevant packages' versions based on the _source_ directory (`src/backend/node_modules/...`) and not the directory where the installation of project occurs (`nodejs-assets/backend/`). 

Abbreviated sequence of what was happening:

1. Copy `src/backend/` to `nodejs-assets/backend/`
2. run `npm ci` in `nodejs-assets/backend/` (populates `nodejs-assets/backend/node_modules/` as a result)
3. Bundle backend and copy relevant files to `nodejs-assets/nodejs-project/`
4. delete `nodejs-assets/backend/`
5. download prebuilds

the main issue is that step 5 ideally gets needed information from something that is erased in step 4. instead, it was pulling in information in a directory that may not exist or may be stale i.e. it would work fine if one ran `npm ci` in `src/backend/` before step 5.

this PR fixes the scripts so that step 4 and 5 are basically switched and that step 5 is sourcing information from the proper directory.  

For anyone reviewing/testing this PR, I would recommend doing the following on your machine to verify that this works:

1. Remove the nodejs assets directory (`rm -rf nodejs-assets`)
2. Remove the backend node modules directory `cd src/backend ; rm -rf node_modules`)
3. In the project root, run `npm run build:backend` - it should not have any notable errors
4. Run the scripts for starting the app as you normally would

### DIFF
--- a/scripts/build-backend.mjs
+++ b/scripts/build-backend.mjs
@@ -1,17 +1,36 @@
 #!/usr/bin/env node
 
-import {parseArgs} from 'util';
+import fs from 'node:fs';
+import {createRequire} from 'node:module';
+import path from 'node:path';
+import {fileURLToPath} from 'node:url';
+import {parseArgs} from 'node:util';
 import {$} from 'execa';
-import path from 'path';
-import fs from 'fs';
+
 import {downloadPrebuilds} from './download-prebuilds.mjs';
+
+const require = createRequire(import.meta.url);
+
+const nodejsAssetsDirectory = fileURLToPath(
+  new URL('../nodejs-assets', import.meta.url),
+);
+const nodejsAssetsBackendDirectory = path.join(
+  nodejsAssetsDirectory,
+  'backend',
+);
+const nodejsAssetsProjectDirectory = path.join(
+  nodejsAssetsDirectory,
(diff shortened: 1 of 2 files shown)


scripts/build-backend.mjs
```diff
@@ -81,24 +106,40 @@ const KEEP_THESE = [
...
-    './nodejs-assets/nodejs-project',
+    nodejsAssetsProjectDirectory,
     name === 'index.bundle.js' ? 'index.js' : name,
   );
 
   fs.cpSync(source, destination, {recursive: true});
 }
 
-await $$`rm -rf ./nodejs-assets/backend`;
+console.log('Downloading native prebuilds...');
```
cimigree, line 118: Do you definitely want to keep this?

cimigree (approved):

scripts/build-backend.mjs
achou11: I think so? Provides a little insight as to what step is occurring

PR #925, opened 2025-01-29
fix: change draw order of remote detection alert layers

rudokemper:
Currently, the layers in `RemoteDetectionAlertsMapLayer` are not in the right draw order. Labels are appearing underneath drawn geometries:

image

This PR fixes the order by ensuring that the `SymbolLayer` is drawn last. Additionally, the Polygon Stroke `LineLayer` is drawn after the Polygon `FillLayer`, which is the common MapGL style convention.

I didn't test this as I don't have a CM / React Native dev environment set up, but I'm pretty sure this will solve the problem 🙂


achou11 (approved):
thanks for the fix!

PR #926, opened 2025-01-29
chore: appium webdriver setup

### LINKED ISSUE #908: Setup Minimal Appium E2E Test for CI Compatibility
Create a minimal e2e set up to test whether appium will work on our app and in our CI. The intention is to test that we can run it in CI before creating the whole suite of tests.

Related to [this](https://github.com/orgs/digidem/projects/37/views/3?pane=issue&itemId=88559768&issue=digidem%7Ccomapeo-mobile%7C862) work done in the UK

## To determine
- [ ] how Webdriver iO or Nightwatch will effect the setup of the CI.
- [ ] how UiAutomator or Esspresso will effect the setup of the CI.

cimigree:
closes #908 

### **Driver:**  
Because we decided to use Browserstack and Appium, that lead naturally to the use of **UIAutomator**. That is what Browserstack uses with Appium.  BrowserStack abstracts the driver implementation, and they implement UiAutomator with Appium automatically. In addition, UIAutomator can theoretically be used with our web app. Espresso is available through Browserstack, but it is its own entity, not the standard Appium implementation.  We may want to test system-level features like GPS and Wi-Fi, which Espresso doesn’t handle well. I decided to go with the most supported, most fleshed out driver available on Browserstack.
### **Client:**  
Basically the same thing. I chose to use **Webdriver IO** because it is well-documented and actively maintained.
Great for multi-platform testing (mobile, web, and Electron). It has built-in integrations for BrowserStack and easier configuration for parallel test runs. It also has a large plugin ecosystem.
### **Test Framework:**  
I chose **Mocha,** a widely-used, flexible, and lightweight JavaScript test framework. It allows you to use any assertion library (e.g., Chai, Assert). It supports hooks like before, after, beforeEach, afterEach. It has simple syntax: describe and it. caveat... I am not sold on Mocha. **Jasmine** might be better and I am open to reconsidering.
### **Test Reporter:**  
I chose **spec reporter**.  It provides quick and easy viewing of test results during local development and debugging with simple, human-readable output. It's pretty basic, imo ideal for getting started and debugging minimal test suites.

## Testing
I wrote a simple test to check the first two screens and that you can get from the first screen to the second. It was not easy to figure out how to select text to test. I got my testing language from [Browserstack's guidelines](https://www.browserstack.com/docs/app-automate/appium/getting-started/nodejs/webdriverio/local-testing). But the same guidelines are available via [WebdriverIO](https://webdriver.io/docs/selectors/#android-uiautomator). It would be much much easier to look for text if we used accessibilityLabels on the text... Something to consider for the future?

## Configuration file
I am really including the almost minimum that one can include in the configuration file. A lot more can be done with that, like testing on multiple devices, increasing number of instances, etc. Here's how the file can be adjusted from the docs: https://webdriver.io/docs/configurationfile/

## Workflow
I built this based on previous work by Erik and Gregor -- Erik for signing into Expo and building using EAS. Gregor for building an apk. That increase to JavaHeapSpace (because I ran into out of memory errors when building) and work with caching and freeing up disc space (this makes the building of the apk at least 50% faster) comes from him. I wanted the test to not only fail in Github if the E2E tests failed, but I wanted an easy way for the PR maker to see why, since we only get 2 logins to Browserstack with our free account. In order to get that to work, each time someone makes a push, there is a file included summarizing the test results (this is where spec reporter comes in handy.

## How to test
Push a new commit or open a PR to feat/appium-webdriver-setup.
GitHub Actions will automatically:
Build the APK using EAS.
Upload the APK to BrowserStack.
Run WebdriverIO tests on BrowserStack.
You should see the results as an artifact on the Workflow. LMK if you want my login to see how things look on Browserstack. With it I think you go to [this link](https://app-automate.browserstack.com/dashboard/v2/builds/59ffe51693ceaecfd613cfa23b034995e91eb24b)

**To test locally**
- Update the capabilities in wdio.conf.ts.
```
capabilities: [
  {
    platformName: 'android',
    'appium:platformVersion': '12.0',
    'appium:deviceName': 'Samsung Galaxy S22 Ultra', // whatever your local device is
    'appium:automationName': 'UIAutomator2',
    'appium:app': '/path/to/local/apk/app-debug.apk', // path to your apk
  },
],

```
Start an Android emulator or connect a real device.
Run Appium in a separate terminal:
`npx appium`
Execute WebdriverIO:
`npx wdio run wdio.conf.ts`
 or 
`npm run wdio`
You will have to install appium locally to do this.

## Next Steps:
- Add multiple devices
- Consider Jasmine instead of Mocha?
- Improve element selection (add accessibilityLabels?)
- Expand tests to cover more screens & flows


expo-config-plugins/increaseJavaMemory.js
ErikSin, line 1: Can we create a readme in the config plugins folder to document what the increase in memory is for?

package.json
```diff
@@ -27,7 +27,8 @@
     "tests": "maestro start-device --platform android && npm run android && maestro test e2e",
     "upgrade-deps:all": "npm-check --update --save-exact",
     "upgrade-deps:production": "npm-check --update --save-exact --production",
-    "upgrade-deps:dev": "npm-check --update --save-exact --dev-only"
+    "upgrade-deps:dev": "npm-check --update --save-exact --dev-only",
+    "wdio": "wdio run ./wdio.conf.ts"
```
ErikSin: i think this should be slightly more semantic, maybe "test:e2e"?

tests/e2e/test.e2e.ts
```diff
@@ -0,0 +1,33 @@
+import {expect, $} from '@wdio/globals';
+
+describe('CoMapeo App Tests', () => {
+  it('should launch the app, grant permissions, and navigate to the next screen', async () => {
+    await browser.pause(15000);
+    const description = await $(
+      'android=new UiSelector().text("Map your world, together")',
```
ErikSin, line 8: should we write reusable functions for these "strings"? Something like

```ts
function selectText(textToFind:string){
     return `android=new UiSelector().text("${textToFind}")`
}
```

tests/e2e/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "compilerOptions": {
+    "types": ["@wdio/globals/types", "@wdio/mocha-framework", "mocha"]
```
ErikSin: this is a personal preference, but i like when we explicitly import the functions instead of setting them as globals.

I know we cant do that for webdriverio, but we can do that for mocha.

Again, this is a personal preference, so if you feel strongly about keeping mocha as a global, happy to discuss further

wdio.conf.ts
ErikSin, line 1: were moving towards having all our config files be js files with [typescript support through jsdoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

wdio.conf.ts
```diff
@@ -0,0 +1,62 @@
...
+        buildName: 'CoMapeo Android Build',
+        sessionName: 'Launch App and Grant Permissions',
+        appiumVersion: '2.12.1',
+        debug: true,
+        networkLogs: true,
+      },
+    },
+  ],
+  logLevel: 'error',
+  waitforTimeout: 20000,
```
ErikSin: if were setting a waitForTimeout here, i think we should take out all the timeouts times in the tests so we dont have redundnant settings

wdio.conf.ts
```diff
@@ -0,0 +1,62 @@
...
+    timeout: 60000,
+  },
+  reporters: [
+    [
+      'spec',
+      {
+        showPreface: false,
+        color: true,
+        addConsoleLogs: true,
+        onlyFailures: false,
```
ErikSin: showPreface, color, and onlyFailures can be taken out as you are setting the default values

wdio.conf.ts
```diff
@@ -0,0 +1,62 @@
...
+      },
+    ],
+  ],
+  outputDir: './test-results',
```
ErikSin: does this need to be added to the gitignore? Or is this only in the ci environment?

wdio.conf.ts
```diff
@@ -0,0 +1,62 @@
...
+      'appium:app': process.env.BROWSERSTACK_APP_URL,
+      'appium:autoGrantPermissions': true,
+      'bstack:options': {
+        projectName: 'CoMapeo',
+        buildName: 'CoMapeo Android Build',
+        sessionName: 'Launch App and Grant Permissions',
+        appiumVersion: '2.12.1',
+        debug: true,
+        networkLogs: true,
+      },
```
ErikSin: where are you getting this syntax from?

ErikSin (commented):
This is not a review per-se. More just questions and comments. I think it would be good for @gmaclennan to take a look at this?

tests/e2e/test.e2e.ts
cimigree: Will do that in next PR when adding more tests. Thanks for suggestion but trying to keep things as simple as possible for this iteration.

tests/e2e/tsconfig.json
cimigree: Not a problem to explicitly import.

wdio.conf.ts
cimigree: Good suggestion! Thanks.

wdio.conf.ts
cimigree: I am pretty sure it is only in the git environment, but I will add it to gitignore just to be sure in case it is added when tests are run locally.

wdio.conf.ts
cimigree: A combination of from here:
https://appium.io/docs/en/2.0/guides/caps/ (specifies the appium prefix for all capabilities)
And here:
https://www.browserstack.com/docs/app-automate/capabilities
click legacy, node.js
Maybe you found syntax for Appium 1? This is for Appium 2+...

expo-config-plugins/increaseJavaMemory.js
cimigree: I think I need Gregor's help with that. I know that the build fails and what the error message is, but I don't exactly know why.

gmaclennan: ha I'm not sure either. Maybe just a doc to say "increase java memory to make builds work on CI"!

gmaclennan (approved):
Great work on this Cindy, I can see that a lot has gone into this, and it's amazing you were able to get all the moving pieces to line up and make this work. The overview about why choices were made is helpful too and the choices make sense to me based on what we know at this time. I think this is ok to merge and start using, and we can evolve and learn from it as we add more tests.

expo-config-plugins/increaseJavaMemory.js
cimigree: Ok done.

cimigree:
@ErikSin, before I merge this, are you good with running it 
```
on:
  pull_request:
    branches:
      - main
      - develop
    types: [opened, synchronize, reopened, ready_for_review]
```
(for the workflow yml file?)

ErikSin:
> on:
>   pull_request:
>     branches:
>       - main
>       - develop
>     types: [opened, synchronize, reopened, ready_for_review]

I think it is good to merge. Perhaps in the future, if it takes a long time and is resource intensive, we could remove `syncronize` and have it only run on `review_requested`. But we can talk about it after it is built

Issue #909, opened 2025-01-20
Upgrade React Navigation


ErikSin:
Upgrade React Navigation following [this guide](https://reactnavigation.org/docs/upgrading-from-6.x).

We should convert from the [dynamic Api to the static Api as well](https://reactnavigation.org/docs/upgrading-from-6.x#static-configuration-api)

ErikSin:
1 PR for breaking changes and minor API changes.

Seperate PR for dynamic Api to the static Api. Timebox to 1 day, and check-in to determine if we should continue this work

PR #931, opened 2025-02-04
chore: upgrade react native navigation

cimigree:
Related to #909 
Because the upgrade and the change to Static API can be done separately, This PR just handles the upgrade to React Navigation 7.x, using this guide https://reactnavigation.org/docs/upgrading-from-6.x.
## What this PR does
- Upgrades packages. 
- Uses popto when going back in the stack, based on this: https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-goes-back-use-popto-instead
- Adjusts logic for showing Auth Screen. 
   - The Auth screen is shown if someone requires a password as the initial screen. I moved it to the initial screen logic. Otherwise, the Auth screen was not able to load correctly because of differing requirements around nested navigators.
- Fixes import of types. (Some type definitions are in @react-navigation/elements and are no longer where they used to be)
-  Adds logic for nested navigators in the drawer (https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-navigates-to-screen-in-a-nested-child-navigator)
- Removing canGoBack as a parameter (got an error that it was not part of HeaderBackButtonProps (and it isn't based on the Type definition) even though I see it in HeaderBackButton? very confused)(https://reactnavigation.org/docs/elements/#headerbackbutton)
- Removes unused props that have been changed anyway (in this file: src/frontend/screens/ObservationsList/ObservationListHeaderLeft.tsx, https://reactnavigation.org/docs/upgrading-from-6.x/#labelvisible-is-removed-in-favor-of-displaymode-in-headerleft-and-headerbackbutton-elements)
- Fixes use of incorrect props (HeaderButtonProps instead of HeaderBackButtonProps)
- Removes unused key property (https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-accepts-a-key-option)
- Updates prop name for test id ([tabBarButtonTestID:](https://reactnavigation.org/docs/upgrading-from-6.x/#the-tabbartestid-option-is-renamed-to-tabbarbuttontestid-in-bottom-tab-navigator-and-material-top-tab-navigator)


src/frontend/Navigation/Drawer.tsx
ErikSin, line 1: There are some visual differences in the drawer:

image

src/frontend/Navigation/Tab/index.tsx
ErikSin, line 1: Some visual differences were introduced: 
image

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -74,20 +73,13 @@ function RootStackNavigatorChild() {
...
   const security = useSecurityContext();
 
-  const {navigate} = useNavigationFromRoot();
-
-  React.useEffect(() => {
-    if (security.authState === 'unauthenticated') {
-      navigate('AuthScreen');
-    }
-  }, [security.authState, navigate]);
-
```
ErikSin: This `useEffect` has to stay.  We want the auth screen to open everytime the app goes in the background (not just on first open).

But this is actually a typing issue and can be fixed like this:

```ts
const {navigate} = useDrawerNavigation();

  React.useEffect(() => {
    if (security.authState === 'unauthenticated') {
      navigate('DrawerHome', {screen: 'AuthScreen'});
    }
  }, [security.authState, navigate]);
```

ErikSin (changes requested):
2 visual changes, and a simple swap around of hooks, but otherwise looks good!

src/frontend/Navigation/Stack/index.tsx
cimigree: OK should be good now, but I made a few tweaks.

cimigree:
Those were a bit tricky to fix! All default styles from react navigation. But should be good now.

---
<image src="https://github.com/user-attachments/assets/bf810e1b-a787-49a1-af15-66b52a0376ba" width="350" />

<image src="https://github.com/user-attachments/assets/c584e120-5f11-4c9a-b29c-50920ffc075a" width="350" />

ErikSin (approved):
you have 2 of the same `navigation` objects being propogated. One through the `DrawerNavigationContext` and one as a prop. Instead of passing it as a prop you can just use `useDrawerNavigation()` directly inside RootStackNavigatorChild. If you can just fix that before merging :)

PR #958, opened 2025-02-10
chore: refactor audio

### LINKED ISSUE #928: Refactor Audio
Refactor Audio following [this architecture.](https://www.notion.so/digidem/Audio-Refactor-18a1b08162d58042be32f8a16d0c6c42)

Should use the new bottom sheet architecture introduced in #927.

ErikSin:
closes #928 

Permissions:
https://github.com/user-attachments/assets/6454f1ae-517e-46ab-87af-959e181257e4

Audio Recording:
https://github.com/user-attachments/assets/89d3c489-3676-47e5-a6e4-8961138e7ec8

Audio Playback:
https://github.com/user-attachments/assets/a3c1adfb-d888-4abd-94b6-93c4014c2117

*** To Note : 11 of the 28 `files changed` are deleted files, and 3 of them are files that have been moved


src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -405,6 +450,18 @@ export const createDefaultScreenGroup = ({
...
+        component={AudioAskPermissionBottomSheet}
+      />
+      <RootStack.Screen
+        name="DeleteAudioBottomSheet"
+        component={DeleteAudioBottomSheet}
+      />
+      <RootStack.Screen
+        name="AudioSavedBottomSheet"
+        component={AudioSavedBottomSheet}
+      />
```
achou11, line 450: more of a nit but would be helpful to have these co-located with the other audio-related screens just to make it easier to find and keep track of later on

src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -389,6 +392,48 @@ export const createDefaultScreenGroup = ({
...
+                  };
+                  addAudio(audioRecording);
+                  navigation.replace('AudioSavedBottomSheet');
+                }}
+                backImage={backImageProps => (
+                  <CloseIcon color={backImageProps.tintColor} />
+                )}
+              />
+            );
+          },
```
achou11: instead of using `useDraftObservation()` at the top-ish level of the app in order to pass `addAudio` from `Navigation/Stack/index.tsx`, I think it would be preferable to have it co-located with where it's actually needed. In order to use it that hook within this `headerLeft` prop, you would need to define an external component that uses the hook and is then passed into this `headerLeft` prop. For example:

Define a custom `AudioHeaderLeft` component (name is not good but whatever):

```tsx
// Would be nice to be able to define a more generic `onPress` prop here such that `uri` and `duration` aren't needed, 
// but that would require setting up the `useDraftObservation()` hook outside of the component to have access to `addAudio`, 
// which is what we're trying to avoid in this case.

function AudioHeaderLeft({
  duration,
  uri,
  ...headerBackButtonProps
}: HeaderBackButtonProps & {
  duration: number;
  uri: string;
}) {
  // Now the usage of this hook is co-located to where it's needed and not unnecessarily set up further up the component tree (which potentially introduces rendering overhead)
  const {addAudio} = useDraftObservation();
  const navigation = useNavigation();

  return (
    <HeaderBackButton
      {...headerBackButtonProps}
      onPress={() => {
        addAudio({
          uri,
          duration,
          createdAt: Date.now(),
        });
        // We use this instead of `navigation.replace()` because TS will complain 
        // since it cannot guarantee that this component is used within a stack navigator 
        // (which is where `navigator.replace()` is available)
        navigation.dispatch(StackActions.replace('AudioSaveBottomSheet'));
      }}
      backImage={backImageProps => (
        <CloseIcon color={backImageProps.tintColor} />
      )}
    />
  );
};
```

Then the usage here will look like this:

```tsx
          headerLeft: props => (
            <AudioHeaderLeft
              {...props}
              uri={route.params.uri}
              duration={route.params.duration}
            />
          ),
```

src/frontend/Navigation/Stack/index.tsx
achou11, line 1: this is a smell to me. i provided a suggestion in `Navigation/Stack/AppScreens.tsx` to have this be set up/co-located with where it's truly needed

src/frontend/screens/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,134 @@
...
+}: NativeRootNavigationProps<'AudioAskPermissionBottomSheet'>) => {
+  const {formatMessage} = useIntl();
+  const {goBack, replace} = navigation;
+  const [permission, setPermission] = React.useState<Audio.PermissionResponse>(
+    route.params.audioPermission,
+  );
+
+  React.useEffect(() => {
+    let isCancelled = false;
+    const handleAppStateChange = async (nextAppState: AppStateStatus) => {
```
achou11: maybe a comment about what this is handling would be helpful. not very clear at an initial glance

src/frontend/screens/AudioPlaybackSaved.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const [shareError, setShareError] = React.useState<Error | null>(null);
+
+  const handleShare = React.useCallback(async () => {
+    setShareLoading(true);
+    try {
+      let fileUri = localUri;
+
+      if (!fileUri) {
+        const tempFileName = `audio_${Date.now()}.m4a`;
+        const localFilePath = `${FileSystem.cacheDirectory}${tempFileName}`;
```
achou11: suggestion: construct the path using `new URL()`

```suggestion
        const localFilePath = new URL(tempFileName, FileSystem.cacheDirectory!).pathname;
```

src/frontend/screens/AudioPlaybackSaved.tsx
```diff
@@ -0,0 +1,198 @@
...
+          localFilePath,
+        );
+
+        fileUri = downloadedUri;
+        setLocalUri(downloadedUri);
+      }
+
+      await Share.open({url: fileUri, failOnCancel: false});
+    } catch (err) {
+      setShareError(err as Error);
```
achou11: slightly more resilient error handling:

```suggestion
      if (err instanceof Error) {
        setShareError(err)
      } else {
        // Replace message with something more useful for debugging
        setShareError(new Error('Unable to share audio file'))
      }
```

src/frontend/screens/DeleteAudioBottomSheet.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </View>
+        <View style={{width: '100%'}}>
+          <Button
+            fullWidth
+            color="dark"
+            style={{backgroundColor: RED, marginTop: 30}}
+            onPress={() => {
+              deleteAudio(
+                route.params.uri,
+                route.params.screenToPopToAfterDelete === 'ObservationEdit',
```
achou11: wondering if it makes sense to have this be derived either from the `uri` e.g. `uri.startsWith('https://')` or just another more explicit nav prop that's provided when navigating to this sheet e.g. `audioIsSaved: boolean`. the latter is probably better than the former in terms of clarity.

src/frontend/sharedComponents/ActionRow.tsx
```diff
@@ -113,18 +83,6 @@ export const ActionsRow = ({fieldRefs}: ActionButtonsProps) => {
   return (
     <>
```
achou11: can remove the fragment here now

src/frontend/sharedComponents/AudioStyles.tsx
achou11, line 1: nit: wouldn't consider this file a shared component. more like shared variables between various audio-related screens. if sticking with this approach in terms of how styles/nav options are shared, one suggestion would be to restore a `Screens/Audio/` directory that roughly looks like this:

```
- screens/Audio/
  |-- shared.ts # <-- put the contents of this file in here
  |-- AudioAskPermissionBottomSheet.tsx
  |-- AudioPlaybackSaved.tsx
  |-- AudioPlaybackUnsaved.tsx
  |-- AudioRecording/
      |-- ...
```

src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
```diff
@@ -36,25 +33,24 @@ export const AudioThumbnail: FC<AudioThumbnailProps> = ({
...
+      type: audioAttachment.type,
+      variant: 'original',
+    });
     setLoading(false);
-    navigation.navigate('Audio', {
-      isEditing: currentRoute?.name === 'ObservationEdit',
+    navigation.navigate('AudioPlaybackSaved', {
       uri,
-      isSavedUri,
+      canDelete: currentRoute?.name === 'ObservationEdit',
```
achou11, line 54: i don't think shared components should have checks like this, as it results in an implicit dependency on the navigation context which makes them less reusable. ideally this is derived based on a explicitly passed prop into this component (e.g. `canDelete?: boolean`) + some checks in this component related to the URI in question (similar to before when `isSavedUri` is calculated).

May potentially require a little rewiring because this is used pretty deeply within a couple of shared components.

achou11 (changes requested):
Overall very promising! Provided a couple of notable suggestions that I'm interested to get your thoughts on.

src/frontend/Navigation/Stack/AppScreens.tsx
ErikSin: I agree that it would be more favorable to co-locate this with the audio related screen, but i dont think we should do that at the expense of taking it out of the `Rootstack.Group` that configure all the modals to be the same. Its just easier to maintain the modals if they are all configured in one spot

achou11: maybe misunderstanding, but my suggestion was to move these lines further up in the file so that they are next to the other declarations of the audio-related screens. not suggesting an actual re-grouping via react navigation `Group`

src/frontend/screens/AudioAskPermissionBottomSheet.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/5a5eac894aab8e4aff5a6456e614b956fd073ddc

src/frontend/sharedComponents/ActionRow.tsx
ErikSin: this component is now just forwarding another component which is not ideal. I think im just going to get rid of it instead

src/frontend/sharedComponents/MediaScrollView/AudioThumbnail.tsx
ErikSin: I agree, but you are right that it is too nested to do anything right now. This component will eventually be refactored (with the the update of the observation hook). I will add a to do note to that when component is refactored

src/frontend/screens/DeleteAudioBottomSheet.tsx
ErikSin: I actually wonder if we should just pass a  `onPressDeleteAudio:()=>void`? I know react navigation suggest passing only serializable objects in params, but that is for deep linking and state persistence. We are not persisting state and if a user deep links to a delete screen, I think we should throw an error (it would be bad ui to deep link to an action)

ErikSin: Looking into this, its only a suggestion by react navigation. So I think it does make sense to pass the onDeleteFunction here. It will make the bottom sheet agnostic, and not rely on the navigation context

achou11: yeah i was thinking about suggesting something like that but then i remembered recommendation from react navigation. i wouldn't suggest going against that recommendation but as you say, it would technically work. i just worry about having to remember about a one-off case where we do this differently compared to other screens.

ErikSin: Agreed and perhaps we should just agree that the bottom sheets should just take a function. Currently ALL the bottom sheets are designed to take an action based on its parent. So if all the bottom sheets are taking a function in params it no longer becomes a one-off case.

This would align well with our principle of maintainability. The bottom sheets actions will be agnostic and allows for their actions to be swapped easier without having to rely on the navigation context

src/frontend/Navigation/Stack/AppScreens.tsx
ErikSin: Oh you can't see it in the github dif. But they are currently in a `RootStack.Group` for all the bottom sheets. So moving it beside the audio screens would take it out of that group

src/frontend/screens/AudioPlaybackSaved.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/cc0959e2ee050f2ad30c957230588a0a075fb6bb

src/frontend/screens/AudioPlaybackSaved.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/5d09e7e5176e12da23a844e045b7851354ac1406

src/frontend/screens/DeleteAudioBottomSheet.tsx
ErikSin: I did this here: https://github.com/digidem/comapeo-mobile/pull/958/commits/c6787289d487b5178d14297e3f18aeebf7afd260

Lets talk about it tomorrow at the app dev meeting

src/frontend/Navigation/Stack/AppScreens.tsx
achou11: ahhh got it - sorry i didn't notice that despite looking at this locally 😅

src/frontend/sharedComponents/AudioStyles.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/e6622e0b12f84b50fa9ff5a83a9156fa61b707de

src/frontend/screens/AudioPlaybackSaved.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/418692282af330638156e5fbc33886b7d30777eb I changed it back because it was throwing an error:

ErikSin:
Addressed all your feedback. I will wait until after we talk at the app dev meeting tomorrow to decide what we do about passing functions as params in navigation

achou11:
> Addressed all your feedback. I will wait until after we talk at the app dev meeting tomorrow to decide what we do about passing functions as params in navigation

will start re-reviewing but make sure to also update with latest develop 👍

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,137 @@
...
+  description: {
+    id: 'screens.AudioPermission.description',
+    defaultMessage:
+      'To record audio while using the app and in the background CoMapeo needs to access your microphone. Please enable microphone permissions in your app settings.',
+    description: 'Screen description for audio permission screen',
+  },
+  notNowButtonText: {
+    id: 'screens.AudioPermission.Button.notNow',
+    defaultMessage: 'Not Now',
+    description: 'Screen button text for not allowed audio permission',
```
achou11:
```suggestion
    description: 'Screen button text for not granting audio permission',
```

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,137 @@
...
+  },
+  allowButtonText: {
+    id: 'screens.AudioPermission.Button.allow',
+    defaultMessage: 'Allow',
+    description: 'Screen button text for allow the audio permission',
```
achou11:
```suggestion
    description: 'Screen button text for granting the audio permission',
```

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,137 @@
...
+  },
+  goToSettingsButtonText: {
+    id: 'screens.AudioPermission.Button.goToSettings',
+    defaultMessage: 'Go to Settings',
+    description:
+      'Screen button text for navigate user to settings when audio permission was denied',
```
achou11:
```suggestion
      'Screen button text for navigating the user to settings when audio permission is denied',
```

src/frontend/screens/Audio/AudioCustomHeaderLeft.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export function AudioCustomHeaderLeft({
+  duration,
+  uri,
+  ...headerBackButtonProps
+}: HeaderBackButtonProps & {
+  duration: number;
+  uri: string;
+}) {
+  // Now the usage of this hook is co-located to where it's needed and not unnecessarily set up further up the component tree (which potentially introduces rendering overhead)
```
achou11: can remove this comment. i mostly added it for the sake of providing context during review 😄 
```suggestion
```

src/frontend/screens/Audio/AudioCustomHeaderLeft.tsx
```diff
@@ -0,0 +1,39 @@
...
+      {...headerBackButtonProps}
+      onPress={() => {
+        addAudio({
+          uri,
+          duration,
+          createdAt: Date.now(),
+        });
+        navigation.replace('AudioSavedBottomSheet');
+      }}
+      // eslint-disable-next-line react/no-unstable-nested-components
```
achou11: i'd prefer to avoid adding eslint ignores for now. allows us to get a full understanding of what errors/warnings to consider once we start addressing ESLint concerns
```suggestion
```

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,137 @@
...
+  );
+
+  // When the user changes their permission in phone settings and returns to the app,
+  // we need to check for updates and navigate accordingly.
+  // Without this, the app won't detect the permission change.
+  React.useEffect(() => {
+    let isCancelled = false;
+    const handleAppStateChange = async (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'active') {
+        // eslint-disable-next-line @typescript-eslint/no-shadow
```
achou11: same comment about not ignoring eslint for now

```suggestion
```

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                marginTop: 20,
+              }}>
+              <UIActivityIndicator
+                color={WHITE}
+                size={PRIMARY_CONTROL_DIAMETER}
+              />
+            </View>
+          ) : (
+            <TouchableOpacity
+              onPress={!isRecording ? startRecording : finishRecording}
```
achou11:
```suggestion
              onPress={isRecording ?  finishRecording : startRecording}
```

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+              style={AudioStyles.basePressable}>
+              {<View style={!isRecording ? styles.record : styles.stop} />}
```
achou11:
```suggestion
              {<View style={isRecording ? styles.stop : styles.record} />}
```

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+            </TouchableOpacity>
+          )
+        }>
+        <View style={AudioStyles.container}>
+          <View style={AudioStyles.timerContainer}>
+            <Text
+              style={[
+                AudioStyles.timerText,
+                {color: !isRecording ? MEDIUM_GREY : WHITE},
```
achou11:
```suggestion
                {color: isRecording ?  WHITE : MEDIUM_GREY},
```

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+              ]}>
+              {Duration.fromMillis(timeElapsed).toFormat('mm:ss')}
+            </Text>
+          </View>
+          <HeaderText variant="header3" style={AudioStyles.message}>
+            {!isRecording
+              ? formatMessage(m.record5Min)
+              : timeElapsed < 240000
+                ? formatMessage(m.lessThan5)
+                : formatMessage(m.lessThan1)}
```
achou11, line 118: aside from flipping the ternary and ideally avoiding using a nested ternary, I would primarily suggest restoring the usage of the [plural syntax](https://formatjs.github.io/docs/core-concepts/icu-syntax#plural-format) for this.

1. Replace `m.lessThan5` and `m.lessThan1` with:

```ts
  minutesRemaining: {
    id: 'screens.AudioRecording.minutesRemaining',
    defaultMessage:
      'Less than {value, plural, one {a minute} other {{value, number} minutes}} left',
  },
```

2. Use it here

```tsx
            {isRecording
              ? formatMessage(m.minutesRemaining, {
                  value: timeElapsed < 240_000 ? 5 : 1,
                })
              : formatMessage(m.record5Min)}
```

src/frontend/screens/Audio/AudioSavedBottomSheet.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <BodyText style={{marginTop: 10, textAlign: 'center'}}>
+            {formatMessage(m.subTitle)}
+          </BodyText>
+        </View>
+        <View style={{width: '100%'}}>
+          <Button
+            fullWidth
+            variant="outlined"
+            onPress={() => {
+              // Should use `popTo` in React Nav v7
```
achou11: nit: Usually mark items like this as actual todo comments to distinguish from documentation-like comments

```suggestion
              // TODO: Should use `popTo` in React Nav v7
```

src/frontend/sharedComponents/ActionsRow/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+              </HeaderText>
+            </TouchableOpacity>
+          ))}
+        </View>
+      ) : (
+        <KeyboardAccessory
+          items={bottomSheetItems}
+          onPress={() => Keyboard.dismiss()}
+        />
+      )}
```
achou11:
```suggestion
      {keyboardVisible ? (
        <KeyboardAccessory
          items={bottomSheetItems}
          onPress={() => Keyboard.dismiss()}
        />
      ) : (
        <View style={[styles.container, styles.containerPadding]}>
          {bottomSheetItems.map(item => (
            <TouchableOpacity
              onPress={item.onPress}
              style={styles.itemContainer}
              key={item.testID}
              testID={item.testID}>
              <View style={styles.itemIcon}>
                <CustomCircleIcon icon={item.icon} />
              </View>
              <HeaderText variant="header6" numberOfLines={1}>
                {item.label}
              </HeaderText>
            </TouchableOpacity>
          ))}
        </View>
      )}
```

achou11 (changes requested):
Mostly minor suggestions in terms of code, but found a potential issue related to dealing with initial audio permissions.

From a fresh installation, the permissions sheet keeps re-appearing after I've already granted permissions while working with the same draft observation. The problem goes away once i exit the existing draft observation flow and start a new draft observation

https://github.com/user-attachments/assets/d17e350d-adee-4aa8-aae1-8e0415aa9971

ErikSin:
> From a fresh installation, the permissions sheet keeps re-appearing after I've already granted permissions while working with the same draft observation. The problem goes away once i exit the existing draft observation flow and start a new draft observation

This is happening because the [audioPermission here] (https://github.com/digidem/comapeo-mobile/blob/2e8dea006c88897f8d47ed74d124c99ada636c48/src/frontend/sharedComponents/ActionsRow/index.tsx#L38) is not reactive, meaning it doesn’t update when the permission is granted in the bottom sheet.

Since it doesn’t re-check permissions automatically, it remains unaware of the change. However, when you close and reopen the page, the permission is retrieved again, so it correctly reflects the new state.

The problem is that React Navigation keeps the screen in memory while it's still in the navigation stack, so the outdated permission status persists instead of updating in real time. Thats why when you navigate back, it thinks there is no permission. I have a fix for it, and I think the cleanest thing to do is pass a function through navigation...so we can discuss it today at the meeting

src/frontend/screens/Audio/AudioSavedBottomSheet.tsx
ErikSin: actually we are already on v7!
https://github.com/digidem/comapeo-mobile/pull/958/commits/0b87115b20b41b08d8a6fae27ad93a295803c9d4

src/frontend/screens/Audio/AudioRecording/index.tsx
ErikSin: This feels like this adds unnecessary complexity without a clear maintainability benefit. Given our agreed-upon principles of keeping things maintainable and avoiding premature optimizations, I’d suggest keeping the existing approach, which is simpler and easier to understand at a glance.

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/a85f1405efd895c370ce20ebdf63ad8126b85e94

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/a85f1405efd895c370ce20ebdf63ad8126b85e94

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/a85f1405efd895c370ce20ebdf63ad8126b85e94

src/frontend/screens/Audio/AudioCustomHeaderLeft.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/fb8a0d3d4f7d5bd1a203ce3106e13cd93ba221ec

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/958/commits/fb8a0d3d4f7d5bd1a203ce3106e13cd93ba221ec

ErikSin:
> > From a fresh installation, the permissions sheet keeps re-appearing after I've already granted permissions while working with the same draft observation. The problem goes away once i exit the existing draft observation flow and start a new draft observation
> 
> This is happening because the [audioPermission here] (
> 
> https://github.com/digidem/comapeo-mobile/blob/2e8dea006c88897f8d47ed74d124c99ada636c48/src/frontend/sharedComponents/ActionsRow/index.tsx#L38
> ) is not reactive, meaning it doesn’t update when the permission is granted in the bottom sheet.
> 
> Since it doesn’t re-check permissions automatically, it remains unaware of the change. However, when you close and reopen the page, the permission is retrieved again, so it correctly reflects the new state.
> 
> The problem is that React Navigation keeps the screen in memory while it's still in the navigation stack, so the outdated permission status persists instead of updating in real time. Thats why when you navigate back, it thinks there is no permission. I have a fix for it, and I think the cleanest thing to do is pass a function through navigation...so we can discuss it today at the meeting

Actually passing things through params was adding an unnecessary complication (i technically needed to pass `canAskAgain` as well). So i thought the simplest approach would be to just useFocusEffect, and just check everytime the page came back into focus. 

https://github.com/digidem/comapeo-mobile/pull/958/commits/3c780d7f6868dd6437958ab3293562701e694238

src/frontend/screens/Audio/AudioRecording/index.tsx
achou11: I agree it's a bit harder to understand at a glance. Not opposed to keeping what you have, but I wouldn't be surprised if the suggested approach is a generally better experience for translators. Can't really confirm this but I would imagine that keeping these as the same message allows the translator to understand that this message is used for the same UI (as in the specific UI in which this is used has multiple variants), whereas having two messages doesn't guarantee that. Purely conjecture though!

ErikSin:
Let's discuss ternary usage in the app dev meeting today. At first glance, this feels like a minor stylistic preference, but I'm open to hearing your reasoning. After discussion hopefully we can come to a consensus whether that it is a thing we would like to do as a team moving forward.

achou11:
> Let's discuss ternary usage in the app dev meeting today. At first glance, this feels like a minor stylistic preference, but I'm open to hearing your reasoning. After discussion hopefully we can come to a consensus whether that it is a thing we would like to do as a team moving forward.

Yeah mostly informed by personal experience (and preference). 

- For simple ternaries, using the negation is unnecessary cognitive overhead
- It's easy to miss the `!` when reading code in general, which can lead to misunderstandings and again, cognitive overhead

gmaclennan (commented):
In reviewing this I noticed a potential race condition in the `useAudioRecording()` hook, in code which is not in this PR, but I think this PR should address. When the user starts recording, we call the async function `await Audio.Recording.createAsync()`, which then calls `setRecording(recording)`. When the user clicks `stop` we change the recording state, and then call stop on it.

Because `await Audio.Recording.createAsync()` is async, there will be a period of time when the user could click `stop`, and the `recording` state is `null`, so `stop` would no nothing, and then the async creation would complete, and audio would start recording, when the user is expecting it to stop.

I am unsure how long `await Audio.Recording.createAsync()` takes, and if it only take milliseconds, then it would be hard to encounter this race condition, but on lower-end devices sometimes these async actions can take longer.

I think the solution is to store the promise in the state, which makes it slightly annoying because you have to await the promise in the state when you want to use it, but we don't actually expose the `recording` state outside this hook, and we only use it in effects, so I think this is ok. It could be ok to leave, but it's one of those little edge-case race conditions that is hard to debug when it actually happens.

gmaclennan (commented):
Did a fly-by review specifically to look at the challenge of how to pass context for confirming deletes etc. on a separate screen (e.g. the "bottom sheets"). I agree with your assessment that the easiest approach for now is to pass a function as a screen parameter. The resulting code seems easy to understand, and I don't see any obvious downsides other than the react navigation docs warning against it. I'm happy with keeping this pattern, until we come across issues with this.

ErikSin:
> In reviewing this I noticed a potential race condition in the `useAudioRecording()` hook, in code which is not in this PR, but I think this PR should address. When the user starts recording, we call the async function `await Audio.Recording.createAsync()`, which then calls `setRecording(recording)`. When the user clicks `stop` we change the recording state, and then call stop on it.
> 
> Because `await Audio.Recording.createAsync()` is async, there will be a period of time when the user could click `stop`, and the `recording` state is `null`, so `stop` would no nothing, and then the async creation would complete, and audio would start recording, when the user is expecting it to stop.
> 
> I am unsure how long `await Audio.Recording.createAsync()` takes, and if it only take milliseconds, then it would be hard to encounter this race condition, but on lower-end devices sometimes these async actions can take longer.
> 
> I think the solution is to store the promise in the state, which makes it slightly annoying because you have to await the promise in the state when you want to use it, but we don't actually expose the `recording` state outside this hook, and we only use it in effects, so I think this is ok. It could be ok to leave, but it's one of those little edge-case race conditions that is hard to debug when it actually happens.

So technically startRecording is a promise already.

We could block the UI:

Something like this:
```ts
function start() {
    setLoading(true);
    startRecording().then(() => {
      setLoading(false);
    });
  }

 //we have to wait for the uri to be produced after the recording is done because it is being passed to the next page. So status?.isDoneRecording allows us to show a loader while the user is waiting for the uri to be made. 
  const showLoader = loading || !!status?.isDoneRecording;
```
On a coding level it doesn't block the stop from being called, but it does block it on a UI level.

Or did you have something else in mind?

gmaclennan:
> Or did you have something else in mind?

I _think_ adding a loading state could work, although it complicates the API of the hook because it means you need to wait for the loading to finish before you call stop. I think it's probably easier to handle the loading state in the start and stop functions, which as you say are promises already. This is one way to solve the issue:

```ts
import {useState, useCallback} from 'react';
import {Audio} from 'expo-av';

  const [recordingPromise, setRecordingPromise] = useState<Promise<Audio.Recording> | null>(null);
  const [status, setStatus] = useState<Audio.RecordingStatus | null>(null);
  const [error, setError] = useState<Error | null>(null);

  const handleError = useCallback((err: unknown) => {
    const newError =
      err instanceof Error ? err : new Error('An unknown error occurred');
    setError(newError);
  }, []);

  const reset = useCallback(async () => {
    try {
      if (recordingPromise) {
        const recording = await recordingPromise
        await recording.stopAndUnloadAsync();
      }
      setRecordingPromise(null);
      setStatus(null);
      setError(null);
    } catch (err) {
      handleError(err);
    }
  }, [recordingPromise, handleError]);

    try {
      const recordingPromise = Audio.Recording.createAsync(
        Audio.RecordingOptionsPresets.HIGH_QUALITY,
        stat => setStatus(stat),
      ).then(({recording}) => recording);
      setRecordingPromise(recordingPromise);
    } catch (err) {
      handleError(err);
    }
  }, [handleError]);

  const stopRecording = useCallback(async () => {
    try {
      if (!recordingPromise) return;
      const recording = await recordingPromise;
      await recording.stopAndUnloadAsync();
      return recording.getURI();
    } catch (err) {
      handleError(err);
    }
  }, [recordingPromise, handleError]);

  return {
    reset,
    startRecording,
    stopRecording,
    status,
    error,
    setError,
  };
}
```

Side-note, unrelated, I think `reset()` should probably just swallow the error (but maybe report to Sentry), since any error there is in the recording object that is being discarded anyway. Not a huge one, can be addressed at a later stage.

src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -389,6 +388,38 @@ export const createDefaultScreenGroup = ({
...
+        component={AudioPlaybackUnsaved}
+        options={({route}) => ({
+          ...sharedAudioNavOptions,
+          headerLeft: props => {
+            return (
+              <AudioCustomHeaderLeft
+                {...props}
+                duration={route.params.duration}
+                uri={route.params.uri}
+              />
```
achou11, line 408: slight style mismatch compared to the other header left buttons for similar screens. the [default custom header left component](https://github.com/digidem/comapeo-mobile/blob/0fc6a2c43e463e84af48cf9b5a6a39b6c56b829e/src/frontend/sharedComponents/CustomHeaderLeft.tsx#L32) applies a `style={{marginLeft: 0, marginRight: 15}}` whereas [AudioCustomHeaderLeft](https://github.com/digidem/comapeo-mobile/blob/cb54cb101d31d10d65928526825ee5d9d2f83925/src/frontend/screens/Audio/AudioCustomHeaderLeft.tsx) doesn't, causing it to be shifted more to the right in this case.

Other screens using the normal custom header left:

image

This screen as currently implemented:

image

This screen after applying the same style as mentioned:

image

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const {formatMessage} = useIntl();
+  const {goBack, replace} = navigation;
+  const [permission, setPermission] = React.useState<Audio.PermissionResponse>(
+    route.params.audioPermission,
+  );
+
+  // When the user changes their permission in phone settings and returns to the app,
+  // we need to check for updates and navigate accordingly.
+  // Without this, the app won't detect the permission change.
+  React.useEffect(() => {
```
achou11: should this be using `useFocusEffect()`? my understanding is that this should only run when this screen is actually focused on (and not hidden behind some other screen, for example).

maybe doesn't have much of difference in impact in practice.

achou11 (approved):
Minor comments but otherwise functionality seems good to me!

PR #960, opened 2025-02-11
fix: observation list icon

### LINKED ISSUE #945: (Correction) Icon fills are not 100%
https://www.notion.so/digidem/Correction-Icon-fills-are-not-100-10b1b08162d580f99a7de240f93f028a

All other navigation icons fill completely to 100% in Mapeo Blue except for the observations list which shows up as blue outlines but grey center. 

To check-in with Erik if it was done?

cimigree:
closes #945 
Modifies svg of observation list icon to take current color so that it is all blue when active instead of a blue outline and grey interior.
<image src="https://github.com/user-attachments/assets/1f7a9ec7-705f-4e18-a424-fc673e14a745" width="350" />


ErikSin (approved):
I think we should do both the stroke and color.

with stroke and color:
image

with just color:
image

PR #963, opened 2025-02-11
fix: do not require passcode entry when app is backgrounded due to share dialog opening

### LINKED ISSUE #955: Share difficult/ confusing with passcode enabled
https://www.notion.so/digidem/Share-difficult-confusing-with-passcode-enabled-1041b08162d580fbb9fee55f60a4a143

Remove passcode screen when sharing

achou11:
Fixes #955 

After setting up a passcode for the app, there are two ways of testing this:

1. Create an observation. Go to the observation and press the share button.
2. Create an observation with an audio attachemnt. Go to the observation and press the audio thumbnail and then press the share button.

The passcode screen should not show up at any point when the share dialog is open.

The fix for this does require us to remember to use the `useOpenShareDialog()` from `hooks/share.ts` instead of using `react-native-share` directly. Couldn't really think of a more elegant way of approaching this and I don't think this is too bad of a caveat.

---

https://github.com/user-attachments/assets/355924e8-87b7-4700-94f0-03d612824e85

### DIFF
--- a/src/frontend/contexts/SecurityContext.tsx
+++ b/src/frontend/contexts/SecurityContext.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {AppState, AppStateStatus} from 'react-native';
 import {usePersistedPasscode} from '../hooks/persistedState/usePersistedPasscode';
+import {useIsShareDialogOpen} from '../hooks/share';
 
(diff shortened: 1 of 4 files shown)


cimigree (approved):
I tested on my device. Looks good!

PR #968, opened 2025-02-13
Adds Univaja languages

luandro:
- Matis
- Matses
- Kanamari
- Marubo


cimigree (approved):
Just clarifying there are no functional changes brought about by this PR.
Should I add these 4 languages to the crowdin project, @ErikSin ?
Then translations can be added?
These languages won't show up in comapeo settings until they are present.
The filtering logic in `intl.ts:41-44` excludes languages without at least one translated string

cimigree:
I added these to Crowdin today, January 27, 2026.

Issue #957, opened 2025-02-10
E2E - Migrate to Maestro tests


ximenabb:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #969, opened 2025-02-13
E2e: onboarding

cimigree:
Related to #957 
- Adjusts the config file timeouts
- Adds a "flow" file containing the tests to be run in order
- Adds tests related to onboarding and editing a device name
- Creates a selector file for abstractions
- Creates a naming file
- Removes some test ids that cause duplicate ids to be present in the app, which is challenging for Appium to make selections
- Edits documentation for onboarding to include data and privacy and privacy policy screens


package.json
```diff
@@ -28,7 +28,7 @@
     "upgrade-deps:all": "npm-check --update --save-exact",
     "upgrade-deps:production": "npm-check --update --save-exact --production",
     "upgrade-deps:dev": "npm-check --update --save-exact --dev-only",
-    "test:e2e": "wdio run ./wdio.conf.js"
+    "test:e2e": "wdio run ./tests/e2e/config/wdio.conf.js"
```
ErikSin: For consistency, i think we should leave the `wdio.conf.js` at the root, like all the other configuration files, unless you have a reason that im unaware of to put it here!

tests/e2e/utils/selectors.ts
```diff
@@ -0,0 +1,8 @@
+export const byResourceId = (id: string) =>
+  `android=new UiSelector().resourceId("${id}")`;
+export const byText = (text: string) =>
+  `android=new UiSelector().text("${text}")`;
+export const byTextMatches = (text: string) =>
+  `android=new UiSelector().textMatches("${text}")`;
```
ErikSin: According to [the docs ](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector#textMatches(java.lang.String)) `textMatches` should be regex expression. They way you are using it here, it is not just the same as `UiSelector().text`?

tests/e2e/specs/project/edit-device-name.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  it('should navigate to project settings and edit the device name', async () => {
+    const drawerIcon = await $(byResourceId('drawer-icon-home'));
+    await drawerIcon.click();
+
+    const projectSettingsItem = await $(byText('Project Settings'));
+    await projectSettingsItem.click();
+
+    const deviceNameListItem = await $(
+      byResourceId('PROJECT.device-name-list-item'),
+    );
```
ErikSin, line 16: out of curiosity, why are you choosing to use `resourceId` instead of text here?

ErikSin (approved):
a couple minor change and questions, but otherwise looks good

tests/e2e/utils/selectors.ts
cimigree: Ok I was passing the regex in the test expression but I changed it to pass it in the selector expression to make it more automatic.

tests/e2e/specs/project/edit-device-name.test.ts
cimigree: Because it gets found a lot faster... .84 seconds as opposed to 20 seconds.

PR #970, opened 2025-02-13
Fix: secondary button color

### LINKED ISSUE #947: (Correction) Correct style for secondary buttons
- secondary button should look like this with these colors: (outlined)
Text: 0066FF (COMAPEO_BLUE)
Border: CCCCD6 (BLUE_GREY)
Fill: white
Stroke width: 1.5
Font: Rubik medium 16pt

https://www.notion.so/digidem/Correction-Correct-style-for-secondary-buttons-10b1b08162d5801eb19be0c1f6b1be80?pvs=4

cimigree:
closes #947 
See notion document for guidelines.
But, in sum, 

> Clarifications: A secondary button is any outlined button. 

> Our secondary button should look like this with these colors:
> 
> - Button Text Color: #0066FF
> - Button Border Color: CCCCD6
> - Button Stroke Width: 1.5
> - Button Fill: White
> - Button Font: Rubik Medium 16pt
> - Button Width: Buttons should be 280px width

So I updated all of the outlined buttons to fit these specifications. I changed the border color as specified. As far as the text,  I did so by passing the color string unless the header text had been added separately, in which case, I added the color to the header text (and the icon).
Passing the string seemed to be in keeping with the expectations of the button component, so that is what I did, rather than setting a default color on outlined buttons in the Button file. But I am not wedded to that idea and would be happy to hear feedback on that choice.
Particular attention was paid to the buttons mentioned in the Notion document. Widths were fixed in those cases as well.

Some screen shots:
<image src="https://github.com/user-attachments/assets/b310e3ee-dcac-487f-8e6c-a2f4e352c8f1" width="250" />
<image src="https://github.com/user-attachments/assets/bfae96d1-4848-4cf9-bfb8-4b7718ec3c88" width="250" />
<image src="https://github.com/user-attachments/assets/5eb6ea67-b381-4ce7-a781-b0b895445c64" width="250"/>
<image src="https://github.com/user-attachments/assets/a440787e-d309-4d60-966c-9e447790d3d8" width="250" />

### DIFF
--- a/src/frontend/screens/HowToLeaveProject.tsx
+++ b/src/frontend/screens/HowToLeaveProject.tsx
@@ -57,6 +57,7 @@ export const HowToLeaveProject = ({
         <Button
           fullWidth
           variant="outlined"
+          color="ComapeoBlue"
           onPress={() => {
             navigation.goBack();
           }}>
--- a/src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
+++ b/src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
@@ -52,7 +52,11 @@ const ObservationEmptyView = ({onPressBack}: Props) => {
         </Text>
       </View>
       <View style={styles.backButton}>
-        <Button onPress={onPressBack} variant="outlined" color="dark">
+        <Button
+          onPress={onPressBack}
+          variant="outlined"
+          color="ComapeoBlue"
+          fullWidth>
           {t(m.backButton)}
         </Button>
       </View>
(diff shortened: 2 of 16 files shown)


ErikSin:
So based on the design system and my conversation with Aldo, it sounds like the buttons styles are quite opinionated. So similar to what you suggested we do with the text, I think we should just create seperate opinionated button components for `primary`, `secondary`, and `destructive` button. And instead of manually setting the styling to each use of the button, we can simply choose the component and determine only the relevant styling (size).

ErikSin:
Created this ticket: #979

Issue #899, opened 2024-12-19
Expo Update


ErikSin:
Update Expo and all related dependencies to @latest

PR #971, opened 2025-02-13
Chore/upgrade dev dependencies

### LINKED ISSUE #942: Upgrade Dev Dependencies
Upgrade all dev dependencies. Upgrade dependencies to @latest and to exact versions. Can use `npm run upgrade-deps:dev`

ErikSin:
closes #942 
Upgrades Dev dependencies.

Every dependency was backwards compatible, so none of the code needed to be changed. While husky is backwards compatible, the api did slightly change so i updated that. 

I did not upgrade anything related to eslint as that is being dealt with in #897 .

I did not upgrade anything related to react native or metro as that was recently covered in the #899.

I did not upgrade glob or rimraf as they are being removed by #914 

Typescript was upgraded. There are changes that can be made to optimize typescript, but I did not include those and worth having a discussion about it.


.husky/pre-commit
```diff
@@ -1,4 +1 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-npx lint-staged
+lint-staged
```
achou11, line 1: I would restore the `npx` here just based on lint-staged's [docs](https://github.com/lint-staged/lint-staged?tab=readme-ov-file#examples):

```suggestion
npx lint-staged
```

package.json
```diff
@@ -156,26 +155,25 @@
...
     "react-native-svg-transformer": "1.5.0",
-    "react-test-renderer": "18.3.1",
     "rimraf": "5.0.5",
-    "semver": "7.6.0",
-    "type-fest": "4.26.0",
-    "typescript": "5.3.3",
+    "semver": "7.7.1",
+    "type-fest": "4.34.1",
+    "typescript": "5.7.3",
     "webdriverio": "9.7.2"
```
achou11: is it worth updating this too?

achou11 (approved):

.husky/pre-commit
ErikSin: its actually a [husky suggestion ](https://github.com/typicode/husky/releases/tag/v9.1.1) to take out npx!

PR #974, opened 2025-02-13
fix: allow importing category files in zip format when creating project

### LINKED ISSUE #966: Custom category - accept Zip files
Via Slack - as report by Luandro : https://awanadigital.slack.com/archives/C060CE2B97C/p1738777860627709

Custom category files must end in .cat , however, the app is adding .zip to the filename, thus generating an error
Goal : Accept zip filed to avoid this error

achou11:
Follow-up to #972 that fully fixes #966 , as the other PR forgot to update the case where we're importing a config when creating a project.

Highlighted some tech debt that I wanted to address related to selecting files, so there's a bit of refactoring in here as well. Touches code that's also used by the custom map import functionality, so I made sure to check that there aren't any regressions there.


src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
```diff
@@ -123,18 +125,23 @@ export const CreateProject: NativeNavigationComponent<'CreateProject'> = ({
     );
   }
 
-  async function importConfigFile() {
-    try {
-      const asset = await selectFile(['comapeocat']);
```
achou11, line 128: This was the line that needed to be updated to fully fix the original issue.

src/frontend/lib/file-system.ts
```diff
@@ -8,26 +8,37 @@ export function convertFileUriToPosixPath(fileUri: string) {
...
+ * @returns The selected file. If `null` it means the selection was cancelled.
+ */
 export async function selectFile({
-  mimeFilters,
+  copyToCacheDirectory,
   extensionFilters,
+  mimeFilters,
 }: {
-  mimeFilters?: Array<string>;
+  copyToCacheDirectory: boolean;
```
achou11, line 27: i made this option required because its value affects the result of the URI that the file has, as documented in the typedoc comment above. makes it slightly clearer about what you can or can't do with the file.

it's kind of an annoying difference that affects us, unfortunately.

src/frontend/hooks/useSelectFileAndImportConfig.ts
```diff
@@ -1,26 +0,0 @@
...
-    mutationFn: async () => {
-      const asset = await selectFile(['comapeocat', 'zip']);
-      if (!asset) return;
-
-      try {
-        return await importProjectConfigMutation.mutateAsync(
-          convertFileUriToPosixPath(asset.uri),
-        );
-      } finally {
-        await FileSystem.deleteAsync(asset.uri).catch((err: unknown) => {
```
cimigree, line 20: Am I correct that the file is no longer deleted after import? Might that leave the temp file in storage?

src/frontend/lib/file-system.ts
```diff
@@ -8,26 +8,37 @@ export function convertFileUriToPosixPath(fileUri: string) {
...
 
   const asset = documentResult.assets[0];
 
+  // Shouldn't happen but just in case
   if (!asset) {
-    throw new Error();
+    throw new Error('Expected document to be selected');
   }
 
   if (extensionFilters) {
```
cimigree: Is this ok that it is case-sensitive?

src/frontend/screens/Settings/Config.tsx
```diff
@@ -58,11 +69,25 @@ export const Config: NativeNavigationComponent<'Config'> = ({navigation}) => {
...
 
-  async function importConfigFile() {
+  async function selectAndImportConfigFile() {
     if (!isCoordinator) return;
-    selectAndImportConfigMutation.mutate();
+
+    selectFileMutation.mutate(
+      {
+        copyToCacheDirectory: true,
+        extensionFilters: ['comapeocat', 'zip'],
```
cimigree: Is there any room here to check the contents of the zip files (in file-system.ts)? to validate that once they are unzipped they are the correct file type? Otherwise it just fails silently, right?

cimigree (commented):
The code looks great in general. Just added a few questions
I like how it uses the existing selectFile hook which then uses the file-system to keep the logic centralized.

I tested with the zip file that [Luandro posted in slack ](https://awanadigital.slack.com/archives/C060CE2B97C/p1738777860627709) and was not able to add it as a config file in either place.

Well, I was able to upload it but then there was nothing in the categories. I did that when first creating a project.

See screenshot
<image src="https://github.com/user-attachments/assets/1867aa71-31b7-4d75-a2c7-eee92afb53c8" width="250" />

If I tried to add it when a project already existed, then the original file (either default or the working, comapeocat file he included) remained in place.

Here is the .comapeocat file categories (seems to have worked)
<image src="https://github.com/user-attachments/assets/b8e736ec-76e0-4281-80a3-5a38c4a4b981" width="250"/>

I also noticed that if I tried to upload a non accepted file type it just says something went wrong instead of a more clear message. Not sure if that is desireable.

src/frontend/hooks/useSelectFileAndImportConfig.ts
achou11: ah yeah should probably restore that. thanks for catching!

src/frontend/lib/file-system.ts
achou11: you're referring to the values that may be passed for this option? probably better to be case-insensitive! can fix

cimigree: Yes the values that can be passed  like zip vs ZIP

src/frontend/screens/Settings/Config.tsx
achou11: unfortunately not really. when using the config import mutation that actually calls the backend, the backend will check the contents and throw an error if invalid, which is propagated to the error callback of the mutation and can be handled there.

achou11:
re: the issue with the categories not loading: very strange! I think i ran into the same issue but it was because i didn't have the code to copy to the cache directory. thought adding that was the fix. may have to try again to see if I can repro the issue.

> I also noticed that if I tried to upload a non accepted file type it just says something went wrong instead of a more clear message. Not sure if that is desireable.

Not desirable by any means but a pre-existing issue that's more of a limitation of the shared component used for surfacing the error. Think it'd be a bigger piece of work to address and probably better to do as a follow-up

achou11:
@cimigree I reliably ran into the same issue that you're describing, but it only happens with the zipped category file that Luandro shared in Slack. I have a suspicion that there's some specific issue with that one in terms of the content because I haven't been to reproduce it with any other category file.

I've been doing the following with the other ones from the QA fixtures:

1. Download the original `.comapeocat` file
2. Rename it so that it has a ZIP extension e.g. `my-file.zip`
3. Import that file in the app

Although as you said, it does seem to work after you've created the project and re-import the file via the config settings 🤔 maybe it's a bug in core? 

EDIT: ~Actually it may be an issue with different invalidations that are happening between the project creation mutation (used in project creation screen) vs the import config mutation (used in config settings screen)~ Nope, tried with more aggressive cache invalidation for the former and it didn't change the result

achou11:
@cimigree confirming that the category file's contents are invalid. if you open the `metadata.json` file within it, you'll notice that it's missing many fields that are required (as indicated by [this check](https://github.com/digidem/comapeo-core/blob/6b1e97180c9e1cd560adc957ce1ffba451eeb039/src/config-import.js#L557) in core).

it's not ideal that the failure to use the file causes this really awful state. pretty sure this is somewhat of a known issue in core, although having this as an example of the consequences is helpful for fixing there.

would suggest trying out this PR using the other files after renaming them to be zip files. hopefully that's a better indication of the fix this PR is trying to introduce 😄

src/frontend/lib/file-system.ts
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/974/commits/6d454158b6d35b2f06be761468691383a990f746

src/frontend/hooks/useSelectFileAndImportConfig.ts
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/974/commits/0b0bbbc83f0123f831e7f1ab2400909bf273fc97

cimigree (approved):
Still not working to upload the zip file. I compressed a file from the QA assets to make a Zip. Probably a back end issue, but thought you should know.
I think the front end is working as it should though. 
See screen recordings.
Non zip file:
https://photos.app.goo.gl/bSrnTrcjgJEp4Qp47

Zip file:
https://photos.app.goo.gl/CZKg7gRGLN92iAX47

Got a something went wrong when I tried to import a video.

achou11:
> I compressed a file from the QA assets to make a Zip.

Ah yeah, worth making a distinction between:

- renaming a category file to have the ZIP extension
- compressing a category file into a ZIP file

My understanding of Luandro's issue is that some services add a `.zip` extension to the file when downloading, but not necessarily recompress it. This PR addresses that issue as it's mostly related to just a renaming of a file.

If the category file is _recompressed_, it'd be a zip within a zip and while the frontend could select that file, the backend will not be able to actually read it properly, and result in an error as you were probably seeing. There are definitely some services that may do this, but this PR is not trying to address that scenario as it's something that'd have to be fixed in core or require user intervention to extract the embedded zip before selecting it.

Basically, what you were testing is expected. The issue that this PR solves is being to select the file in the first place just based on the extension name. Any issues with the actual contents of the selected file are concerning core.

PR #975, opened 2025-02-13
feat: indicate tracks from other users

### LINKED ISSUE #936: Exchange Tracks/ No blue line in obvs list
- When exchanging data, received observations will appear with a blue line on the side
- For tracks, it is not the case. so user doesnt know which Track came from another device

ErikSin:
closes #936 

image


cimigree (changes requested):
@ErikSin I noticed this is failing CI. Can you fix that before I review, please?

src/frontend/screens/ObservationsList/TrackListItem.tsx
```diff
@@ -27,12 +30,24 @@ const TrackObservationItemNotMemoized = ({
   onPress,
 }: ObservationListItemProps) => {
   const {formatMessage} = useIntl();
+  const {data: createdByDeviceId} = useOriginalVersionIdToDeviceId(
+    track.originalVersionId,
+  );
+  const {data: deviceInfo} = useDeviceInfo();
+
+  const isMine =
+    createdByDeviceId && createdByDeviceId === deviceInfo?.deviceId;
```
cimigree, line 39: would it make any sense to put this in a hook or function and returns a simple boolean?
useIsMine(originalVersionId) or something?
Then the render function in the list item is more about layout and style?

src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -38,18 +39,17 @@ function ObservationListItemNotMemoized({
...
-  const {data: deviceInfo, status: deviceInfoQueryStatus} = useDeviceInfo();
 
   const photos = observation.attachments.filter(isSavedPhoto);
 
-  const {
-    data: createdByDeviceId,
-    status: originalVersionIdToDeviceIdQueryStatus,
-  } = useOriginalVersionIdToDeviceId(observation.originalVersionId);
-  const isMine = createdByDeviceId === deviceInfo?.deviceId;
-  const queriesSucceeded =
```
cimigree, line 50: Is it worth still checking if deviceInfo or createdByDeviceId is still loading? If so, would the isMine be incorrect (in the new file) and thus apply the incorrect styling? (would default to not isMine when I am not sure that is the more likely scenario)

cimigree (approved):
Some suggestions and questions but otherwise seems fine.

PR #976, opened 2025-02-14
fix: Removes excess shadow from observation list

### LINKED ISSUE #950: (Correction) Inconsistent shadow on Observations List top nav bar
The shadow for Observation List’s top navigation bar is inconsistent with how the shadow is treated everywhere is.

https://www.notion.so/digidem/Correction-Inconsistent-shadow-on-Observations-List-top-nav-bar-10c1b08162d580da9781d0eebec2b615?pvs=4

cimigree:
closes #950 (see Notion issue linked in Github issue)
There were excess shadows being applied in the observation list -- some from the header style and some from the container

The design team wanted the shadow to look like the rest of the app. The rest of the app just uses the default header style that comes from react navigation elements and is never specified. The shadow is applied by default as long as the header is not transparent. 

So, I took out any defining of shadows in the Observation List and now it matches the rest of the app.

The Sync Screen. What the Shadow is supposed to be.
<image src="https://github.com/user-attachments/assets/1ded8131-5dc6-4e12-b96a-b3d363a301f5" width="250" />

The new Observation List (empty)
<image src="https://github.com/user-attachments/assets/42bd3ef2-4e5f-499b-8b6b-4a51bb19a8e7" width="250" />

The new Observation List with a list item
<image src="https://github.com/user-attachments/assets/21a29141-15c9-417a-9d92-06d4fb941f2f" width="250" />

### DIFF
--- a/src/frontend/screens/ObservationsList/index.tsx
+++ b/src/frontend/screens/ObservationsList/index.tsx
@@ -10,7 +10,7 @@ import {ObservationsListBarIcon} from '../../Navigation/Tab/TabBar/ObservationsL
 import {ObservationListHeaderLeft} from './ObservationListHeaderLeft';
 import {NativeHomeTabsNavigationProps} from '../../sharedTypes/navigation';
 import {NoProjectWarning} from './NoProjectWarning';
-import {LIGHT_GREY, WHITE} from '../../lib/styles';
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
Looks Good!

Issue #933, opened 2025-02-05
Update useObservationHook


ErikSin:
Implement the new `useObservation` hook from #860.

To Do:
- [ ] update hook to follow this [Zustand store pattern](https://github.com/digidem/comapeo-desktop/issues/66)
- [ ] Add hook (but don't use in the app)
- [ ] Write tests for hooks

ErikSin:
closed via https://github.com/digidem/comapeo-mobile/pull/1641

Issue #910, opened 2025-01-20
Set Up React Testing Library


ErikSin:
Set Up React Testing Library, including mocking the necessary the contexts.

achou11:
Going to close this as a result of #1217 getting merged. Think the basics of what we needed have been addressed. Just a matter of adding more tests as needed and filling the gaps!

PR #980, opened 2025-02-17
chore: refactors useObservation Hook

ErikSin:
Closes partially #933.

Adds the new hook. The hook is not being used anywhere, and there are no tests. I will write tests in a follow up PR, I just want to make sure that were agreed to how it is being implemented


src/frontend/App.tsx
```diff
@@ -78,6 +73,10 @@ TaskManager.defineTask(
   },
 );
 
+const persistedDrafObservationStore = createDraftObservationStore({
```
gmaclennan: `persistedDrafObservationStore` -> `persistedDraftObservationStore`

src/frontend/contexts/AppProviders.tsx
```diff
@@ -34,6 +36,7 @@ type AppProvidersProps = {
   mapeoApi: MapeoClientApi;
   appMetrics: AppDiagnosticMetrics;
   deviceMetrics: DeviceDiagnosticMetrics;
+  persistedDrafObservationStore: DraftObservationStore;
```
gmaclennan, line 39: Small naming nitpick: the provider will accept a store whether it is persisted or not, so naming the prop just `draftObservationStore` will make this clearer.

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -0,0 +1,179 @@
...
+  DraftState,
+  DraftObservationStore,
+  convertPosition,
+} from './PersistedStores/DraftObservationStore.ts';
+
+export function useDraftObservationInstance(): DraftState;
+export function useDraftObservationInstance<T>(
+  selector: (state: DraftState) => T,
+): T;
+export function useDraftObservationInstance<T>(
```
gmaclennan: Naming: As I understand it, you are using `instance` to refer to an instance of the store. What this hook is returning is the `state` of the store, not actually a store instance. I know the Zustand hook is called `useStore()`, but it does take a store instance as the parameter, so it makes a bit more sense. I don't feel strongly about this, feel free to ignore.

docs/PersistedState.md
```diff
@@ -0,0 +1,93 @@
...
+// No need for a selector because this is stable across renders
+export function useBearActions() {
+  const {actions} = useBearContext();
+  return actions;
+}
+```
+
+Main things to note:
+
+- We are returning the store state as `instance` and the actions that manipulate the state as `actions`
```
gmaclennan: Naming nit: We are returning a store _instance_ as `instance`, but in the hooks we return the store _state_ (as returned from `storeInstance.getState()`)

gmaclennan (approved):
All looks ok to me. Just a few comments on naming, which can be ignored, because naming is hard, except for `Draf` -> `Draft`

src/frontend/contexts/AppProviders.tsx
ErikSin: The store used here in `AppProvider` should always be the persisted store, no? My understanding is that in testing, we should create another instance of the provider and store.

src/frontend/contexts/DraftObservationContext.tsx
achou11: personally would find it a bit confusing to keep the `instance` naming here given that it returns the state of the store instance and not the instance itself. It's helpful to clearly distinguish between a store instance and its state given the updated approach to zustand we're trying to adopt.

Would prefer renaming this to avoid the usage of `instance` to avoid confusion e.g. `useDraftObservation()` or `useDraftObservationState()`.

src/frontend/contexts/AppProviders.tsx
achou11: > The store used here in AppProvider should always be the persisted store, no?

In the case of unit-like tests, it may not be desirable nor necessary to use a persisted store. using a persisted store in jest tests would require mocking out mmkv, which may be unnecessary work depending on what you're trying to test.

unless you're doing some end-to-end test where something needs to survive app restarts, the app provider and anything using that store state within the app shouldn't be affected by whether the state is persisted or not.

ErikSin: > In the case of unit-like tests, it may not be desirable nor necessary to use a persisted store. using a persisted store in jest tests would require mocking out mmkv, which may be unnecessary work depending on what you're trying to test.

I understand that we do not want to use a persisted store in our tests. 

I was saying that in testing environments, I was under the impression that we are going to create a different instance of the provider and create a store that is not persisted to pass to that provider.

Example
```tsx
const draftObservationStore = createDraftObservationStore({persisted:false})

const TestingProviders = ({children}) => {
  return (
    <OtherProviders>
      <DraftObservationProvider store={draftObservationStore}>
        {children}
      </TranslationProvider>
    <Other Providers>
  )
}

// custom render used in testing environment
const customRender = (
  ui: ReactElement,
  options?: Omit<RenderOptions, 'wrapper'>,
) => render(ui, {wrapper: TestingProviders, ...options})

//Aka, this is what i thought we were not doing:
const draftObservationStore = createDraftObservationStore({persisted:false})
const TestingProviders = ({children}) => {
  return (
    <AppProvider {...rest} store={draftObservationStore}>
        {children}
    <AppProvider>
  )
}

```

ErikSin: re: [this comment](https://github.com/digidem/comapeo-desktop/pull/73#discussion_r1892462346)

gmaclennan: It's fine to keep the naming if you like, it doesn't really matter. I think Andrew and I are just a bit OCD about naming, and the point we are making is that this is like having a component that takes a background image, and calling the prop "blueBackgroundImage" because we always use a blue background in the app. But that's just personal preference and doesn't have a consequence for the app.

achou11: yeah agree that i think the naming won't matter in the long run in this case

ErikSin: Just to be clear, im not caught up on the naming, Im trying to get at the root of the architecture here.

We are moving the persisted stores from global state to context. So what I am saying with this naming is that this context should explicitly always take a persisted store because it is what the app is using to persist state.

When we are not using the persisted store (in testing environments) we are creating an entirely different provider and entirely different store.

So perhaps the deeper question i am trying to get at is whether this app provider should ever take a non persisted store if it is not going to used in testing environments?

achou11: > When we are not using the persisted store (in testing environments) we are creating an entirely different provider and entirely different store.

Agree with using an entirely different store. That's the main benefit of the updated approach to how we're setting up zustand.

I think the question about how providers are set up in tests can vary based on the level of isolation you're trying to test. If we're not planning to use this `AppProviders` component in setting up tests, I see little reason to have it defined in the first place - why not just move the providers used within it directly into `App.tsx` in that case and avoid the props naming conversation altogether?

I think the potential utility of this `AppProviders` component is having consistency in the setup of these providers regardless of if you're using them in the running app or in a test, as the ordering of providers can make a difference. If that's the case, having the naming of certain props be less tied to implementation details is more aligned with that purpose.

achou11: > I think the potential utility of this AppProviders component is having consistency in the setup of these providers regardless of if you're using them in the running app or in a test, as the ordering of providers can make a difference. If that's the case, having the naming of certain props be less tied to implementation details is more aligned with that purpose.

To be clear on this note: not saying that tests should actually be using this as opposed to using the "minimally required" providers for them to properly test a specific thing. It may not make sense to, but not super sure off the top of my head.

gmaclennan: Ah ok, yes, I understand the confusion. I can try to explain here, happy to chat on Slack too if that helps.

In the examples you share, both are doing the same thing: they are passing a single instance of the store to all tests. The AppProvider is just a wrapper around the DraftObservationProvider, and it doesn't really matter which you use in tests. If it's convenient in tests, you could use the AppProvider, or if you only need the draft observation context, then you can use only the DraftObservationProvider in the wrapper.

The clunky thing in the tests, because the wrapper that you pass to the render function can't take any props, means that if you want to pass different instances for each test, then you need to create a new instance of the wrapper for each test. That means that a custom render function would look like this:

```ts
function createWrapper () {
  return function ({children}) {
    const store = createDraftObservationStore()
    return (
      <OtherProviders>
        <DraftObservationProvider store={store}>
          {children}
        </TranslationProvider>
      <Other Providers>
    )
  }
}

// custom render used in testing environment
const customRender = (
  ui: ReactElement,
  options?: Omit<RenderOptions, 'wrapper'>,
) => render(ui, {wrapper: createWrapper(), ...options})
```

Alternatively if you wanted to mock certain methods on the store in your test (not saying this is a good idea) then you might do:

```ts
function createWrapper ({ store }) {
  return function ({children}) {
    return (
      <OtherProviders>
        <DraftObservationProvider store={store}>
          {children}
        </TranslationProvider>
      <Other Providers>
    )
  }
}

// custom render used in testing environment
const customRender = (
  ui: ReactElement,
  options?: Omit<RenderOptions, 'wrapper'> & { store: DraftObservationStore },
) => render(ui, {wrapper: createWrapper({ store }), ...options})
```

then in the test do:

```ts
test('testing with mocked store', () => {
  const store = createDraftObservationStore()
  jest.mock(store.getState, () => 'foo') // can't remember the syntax here
  render(ui, { store })
```

If you liked in the test you could use the AppProvider (if you needed all the providers for the test, and the AppProvider is convenient - you don't need to do this, but it would work):

```ts
function createWrapper () {
  return function ({children}) {
    const store = createDraftObservationStore()
    return (
      <AppProvider {...rest} draftObservationStore={store}>
          {children}
      <Other Providers>
    )
  }
}

// custom render used in testing environment
const customRender = (
  ui: ReactElement,
  options?: Omit<RenderOptions, 'wrapper'>,
) => render(ui, {wrapper: createWrapper(), ...options})
```

TL;DR whether you use the AppProvider or just the DraftObservationProvider in the tests, it doesn't matter. What matters is that for each test you pass a new instance (or instances, if multiple providers) to each test.

ErikSin: > I think the potential utility of this `AppProviders` component is having consistency in the setup of these providers regardless of if you're using them in the running app or in a test, as the ordering of providers can make a difference. 

I agree, and I think this is what we should be doing: using a reusable component means testing will more closely resemble the app, and we only have to maintain it in one place.

And if `AppProviders` is used in the testing environment, I 100% agree that the prop should be agnostic (it shouldn’t care whether it’s persisted or not).

What I’m confused about is that I previously tried to do this in desktop—creating a reusable provider for both testing and the app—but at the time, I was told it wasn’t necessary. Could you clarify what changed?

achou11: > What I’m confused about is that I previously tried to do this in desktop—creating a reusable provider for both testing and the app—but at the time, I was told it wasn’t necessary. Could you clarify what changed?

Mostly not having a full understanding of the benefit of having a reusable provider wrapper component at the time. It's clearer to me now that it _can_ be helpful, but at the time it felt premature.

gmaclennan: > When we are not using the persisted store (in testing environments) we are creating an entirely different provider and entirely different store.

And just to clarify here, we are using exactly the same provider (`DraftObservationProvider`) in tests as in the app. Wrapping it is `AppProvider` doesn't change the provider, it's just a syntax convenience to wrap the DraftObservationProvider and all the other providers. We are using an entirely different store, but with the same provider.

I think this conversation is getting into questions about how we setup tests and ensure isolation between tests, which would be more useful to have in a different place for future reference.

ErikSin: > And just to clarify here, we are using exactly the same provider (`DraftObservationProvider`) in tests as in the app. Wrapping it is `AppProvider` doesn't change the provider, it's just a syntax convenience to wrap the DraftObservationProvider and all the other providers. We are using an entirely different store, but with the same provider.

In [a previous PR](https://github.com/digidem/comapeo-desktop/pull/73#discussion_r1892462346), it was implied that I **should not** use `AppProviders` in testing. Instead, I should create a new instance of all the providers separately and use those new instances in tests.  

Based on that, I was suggesting that we should create a completely **new** instance of `DraftObservationProvider` in tests—rather than using `AppProviders`. Given this approach, my argument was that `AppProviders` should **always** take a persisted store explicitly, since its only function would be to pass persisted state to the app (and **not** be used in testing).  

That said, I now think we **should** use `AppProviders` in tests—or at least have a reusable provider that can be used in both the app and tests.  

So, I won’t change the prop name in this PR. Instead, in ticket #910, I’ll create a reusable provider. In that provider, I’ll use a more agnostic prop name since it should support both persisted and non-persisted stores.

PR #981, opened 2025-02-19
fix: workaround for file lock errors crashing app

### LINKED ISSUE #982: Fatal error occurs when opening app after installation
**Describe the bug**

Basically the user-facing error that's related to https://github.com/digidem/comapeo-core/issues/995. This has only been noticed on certain devices but basically makes the app unusable for those devices, as pressing the `Restart App` button eventually leaves you in the same state.

**To Reproduce**

1. Install the app
2. Open it
3. Screen with message `Something went wrong` and a `Restart App` button appears.

**Expected behavior**

Should be able to open the app and reach the onboarding, at least.

gmaclennan:
This is a workaround for https://github.com/digidem/comapeo-core/issues/995. For affected devices, it means that the hypercore storage files are not actually locked (the attempt to lock the file is silently ignored). The files in question are the `oplog` files for each Hypercore. This should not be an issue in normal operation, because there should not be any other process which is trying to read or write to these files while CoMapeo is running.

Fixes #982

### DIFF
--- a/src/backend/patches/README.md
+++ b/src/backend/patches/README.md
@@ -7,3 +7,9 @@ These patches use [patch-package](https://github.com/ds300/patch-package) to upd
 ### [Fix shim insertion](./@rollup+plugin-esm-shim+0.1.4+001+fix-shim-insertion.patch)
 
 This is a workaround for [a bug in the plugin](https://github.com/rollup/plugins/issues/1709).
+
+## `random-access-file`
+
+### [Ignore EINVAL errors](./random-access-file+4.0.7.patch)
+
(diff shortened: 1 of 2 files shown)


gmaclennan:
I have tested this workaround on these devices and I was able to reach the onboarding screens without experiencing a crash:

- Samsung Galaxy A11 (Android 10.0)
- Samsung Galaxy A10 (Android 9.0)
- Moto G7 Play (Android 9.0)

achou11 (approved):
~have you tested this already? if not, would you prefer I try it out before you merge?~ EDIT: nevermind, didn't see your follow-up comment about having tested it already

Issue #983, opened 2025-02-19
App crashes on Android 7


achou11:
**Describe the bug**

The app crashes upon opening when using a device that is running Android 7. Was able to test this using a the `Samsung Galaxy S8` device from Browserstack App Live.

**To Reproduce**

1. Use device that's on Android 7
2. Open app

**Expected behavior**

Android 7 is the oldest supported Android version that our app can run on, so it shouldn't crash like this upon opening.

**Additional context**

Stacktrace of the crash:

```console
02-18 21:55:43.511 E/AndroidRuntime(31453): FATAL EXCEPTION: AsyncTask #1
02-18 21:55:43.511 E/AndroidRuntime(31453): Process: com.comapeo.rc, PID: 31453
02-18 21:55:43.511 E/AndroidRuntime(31453): java.lang.NoSuchMethodError: No virtual method toPath()Ljava/nio/file/Path; in class Ljava/io/File; or its super classes (declaration of 'java.io.File' appears in /system/framework/core-oj.jar)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at org.apache.commons.io.FileUtils.newOutputStream(FileUtils.java:2552)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at org.apache.commons.io.FileUtils.copyToFile(FileUtils.java:1077)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at org.apache.commons.io.FileUtils.copyInputStreamToFile(FileUtils.java:986)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at expo.modules.updates.UpdatesUtils.verifySHA256AndWriteToFile(UpdatesUtils.kt:115)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at expo.modules.updates.loader.LoaderFiles.copyContextAssetAndGetHash$expo_updates_release(LoaderFiles.kt:49)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at expo.modules.updates.loader.LoaderFiles.copyAssetAndGetHash(LoaderFiles.kt:33)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at expo.modules.updates.loader.EmbeddedLoader.loadAsset(EmbeddedLoader.kt:84)
02-18 21:55:43.511 E/AndroidRuntime(31453): 	at expo.modules.updates.loader.Loader.downloadAllAssets(Loader.kt:242)
[...]

PR #984, opened 2025-02-19
fix: disable expo updates (fix Android 7 crash)

gmaclennan:
Expo updates is crashing Android 7 devices in the latest version (https://github.com/digidem/comapeo-mobile/issues/983)

Expo updates allow over-the-air updates of bundled JS code. We currently do not use this feature, but it is enabled by default (https://docs.expo.dev/versions/latest/config/app/#enabled). Disabling the option should have no consequences for the app, it avoid unnecessary network traffic, and reduces the bundle size.

### DIFF
--- a/app.json
+++ b/app.json
@@ -8,6 +8,9 @@
     "icon": "./assets/icon.png",
     "userInterfaceStyle": "light",
     "assetBundlePatterns": ["**/*"],
+    "updates": {
+      "enabled": false
+    },
     "plugins": [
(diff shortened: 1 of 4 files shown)


achou11 (approved):
Can we also remove `expo-updates` as a dep then?

gmaclennan:
> Can we also remove `expo-updates` as a dep then?

I think so, let's see if tests pass with it removed.

PR #986, opened 2025-02-19
fix: trims text when creating project name

### LINKED ISSUE #940: Project name can be set to just space characters
https://www.notion.so/digidem/Project-name-can-be-set-to-just-space-characters-1581b08162d580f284e3f2fca738877c

When entering a name for your project, the validation for the name does not account for using just space characters. Results in having a project name that appears to be unset or blank, which is confusing.

ErikSin:
closes #940


cimigree (approved):
Works great. Should we apply the same rule to device name? That can also be all spaces...

PR #987, opened 2025-02-19
fix: show offline map in tracks view

### LINKED ISSUE #961: Tracks information doesn't show on default map when offline
https://www.notion.so/digidem/Tracks-information-doesn-t-show-on-default-map-when-offline-1971b08162d580288013debf5bccfe0d

ErikSin:
closes #961

you can run `npx expo start --localhost --android` to run the app in dev mode while the wifi is off.


cimigree (approved):
The offline map is very basic, but I guess it is better than the black screen!

PR #1005, opened 2025-02-26
chore: add e2e project tests

cimigree:
related to #957 
- Adds 6 more tests, most related to projects
- Adds some accessibilityLabels to make the finding of elements quicker and more accurate
- Uses those updated labels in existing tests
- adds a test profile in EAS, primarily now to be able to use an ENV variable
- Uses that env variable to set the style for the map. 
    - For some reason that I cannot determine, Browserstack has not been able to read our local urls and they are returning 502s, see https://github.com/digidem/comapeo-mobile/issues/1008
    - The mapbox style is one of those URLs and not having it messes everything up. It makes all of the queries in tests take forever and makes us unable to test anything related to observations.
    - So if the app is being built for E2E we now add an env variable to specify as such so that we can use the public mapbox fallback style
    - Using that default style makes the tests run 10X faster!!!
  
I am going to use that test profile in future experimentation to see if I can build that apk in such a way that it will read the urls from the fastify server on Browserstack.


src/frontend/screens/MapScreen/index.tsx
```diff
@@ -52,6 +52,10 @@ export const MapScreen = () => {
 
   const styleUrlQuery = useMapStyleJsonUrl();
 
+  const styleUrl = process.env.EXPO_PUBLIC_E2E_TEST
```
achou11: definitely worth adding a comment to explain the context for this

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -52,6 +52,10 @@ export const MapScreen = () => {
...
+    ? 'mapbox://styles/mapbox/streets-v11'
+    : (styleUrlQuery?.data ?? 'mapbox://styles/mapbox/streets-v11');
```
achou11: not sure if having the fallback here is desirable, as it could mean a rather drastic change in the map style once the query does load e.g. if the app is using a custom map, the user would see the mapbox fallback you've specified first and then switch to the custom map (which is very likely to be visually different)

```suggestion
    : styleUrlQuery.data;
```

src/frontend/screens/MapScreen/index.tsx
achou11: also, maybe not necessary to do right now but wondering if you think it'd be worth moving this to into `useMapStyleJsonUrl()` so you don't have to worry about doing this for other parts of the app that display a map e.g.

```ts
export function useMapStyleJsonUrl() {
  const api = useClientApi();
  const refreshToken = useRefreshToken();

  return useQuery({
    queryKey: [MAPS_QUERY_KEY, 'stylejson-url', refreshToken],
    queryFn: async () => {
      // 👇 Add this
      if (process.env.EXPO_PUBLIC_E2E_TEST) {
        return 'mapbox://styles/mapbox/streets-v11'
      }
      const result = await api.getMapStyleJsonUrl();
      return result + `?refresh_token=${refreshToken}`;
    },
  });
}
```

achou11 (changes requested):
couple of small requested changes but otherwise looks good! the tests read quite nicely 👍

src/frontend/screens/MapScreen/index.tsx
cimigree: Good point about moving it to the hook...I did that here. Hope that is ok.

src/frontend/screens/MapScreen/index.tsx
cimigree: OK took out fallback (but irrelevant now that I moved to the hook).

src/frontend/hooks/server/maps.ts
```diff
@@ -34,6 +34,11 @@ export function useMapStyleJsonUrl() {
   return useQuery({
     queryKey: [MAPS_QUERY_KEY, 'stylejson-url', refreshToken],
     queryFn: async () => {
+      // If we're running E2E tests (e.g. on BrowserStack), fall back to a
+      // public Mapbox style rather than our local style server to avoid 502 errors.
```
achou11, line 38: would add a link to the github issue in this comment for more direct access to additional details

```suggestion
      // If we're running E2E tests (e.g. on BrowserStack), fall back to a
      // public Mapbox style rather than our local style server to avoid 502 errors.
      // (see https://github.com/digidem/comapeo-mobile/issues/1008)
```

achou11 (approved):
one non-blocking suggestion but lgtm otherwise!

PR #1014, opened 2025-03-03
fix: disallow space only device names

### LINKED ISSUE #973: Device name can be set to just space characters
When entering a name for your device, the validation for the name does not account for using just space characters. Results in having a device name that appears to be unset, which is confusing. Happens in both the onboarding step of setting a device name and the screen for editing device name later on.

See details : https://www.notion.so/digidem/Device-name-can-be-set-to-just-space-characters-1581b08162d5808bab08e3ce9b41b85b?pvs=4

Can be linked to : https://github.com/digidem/comapeo-mobile/issues/940

cimigree:
Fixes #973  
Makes sure need to enter some kind of text for device name by using trim()

The Edit Device Name did not need to be adjusted as it already does not allow just a blank space name.

### DIFF
--- a/src/frontend/screens/Onboarding/DeviceNaming.tsx
+++ b/src/frontend/screens/Onboarding/DeviceNaming.tsx
@@ -36,7 +36,6 @@ export const DeviceNaming = ({
 }: NativeStackScreenProps<OnboardingParamsList, 'DeviceNaming'>) => {
   const [name, setName] = React.useState('');
   const [errorTimeout, setErrorTimeout] = useTemporaryError();
-  const invalidName = name.length === 0 || name.length > 60;
   const {formatMessage: t} = useIntl();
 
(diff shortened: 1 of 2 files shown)


achou11 (approved):
would be a good case to cover via e2e tests, although maybe that's a follow-up since the current onboarding e2e test isn't ported yet?

cimigree:
> would be a good case to cover via e2e tests, although maybe that's a follow-up since the current onboarding e2e test isn't ported yet?

@achou11 I can add it. I have that test ported. Want me to add it before I merge this in? You want to approve it?

achou11:
> @achou11 I can add it. I have that test ported. Want me to add it before I merge this in? You want to approve it?

Ah great! No strong preference if done here or as a follow-up

achou11 (approved):

PR #1015, opened 2025-03-03
fix: simplify app check

### LINKED ISSUE #1004: Simplify opening app check for exisiting observation
Currently when a user opens the app we check for the existence of an observation in persisted state. If an observation exists, we use getInitailRouteName to navigate the user of the Observation Create/Edit Screen.

Due to this, we hve to derive the navigation stack to have the `home` screen in it so the user can go back to the home screen.

To simplify this, we can just check for the existence of an observation in the `home` screen. If there is an observation, then navigate to the Observation Create/Edit Screen.

## To Do:
- [ ] Add useEffect to home (Map) Screen
- [ ] add appropriate logic from`getInitialRouteName` into this useEffect
- [ ] remove `initialRouteName` from RootStackNavigator (the onboarding screen are conditionally rendered anyways)
- [ ] remove the derived navigation states

cimigree:
closes #1004 

- [x] Adds useEffect to home (Map) Screen
- [x] Adds appropriate logic from getInitialRouteName into this useEffect
- [x] Removes initialRouteName from RootStackNavigator
- [x] Removes the derived navigation states 

Also adds an E2E test to check if it works
Adds documentation for the test


ErikSin (changes requested):
We need to remove the derived nav states that we were creating before to deal with the cases where the user was navigated to `ObservationCreate` and `ObservationEdit` Screens initially, but had no screen to go "back" to.

See 
https://github.com/digidem/comapeo-mobile/blob/5c1ae450fda5519e7a4300b2bafa01009a22f06c/src/frontend/screens/ObservationCreate/index.tsx#L167  
https://github.com/digidem/comapeo-mobile/blob/5c1ae450fda5519e7a4300b2bafa01009a22f06c/src/frontend/screens/ObservationEdit/index.tsx#L112
https://github.com/digidem/comapeo-mobile/blob/5c1ae450fda5519e7a4300b2bafa01009a22f06c/src/frontend/screens/ObservationEdit/HeaderLeft.tsx#L68
https://github.com/digidem/comapeo-mobile/blob/5c1ae450fda5519e7a4300b2bafa01009a22f06c/src/frontend/screens/ObservationCreate/HeaderLeft.tsx#L69

User should now just be able to goBack when an observation is created/Edited.

cimigree:
@ErikSin I had to add some logic in the Map screen to only go to the observation screens when the app is restored from the background/ reopened
I had to leave the navigation reset in the Observation Edit (success) and ObservationEdit Header Left on dismiss otherwise the param for the previously clicked on Observation (from the Observation List) remained in the navigation stack and couldn't be cleared
I did a lot of tweaks to the tests to try to make them less flaky

ErikSin:
>  I had to leave the navigation reset in the Observation Edit (success) and ObservationEdit Header Left on dismiss otherwise the param for the previously clicked on Observation (from the Observation List) remained in the navigation stack and couldn't be cleared I did a lot of tweaks to the tests to try to make them less flaky

Is this a problem because in `AudioPlaybackSaved` we try to popTo observationEdit without a param? I think it would be better to make the `{observationId: string}` mandatory (take out `undefined` in the navigation types). It would require us to pass the `observationId` to `AudioPlaybackSaved`, but I think thats feels like a clearer solution and makes everything very explicit.

ErikSin:
>I had to add some logic in the Map screen to only go to the observation screens when the app is restored from the background/ reopened 

Why is that?

cimigree:
> > I had to add some logic in the Map screen to only go to the observation screens when the app is restored from the background/ reopened
> 
> Why is that?

I think you are right. I just needed to take the existingObservation out of the dependency array.

cimigree:
I really don't think the changes I made have anything to do with whether or not the observation Id would not show up when deleting audio, as the code I wrote only runs when the observation editing is saved or discarded.
But, since you requested it, I did it. The only way to get the observation id to the audio playback saved by passing, is to pass from the ObservationEdit (to the Editor, to the MediaScroll View, to the Audio thumbnail and then to the audioplaybacksaved), which didn't make a lot of sense to me (as you wanted to make sure the param was in observation edit, so it didn't make sense to me to pass it FROM the observation edit through all of those components/ screens). So, I got it out of the store in the audio playback saved. Hope that is satisfactory to you.

ErikSin:
> I really don't think the changes I made have anything to do with whether or not the observation Id would not show up when deleting audio, as the code I wrote only runs when the observation editing is saved or discarded. But, since you requested it, I did it. The only way to get the observation id to the audio playback saved by passing, is to pass from the ObservationEdit (to the Editor, to the MediaScroll View, to the Audio thumbnail and then to the audioplaybacksaved), which didn't make a lot of sense to me (as you wanted to make sure the param was in observation edit, so it didn't make sense to me to pass it FROM the observation edit through all of those components/ screens). So, I got it out of the store in the audio playback saved. Hope that is satisfactory to you.

Just to clarify. I wasn't requesting that you do that. I was just trying to understand why we having the `observationId` set in params was causing problems. When i was looking into the code I assumed that it was due to the audio screen and was suggesting that we should fix the audio screen if that was the case.

ErikSin (changes requested):
Sorry, I think there is some confusion happening and let me know if you want to meet on a call to talk about. The main issue that we are trying to solve is the complexity around having to manually derive and set the navigation state. The screens should not have to worry about what state the navigation in is as there is a high maintenance cost as any update to navigation could impact these screen.

Were still setting the navigation state in Observation Edit, so let me know if you need support in trying to take that out

cimigree:
Is clearing the params still manually deriving and setting the navigation state? This also fixes the issue I was having.
That is the issue. I am not sure of another way and would be up for hearing ideas. 
If you don't do clear the params somehow, then the param of whichever observation you had clicked on earlier to edit stays in the params, so whenever you go to edit a different observation, it is that first one you went to edit instead of the current one you are trying to edit.

Also, this is failing CI because apparently I have to put the Observation and the preset in the dependency array, but if I do that, then It wrongly forced navigation every time a user added a detail to an observation, and it created a loop where the user cannot exit ObservationEdit properly (either by saving or discarding).

So could I disable the lint warning? Or do you have another idea?

ErikSin:
> If you don't do clear the params somehow, then the param of whichever observation you had clicked on earlier to edit stays in the params, so whenever you go to edit a different observation, it is that first one you went to edit instead of the current one you are trying to edit.

Im a little confused why this would happen. If you navigate by doing `navigate('observationEdit' {observationId:newId})`, `newId` should overwrite the previous id? Otherwise all navigation params wouldn't work and would only work the first time

ErikSin:
> Also, this is failing CI because apparently I have to put the Observation and the preset in the dependency array, but if I do that, then It wrongly forced navigation every time a user added a detail to an observation, and it created a loop where the user cannot exit ObservationEdit properly (either by saving or discarding).

if you use useFocusEffect, it will only run the when the screen is focused, not when it is in the background

ErikSin (changes requested):
There is a nav error happening when the user has changed their preset before saving. Because the preset screen is previous screen, `goBack` goes to that screen.

In observation create, instead of goBack, lets `popTo` to `"Home", {screen:"Map"}`, for both the discard observation and mutation success. (see note below)

And in observation edit, instead of goBack, lets `popTo` to `"ObservaitonEdit", {observationId:observationId}`, again for both the discard observation and mutation success.

https://github.com/user-attachments/assets/d364878e-f8fd-4990-90ac-9d4452457a8a

Also, i know that in e2e tests after creating an observation we navigate to the observation list. The original reason we did this was for consistency between certain screens. And then those screen diverged, and we never updated it. So I think we should just go back to the original behaviour of just always popping to the map screen. would that make things too complicated for e2e test? Or is it just a matter of updating a couple lines in e2e? If it does add a lot of complication let me know and it might be good to just keep the behaviour as is.

cimigree:
@ErikSin Some added complications in ObservationEdit because there is the very, very slight chance there won't be an id in there (and it has to now be passed to the header left) but I think the alternative of navigating to the map in that case is a good fall back. What do you think?

ErikSin:
> @ErikSin Some added complications in ObservationEdit because there is the very, very slight chance there won't be an id in there (and it has to now be passed to the header left) but I think the alternative of navigating to the map in that case is a good fall back. What do you think?

I actually think this goes back to the audio comment that I was making earlier (aka we should just make the observation id mandatory for the observation edit screen). BUT i think it is fine for now. Im actually going to update some navigation stuff, so I will fold it into that work

ErikSin (approved):

PR #1017, opened 2025-03-04
chore: replace drawer navigator with animated component

### LINKED ISSUE #1001: Simplify Drawer Navigator with Drawer Layout
Currently we are using the [Drawer Navigator](https://reactnavigation.org/docs/drawer-navigator) as the parent navigator for the entire app. This complicates the navigation as we now have 2 nested navigators (root stack and home tabs). We do not actually use the Drawer Navigator, as its actions are to navigate within the rootstack. We can replace the Drawer Navigator simply with the [Drawer Layout UI](https://reactnavigation.org/docs/drawer-layout)

As a part of this work we can also move the `React.Suspense` that is wrapping the Drawer Navigator into the  RootStack [Navigator Layout](https://reactnavigation.org/docs/navigator#layout)

### To Do
- [ ] Remove the Drawer Navigator
- [ ] Add https://reactnavigation.org/docs/drawer-layout in the map screen
- [ ] Move React Suspense to RootStack Navigator Layout

ErikSin:
closes #1001 

I attempted to replace the drawer navigator with [Drawer Layout](https://reactnavigation.org/docs/drawer-layout/), but I was not able to prevent the behaviour of closing the drawer when the user clicks outside the drawer. Instead i just replaced the drawer with an animated component:

image


cimigree (commented):
Everything looks good except there is some different/ weird back button behavior that I am noticing in the Passcode settings.

It may be occurring elsewhere, but I see it in the passcode settings.

If you go back, you are going back through all of the screens that you went to. So once you set your passcode, you go back through the set passcode screen.

That is causing one of the tests to fail.

Here is a screen recording. Sorry it is long. It doesn't happen if you just straight up set the passcode, but it does occur if you are in and out of those screens a bit, like is done in the test, which is why I am showing the video.
If you want to follow along, it is this test:
tests/e2e/specs/passcode/set-passcode.test.ts

https://photos.app.goo.gl/S3k5MX36nHe7Q9d47

cimigree:
@ErikSin The test is failing because something is wrong... When I close the app (send to background) and open it again, the screen is blank. In the test this is happening with the passcode, but when I test locally, that is happening on any screen. Going to get a screen recording. If you want to see what is happening, go to here: https://app-automate.browserstack.com/dashboard/v2/builds/793479130d296f811a3d0663985e64301986868d/sessions/86ff72cf7a5351b0fdee14e24ec966888985fc4d
Play the video, right around 6:21

ErikSin:
yeah, im quite stumped why this is going on...The entire map screen is not being mounted, including the header, which means something is happening with the stack navigation....but that is as far as I have gotten...

ErikSin:
I figured it out, suspense was causing problems and dismounting the navigator. Im going to do a follow up PR to deal with suspense and simplify navigation even further

cimigree (approved):
Glad you figured it out. Can't say I understand it... But glad it works now.

PR #1021, opened 2025-03-06
chore: update track bottom sheet with animated view

### LINKED ISSUE #1006: Update Tab Bar with animated component for tracks
Remove the deprecated `BottomSheetModal` that is being used on the track tab. Replace with a new component that uses [React Native Reanimated](https://docs.swmansion.com/react-native-reanimated/) to slide up and down from the tab bar

## To Do
- [ ] Remove Bottom Sheet Modal
- [ ] Create Custom Animated Component
- [ ] Refactor content of modal to be used with this custom component

ErikSin:
Previously the track bottom sheet was quite complicated.

Previously the bottom tab bar was created by react navigation. This was done by having each tab screen set a [tabBarIcon](https://reactnavigation.org/docs/bottom-tab-navigator#tabbaricon). By having each tab set their own icon, react navigation handles creating the tab bar including all the navigation actions. The complication was that in order to have 4 tabs we needed 4 tab screens and the fourth tab (tracks) was not its own screen. Tracks is actually just the map screen with the bottoms sheet open. So to get around this, we had the track screen render an empty component and then we intercepted the tab touch event, navigated to the map tab screen, and opened the bottom sheet. But when the bottom sheet was open, we wanted the track tab icon to be highlighted. Since we were technically on the map screen, we had to manually keep track of which tab was clicked. We did this by using zustand to keep track of a global state of which tab the user had selected. And we had to sync that state with the navigation (aka, if the user was on the map screen, then the zustand store needed to indicate the map icon was clicked).

The other complication was that when tracking was cancelled we wanted to close the tracks bottom sheet. Since the tracking is handled in persisted state, we needed to propogate the bottom sheet ref into a context so that when the useTracking hook called 'cancelTracking` it could close the bottom sheet using that ref. This created an extra layer of hoisting the bottomSheetRef into another context.

This Pr simplifies the logic by creating a [custom tab bar.](https://reactnavigation.org/docs/bottom-tab-navigator#tabbar). Because this tab bar is NOT controlled by React Navigation, it means that the Tab Navigator only needs to have 3 screens, and does not need a 4th empty screen to create a 4th tab bar item. Of course, it means that we need to maintain the tab bar and make sure that each icon in the tab bar does the appropriate navigation action. But I think the the tradeoff of manually setting the onPress actions is worth not having to maintain a empty navigation screen that intercepts the touch event

Secondly, I removed the zustand store that kept track of which icon was clicked. I simply had the state of whether an icon is selected be based off navigation. In otherwords if currentRoute.name === 'Camera', then the camera icon was highlighted as the selected icon. This declarative approach means we do not manually have to keep track and sync that state with navigation

Obviously there is the matter of the Track Icon, which does not have its own screen (so we cannot do currentRoute.name === 'Track'). But I simply moved whether the track bottom sheet was open into the `Map` screen params. Remember that tracks is the map screen, but with a bottom sheet open. So when the user selects the tracks icon, I simply navigated to the `Map` screen and set the param `trackinOpen:true`. When `trackinOpen===true`, the bottom sheet is animated open. Otherwise it is closed (and animated on close). And we are able to derive whether the Track Tab should be highlight by checking the route name and param. 

Since the the bottom sheet being open is set by Nav `Params` the useTracking hook simply needs to `setParam` of `trackingOpen` to false in order to close the sheet. And this means we do not need to propgate the sheetRef and we can get rid of the context completely.

Lastly I removed the use of Gorham Bottom Sheet. I simply used an Animated.View since all we needed was the animation of the bottom sheet going up and down.

** Note that 8 of the 22  'file changed` are deleted filed, and few of the files simply update styling

closes #1006


src/frontend/Navigation/Tab/TabBar.tsx
```diff
@@ -0,0 +1,83 @@
...
+            currentTab?.name === 'ObservationsList' ? COMAPEO_BLUE : MEDIUM_GREY
+          }
+          height={BUTTON_SIZE}
+        />
+      </TouchableOpacity>
+      <TouchableOpacity hitSlop={HIT_SLOP} onPress={onPressMap}>
+        <MaterialIcons
+          color={
+            //@ts-expect-error known type
+            currentTab?.name === 'Map' && !currentTab?.params?.trackingOpen
```
ErikSin, line 52: the alternative to this is :
```ts
            currentTab?.name === 'Map' &&
            (!currentTab?.params ||
            'trackingOpen' in currentTab.params &&
            currentTab.params.trackingOpen !== true)
              ? COMAPEO_BLUE
              : MEDIUM_GREY

```

so the ts ignore feels alot easier to reason out...

src/frontend/Navigation/Tab/TabBar.tsx
```diff
@@ -0,0 +1,83 @@
...
+          name="photo-camera"
+        />
+      </TouchableOpacity>
+      <TouchableOpacity hitSlop={HIT_SLOP} onPress={onPressTracks}>
+        <MaterialIcons
+          color={
+            currentTab?.name === 'Map' &&
+            currentTab?.params &&
+            'trackingOpen' in currentTab.params &&
+            currentTab.params.trackingOpen === true
```
ErikSin: arguably a ts-ignore here would look cleaner, but this logic is a little more staigtforward as there are no `||`

src/frontend/Navigation/Tab/TabBar.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  function onPressMap() {
+    navigation.navigate('Map', {trackingOpen: false});
+  }
+
+  function onPressTracks() {
+    navigation.navigate('Map', {trackingOpen: true});
+  }
+  return (
+    <View style={styles.container}>
```
cimigree, line 26: Can you please add accessibilityLabels and testIDs in here so all of the E2E tests 1. don't break and 2. goes as fast as is possible?

cimigree (changes requested):
There used to be a green dot indicator with a time indicator when Tracks was active but there no longer is. Is that intentional?

And can you please add back the accessibilityLabels and test IDs?

Aside from those two things, this PR looks like a nice refactor: fewer contexts, simpler navigation, and direct param = open or closed logic.

ErikSin:
> There used to be a green dot indicator with a time indicator when Tracks was active but there no longer is. Is that intentional?

Interesting...I see it on my screen:
image

ErikSin:
> > There used to be a green dot indicator with a time indicator when Tracks was active but there no longer is. Is that intentional?
> 
> Interesting...I see it on my screen: image

oh, you know what, the tab bar is missing the timer. Is that what you are referring to?

cimigree:
> > > There used to be a green dot indicator with a time indicator when Tracks was active but there no longer is. Is that intentional?
> > 
> > 
> > Interesting...I see it on my screen: image
> 
> oh, you know what, the tab bar is missing the timer. Is that what you are referring to?

Yes! There is a dot but the timer is cut off on my phone:
<image src="https://github.com/user-attachments/assets/4cd2fb08-3a75-4f24-94d8-528355ac3096" width="250" />

ErikSin:
ok, i updated the tab bar!
image

cimigree (approved):
Looks fantastic now! And all the tests pass! Yay!

PR #1024, opened 2025-03-06
chore: remove luxon

### LINKED ISSUE #935: Remove Luxon Dependency
Remove [Luxon](https://github.com/moment/luxon) as it is one of the bigger dependencies contributing to our bundle size.

There are 3 instances (2 functions) where it is being used.

ErikSin:
closes #935


src/frontend/lib/millisecondsToFormattedTime.ts
```diff
@@ -0,0 +1,21 @@
+export function millisecondsToHHMMSS(ms: number) {
```
cimigree, line 1: This is a nit, to be sure, but when I see the caps, I think of different things. Like MM is a double month digit. I know we are not using SQL, but we used to use the lower case convention and that is what is used in many formats these days. Is there a reason to change to all upper case?

src/frontend/screens/SaveTrack/SaveTrackButton.tsx
```diff
@@ -27,7 +26,7 @@ export const SaveTrackButton: FC = () => {
               longitude: loc.longitude,
             },
             mocked: false,
-            timestamp: DateTime.fromMillis(loc.timestamp).toISO()!,
+            timestamp: new Date(loc.timestamp).toISOString()!,
```
cimigree, line 32: This changes the format of what is saved.
DateTime.fromMillis(loc.timestamp).toISO()! gives you
"2025-03-13T16:01:52.620-04:00"
(offset for local time with the -04:00)

Whereas 
Date(loc.timestamp).toISOString()
" 2025-03-13T20:01:52.620Z"
gives you just UTC -- no offset. I don't know exactly when/ where this is being used, but if the time in the local zone matters, or the backend is expecting that time, this would not be an ideal change.

cimigree (commented):
One comment that you can ignore if you want.
One that you should think about, imo.
I love having the tests! It is nice to see how things shake out and what you are testing so I can look at the UI to test whatever else I think I need to (which I did).

src/frontend/lib/millisecondsToFormattedTime.ts
ErikSin: main reason was that `millisecondsTohhmmss` reads as `milli-seconds Tohhmmss` to me. But im not particulary attached to it either way.

ErikSin: actually `millisecondsTommss` feels unclear now. Im going to leave both as is.

src/frontend/screens/SaveTrack/SaveTrackButton.tsx
ErikSin: the second format (without the UTC) is how we save it observations. Both formats are valid in JS and will produce the same results when processed into a human readable date. Ttechnically `JS Date` just converts everything to UTC. 

In other-words it shouldn't cause a problem.

cimigree: ok

cimigree (approved):

Issue #1019, opened 2025-03-05
Port zustand-based states to use preferred setup


achou11:
Basically what's described in https://github.com/digidem/comapeo-desktop/issues/66

We have the following:

- [x] draft observation
- [ ] locale ~(see #1011)~ (~see #1027~) (see #1063)
- [x] settings (~see #1027~) (see #1060)
- [x] track (see #1031)
- [x] project id (see #1061)
- [x] passcode (see #1058)
- [x] obscure code (see #1058)
- [x] metrics (see #1064)

~Settings and locale are detailed in #1018~

Passcode and obscure code are detailed in #1036

PR #1027, opened 2025-03-10
refactor: consolidate settings-related state

### LINKED ISSUE #978: App does not use system language preference when expected

### LINKED ISSUE #1018: Consolidate persisted settings state
Currently we have various settings-related values that we store as persisted state:

- language
- passcode
- obscure code
- metrics diagnostics permissions
- coordinate format
- manual coordinate entry format

under the hood, each of these are stored as separate entries in the local storage in the following ways:

- language
- passcode + obscure code
- metrics diagnostics permissions
- coordinate format + manual coordinate entry format

The proposed idea would be to have these all grouped in the same storage entry, which would make it easier to discover and manage persisted settings-related values. Access to these values can still be granular by implementing custom hooks that access the desired values using state selectors.

An idea for what the shape of the persisted state can look like:

```ts
type SettingsState = {
  locale: {
    languageTag: string | null
  },
  security: {
    passcode: string | null
    obscureCode: string | null
  },
  metrics: {
    diagnosticsPermissionsEnabled: boolean 
  },
  // Ideally these two are nested but can't think of a good name...
  coordinateFormat: CoordinateFormat;
  manualCoordinateEntryFormat: manualCoordinateEntryFormat;
}
```

The nesting of values might be overkill and I'm open to making things flatter.

Questions:

- how much do we care about preserving backwards compatibility? the answer may depend on which value we're talking about e.g. not preserving the previous language code is not as detrimental compared to not preserving the diagnostics permissions. I think ideally we preserve and move existing values from the previous storage locations to this new one, but will probably mean some additional code to handle that gracefully.

achou11:
Fixes #978
Closes #1018
Towards #1019 

Couldn't figure out a great way of doing this incrementally so apologies in advance for the large scope of this PR. Would generally recommend following the commits to get a somewhat sequential idea of how changes were applied.

Implementation notes:

- Consolidates the following states into a single state:

    - ~passcode and obscure code~ EDIT: reverted. see 368df00870ad6e7ffbdfdff5ae5768db0c2e52be
    - language
    - displayed coordinate format and manual entry coordinate format
    - metrics permissions
    
    As of this writing, the implementation is re-using the persistence key associated with the coordinate format and manual entry coordinate format, meaning that those are the only values that are _explicitly migrated/preserved_ as part of this PR. During my discussion with @ErikSin about migration of the persistence, we decided that at least as part of the initial implementation, migrating previously stored values wasn't a hard requirement given the nature of the app's current usage. There may still be a path where we try to migrate the previous values anyways, but not a focus for now.
    
- For all language-related changes, please refer to the details written in #1011 to understand the changes. I basically ported the changes from there into this PR because an incremental approach was too tricky. There are some small implementation differences (mostly related to naming) but aside from that, what's written in that PR is still very relevant here.

- Not in love with the naming, but made an effort to distinguish between setting state that acts as source of truth and the "resolved" more granular states that should be used within the app. Basically, any settings-related state that may have a default/fallback value when used in the app falls under this classification. Generally, the resolution of these settings happens in the following order:

    1. Check the relevant fields in the settings state. The settings state should always reflect what was explicitly set by the user and should always be updated via explicit interactions. If a field in the settings state is `null`, it generally means that the user hasn't done anything to explicitly choose a value for that field.
    2. Use a default/fallback value. In the case that the user has not explicitly set a value (or they have explicitly unset it), then it becomes an app-specific decision to decide what to fallback to.
    
    Using the displayed coordinate format as an example of the above:
    
    1. The relevant settings state is a valid value (e.g. `'utm'`, `'dd'`, `'dms'`), use that value.
    2. The relevant settings state is `null`, so we default to `'utm'`.
    
    The relevant settings with some form of this approach are:
    
    - displayed coordinate format
    - manual entry coordinate format
    - language (this is more involved since we also account for system preferences as part of the resolution process)
    - metrics permissions
    
    ~It is HIGHLY recommended to use the hooks from `resolvedSettings/` when reading settings-related values for the above. Reading from the `useSettingsState()` hook requires more work and is prone to inconsistency since it doesn't handle resolution of values when they are not explicitly set. However, reading the passcode and obscure code from the `useSettingsState()` directly is fine, as those will never have a fallback value when not set by the user.~ EDIT: No longer relevant. See https://github.com/digidem/comapeo-mobile/pull/1027/commits/061458e53fd2e4d72fe205cdde54f5fd9a851d98
    
- General features that are affected by the changes:

    - app onboarding (metrics permission in privacy section)
    - ~Passcode setup and usage~ EDIT: No longer the case due to reverting changes to passcode + obscure code
    - Language settings
    - Display of coordinates (displayed observations)
    - Inputs provided when manually entering a coordinate (observation creation)
    
    The language settings is the only feature that should see different behavior as a result of the changes (since it fixes an existing bug). The other features should not exhibit any notable change in behavior.


src/frontend/lib/utils.test.ts
achou11, line 1: moved to a more specific file in this directory

src/frontend/lib/utils.ts
achou11, line 1: moved to `coordinateFormat.ts`

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,168 @@
...
+    v.object({
+      languageTag: v.string(),
+    }),
+    v.null(),
+  ]),
+  manualCoordinateEntryFormat: v.union([CoordinateFormatSchema, v.null()]),
+  metricsDiagnosticsPermissionsEnabled: v.union([v.boolean(), v.null()]),
+  obscureCode: v.union([v.string(), v.null()]),
+  passcode: v.union([v.string(), v.null()]),
+});
```
achou11: welcoming input on whether this feels okay in terms of structure. Was thinking about doing some nesting to group related fields together, but that probably satisfies my OCD more than an actual need 😅

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+export type SettingsState = v.InferOutput<typeof SettingsStateSchema>;
+
+// NOTE: Do not change!
+export const STORAGE_KEY = 'Settings' as const;
```
achou11, line 45: wondering if it's really worth using the same key as before as opposed to starting fresh and using a completely new key. The only pre-existing values associated with this storage key are the coordinate format and manual entry coordinate format.

ErikSin:
@gmaclennan Andrew and I would like you opinion on this. We discussed this on slack, and we disagree with the best approach. So I am going to write down my thoughts, and he is going to further summarize his thoughts and we were hoping to have you help us come to a consensus.

I believe that having a separate default value and persisted state adds unnecessary technical complexity to solve a problem that doesn’t need solving.  

### Current Approach in This PR  
- The setting state has a **default value** (not persisted).  
- When reading the setting state, we check if a **persisted value** exists.  
- If a persisted value **does not exist**, we return the default value.  

#### **Pros**  
- The default value can change and be reflected in the app.  
- This was necessary for **language settings**, since the default was based on app settings, ensuring updates were reflected.  

#### **Cons**  
- Adds technical complexity and overhead as we are constantly checking for a persisted state and **reconciling two sources of truth**.  

### **Proposed Alternative**  
- On **initialization**, set the **persisted value** to the default.  
- When reading the value, **always read from persisted state**—no need to check if it exists as we know it is being set on initialization.

#### **Pros**  
- **Technically simpler** to maintain—only one source of truth.  
- Avoids ongoing logic to **reconcile values** within the hook.  

#### **Cons**  
- The **default value cannot change dynamically**.  
- Since it’s persisted, we **can't distinguish** between user-set values and defaults, meaning changes to defaults won’t be reflected unless users manually update their settings.  

While dynamic defaults were necessary for **language settings**, I don’t think it’s worth maintaining this logic for other settings. I dont think default value will change for setting, and even if they do, these values are not detrimental to the user experience.

Following our principles of **test ability** and **maintainability**, we should avoid unnecessary complexity. While I see why this approach was introduced for **language settings**, I don't think it’s needed for other settings.

ErikSin:
On a side note, I hadn’t seen the language-related changes earlier, but I think the technical overhead of maintaining the default language might be unnecessary. I understand that the goal is for the app language to follow the device language if the user hasn’t explicitly set a preference, and while I agree that’s the ideal behavior, I’m not sure it was worth the added complexity of introducing a default state. That said, since the work is already done, it may not matter at this point—sorry that we didnt do this architecture exploration before hand.

achou11:
To address some implementation concerns that have been raised:

### Why not specify the default values for certain fields when setting up the store state?

Given that the store state can be persisted, the main arguments I will make are:

1. default values should only be applied when _reading_ persistable state
2. stores should never implicitly persist values (default or not)

In the context of the store setup, this is what is meant by using default values:

```ts
function createInitialState() {
  return {
    // Use the default value instead of `null` 
    coordinateFormat: 'utm',
    // Other fields...
   }  
}
```

#978 is a good example of this of why we shouldn't do this. Using a simplified example, this is a sequence that can occur if you made the above change:

1. The app starts. The initial settings state is `{ coordinateFormat: 'utm', someOtherField: null }`, where the `'utm'` is the default value we want to use for the coordinate format. Due to how Zustand's persistence middleware works, this is NOT yet persisted into storage.

2. The user performs an action that updates `someOtherField` via `store.setState({ someOtherField: 'foo' })`. The settings state is now `{ coordinateFormat: 'utm', someOtherField: 'foo' }`. This DOES get persisted to storage. Now we've committed the default value of `coordinateFormat` to storage, despite the user performing an action that had nothing to do with it!

3. Turns out that `'utm'` is actually not what we want to use as the default value. We update the code such that `'dd'` is now the desired default value when the user has not explicitly made a choice.

4. However, the `coordinateFormat` with value `'utm'` is persisted already, so we defer to that value as opposed to considering the new default value, as we do not know if `'utm'` was the old default value that was _unintentionally stored_ or if it was an explicit choice that the user made. As a result, the new default value of `'dd'` can never be applied unless the storage is cleared.

In this example, we're talking about a setting that is relatively inconsequential in the grand scheme of things. However, there's nothing that prevents this happening for data that has much larger consequences or causes more problematic issues for users.

I would like to effectively communicate that the `SettingsState` ideally always represents what's been explicitly determined based on user interactions. Happy to better document this with some guidance.

### There is an implicit knowledge that one needs to use the more granular hooks instead of `useSettingsState()` in order to get resolved settings values

Yes, this is not ideal and a fair criticism. I mainly took this approach because it was very easy to write and test, and felt natural to do with zustand. As currently implemented, I don't really have a great suggestion to address this concern other than sufficient documentation, which probably still is prone to misuse.

An alternative approach could be to do the resolution within the `useSettingsState()` hook, such that now the app can always just use that hook to access resolved values. Rough first idea of an implementation would look kind of like this:

```ts
// Actual implementation might look more complicated to appease TS...
function useSettingsState(selector?: <T>(state: SettingsState) => T) {
  let actualSelector 
  
  if (selector) {
    actualSelector = (state: SettingsState) => selector(stateWithDefaults(state))
  } else {
    actualSelector = (state: SettingsState) => stateWithDefaults(state)
  }
  
  const {instance} = useSettingsStoreContext();
  
  return useStore(instance, actualSelector)
}

function stateWithDefaults(state: SetttingsState) {
  return {
    ...state,
    coordinateFormat: state.coordinateFormat === null ? 'utm' : state.coordinateFormat,
    // Other fields with default values...
  }
}
```

There are some types complications with this as written, but those are potentially addressable. A couple of drawbacks that come to mind:

- we lose the ability to read the settings state and basically know what's actually persisted in storage (assuming that's been enabled). probably not a big issue since we don't really have cases where that's actually useful in the app (i think?). one could work around this by exporting the `useSettingsStoreContext()` hook, but that's inconsistent with our general approach for zustand-based state.

- it doesn't work well in the case of values whose resolution relies on external factors. The language is a good example of this, where we need to account for system preferences as well. In this case, one could either:

1. update the above implementation to use `useLocales()` in `useSettingsState()`

2. have a separate hook that does the full resolution of the language setting (basically what `useLanguageTag()` does right now)

(1) is not ideal because now we lose the benefit of granular updates to the settings state i.e. with a naive implementation, any changes to the system-preferred locales updates the _whole_ settings state, despite it only being relevant for one field within that state.

(2) is not ideal because the resolution of the field (and the application of the default if needed) is happening elsewhere. 

Given a choice, I think the proper technical choice is (2), but then we run into the same issue related to having to know when to use the whole state hook vs a specialized hook. I guess only one exception is better than _N_ exceptions, but it's still an exception nonetheless.

achou11:
> Current Approach in This PR
> 
>     - The setting state has a default value (not persisted).
>     - When reading the setting state, we check if a persisted value exists.
>     - If a persisted value does not exist, we return the default value.

To clarify, certain _fields_ within the settings state will have defaults defined by app/design requirements. When we read the settings state and check each field, if the field does not have a meaningful persisted value (we use `null` to communicate this), then the relevant hook that processes that field will use the relevant default value.

EDIT: possible that by "default state", you mean "initial state"? if that's the case, this PR is introducing an initial state that is more resilient/reflective to how user settings works from a design perspective.

> Adds technical complexity and overhead as we are constantly checking for a persisted state and reconciling two sources of truth.

Not sure what the "constantly" refers to? We read _persisted_ state when the app starts. If you're referring to resolution for certain fields, yes that happens on every update of the store state but that's because settings can/should be un-settable by users throughout the app's lifetime, in order to allow defaults to be applied if necessary.

re: "reconciling two sources of truth", i could be misunderstanding what this refers to, but from my perspective, the store state remains the sole source of truth for settings that are actually defined by the user. We still have to account for the case where they don't specify some settings in order for the app to be usable, which should be done at runtime without implicitly persisting that choice for the user. 

Alternatively, if this is referring to the existence of the more granular hooks, nothing stops us from doing something like this in multiple places:

```ts
function Foo() {
  const coordinateFormat = useSettingsState(state => state.coordinateFormat || 'utm')
}

function Bar() {
  const coordinateFormat = useSettingsState(state => state.coordinateFormat || 'utm')
}
```

While this works fine, it's prone to inconsistency (what if you did `'dd'` in `Bar` by accident?). The implementation of the granular hooks is mostly to mitigate this concern, with the overhead of being another hook you have to know about. I find the overhead to be okay because it's easy to make mistakes in the way shown above.

> I’m not sure it was worth the added complexity of introducing a default state.

I think this comment is misunderstanding the language issue a bit. The changes to language settings in this PR is NOT introducing a default state - we already had that but the way it is done is flawed and causes user-facing issues. Instead, what this PR is introducing is (more) properly representing when the user has not specified a language to use, and it's updating the calculation of the default value to use more thoroughly.

gmaclennan:
I had a look through this today. I share concerns that this is adding a lot of technical overhead, and I also see that a possible problem without the proposed changes is that if we change any defaults, then the behaviour is indeterminate: if a user has changed _any_ setting, then changing a default will not affect them, but if they have changed no settings, and we update the defaults, then they will see the new setting when they update the app.

I'm not sure it is a huge issue because I can't see a reason why we would change a default, and if we did we _may_ actually want it _not_ to change for existing users. However I can see the argument for "doing this right" and only persisting a value when the user has selected it.

I think for the language setting, this is not a default vs. persisted issue, but rather there is an additional "setting" for language, which is "use system language", e.g. the user can choose a specific locale for the app, or they can choose to use the system locale.

I think the implementation here, with multiple hooks and a separate `useSettingsStore()` hook that shouldn't really be used, in multiple places, does add technical overhead that we don't really want. It also creates a very opinionated abstraction with an unclear boundary between what is in which hook.

I think there is a simpler way to implement what is done here, which could be more flexible and easier to maintain:

1. Rather than null values, make all setting properties optional. This way adding a new settings doesn't require migration (e.g. adding a null value in the migration funciton).
2. Drop the `useSettingsStore()` hook, and instead make a simple wrapper `useSetting(settingName)` to assign defaults
3. Only add a wrapper hook for a specific setting where strictly necessary, e.g. language, which needs to read the system language.

This way the settings store and `useSettingState()` hook have very limited and clear responsibilities: for storing and setting values in a "key-value" store, according to a schema.

```ts
const SettingsSchema = v.object({
  coordinateFormat: CoordinateFormatSchema,
  locale: v.union([
    v.object({
      languageTag: v.string(),
    }),
    v.null(),
  ]),
  manualCoordinateEntryFormat: CoordinateFormatSchema,
  metricsDiagnosticsPermissionsEnabled: v.boolean(),
  obscureCode: v.union([v.string(), v.null()]),
  passcode: v.union([v.string(), v.null()]),
});

type Settings = v.InferOutput<typeof SettingsSchema>;

const SETTINGS_DEFAULTS: Settings = {
  coordinateFormat: 'dd',
  locale: null,
  manualCoordinateEntryFormat: 'dd',
  metricsDiagnosticsPermissionsEnabled: false,
  obscureCode: null,
  passcode: null,
};

const SettingsStateSchema = v.partial(SettingsSchema);

export type SettingsState = v.InferOutput<typeof SettingsStateSchema>;

// NOTE: Do not change!
export const STORAGE_KEY = 'Settings' as const;

function createInitialState(): SettingsState {
  return {};
}

export function createSettingsStore({persist} = {persist: false}) {
  let store: StoreApi<SettingsState>;

  if (persist) {
    store = createStore(
      createPersistedState(createInitialState, {
        name: STORAGE_KEY,
        storage: createJSONStorage(() => MMKVZustandStorage),
        version: 1,
        migrate,
      }),
    );
  } else {
    store = createStore(createInitialState);
  }

  function setState<T extends keyof SettingsState>(
    key: T,
    value: SettingsState[T],
  ) {
    store.setState(state => {
      const newState = {...state};
      newState[key] = value;
      return newState;
    });
  }

  return {
    instance: store,
    setState,
  };
}

export type SettingsStore = ReturnType<typeof createSettingsStore>;

export const SettingsStoreContext = createContext<SettingsStore | null>(null);

export const SettingsStoreProvider = SettingsStoreContext.Provider;

function useSettingsStoreContext() {
  const value = useContext(SettingsStoreContext);

  if (!value) {
    throw new Error('Must set up SettingsStoreProvider first');
  }

  return value;
}

export function useSettingState<T extends keyof Settings>(
  key: T,
): [Settings[T], (value: Settings[T]) => void] {
  const store = useSettingsStoreContext();
  const state = useStore(store.instance, state => {
    return {...SETTINGS_DEFAULTS, ...state}[key];
  });
  const setState = React.useCallback(
    (value: Settings[T]) => store.setState(key, value),
    [key, store.setState],
  );
  return React.useMemo(() => [state, setState], [state, setState]);
}
```

When we want to add a new setting, we update the schema and add a default to `SETTINGS_DEFAULTS`. Other than the schema, the settings hook is agnostic to what's in it.

Long-term the passcode probably does not belong in here, and needs to be in a secure store, but I think ok to leave it here for now. If we want to strictly type needing a passcode in order to have an obscure code, then we could merge them e.g.

```ts
const SettingsSchema = v.object({
  //...
  passcodes: v.union([
    v.object({passcode: v.null(), obscureCode: v.null()}),
    v.object({passcode: v.string(), obscureCode: v.string()}),
    v.object({passcode: v.string(), obscureCode: v.null()}),
  ]),
});
```

then:

```ts
const [{ passcode, obscureCode }] = useSettingState('passcodes')
```

I also recommend using the valibot schema in the migration function, and not setting the value (e.g. reverting to defaults) if an invalid value is found, in case we mess up migration somehow.

I think this achieves similar to what @achou11 is looking for, without all the technical overhead and maintenance cost, and the problem of having two hooks which can return different values for the same setting, which could easily not be noticed by new devs on the team.

achou11:
@gmaclennan thanks for this response, it's helpful to see a suggested alternative implementation that most likely addresses concerns for both me and Erik.

From how I understand it, the main points based on feedback from both of you boil down to:

- the reasoning for the change in how settings state is represented is generally valid

- the current implementation needs a lot of work

> Long-term the passcode probably does not belong in here, and needs to be in a secure store, but I think ok to leave it here for now

This is a good point. Sticking to my "doing it right" mindset, I'm wondering if we should keep the security things out of this PR and address those separately instead? It would be nice to reduce the surface area of this PR. Nothing is forcing us to move the security-related fields into the settings state as this PR does - it was mostly with the intention of having a vehicle to update their setup to align with our preferred zustand state approach. If possible, it would be nice to avoid having to port the security stuff twice (i.e. moving to consolidated settings, then moving to secure store)

gmaclennan:
Yes I think maybe simpler to just leave passcode untouched for now. And I think you summarize the main points from my feedback at least.

Just one note as I was thinking about this after posting, might be worth updating the setState implementation to (requires latest valibot):

```ts
const setState = React.useCallback(
    (value: Settings[T]) => {
      v.assert(SettingsSchema.entries[key], value);
      store.setState(key, value);
    },
    [key, store.setState],
  );
```

That way you can write a stricter schema, e.g. add a regex validation for language, or other [actions](https://valibot.dev/api/#actions) in the schema, to avoid the need to write extra wrapper hooks.

gmaclennan:
One final thought that came into my head today: using `null` as a "special" value for language to mean "use system language" is a bit opaque. It may be better if the setting is:

```ts
type Settings = {
  locale: { languageTag: null, useSystemLanguage: true } | { languageTag: string, useSystemLanguage: false }
  // ...
}
```

to be more explicit.

achou11:
> One final thought that came into my head today: using `null` as a "special" value for language to mean "use system language" is a bit opaque. 

As I'm trying the suggestion out, I have some concerns about it. From my perspective, `null` is meant to represent that the user has not chosen a specific language. The app can do whatever it wants with this information. In our case, we happen to try to use the system preferences, which is a runtime decision.

Maybe part of the misalignment I'm feeling is due to the distinction I make between a _state that's stored_ and a _state that's used_. For example, I agree that `null` as state that's used isn't the clearest (although it is the least opinionated), which is where your suggestion makes sense. However, I don't agree that suggestion makes sense as a state that's stored. Maybe this is what you meant, but it's hard to tell with the context I have.

It also doesn't help that in the case of language, #1028 isn't yet implemented. Probably would make figuring this out a lot clearer for me

EDIT: Maybe my concerns are less valid given the future UI that allows an explicit choice of system language. Think a lot of my thinking around this has been assuming its absence.

gmaclennan:
> EDIT: Maybe my concerns are less valid given the future UI that allows an explicit choice of system language. Think a lot of my thinking around this has been assuming its absence.

I had been assuming we were fixing that, so that's probably why we're diverged in thinking. Treating following system language as an option only available when the user has not selected a language, and once they have selected a language they can never go back to that "follow system language" is a bug. I don't think we should be following a pattern with settings where we have a special setting with different behaviour when a user has not selected an option vs. when they have selected one. I think we should have settings defaults, and the user can change away from them, or return to them.

In terms of persistence, there is no perfect answer, but I think the one that makes most sense is:

1. If a user has never changed an option, it's not persisted, so if the defaults change, then their setting changes
2. If a user has changed an option, even if they have changed it back to the default, then it is persisted, so if defaults change, their setting does not change.

I don't really think that handling persistence and defaults really matters. It's unlikely we would change a default. If we did, then the negative user experience is minor - it might just be that it changes unexpectedly and they have to change it back.

We can use `null` in storage to represent "has not selected an option yet", however it does take more code to implement than just not persisting anything (e.g. no property or value in the stored JSON) until the user changes a particular setting (as I shared in my example code).

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,170 @@
...
+    v.union([
+      v.object({
+        languageTag: v.string(),
+      }),
+      v.literal('system'),
+    ]),
+  ),
+  manualCoordinateEntryFormat: v.optional(CoordinateFormatSchema),
+  metricsDiagnosticsPermissionsEnabled: v.optional(v.boolean()),
+});
```
achou11, line 25: changed this to use optional fields instead of null unions

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,170 @@
...
+                persistedState.manualCoordinateEntryFormat,
+              )
+            ) {
+              migratedState.manualCoordinateEntryFormat =
+                persistedState.manualCoordinateEntryFormat;
+            }
+          }
+
+          return migratedState;
+        },
```
achou11, line 90: > I also recommend using the valibot schema in the migration function, and not setting the value (e.g. reverting to defaults) if an invalid value is found, in case we mess up migration somehow.

@gmaclennan can you elaborate on this? As currently implemented, this migration function does the following:

- ports the existing `coordinateFormat` from v0 settings
- ports the existing `manualCoordinateEntryFormat` from v0 settings

for the other fields in v1, it applies the initial state i.e. omitting the field entirely (since initial settings state is now `{}`).

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,170 @@
...
+          SETTINGS_DEFAULTS.manualCoordinateEntryFormat,
+        metricsDiagnosticsPermissionsEnabled:
+          state.metricsDiagnosticsPermissionsEnabled ??
+          SETTINGS_DEFAULTS.metricsDiagnosticsPermissionsEnabled,
+      };
+
+      return selector ? selector(resolvedState) : resolvedState;
+    }),
+  );
+}
```
achou11, line 153: noting that I haven't yet implemented the API that Gregor suggested. wanted to see if i could have this so that it's more similar to how our other zustand state hooks are implemented. with this approach, state and actions remain as separate hooks as opposed to the suggestion of unifying them per settings field.

Basically comparing the following:

```ts
const coordinateFormat = useSettingsState(state => state.coordinateFormat)
const {setSetting} = useSettingsActions()

function onPress() {
  setSetting('coordinateFormat', 'dd')
}
```

vs

```ts
const [coordinateFormat, setCoordinateFormat] = useSettings('coordinateFormat')

function onPress() {
  setCoordinateFormat('dd')
}
```

if there's a strong preference for the latter, I can update this to align with that approach.

src/frontend/metrics/AppDiagnosticMetrics.ts
achou11, line 1: not overly excited about how i've done things here, so definitely looking for thoughts on how to improve this. feels like there's a separate resolution implementation that's happening here compared to the SettingsStoreContext but I didn't want to introduce some kind of dependency injection for the class

src/frontend/contexts/SettingsStoreContext.tsx
```diff
@@ -0,0 +1,158 @@
...
+export const SettingsStateSchema = v.object({
+  coordinateFormat: v.optional(CoordinateFormatSchema),
+  locale: v.optional(
+    v.union([
+      v.object({
+        languageTag: v.string(),
+      }),
+      v.literal('system'),
+    ]),
+  ),
```
achou11, line 22: thoughts on this shape as opposed to what was suggested by Gregor?

```ts
locale?: 'system' | { languageTag: string }
```

vs

```ts
locale?: { languageTag: null, useSystemLanguage: true } | { languageTag: string, useSystemLanguage: false }
```

achou11 (commented):
Highlighting some areas now that I've implemented many changes based on feedback. Thanks for raising concerns and providing suggestions. I think there are still a couple of smaller details to figure out but overall feeling much better about how this looks.

achou11:
Closing this PR in light of the reasoning I provide in https://github.com/digidem/comapeo-mobile/issues/1018#issuecomment-2741430448 for NOT consolidating. Our discussion identified that consolidating unrelated values is one of the main issues for the technical complexities that have been raised in this PR (e.g. actions accidentally persisting unrelated values).

PR #1030, opened 2025-03-11
chore: simplify navigation

ErikSin:
With the upgrade of react navigation to version 7, the removal or the drawer navigator, and the update of the tab bar, we can greatly simplify the navigation.

1. With the addition of the [Layout Prop](https://reactnavigation.org/docs/upgrading-from-6.x#new-layout-props) in navigators and screen, we can wrap entire screen without wrapping the navigator. Wrapping the navigator was causing problems with suspense. When suspense was being mounted, it would unmount the entire navigator, and then remount when suspense was resolved. This was causing errors as Navigators are not meant to be unmounted. This PR uses the `StackNavigator` Screen Layout, which wraps each screen inside a suspense, allowing use to remove the suspense wrapping the navigator. This take out all the suspenses that were wrapping navigators
~~2. We previously had a listener at the AppNavigator level that listened for changes in navigation state. This listener would update a boolean to determine if the project invite sheet should be disabled/enabled. This mean that on every nav state change, we would cause a re-render of the app navigator, effectively doubling the amount of re-renders. This removes that listener, and instead just passes the route name to the project invite bottom sheet listener, where it can internally decide if it should be disabled/enabled.~~ I remember why I did this in the first place. There is some added complications with how the navigatorRef initializes. It initializes async, so by passing down properties from it, sometime we can pass undefined/uninitialized properties. There is probably a better way to do this, but I left it as it was previously because it felt out of scope for this PR 
3.  #1015 accidentally introduced unwanted behaviour where when an observation was created, the user navigated to the map screen, instead of popping back to the nav screen. This added the map screen to the nav stack, instead of going back into the stack, allowing users to go back into the observation create screen after they had already saved the observation. This PR changes `navigate` to `popTo`
4. The Observation edit screen would navigation back to an observation when it was completed. However because the `observationId` was not a mandatory param for the Observation edit screen, but is a mandatory param for the Observation Screen, we had to check for the existence of an observationId before we could navigate. This PR makes `observationId` a mandatory param for the observationEdit screen to simplify the nav behaviour.
5. In security, when a user is setting their password, the cancel button was using "navigate" instead of "popTo" causingthe same problem as 2 above. This PR fixes that


cimigree:
@ErikSin Looks like the security stuff (passcode) not working

ErikSin:
Invites is broken now...so trying to fix that

ErikSin:
OK, everything should be good now!

cimigree (approved):
Everything else looks like a solid improvement and glad we have the e2e tests to check everything. I did some additional poking around and testing and all worked as expected.

ErikSin:
Yeah! the e2e has been a god send for catching these bugs

PR #1031, opened 2025-03-11
refactor: update setup for track state

achou11:
Towards #1019 

Updates the track state to align with our preferred approach for Zustand-based state.


src/frontend/contexts/TrackStoreContext.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+      isTracking: false,
+      locationHistory: [
+        {latitude: 0, longitude: 0, timestamp: timestamp1},
+        {latitude: 1, longitude: 1, timestamp: timestamp2},
+      ],
+      observationRefs: [],
+      trackingSince: null,
+      distance: expect.any(Number),
+    });
+    expect(stateHook.result.current.distance).toBeGreaterThan(0);
```
achou11, line 210: might be preferable to test actual values instead of using a generalization, but felt like I was reaching into internal details too much

src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -0,0 +1,186 @@
...
+  createJSONStorage,
+  persist as createPersistedState,
+} from 'zustand/middleware';
+import * as v from 'valibot';
+
+import {MMKVZustandStorage} from '../hooks/persistedState/createPersistedState';
+import {LocationHistoryPoint} from '../sharedTypes/location';
+import {calculateTotalDistance} from '../utils/distance';
+
+const TrackStateSchema = v.intersect([
```
achou11, line 13: curious to get thoughts on more frequently using a schema-oriented approach to defining (some) types. i like it a lot when there's a need to perform some validation on data

src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -0,0 +1,186 @@
...
+          const newState = createInitialState();
+
+          if (version === 0) {
+            try {
+              // Even if persisted state has fields that are not part of the schema,
+              // Valibot will only extract the relevant fields (assuming all relevant ones pass validation).
+              // https://valibot.dev/guides/objects/
+              return v.parse(TrackStateSchema, persistedState, {
+                abortEarly: true,
+              });
```
achou11, line 80: not in love with having to remember this detail, but it definitely is convenient in this case.

src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -0,0 +1,186 @@
...
+      isTracking: v.literal(false),
+      trackingSince: v.null(),
+    }),
+  ]),
+]);
+
+export type TrackState = v.InferOutput<typeof TrackStateSchema>;
+
+// NOTE: Do not change!
+const STORAGE_KEY = 'MapeoTrack' as const;
```
achou11, line 48: once again posing the question of whether it makes sense to continue using the previous key or if we should start clean with a new key 🤔 Not sure how many people are using tracks at the moment

src/frontend/contexts/TrackTimerContext.tsx
```diff
@@ -9,7 +9,7 @@ interface TrackTimerContext {
 const TrackTimerContext = createContext<TrackTimerContext | null>(null);
 
 const TrackTimerContextProvider = ({children}: {children: React.ReactNode}) => {
-  const trackingSince = usePersistedTrack(state => state.trackingSince);
+  const trackingSince = useTrackState(state => state.trackingSince);
```
cimigree, line 12: would using a useMemo make sense here to get rid of this lint warning? Seems like now might be the time to do it...

src/frontend/contexts/TrackStoreContext.tsx
cimigree: I really doubt many people are using Tracks right now. I wonder if there is a way to find out?
That being said,
How weird would the old data be if we kept the same key? And if there are no users really, what harm is there in keeping the old key?

src/frontend/contexts/TrackStoreContext.tsx
cimigree: It does add overhead, but it does provide extra safety especially as we are changing shapes etc. So if we anticipate continuing shape changes and/ or partial data, then we should keep it. If everything is all set, them it seems like extra. So, I don't know. Hard to decide. This might be especially useful if you keep the old key, though.

src/frontend/contexts/TrackStoreContext.test.tsx
cimigree: Yeah that seems like it would be checking cheapruler? Or am I misunderstanding how that would work with exact numbers. Your test basically seems to check that things are being called and updated which I think is what it should, right?

cimigree (approved):
I wish we had E2E tests for this but your added tests are very thorough.
I did test creating, starting, stopping tracks, etc. and all worked as expected.
I do think it would make sense to deal with the lint warning now while you are touching that specific file. If we are never going to deal with those warnings, or that one in particular, might make sense to get rid of that check.

src/frontend/contexts/TrackTimerContext.tsx
achou11: I have a local change that will remove this context in favor of a hook, which make the warning irrelevant. planning to do as a follow-up PR because it felt a bit out of scope

src/frontend/contexts/TrackStoreContext.tsx
achou11: > How weird would the old data be if we kept the same key?

As currently implemented, using the same key will trigger the `migrate` function that I've specified for the zustand persistence middleware since the version that's stored (`0`) since it wasn't specified in first implementation) is different from the version I've specified in this new implementation (`1`). So unless the user never upgrades the app to a version with the new implementation, the persisted store state will reflect what I've implemented in this PR.

> And if there are no users really, what harm is there in keeping the old key?

Yeah i think for now it's better to keep the same key when possible. Forces us to get into the habit of thinking about migration more thoroughly!

src/frontend/contexts/TrackStoreContext.tsx
achou11: yeah I'm leaning towards keeping it for this case, mostly because the state is relatively complex. There will be cases where our store state is very simple (e.g. active project id) and the usage of valibot isn't as helpful.

src/frontend/contexts/TrackStoreContext.test.tsx
achou11: > Yeah that seems like it would be checking cheapruler?

Yeah essentially this. I think ideally we'd have unit tests for the underlying mechanism that actually calculates the distance. There's a very tricky edge case that makes the `addNewLocations()` more complex than one would imagine, due to how CheapRuler - and geographic coordinate systems in general - works.

The TL;DR is that depending on the first point that's provided, the resulting distance calculation from the `calculateTotalDistance()` helper (which uses CheapRuler underneath) can vary. For example:

This:
 
```ts
calculateTotalDistance([point_1, point_2, point_3])
```

and

```ts
calculateTotalDistance([point_1, point_2])  + calculateTotalDistance([point_2, point_3])
```

can produce different results depending on how large the difference between the latitudes of `point_1` and `point_2` is.

The `distance` field uses both approaches depending on the inputs and the existing points that were recorded. I've kept the implementation the same for now but would be interesting to further document if needed.

Again, I think this is an edge case because most usages of tracks won't traverse very large areas, where "large" means globally significant distances.

achou11:
> I wish we had E2E tests for this

Same! I do wonder how hard it would be to write it, but maybe more appropriate to figure out as a follow-up.

> I do think it would make sense to deal with the lint warning now while you are touching that specific file. If we are never going to deal with those warnings, or that one in particular, might make sense to get rid of that check.

In this case, I plan to circumvent the lint warning by replacing that file with a more appropriate hook-based implementation, but can address it for the time being. Agree with you in general though!

src/frontend/contexts/TrackStoreContext.test.tsx
cimigree: Documenting somewhere more findable than a PR comment sounds good.

achou11: documented this where the function lives, see 6ef46d8d2ed9fd379ee363656987e314f1c6cf6c

PR #1035, opened 2025-03-13
chore: e2e testing updates

cimigree:
1. Adds guidelines for running end to end tests with Appium etc.
2. Updates the config file to include running the tests on a fast(er) device than the Samsung. 
3. Updates some tests to be less brittle.


docs/EndToEndTests/E2EWithAppium.md
```diff
@@ -0,0 +1,166 @@
...
+     runner: 'local',
+     specs: [path.resolve(__dirname, 'tests/e2e/specs/flow.test.ts')],
+     maxInstances: 1,
+      capabilities: [
+      {
+      platformName: 'Android',
+      'appium:deviceName': 'emulator-5554', // Replace with your device name/ device ID (from adb devices)
+      'appium:platformVersion': '13.0', // Replace with your platformVersion
+      'appium:automationName': 'UIAutomator2',
+      'appium:app': 'build-1741119529300.apk', // Replace with where/ what your apk is
```
ErikSin: I assume the location should be relative to the comapeo folder?

wdio.config.js
```diff
@@ -28,16 +28,15 @@ const config = {
...
       'appium:automationName': 'UIAutomator2',
       'appium:app': process.env.BROWSERSTACK_APP_URL,
       'appium:autoGrantPermissions': true,
-      'appium:autoDismissAlerts': true,
       'bstack:options': {
         projectName: 'CoMapeo',
-        buildName: 'CoMapeo Android Build',
-        sessionName: 'Run Android e2e tests',
+        buildName: 'Parallel Android Build',
+        sessionName: 'Pixel 7 - Android 13',
```
ErikSin: does this build name and session name need to be changed?

ErikSin (approved):
Approved with a couple non blocking comments.

I am wondering if in the `describe` part of the tests we can add the name of the folder. For example the test "Add Details Flow" would be "Observations-Add Details Flow". The Reason I ask is i am trying to debug some problems with the e2e and I see this:

image

It would be easier to know that if I want to looks at the test "Map Screen Component Test" I should look under the "Main" folder to find the test.

Let me know if that makes sense

docs/EndToEndTests/E2EWithAppium.md
cimigree: Yep. Changed text.

wdio.config.js
cimigree: Yep! Sorry. Missed that.

cimigree:
> Approved with a couple non blocking comments.
> 
> I am wondering if in the `describe` part of the tests we can add the name of the folder. For example the test "Add Details Flow" would be "Observations-Add Details Flow". The Reason I ask is i am trying to debug some problems with the e2e and I see this:
> 
> image
> It would be easier to know that if I want to looks at the test "Map Screen Component Test" I should look under the "Main" folder to find the test.
> 
> Let me know if that makes sense

Ok sure. If that is helpful. I have got everything memorized which is why I didn't think of it!

PR #1037, opened 2025-03-13
chore(deps): update production deps

### LINKED ISSUE #996: Update Production Dependencies

achou11:
Closes #996 

Notes:

- Had to pin to one patch version before latest version of `@rnmapbox/maps` due to https://github.com/rnmapbox/maps/issues/3796

- For Expo-related modules, I only updated the ones that required updating patch versions. Did not update the following as they are already outside of Expo's preferred/listed compatibility and would require updating at least a minor version in order to reach latest:
    
    - `@sentry/react-native`
    - `react-native-gesture-handler`
    - `react-native-reanimated`
    - `react-native-safe-area-context`
    - `react-native-screens`
    
    Open to attempting but figured it's probably not worth spending too much time on for now.


package.json
```diff
@@ -170,7 +170,7 @@
     "globals": "16.0.0",
     "husky": "9.1.7",
     "jest": "29.7.0",
-    "jest-expo": "52.0.3",
+    "jest-expo": "52.0.6",
```
achou11, line 173: technically a dev dep so probably could've left this out. can omit if desired

cimigree (approved):
It took me a while to get this running for some unknown reason? But once I did everything seems to work fine.
We should probably make a note somewhere? To update the expo related modules that are out of compatibility to align with Expo. And to keep track of the mapbox bug. Not sure where that would go?

achou11:
> It took me a while to get this running for some unknown reason? But once I did everything seems to work fine.

Can you share any details on what you mean here? e.g. installation of deps, specific build steps, running the app, etc

> We should probably make a note somewhere? To update the expo related modules that are out of compatibility to align with Expo. 

The modules in question are more recent than what Expo's compatibility lists. Some of these were cases of them fixing issues introduced with the Expo SDK version but Expo not updating its compat list to include. Some of the others were  judgment calls from myself based on the usefulness of fixes they introduced without causing issues with Expo (some context in PR description in #902).

> And to keep track of the mapbox bug. Not sure where that would go?

If only `package.json` supported actual comments...Open to having a living document (whether an issue, wiki doc, git-managed markdown file, etc) that elaborates on some of these deps issues. Otherwise, not too sure about where such things should live.

cimigree:
@achou11 
> > It took me a while to get this running for some unknown reason? But once I did everything seems to work fine.
> 
> Can you share any details on what you mean here? e.g. installation of deps, specific build steps, running the app, etc
> 

I did my typical steps to get an emulator running when updating dependencies:
rm -rf node_modules && npm install
rm -rf ~/.gradle/caches/
npx expo prebuild --clean -p android
cd android && rm -rf app/build && ./gradlew --stop && ./gradlew clean && cd ../
npm run extract-messages
npx expo prebuild --clean -p android
npm start --clear

But at that point, with the npm start, I got an error. I am so sorry but I did not make a note of what the error was. I figured it was some kind of caching clash. I uninstalled and reinstalled node modules again. clean build again. Still got an error with npm start.

Then I tried this: cd src/backend && npm install && cd ../, which didn't work. Again, so sorry but I did not make note of the error.

I can't remember but I think I might have done the whole thing again with all of the steps. Eventually it did start and ran fine, as I said.


> > We should probably make a note somewhere? To update the expo related modules that are out of compatibility to align with Expo.
> 
> The modules in question are more recent than what Expo's compatibility lists. Some of these were cases of them fixing issues introduced with the Expo SDK version but Expo not updating its compat list to include. Some of the others were judgment calls from myself based on the usefulness of fixes they introduced without causing issues with Expo (some context in PR description in #902).
> 

Ah ok. I was misunderstanding. Sounds good.

> > And to keep track of the mapbox bug. Not sure where that would go?
> 
> If only `package.json` supported actual comments...Open to having a living document (whether an issue, wiki doc, git-managed markdown file, etc) that elaborates on some of these deps issues. Otherwise, not too sure about where such things should live.

I think an issue might be a good place? But not positive that that is best.

achou11:
@cimigree thanks for the additional details! Since you were able to eventually get it working, let's keep an eye on it for now. Can always revisit if similar issues come up again (from you or anyone else).

I'll create an issue to track the maps dep issue

PR #1034, opened 2025-03-12
refactor: update error bottom sheet

ErikSin:
closes #1002 

image

PR #1038, opened 2025-03-14
Chore: use core react

### LINKED ISSUE #1000: Update all CoMapeo core server hooks with comapeo-core-react hooks
Api docs: https://github.com/digidem/comapeo-core-react/blob/main/docs/API.md

All `read` operations now use ReactSuspense, so loading states can be removed

### Files to be replaced:
- [x] deviceInfo.ts
- [x] fields.ts
- [x] icons.ts
- [x] maps.ts
- [x] media.ts
- [x] observations.ts
- [x] presets.ts
- [x] projects.ts
- [x] remoteDetectionAlerts.ts
- [x] track.ts

cimigree:
closes #1000 

### Summary
This PR removes most usages of `tanstack/react-query` and direct `clientApi` calls in favor of using the hooks provided by [`comapeo/core-react`]. By doing this, our server/hooks code is largely simplified or removed, and many screens now call `comapeo/core-react` hooks directly. When a hook was only needed in one or two places, I used the core-react hook directly in the screen; if it was used by multiple files or required extra logic, I wrapped it in a small hook.

### Notable Changes
- **ActiveProjectContext** now suspends on project loading. In the future, we may remove the `projectApi` from the context entirely, as recommended by [[@achou11](https://github.com/digidem/comapeo-core-react)](https://github.com/digidem/comapeo-core-react) so we only store the `projectId` and then call hooks like `useSingleProject({ projectId })` where needed.
- A few legacy hooks remain:
  - **`useOriginalVersionIdToDeviceId`**: There’s no direct core-react equivalent, so it’s left as is.
  - **`mediaSyncSettings`**: We still rely on a bit of React Query state to track optimistic updates to the user’s sync preference.
  - **Some map-related hooks**: I couldn’t find replacements for certain map-specific or custom-file logic, so they are still referencing local React Query or direct `clientApi` calls.

### Future Work
- Consider dropping `projectApi` usage in `ActiveProjectContext`. Instead, rely on `useSingleProject({ projectId })` throughout the codebase, removing the need for a suspense boundary in the top-level provider.
- Decide how to handle the legacy hooks such as `useOriginalVersionIdToDeviceId` or custom map logic once `comapeo/core-react` expands or changes its API.
- See below about AppNavigator

### Testing & Verification
- Existing screens and logic reliant on these hooks that aren't covered by E2E have been checked manually (media manager, invite flow, tracks, editing observations, etc., importing config) except adding a remote archive (not sure how to do that) and remote detection alerts (not sure how to do that either).
- I _think_ I got everything, but could use extra eyes on that stuff.
- There is a suspense boundary in place for the query that was outside of our suspense wrapper.

**Note on `useOwnDeviceInfo` in `AppNavigator`:**  
I initially replaced the `tanstack/react-query` call with `useOwnDeviceInfo`, but perhaps that caused unexpected refetch behaviors, which broke certain E2E tests and behavior (for example, could not navigate to Device Name without immediately navigating away). As a result, I reverted to the original `useQuery` approach in `AppNavigator`. I can revisit this if you have any ideas how @achou11


achou11:
> In the future, we may remove the projectApi from the context entirely ... so we only store the projectId and then call hooks like useSingleProject({ projectId }) where needed.

Yeah I would replace "may" with "should", assuming it doesn't result in any noticeable issues. There's really no reason to hold a project api instance in context given how react-query works.

> useOriginalVersionIdToDeviceId: There’s no direct core-react equivalent, so it’s left as is.

The equivalent hook is [`useDocumentCreatedBy`](https://github.com/digidem/comapeo-core-react/blob/main/docs/API.md#usedocumentcreatedby)

src/frontend/AppNavigator.tsx
```diff
@@ -8,18 +8,18 @@ import {ProjectInviteBottomSheet} from './sharedComponents/ProjectInviteBottomSh
...
 
 export const rootNavigationRef =
   createNavigationContainerRef<AppStackParamsList>();
 
 export const AppNavigator = ({permissionAsked}: {permissionAsked: boolean}) => {
-  const mapeoApi = useClientApi();
   const [inviteSheetEnabled, setInviteSheetEnabled] = React.useState(() => {
     return false;
   });
+  const DEVICE_INFO_KEY = 'deviceInfo';
```
achou11: using this will lead to issues given when it comes to cache invalidation because the hooks from core-react have a specific key structure that they adhere to. In this case, if you really need to keep this usage, it's preferable to use the same query key as the corresponding hook from core-react:

https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/client.ts#L20-L22

where [`ROOT_QUERY_KEY`](https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/shared.ts#L6): is `'@comapeo/core-react'`

src/frontend/hooks/server/invites.ts
achou11, line 1: `INVITE_KEY` needs to be updated to be aligned with what's used in core-react so that it gets invalidated properly when using the write hooks:

https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/invites.ts#L23-L25

You can either copy that directly or define one that's similar. All of the write hooks currently invalidate this key, which cascades to the other key I linked to:

https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/invites.ts#L19-L21

src/frontend/hooks/server/mediaSync.ts
```diff
@@ -18,31 +12,24 @@ export function isArchiveDevice(value: MediaSyncSetting): boolean {
...
+
+  function mutate(newSetting: MediaSyncSetting) {
+    setIsArchive.mutate({isArchiveDevice: isArchiveDevice(newSetting)});
+  }
+
+  return {
+    mutate,
+    status: setIsArchive.status,
+    reset: setIsArchive.reset,
+  };
```
achou11: I think the changes to this function will cause issues with [this](https://github.com/digidem/comapeo-mobile/blob/chore/use-core-react/src/frontend/screens/Settings/ProjectSettings/MediaSyncSettings/index.tsx#L49-L53).

One option is to reconsider our usage of `useMutationState()` and whether it's really needed there. If it's not, then there may not be a need to have this custom mutation at all.

However, if we're interested in preserving compatibility, the other option is to update this function's implementation:

```ts
export function useSetMediaSyncSetting() {
  const setIsArchive = useSetIsArchiveDevice();

  const {mutate, status, reset} = useMutation({
    mutationKey: [UPDATE_MEDIA_SETTING],
    mutationFn: (newSetting: MediaSyncSetting) => {
      return new Promise((res, rej) => {
        setIsArchive.mutate(
          {isArchiveDevice: isArchiveDevice(newSetting)},
          {
            onError: rej,
            onSuccess: res,
          },
        );
      });
    },
  });

  return {mutate, status, reset};
}
```

(we don't re-export the `mutateAsync` method in core-react, hence the need for the `new Promise(...)` approach).

src/frontend/contexts/ActiveProjectContext.tsx
```diff
@@ -1,16 +1,17 @@
...
-import {useProject, useCreateProject} from '../hooks/server/projects';
+import {useSingleProject} from '@comapeo/core-react';
 import {Loading} from '../sharedComponents/Loading';
 
 const ActiveProjectContext = React.createContext<
   {projectId: string; projectApi: MapeoProjectApi} | undefined
 >(undefined);
 
-export const ActiveProjectProvider = ({
+export const ActiveProjectIdProvider = ({
```
achou11: highly subjective, but I'd suggest keeping the name of this to `ActiveProjectProvider` and updating the internal wrapper to something like `ProjectApiLoader` (or something to make it clear that it's internal).

Then again, this whole file is probably going to be removed as a follow-up, in favor of using the `useSingleProject()` directly

src/frontend/hooks/server/icons.ts
```diff
@@ -1,24 +1,20 @@
...
-    queryFn: async () => {
-      return projectApi.$icons.getIconUrl(iconId, {
-        mimeType: 'image/png',
-        size,
-        pixelDensity: 3,
-      });
-    },
-    // Defaults to 3 if not specified but due to an edge case when pausing and resuming the app
-    // (see https://github.com/digidem/comapeo-core/issues/821#issuecomment-2344231495),
-    // we reduce the retry count for the benefit of more immediate UI feedback in that scenario
```
achou11, line 21: is this issue still relevant? core-react completely disables retries and provides no way of specifying otherwise.

if we need be able to configure retries, then there needs to be some work to preserve this retry behavior. Ideally we don't need to surface some `retry` option in core-react (especially for one-offs), but that may be the solution that requires the least work...

Not sure how one could implement a wrapper hook uses the existing core-react hook internally 🤔 The idiomatic React Query approach is to wrap the _options_ that you pass, but we don't support that for the sake of being opinionated.

src/frontend/hooks/server/maps.ts
```diff
@@ -27,23 +27,33 @@ export type CustomMapInfo = v.InferOutput<typeof CustomMapInfoSchema>;
...
-      }
-      const result = await api.getMapStyleJsonUrl();
-      return result + `?refresh_token=${refreshToken}`;
-    },
+  const {
+    data: baseUrl,
+    error,
+    isRefetching,
+  } = useMapStyleUrl({
+    refreshToken: refreshToken ? refreshToken.toString() : undefined,
```
achou11: Is there a reason for conditionally setting this?

```suggestion
    refreshToken: refreshToken.toString(),
```

src/frontend/hooks/server/media.ts
```diff
@@ -1,162 +1,159 @@
...
-      switch (attachment.type) {
-        case 'UNRECOGNIZED': {
-          throw new Error('Cannot get URL for unrecognized attachment type');
-        }
-        case 'video':
-        case 'audio': {
-          if (variant !== 'original') {
-            throw new Error('Cannot get URL of attachment for variant');
-          }
+export function useCreateBlobMutation() {
```
achou11: this should use React Query's idiomatic approach for wrapping mutations:

```ts
// 🟡 Would reconsider naming this to something like `useCreateAttachment()` since it resolves with an attachment
export function useCreateBlobMutation() {
  const createBlobMutation = useCreateBlob({projectId});

  const wrappedMutation = useMutation({
    mutationFn: (
      file: ProcessedDraftPhoto | UnsavedAudio,
    ): Promise<Attachment> => {
      return new Promise((res, rej) => {
        createBlobMutation.mutate(createBlobArgs(file), {
          onError: rej,
          onSuccess: data => {
            res({
              driveDiscoveryId: data.driveId,
              name: data.name,
              type: data.type,
              hash: data.hash,
            });
          },
        });
      });
    },
  });

  // 🟡 Returning the whole thing is probably more useful in this case
  // but can be more selective if preferred
  return wrappedMutation;
}
```

src/frontend/hooks/server/observations.ts
achou11, line 1: going to run to into issues related to cache invalidation if `OBSERVATION_KEY` is left as is. Needs to be updated be similar to what core-react defines:

https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/documents.ts#L26

(where `docType` is `'observation'`)

Looking at the usages of that variable:

1. `CreateTestData.tsx`: used for invalidating the project's observations after creating test observations. This file should instead be updated to use the relevant core-react hooks, which will properly handle the invalidation internally

2. `ProjectSyncDisplay`: used for invalidating the project's observations when leaving the screen. there's a todo comment that mentions invalidating all project-related queries instead of just its observations.

Given the above two, I think this variable isn't necessary at all. Would suggest making the relevant changes in this PR to get rid of this variable completely.

src/frontend/hooks/server/presets.ts
achou11, line 1: `PRESETS_KEY` variable can be removed (no longer used anywhere)

src/frontend/hooks/server/presets.ts
```diff
@@ -1,37 +1,36 @@
...
-      return await projectApi.preset.getMany({lang: locale});
-    },
+  return useManyDocs({
+    projectId,
+    docType: 'preset',
+    lang: locale,
   });
 }
 
 export function usePresetsMutation() {
```
achou11: this function isn't being used anywhere. can remove if that's the case

src/frontend/hooks/server/projects.ts
```diff
@@ -16,79 +19,9 @@ export const ORIGINAL_VERSION_ID_TO_DEVICE_ID_KEY =
...
-    queryKey: [PROJECT_SETTINGS_KEY, projectId],
-    queryFn: () => {
-      return projectApi.$getProjectSettings();
-    },
-  });
+  const {projectId} = useActiveProject();
+  return useComapeoProjectSettings({projectId});
 }
 
 export const useOriginalVersionIdToDeviceId = (originalVersionId: string) => {
```
achou11: Should be able to replace the `useQuery` in here with the relevant core-react hook:

https://github.com/digidem/comapeo-core-react/blob/main/docs/API.md#usedocumentcreatedby

src/frontend/hooks/server/projects.ts
```diff
@@ -106,87 +39,30 @@ export const useOriginalVersionIdToDeviceId = (originalVersionId: string) => {
...
-        }),
-        queryClient.invalidateQueries({
-          queryKey: [PRESETS_KEY],
-        }),
-      ]);
-    },
-  });
-}
-
 export function useGetOwnRole() {
```
achou11, line 26: ah looks like I missed a hook to implement in core-react. i have the [foundations](https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/projects.ts#L183) for it defined but guess i never actually implemented the hook 😅 will address and create a new release promptly

src/frontend/hooks/server/projects.ts
```diff
@@ -106,87 +39,30 @@ export const useOriginalVersionIdToDeviceId = (originalVersionId: string) => {
...
+  const {data: members, error, isRefetching} = useManyMembers({projectId});
+
+  const archives = useMemo(() => {
+    return members?.filter(m => m.deviceType === 'selfHostedServer') ?? [];
+  }, [members]);
+
+  return {
+    data: archives,
+    error,
+    isPending: isRefetching,
```
achou11: would suggest using `isRefetching` as the name of the field. `isPending` is subtlety different than `isRefetching` in react query so could easily get a bit confusing if renamed.

```suggestion
    isRefetching,
```

src/frontend/hooks/server/track.ts
achou11, line 1: can remove `TRACK_KEY` variable since it's no longer used.

src/frontend/hooks/server/remoteDetectionAlert.ts
achou11, line 1: similar comment about this as the `OBSERVATION_KEY`. The only place this is used is in `ProjectSyncDisplay` in order to invalidate a project's tracks before leaving the screen.

Would recommend updating that file to invalidate the whole project (which cascades down to the relevant observations and tracks queries) and remove this variable entirely.

src/frontend/hooks/server/maps.ts
achou11, line 1: similar comment about how `MAPS_QUERY_KEY` needs to be update for proper cache invalidation.

https://github.com/digidem/comapeo-core-react/blob/9c70b2fcee1a93b85fcca5e3cefcd4bc7d9b1b06/src/lib/react-query/maps.ts#L7

achou11, line 1: `useCustomMapInfo()` can be updated to avoid needing to use the client api directly:

```ts
export function useGetCustomMapInfo() {
  const {data: styleUrl} = useCustomMapStyleUrl();

  return useQuery({
    queryKey: [MAPS_QUERY_KEY, 'custom', 'info', {styleUrl}],
    queryFn: async () => {
      const response = await fetchCustomMapInfo(styleUrl);

      if (response.status === 404) {
        return null;
      }

      if (!response.ok) {
        throw new Error(`Cannot get custom map info: ${response.statusText}`);
      }

      return v.parse(CustomMapInfoSchema, await response.json());
    },
  });
}
```

achou11 (changes requested):
Overall a good start! This is an initial review that mostly focuses on the bits where core-react hooks are being integrated. Probably missed a few things related to this but hopefully will catch on subsequent reviews. Didn't get a chance to try out the changes in the app or how UI has been updated yet.

There are a couple of small tasks that I need to do for core-react in order for some of the suggestions to be applicable. Will try to get those promptly!

Couple of additional notes that I couldn't leave as comments:

1. Remove `useDeviceInfo.ts`. Don't think it's needed anymore.

2. Regarding the AppNavigator issue: I already mentioned a potential solution on Slack but basically need to move the device info query into the `RootStackNavigator` component and then add a suspense boundary around it in the `AppNavigator`.

src/frontend/hooks/server/projects.ts
achou11: exposed as `useOwnRoleInProject` in a new release of core react!

https://github.com/digidem/comapeo-core-react/releases/tag/v3.1.0

src/frontend/hooks/server/mediaSync.ts
cimigree: I think it is probably best (for now?) to keep the optimistic update and I changed to your code above.

src/frontend/contexts/ActiveProjectContext.tsx
cimigree: Ok I was doing that because I thought that was what you wanted. I misunderstood.

src/frontend/hooks/server/icons.ts
cimigree: Ok. Removed.

src/frontend/hooks/server/presets.ts
cimigree: Hmm. I think it's used in 7 different places?

achou11: doing a codebase search for `usePresetsMutation` comes up empty for me 🤔

src/frontend/screens/HowToLeaveProject.tsx
```diff
@@ -71,20 +72,26 @@ export const HowToLeaveProject = ({
...
   );
 };
 
+export const HowToLeaveProject = (
+  props: NativeStackScreenProps<AppStackParamsList, 'HowToLeaveProject'>,
+) => (
+  <React.Suspense fallback={<Loading />}>
+    <HowToLeaveProjectContent {...props} />
+  </React.Suspense>
+);
```
achou11: Is the suspense wrapper needed? my understanding is that it's not necessary because this screen is used in the `RootStackNavigator` and will be automatically wrapped due to this:

https://github.com/digidem/comapeo-mobile/blob/43e7dff691f6ebf018c0b278d2154a2c6bf625b5/src/frontend/Navigation/Stack/index.tsx#L34

src/frontend/hooks/server/projects.ts
cimigree: I accidentally already did that and forgot to delete useOriginalVersionIdToDeviceId
Or did you mean useFindRemoteArchive? I have a feeling not, but just asking...

src/frontend/hooks/server/maps.ts
```diff
@@ -27,23 +27,33 @@ export type CustomMapInfo = v.InferOutput<typeof CustomMapInfoSchema>;
 
 const {useRefreshToken, useRefreshTokenActions} = createRefreshTokenStore();
 
-export function useMapStyleJsonUrl() {
-  const api = useClientApi();
+export function useCustomMapStyleUrl() {
```
achou11: I think the old name is more accurate in terms of what this does (the url doesn't always point to a custom map style)

```suggestion
export function useMapStyleJsonUrl() {
```

src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -108,8 +111,14 @@ export const ButtonFields = ({
...
-          deleteObservationMutation.mutate({id: observationId});
-          navigation.pop();
+          deleteObservationMutate(
+            {docId: observationId},
+            {
+              onSuccess: () => {
+                navigation.pop();
+              },
+            },
+          );
```
achou11: probably out of scope but noticing that we don't have any error handling for this component. probably would make sense to at least capture that in sentry and/or add the error bottom sheet UI

src/frontend/hooks/server/presets.ts
cimigree: Ah. I thought you were referring to the function that is highlighted. My bad. Yes. You are right. I deleted it.

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -88,13 +88,27 @@ export const ObservationCreate = ({
...
   const {formatMessage} = useIntl();
+  const {projectId} = useActiveProject();
   const {usePreset} = useDraftObservation();
   const preset = usePreset();
   const value = usePersistedDraftObservation(store => store.value);
   const attachments = usePersistedDraftObservation(store => store.attachments);
   const {updateTags, clearDraft} = useDraftObservation();
-  const createObservationMutation = useCreateObservation();
-  const createBlobMutation = useCreateBlobMutation();
+  const [error, setError] = React.useState<Error | null>(null);
```
achou11: is this because the core react hooks are not returning react query's `error` field? doesn't seem ideal to do this as now there's an error state that needs to be in sync with the internal error state of the react-query-based hook(s).

Maybe it'd be worth updating core-react to expose that error field? I'd prefer not to if we can avoid it, but I can also see it being generally useful. 

Wouldn't be necessary to if our new UI approach to handling mutation errors were available (blocked by https://github.com/digidem/comapeo-mobile/pull/1034)

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -178,54 +194,44 @@ export const ObservationCreate = ({
...
-          type,
-          name,
-          hash,
-        }),
-      );
+    const newAttachments = [];
+    for (const file of [...unsavedPhotos, ...unsavedAudioRecordings]) {
+      const result = await createBlobAsync(file);
+      newAttachments.push(result);
+    }
```
achou11: preexisting behavior of this was creating the blobs in parallel. this changes it to be done serially. may not have a noticeable impact but wondering if this is intentional or not.

src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -69,6 +73,7 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> = ({
         defaultMessage: preset.name,
       })
     : formatMessage(m.observation);
+  const [error, setError] = React.useState<Error | null>(null);
```
achou11: just raising another case where maybe exposing an `error` field could be useful in core-react's write hooks 🤔

achou11 (changes requested):
Another run through that takes a closer look at changes in the UI parts

src/frontend/hooks/server/projects.ts
achou11: nope, was referring to `useOriginalVersionIdToDeviceId`!

src/frontend/screens/ObservationCreate/index.tsx
cimigree: Yes that is exactly why.

src/frontend/screens/ObservationCreate/index.tsx
cimigree: I was not able to get it to work doing it in parallel. If I took more than one photo, only the second photo would save.

achou11: ah interesting - wonder if that's a bug with the preexisting implementation or something in core (guessing the former)

src/frontend/screens/ObservationCreate/index.tsx
achou11: created a new release of core-react to expose the `error` field!

https://github.com/digidem/comapeo-core-react/releases/tag/v3.2.0

src/frontend/screens/HowToLeaveProject.tsx
cimigree: I have no idea why I did that... Fixed.

src/frontend/screens/ObservationCreate/index.tsx
cimigree: OK.

src/frontend/screens/Observation/Buttons.tsx
cimigree: Ok I did Sentry...

cimigree:
@achou11 Ok I spent the day trying to implement these comments. I am not sure if I did everything the way you intended. 
I don't think I can remove useDeviceInfo because it is used in About.tsx.
I moved the device query into RootStackNavigator as you requested.
I tried to fix the query keys but I am not sure if that is what you intended with them. 
Is there anything additional that needs to be done with attachments and the serial saving?
This might could use some additional QA type testing. I did some (again).

src/frontend/hooks/server/maps.ts
achou11, line 1: Maybe move `ROOT_QUERY_KEY` to `constants.ts` so that all of the files that are needing this are sharing the same value, to reduce room for error?

src/frontend/hooks/server/mediaSync.ts
```diff
@@ -18,31 +13,31 @@ export function isArchiveDevice(value: MediaSyncSetting): boolean {
...
-  return useSuspenseQuery({
-    queryKey: [MEDIA_SYNC_SETTING_KEY],
-    queryFn: async () => {
-      const isArchive = await api.getIsArchiveDevice();
-      return convertMediaSyncSetting(isArchive);
-    },
-  });
+  return {
+    data: setting,
+  };
```
achou11: maybe more stylistic but can reduce this a bit if this is the only value that's needed

```suggestion
  return setting;
```

src/frontend/hooks/server/media.ts
achou11, line 1: the changes to `useAttachmentUrlQuery`  causes it to have subtlety different behaviors from before. the main difference is that errors related to unsupported attachment types are thrown outside of the `queryFn` option, which could lead to issues when it comes to error handling when using the queries.

I think cleaning this up is out of scope for this PR but worth doing as a follow-up since it can simplify the usage quite a bit. The basic recommendation here would be to use the hooks from core-react more directly where they're needed as opposed to having this wrapper hook that attempts to account for various kinds of attachments.

src/frontend/screens/ObservationCreate/index.tsx
achou11: i took a look at this and am kind of perplexed. wasn't able to get it to work either but not really sure where the issue is 🤔 the implementation on develop seems to work fine so i guess something that's been introduced in this PR i guess? not a blocker but definitely has me wondering about potential issues coming up later

separately and prior to changes you've introduced, this whole implementation should probably be cleaned up to make use of nested mutations better. out of scope but noticing the lots of unnecessary complexity due to it 😅

src/frontend/hooks/useProjectInvitesListener.ts
```diff
@@ -1,8 +1,8 @@
 import {useCallback, useEffect, useState} from 'react';
 import {useQueryClient} from '@tanstack/react-query';
-import {INVITE_KEY} from './server/invites';
 import {Invite, InviteRemovalReason} from '@comapeo/core/dist/invite-api';
 import {useClientApi} from '@comapeo/core-react';
+const ROOT_QUERY_KEY = '@comapeo/core-react';
```
achou11: similar comment about using this from a shared file

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -88,13 +88,27 @@ export const ObservationCreate = ({
...
+    mutate: createObservationMutation,
+    reset: resetObservation,
+    status: observationStatus,
+    error: observationError,
+  } = useCreateDocument({
+    docType: 'observation',
+    projectId,
+  });
+  const {
+    mutateAsync: createBlobAsync,
```
achou11: update name to be more accurate:

```suggestion
    mutateAsync: createAttachmentAsync,
```

src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -52,15 +52,19 @@ export const ObservationEdit: NativeNavigationComponent<'ObservationEdit'> = ({
...
     useDraftObservation();
   const preset = usePreset();
-  const editObservationMutation = useEditObservation();
+  const {mutate, status, reset, error} = useUpdateDocument({
+    docType: 'observation',
+    projectId,
+  });
   const attachments = usePersistedDraftObservation(store => store.attachments);
-  const createBlobMutation = useCreateBlobMutation();
+  const {mutateAsync: createBlobAsync, reset: resetAttachment} =
```
achou11: update name to be more accurate:

```suggestion
  const {mutateAsync: createAttachmentAsync, reset: resetAttachment} =
```

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -117,6 +115,7 @@ export const ProjectSyncDisplay = ({
   const navigation = useNavigationFromRoot();
   const {projectApi, projectId} = useActiveProject();
   const progress = useDataSyncProgress();
+  const ROOT_QUERY_KEY = '@comapeo/core-react';
```
achou11: similar comment about moving this to a shared file

src/frontend/hooks/server/invites.ts
```diff
@@ -1,88 +1,17 @@
...
+const ROOT_QUERY_KEY = '@comapeo/core-react';
 
-import {useActiveProject} from '../../contexts/ActiveProjectContext';
-import {useActiveProjectIdActions} from '../../contexts/ActiveProjectIdStoreContext';
-import {ALL_PROJECTS_KEY, PROJECT_MEMBERS_KEY} from './projects';
-
-export const INVITE_KEY = 'pending_invites';
+function getPendingInvitesQueryKey() {
+  return [ROOT_QUERY_KEY, 'invites', {status: 'pending'}] as const;
+}
```
achou11, line 10: would add a comment that mentions that this is copied from [here](https://github.com/digidem/comapeo-core-react/blob/v3.2.0/src/lib/react-query/invites.ts#L23-L25)

src/frontend/hooks/server/invites.ts
```diff
@@ -1,88 +1,17 @@
 import {useClientApi} from '@comapeo/core-react';
-import {
-  useMutation,
-  useQueryClient,
-  useSuspenseQuery,
-} from '@tanstack/react-query';
+import {useSuspenseQuery} from '@tanstack/react-query';
+const ROOT_QUERY_KEY = '@comapeo/core-react';
```
achou11: similar comment about moving this to a shared file

src/frontend/screens/ObservationFields/index.tsx
```diff
@@ -37,10 +38,17 @@ export const ObservationFields = ({
...
   route,
 }: NativeRootNavigationProps<'ObservationFields'>) => {
+  const {projectId} = useActiveProject();
+  const languageTag = useAppLanguageTag();
+
+  const {data: fields} = useManyDocs({
+    projectId,
+    docType: 'field',
+    lang: languageTag,
+  });
```
achou11, line 48: great example of how to use these hooks when there's no custom/shared logic that's needed across the app. there are definitely more places in the app where we can get rid of wrapper hooks in favor of something like this, but probably best to do as a follow-up.

src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -62,11 +66,21 @@ export const Success = ({
...
         onPress={() => {
-          setDeviceName.mutate(deviceName);
+          mutate(
+            {
+              name: deviceName,
+              deviceType: expoToCoreDeviceType(deviceType),
+            },
+            {
+              onSuccess: () => {
+                queryClient.invalidateQueries({queryKey: ['deviceInfo']});
```
achou11: i don't think this is needed? the mutation here is updating the device info related to core. it's kind of confusing because there's this [`useDeviceInfo()`](https://github.com/digidem/comapeo-mobile/blob/3a63a10fc3006309822e079c834249264b63ce44/src/frontend/hooks/useDeviceInfo.ts#L28) hook that's related to the react device info module, which in unrelated to core. that uses the query key that starts with `'deviceInfo'`.

doesn't seem like updating those queries as a result of this mutation is necessary, so i'd suggest removing this `onSuccess` handler entirely if that's the case (since the mutation invalidates the relevant caches internally already)

achou11 (changes requested):
after you address this round of feedback, I feel pretty good about the code you've implemented! think there are lots of opportunities for cleanup and removing lots of "wrapper" hooks but think those should be done as follow-ups.

> I tried to fix the query keys but I am not sure if that is what you intended with them.

yeah roughly! made some suggestions to avoid issues around inconsistency and to document the contexts

> Is there anything additional that needs to be done with attachments and the serial saving?

i don't think so for now. as mentioned in one of the comments, I also ran into issues attempting to use the parallel approach but can't fully figure out why. given that the functionality seems to work fine, I'm okay with it being looked into separately (that whole component could probably use a decent amount of cleanup anyways).

src/frontend/hooks/server/mediaSync.ts
cimigree: ok done. Hope it's ok that I reduced it even more.

src/frontend/hooks/server/media.ts
cimigree: Ok. What would you like me to do at this point, though?

achou11: Sorry should've been clearer: i don't think there's a need to change what's here since I understand why the changes were done. Mostly commented to identify the change in case we needed to refer back to it, since i think it could be a candidate for follow-up cleanup work.

src/frontend/screens/Onboarding/Success.tsx
cimigree: Yes I think you are right. Done.

achou11:
@cimigree just released `@comapeo/core-react@3.3.0`, which includes this change:

https://github.com/digidem/comapeo-core-react/pull/66

The PR description has more context but basically we're now exposing React Query's `mutateAsync` method for the write hooks. This means:

- no more need to do the tedious Promise wrapping of `mutate`. Should be able to just use `mutateAsync` directly.
- the places that used mutations in parallel but were changed in this PR can be restored (e.g. the observation attachments code). the two places where parallel mutations were being used are:
  1. Creating observation attachments. this was changed to be done in serial in this PR since it wasn't working with core-react.
  2. Creating observations in the test data screen. this was not changed to be serial but i found that this feature was not working as intended when trying out this PR (detailed a bit in the core-react PR).
  
Apologies for the extra work but I'm grateful for the issues + limitations that have come up in this PR!

cimigree:
@achou11 Ok I attempted to make the changes you requested.
Was able to save multiple photos and audios when saving an observation and editing an observation.
Creating test data also went smoothly.

src/frontend/hooks/server/mediaSync.ts
```diff
@@ -18,31 +13,27 @@ export function isArchiveDevice(value: MediaSyncSetting): boolean {
...
-        queryKey: [MEDIA_SYNC_SETTING_KEY],
+    mutationFn: (newSetting: MediaSyncSetting) => {
+      return new Promise((res, rej) => {
+        setIsArchive.mutate(
+          {isArchiveDevice: isArchiveDevice(newSetting)},
+          {
+            onError: rej,
+            onSuccess: res,
+          },
+        );
```
achou11: no longer need the promise wrapping and can use `mutateAsync` instead

src/frontend/constants.ts
```diff
@@ -16,3 +16,5 @@ export const EDITING_SCREEN_NAMES:
   'TrackEdit',
   'Camera',
 ];
+
+export const ROOT_QUERY_KEY = '@comapeo/core-react';
```
achou11, line 23: maybe add a comment here about this coming directly from core-react? 

https://github.com/digidem/comapeo-core-react/blob/59a80cf0a1b9dad13e5f066233dae2465d2d20b1/src/lib/react-query/shared.ts#L6

src/frontend/hooks/server/maps.ts
```diff
@@ -7,8 +7,11 @@ import {DOCUMENT_DIRECTORY} from '../../lib/file-system';
 
 import {createRefreshTokenStore} from '../refreshTokenStore';
 import noop from '../../lib/noop';
+import {ROOT_QUERY_KEY} from '../../constants';
 
-export const MAPS_QUERY_KEY = 'maps';
+export function getMapsQueryKey() {
+  return [ROOT_QUERY_KEY, 'maps'] as const;
+}
```
achou11: also could add a comment here about where in core-react this comes from

also no need to export this:

```suggestion
function getMapsQueryKey() {
  return [ROOT_QUERY_KEY, 'maps'] as const;
}
```

src/frontend/hooks/server/media.ts
```diff
@@ -1,162 +1,134 @@
...
-        enabledByDefault,
-      ),
-    ),
+  const {data: rawUrl, error} = useAttachmentUrl({
+    projectId,
+    blobId,
   });
+
+  return {
+    data: rawUrl ? {...validAttachment, url: rawUrl} : undefined,
```
achou11: are the other fields from `validAttachment` needed as opposed to just exposing `data: rawUrl`? Based on where this hook is used, the url seems to be the only thing that's ever used. Can remove some unnecessary code if that's the case.

src/frontend/hooks/server/mediaSync.ts
cimigree: Ok done: [a3d4b7f](https://github.com/digidem/comapeo-mobile/pull/1038/commits/a3d4b7f4f9ff412bdf1a31df23c139faca9e455f)

src/frontend/hooks/server/media.ts
```diff
@@ -1,162 +1,134 @@
...
+    error,
+    isPending: !rawUrl && !error,
```
achou11: not sure if this works as you're expecting it to as this calculation will always be `false`. this hook is essentially a wrapper around a suspense query so the pending state is handled by the consumer of this hook and reflected in the nearest suspense boundary. basically, reaching the contents in this hook indicates that we're not in the pending state but rather a success or failure state. the `isRefetching` field from the read hook in core-react can be used to determine subsequent "pending" states, but this refers to states after the initial pending state.

the proper fix is to update the consumers of this hook such that they define a suspense boundary where needed. basically instead of this:

```tsx
function Component() {
  const {isPending} = useAttachmentUrlQuery()
  
  return isPending ? <SomeLoadingIndicator /> : <ActualRenderedThing />
}
```

you'd essentially do this, where the hook would be used within `ActualRenderedThing`:

```tsx
function Component() {
  return (
    <React.Suspense fallback={<SomeLoadingIndicator/>}>
      <ActualRenderedThing />
    </React.Suspense>
  )
}  
```

As highlighted as one of the tradeoffs of the suspense-based approach, this sometimes requires creating "internal" components or moving suspense-based hooks in order to establish the suspense boundary correctly. react query has a feature that makes this less common but it's an opt-in experimental feature so not really something we can consider just yet.

I have some local changes that update `PhotoAttachmentView` and `PhotoUnpreparedView` to align with this approach. Can share the patch if it's helpful to get a fuller picture of the nature of the changes. I haven't updated `PhotoThumbnail` because it's a more complex case to fix and didn't want to spend too much time on it.

achou11: in terms of suggestions, I would say either:

1. make the changes related to my explanation above

2. update the return value of this hook to basically `return useAttachmentUrl({ projectId, blobId })` and remove all of the `isPending` code from consumers of this hook

(1) probably requires a non-trivial number of changes that can easily become a rabbit-hole.

(2) means that relative to the behavior on the develop branch, the more granular loading states will be lost. however, relative to the current implementation in this PR, it removes unnecessary code that is never reached.

Given this, I would suggest doing (2) as part of this PR. I'd generally consider this non-blocking unless the difference between what's on develop and here is notably undesired (which I'm not sure it is).

(1) feels more appropriate as a follow-up.

achou11 (approved):
Couple of non-blocking suggestions but overall I think this is ready to be merged when you feel confident about it. I've done basic tests manually on my device and haven't noticed any significant issues yet.

I think there are a TON of follow-ups that would be worth doing, and ideally before the next release. I can take the time to try to identify them and create issues if possible, but feel free to open your own if there's anything you've noticed as you've worked on this!

ErikSin:
wooohooo! thanks for updating this

Issue #901, opened 2025-01-13
Implement lockfile checks for backend


achou11:
Introduced lockfile checks for version and dep changes at the app level via https://github.com/digidem/comapeo-mobile/pull/892. Ideally we have something similar for the backend.

- for lockfile version, not sure if there's a straightforward way of getting it to work given the tooling we're using. may need to punt on this (not really as useful or important in all fairness)

- for deps changes, basically can do what Gregor explained in https://github.com/digidem/comapeo-mobile/pull/892#issuecomment-2554023492, but needs https://github.com/digidem/npm-lockfile-changes/pull/2 to be merged first I think?

achou11:
Closed via https://github.com/digidem/comapeo-mobile/pull/1040

PR #1040, opened 2025-03-14
chore: check backend lockfile changes in CI

gmaclennan:
I noticed that our lockfile checks were not checking the backend lockfile. Also required forking the lockfile-version check script to add an option to specify path.


achou11 (approved):
Awesome! Very useful to have this

Just making a note that this should close #901

PR #1042, opened 2025-03-15
Proposal: Update Location to utilize zustand and context

ErikSin:
# Proposal PR: Centralizing Location Management in Zustand

This is a proposal PR. The implementation details are less important at this stage—I am primarily looking for feedback on the architecture.

## Main Proposal

Move `location`, `locationProviderStatus`, and `locationPermissions` into a Zustand store and [propagate those values via a context](https://tkdodo.eu/blog/zustand-and-react-context)](https://tkdodo.eu/blog/zustand-and-react-context).

### How `useLocation` Currently Works
- `useLocation` subscribes to location changes and updates React state.
- Each use of `useLocation` creates a new subscription.
- Example: The map screen and the GPS pill (in the header) both use `useLocation`, leading to two separate subscriptions when the map is open.

### Why We Took This Approach
- **High-frequency updates:** GPS data updates multiple times per second, even when stationary.
- **Battery performance concerns:** Frequent updates were causing excessive re-renders, particularly in the map screen.
- **Throttling strategy:**
  - The map screen receives a throttled location to reduce re-renders.
  - The GPS pill and observation location use the most precise location.
  - Even though these use the same subscription, we needed to throttle it for the map but not for the other components. We decided to use seperate subscriptions for each for that reason
- **Context concerns:**
  - Initially avoided using Context because it would require wrapping the entire navigator meaning the subscription would always be running even if subcription weren't being used

### `locationProviderStatus`
- Tracks whether the user has disabled location services.
- Initially considered Context but rejected due to persistent background subscriptions.

### `locationPermissions`
- Currently a mix of Expo’s `useLocationPermission` hook and Android’s permissions API.
- `useLocationPermission` is not fully reactive (doesn't detect permission revocations from system settings).

## What Has Changed

- **Subscriptions are always running anyway.**
  - Due to React Navigation’s behavior, the first screen in a stack remains mounted.
  - Since the first screen is the **map**, we are already running both the location and provider status subscriptions.
  - Additionally, the GPS pill has its own subscription, leading to duplication.
- **Proposal:** Centralize the subscriptions into a Zustand store.
  - Run them once instead of multiple times.
  - Use Zustand to ensure only consuming components re-render instead of the entire app.

## Proposed Solution

1. **Use Zustand for State Management:**
   - Centralize `location`, `locationProviderStatus`, and `locationPermissions`.
   - Prevent redundant subscriptions by managing state in a single place.
2. **Control Subscriptions with Context:**
   - Use a React context to start/stop subscriptions via lifecycle methods.
   - This avoids memory leaks, which could occur if subscriptions were directly managed in Zustand outside React’s lifecycle.
3. **Optimize Location Updates:**
   - Maintain two separate location values:
     - **Throttled location** for the map (reduces re-renders).
     - **Precise location** for GPS pill and observations.
   - Both values derive from the same subscription but update separate properties in the Zustand store.
4. **Unify Permission Handling:**
   - Subscribe to app state changes (already done in the map screen) to detect permission changes.
   - Store the permission status in Zustand for a simpler, centralized approach.

## Benefits

- **Single source of truth:** Eliminates redundant subscriptions.
- **Better performance:** Only relevant components update, preventing unnecessary re-renders.
- **Cleaner code:** Simplifies permission handling and avoids multiple subscription mechanisms.
- **Easier debugging & maintenance:** Centralized state management improves clarity.

@cimigree and @achou11 I will set up a meeting for us to talk about this further. But feel free to add comments here before hand

### DIFF
--- a/src/frontend/contexts/DraftObservationContext.tsx
+++ b/src/frontend/contexts/DraftObservationContext.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 
 import {StoreApi, useStore} from 'zustand';
 import CheapRuler from 'cheap-ruler';
-import * as Location from 'expo-location';
 import {
   DraftState,
   DraftObservationStore,
@@ -110,11 +109,7 @@ function useCreateDraftObservationLocationUpdator(
           if (!providerStatus?.locationServicesEnabled) return;
 
           // If manual location, do not update
-          if (
-            draftObservationStore.instance.getState().value?.metadata
-              ?.manualLocation
-          )
-            return;
+          if (currentDraft.metadata?.manualLocation) return;
 
           // if no initial position, set initial position
           if (!initialPosition) {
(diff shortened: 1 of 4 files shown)


src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -74,94 +76,123 @@ const MOVED_AWAY_THRESHOLD_METERS = 100;
...
-    if (isManualLocation) return;
-
-    const isStale =
-      Date.now() - location.timestamp > STALE_LOCATION_THRESHOLD_MS;
-    if (isStale) return;
-
-    instance.setState(prev => {
-      if (prev.initialPosition || !prev.value) return prev;
-      return {...prev, initialPosition: convertPosition(location)};
+function useCreateDraftObservationLocationUpdator(
```
ErikSin, line 78: Updating this function might be out of scope. But this architecture does make updating the draft observation location easier and more straight forward to manage now.

Most notably, we can subscribe to location updates from the context while reading the location value from the Zustand store instead of React state. This means that when the location changes we can update the draft observation store location. And since we are using zustand store state, we do not need to worry about it causing unnecessary re-renders.

PR #1046, opened 2025-03-17
fix: align tracks with current location marker

### LINKED ISSUE #938: Tracks / Current location not accurate or consistant while recording a track
https://www.notion.so/digidem/Tracks-Current-location-not-accurate-or-consistant-while-recording-a-track-14c1b08162d580488b02ede631348e0d

**Via Jen**

When a track is active the blue current location dot get’s stuck at startpoints (and sometimes at midpoints on the track line.

ErikSin:
closes #938 

The main issue was that the `currentTrackMapLayer` was recording a location more often than the user location was being updated. In order to align the two, setting there displacment to the same value guarantees that they update at the same rate.


src/frontend/screens/MapScreen/CurrentTrack/CurrrentTrackMapLayer.tsx
```diff
@@ -1,30 +1,18 @@
...
-  const finalLocationHistory = location?.coords
-    ? [
-        ...locationHistory,
-        {
-          latitude: location.coords.latitude,
-          longitude: location.coords.longitude,
-          timestamp: new Date().getTime(),
-        },
-      ]
-    : locationHistory;
```
ErikSin, line 21: because tracks was previously not updating often, I think this was done to keep the line up to date with the most current location. This is no longer a problem as the tracks are updating more often.

src/frontend/hooks/useTracking.ts
```diff
@@ -3,6 +3,11 @@ import {useCallback, useState} from 'react';
 import {usePersistedTrack} from './persistedState/usePersistedTrack';
 import {LOCATION_TASK_NAME} from '../sharedTypes/location.ts';
 
+/**
+ * This constant should be used for tracks and for the userlocation dot. In order for the user location dot to align with the track on the map, they need to update at the same frequency.
+ */
+export const TRACKING_DISTANCE_INTERVAL = 2;
```
achou11: feels like this should live somewhere more discoverable, especially since it's used in various places

achou11 (changes requested):
The PR title should be be marked as "fix" instead of "chore"

Unfortunately kind of hard to test this one 😅

I _think_ this needs to also be updated, otherwise the placement of the track timer tooltip can be misaligned:

https://github.com/digidem/comapeo-mobile/blob/312c140674979caae5d0c6931972f5aadf53d510/src/frontend/contexts/SharedLocationContext.tsx#L23

Another thing that could be contributing to the issue is that the placement of the user location marker is handled internally by the map library via their `UserLocation` component. From what I can tell, its placement is done completely separately from the coordinates provided to Camera or Mapview components. Not sure what it does under the hood but if it differs from how we calculate the track location, I wonder if it could potentially be a source of misalignment that this PR doesn't currently address.

achou11:
> Another thing that could be contributing to the issue is that the placement of the user location marker is handled internally by the map library via their UserLocation component. From what I can tell, its placement is done completely separately from the coordinates provided to Camera or Mapview components. Not sure what it does under the hood but if it differs from how we calculate the track location, I wonder if it could potentially be a source of misalignment that this PR doesn't currently address.

Upon very minimal testing, this doesn't seem like it's a notable issue. Might be okay to punt on looking into it for the sake of at least improving the existing behavior

src/frontend/screens/MapScreen/CurrentTrack/UserTooltipMarker.tsx
```diff
@@ -3,21 +3,21 @@ import {StyleSheet, Text, View} from 'react-native';
...
-import {useSharedLocationContext} from '../../../contexts/SharedLocationContext';
+import {useLocation} from '../../../hooks/useLocation';
+import {TRACKING_DISTANCE_INTERVAL} from '../../../constants';
 
 export const UserTooltipMarker = () => {
   const {timer} = useTrackTimerContext();
-  const {locationState} = useSharedLocationContext();
+  const {location} = useLocation({
+    maxDistanceInterval: TRACKING_DISTANCE_INTERVAL,
+  });
```
ErikSin, line 13: > I _think_ this needs to also be updated, otherwise the placement of the track timer tooltip can be misaligned:
> 
> https://github.com/digidem/comapeo-mobile/blob/312c140674979caae5d0c6931972f5aadf53d510/src/frontend/contexts/SharedLocationContext.tsx#L23
> 

Unfortunately I couldn't just update the location context as that would effect how often the map gets optimized. So i am just using this hook instead.

That being said, I am proposing a new location architecture that will fix all of this (this PR is why I am proposing that) so they can all used a shared location context

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -148,13 +151,15 @@ export const MapScreen = ({
         />
 
         {coords && locationServicesEnabled && (
-          <UserLocation minDisplacement={MIN_DISPLACEMENT} />
+          <UserLocation minDisplacement={TRACKING_DISTANCE_INTERVAL} />
         )}
 
+        {isTracking && <UserTooltipMarker />}
```
achou11, line 157: potentially out of scope: wondering if the rendering of these should be placed after the rendering of the different map layers as in move these lines further down in the file.

maybe something smart happens under the hood but otherwise the typical behavior is that things applied later overlap on top of previously applied UI elements in the map. Generally seems more reasonable for the user location and the tooltip to be displayed on top of everything else.

achou11 (commented):
Not really sure what the fix should look like but wondering if this is considered an improvement or not:

image

Think this may be related to the issue I mentioned in https://github.com/digidem/comapeo-mobile/pull/1046#pullrequestreview-2696027091, where the user location marker's placement is determined by something internal to the map library and not aligned with the location we calculate in our location hooks (or via expo-location in general)

In all fairness, I'm not really trying to move around due to development limitations but this seems aligned with the original issue.

ErikSin:
> Not really sure what the fix should look like but wondering if this is considered an improvement or not:
> image
> 
> Think this may be related to the issue I mentioned in [#1046 (review)](https://github.com/digidem/comapeo-mobile/pull/1046#pullrequestreview-2696027091), where the user location marker's placement is determined by something internal to the map library and not aligned with the location we calculate in our location hooks (or via expo-location in general)
> 
> In all fairness, I'm not really trying to move around due to development limitations but this seems aligned with the original issue.

LOL apparently i didn't fix the issue at all....it works in my testing, but I think thats more of a coincidence after seeing this picture. 

If the tracks and tooltip shared the same location object (instead of instantiating its own location subsctiption) those would be aligned. But the user location dot does internally determine its own location...

I dont know if its worth the effort, but the best solution would be to create our own user location dot that just shares the location from the tracks and tool tip...I will talk to design and see if it worth it.

PR #1058, opened 2025-03-24
refactor: move security settings to secure store

### LINKED ISSUE #1036: Use secure storage for security-related values
For things like the passcode and obscure code, we should be storing these using secure store, as these are sensitive values.

(see https://github.com/digidem/comapeo-mobile/pull/1027#issuecomment-2721075154 for more context)

The relevant values are:

- passcode
- obscure code

Questions:

- currently we store the actual obscure code. however, the app defines a hardcoded value that the user cannot change, so the UX is more like a toggle. does it still make sense to store the actual obscure code as opposed to some boolean-like? i think the answer will depend on if we anticipate allowing users to specify their own obscure code at some point. not really sure. in terms of code, the question comes down to some variation of:

    ```ts
    { passcode?: string, obscureCode?: string }
    ```

    vs

    ```ts
    { passcode?: string, obscureCode: boolean }
    ```

achou11:
Closes #1036 
Towards #1019 

Moves the storage of the passcode value and the obscure code flag to [secure store](https://docs.expo.dev/versions/latest/sdk/securestore/). As part of the architecture discussion, this PR does NOT migrate existing values related to the passcode or obscure code from the old storage, so users that may have used this feature prior to this change will have to set this up again if desired.

No additional security measures for accessing these values are enforced yet. This PR is focused on their storage and using the additional capabilities of the secure store API can be decided on as a follow-up. 

Other important changes:

- Renames the `SecurityContext` to the `AuthContext` since the naming was too similar to the `SecurityStoreContext` that this PR introduces.

- Updates the implementation of the `AuthContext` so that it strictly handles the calculation and management of auth state. Before, it was forwarding values that can be derived by using `useSecurityState()` directly.


cimigree (approved):
Looks good! I tested a little manually although we have good E2E tests for passcode.
Your unit tests look good!
It's good we now have a secure store.
And also glad that the auth logic is no longer in persisted state.
It's good we got this in before we had too many users as folks will have to reset their passcode.
I am about to embark on obscure passcode, so it is good to get this in first.

PR #1059, opened 2025-03-24
chore: update buttons

### LINKED ISSUE #947: (Correction) Correct style for secondary buttons

### LINKED ISSUE #979: Create Opinionated Button
Create new Button components with the [design system](https://www.notion.so/digidem/091e537955f7482f883962b4df8fb84b?v=d6158bd386064246892bae58398896ab&p=eec60615a2ac4317b52ca2ea01f2465d&pm=c) opinionated styles.

# To To
- [ ] add [jsdoc deprecated tag](https://jsdoc.app/tags-deprecated) to the original button component
- [ ] Create Primary button according to design system
- [ ] Create Secondary button according to design system
- [ ] Create Destructive button according to design system

## Example of component 
```ts
import type {IconProps} from 'react-native-vector-icons/Icon';

type RestrictedIconProps = Omit<IconProps, 'size' | 'color' | 'style'>;

type IconButtonProps = {
  label: string;
  onPress: () => void;
  fullWidth?:boolean
  icon?: React.ReactElement<RestrictedIconProps>; //This enforces a react-native-vector-icon, but the size, color, and style cannot be set
};

export const PrimaryButton = ({
  label,
  onPress,
  icon,
}) => {
  return (
    <TouchableOpacity style={...opinionated style} onPress={onPress}>
      {icon && <Text style={...opinionated color and size}>{icon}</Text>} //vector icons are built off text, so you can pass the size and color to the parent text icon, and the icon will inherit it.
      <HeaderText style={...opinionated color} variant="header5">
        {label}
      </HeaderText>
    </TouchableOpacity>
  );
};

```

ErikSin:
closes #947 
closes #979 
- create new button component using design opinionated style
- adds jsdoc deprecated tag to old button
- replaces some buttons as identified [here](https://www.notion.so/digidem/Correction-Correct-style-for-secondary-buttons-10b1b08162d5801eb19be0c1f6b1be80)

### Old Button:
image

### New Button:
image


cimigree:
There is also issue #979, right?
@ErikSin Can you update from develop and fix the merge conflict, please?
it looks like this is going to be an improvement and fit with design guidelines, but, it’s been hard to tell which screens still use the old Button and might need updating. Could you include a list of the specific screens or areas where you changed the button, and maybe point out which ones are left to tackle? Also, a couple of before/after screenshots would be super helpful for ensuring everything meets the design specs. Just thinking of our future selves searching back through completed PRs.

ErikSin:
> @ErikSin Can you update from develop and fix the merge conflict, please?

done

> Could you include a list of the specific screens or areas where you changed the button

Its in the notion doc that is linked in the description

> maybe point out which ones are left to tackle?

Every button needs to be replaced, so I don't think it would be helpful to list them all out.

src/frontend/screens/Onboarding/DataPrivacy.tsx
```diff
@@ -139,8 +137,4 @@ const styles = StyleSheet.create({
     alignItems: 'center',
     gap: GAP,
   },
-  learnMoreButton: {
-    borderColor: BLUE_GREY,
-    borderWidth: 2,
-  },
 });
```
cimigree, line 139: I think the 90% in the buttonContainer is wrong here based on a Slack convo I had with Aldo earlier as these kinds of buttons are supposed to be full width.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
```diff
@@ -70,23 +73,28 @@ export const ProjectCreated = ({
...
       </View>
       <View style={{width: '100%'}}>
-        <Button fullWidth variant="outlined" onPress={handleGoToInviteScreen}>
-          {t(m.inviteDevice)}
-        </Button>
-        <Button style={{marginTop: 20}} fullWidth onPress={handleGoToMap}>
-          {t(m.goToMap)}
-        </Button>
+        <SecondaryButton
+          fullWidth
```
cimigree: How come sometimes you pass fullWidth and sometimes you pass fullWidth={true} ?

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
```diff
@@ -197,19 +204,18 @@ export const CreateProject: NativeNavigationComponent<'CreateProject'> = ({
               </TouchableOpacity>
               {advancedSettingOpen && (
                 <View style={styles.importConfigContainer}>
-                  <Button
-                    fullWidth
-                    variant="outlined"
+                  <SecondaryButton
+                    fullWidth={true}
```
cimigree: Here is an example of fullWidth={true}

src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
```diff
@@ -2,11 +2,11 @@ import {defineMessages, useIntl} from 'react-intl';
 import {StyleSheet, View} from 'react-native';
 
 import {WHITE} from '../../lib/styles';
-import {Button} from '../../sharedComponents/Button';
 import {ObservationListIcon} from '../../sharedComponents/icons';
 import {ScreenContentWithDock} from '../../sharedComponents/ScreenContentWithDock';
 import {BodyText} from '../../sharedComponents/Text/BodyText';
 import {HeaderText} from '../../sharedComponents/Text/HeaderText';
+import {SecondaryButton} from '../../sharedComponents/Buttons';
```
cimigree, line 9: I think the padding needs to be removed on this screen so that the button can appear full width
<image src="https://github.com/user-attachments/assets/92c20e16-1702-43f0-8beb-9da714ec9a9b" width="250" />

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
```diff
@@ -262,18 +251,15 @@ export const ProjectSyncDisplay = ({
...
             projectApi.$sync.stop();
-          }}>
-          <View style={styles.buttonContentContainer}>
-            <StopIcon size={20} color={BLACK} />
-            <HeaderText variant="header5">{t(m.stop)}</HeaderText>
-          </View>
-        </Button>
+          }}
+          text={t(m.stop)}
+          renderIcon={({size, color}) => <StopIcon size={size} color={color} />}
```
cimigree, line 261: How did you know to change the size to 25? Here I see it is 20...

src/frontend/sharedComponents/Buttons.tsx
```diff
@@ -0,0 +1,122 @@
...
+      testID={testID}
+      style={[
+        style,
+        styles.base,
+        {backgroundColor: COMAPEO_BLUE, width: fullWidth ? `100%` : 280},
+      ]}
+      onPress={onPress}>
+      {renderIcon && (
+        <View style={styles.iconSpacing}>
+          {renderIcon({color: WHITE, size: 25})}
```
cimigree, line 36: Just curious... how did you know to change the icon size to 25?

src/frontend/sharedComponents/Buttons.tsx
```diff
@@ -0,0 +1,122 @@
...
+  text,
+  fullWidth,
+  testID,
+  style,
+}: ButtonProps) => {
+  return (
+    <TouchableOpacity
+      testID={testID}
+      style={[
+        style,
```
cimigree, line 28: Shouldn't the passed style be last to override the default styles?

src/frontend/sharedComponents/Buttons.tsx
```diff
@@ -0,0 +1,122 @@
...
```
cimigree, line 58: style is here twice

src/frontend/sharedComponents/Buttons.tsx
```diff
@@ -0,0 +1,122 @@
...
```
cimigree, line 92: see comment above about style being last

cimigree (approved):
This will make it so much easier to use the different buttons. Thanks. I included some comments that I hope you will address but they are non blocking.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
ErikSin: purely stylistic. sometimes i copied and pasted so i just forgot to put `true`

src/frontend/sharedComponents/Buttons.tsx
ErikSin: Good question. I did this on purpose because the styling of background color and text color is opionated. So If a dev tries to overwrite that they cannot...I don't know if this the best strategy, but it was the simplest.

src/frontend/screens/Sync/ProjectSyncDisplay.tsx
ErikSin: Because we didn't have a clear standard for icons, I assume the original sizing was eyeballed (or maybe based off figma). The visual difference between 20 and 25 was very minor, so I made the decision to just remain consistent and use the size that is being returned by the button.

src/frontend/sharedComponents/Buttons.tsx
ErikSin: Yeah similar to the comment above. Whenever an icon was added to a button, the developer put a size in with no clear reason. I reviewed the code and it seemed anywhere between 20 - 30. So I just chose the in between so that we can have a consistent sizing throughout

src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
ErikSin: @achou11 I know you updated this screen recently. Was the `48` padding intentional?

achou11: Probably but most likely based on my own instincts or other similar existing examples, as opposed to following design guidelines. feel free to update however necessary!

ErikSin:
I think we are doing full size button wrong. I am going to confirm with Aldo, but I think the max size of a button is `280`. Im going to leave this PR open until I confirm that

ErikSin:
@achou11 and @cimigree just a good to know moving forward. Buttons max out at 280px. So when updating we just need to make sure that we center them. previously we didnt have to worry about that because they were the full width.

I also changed the prop name to be `fullSize` instead of `fullWidth`

ErikSin:
@cimigree can you do quick look through again due to the width changes? 

I also added the screen shots as you requested :)

cimigree:
>  can you do quick look through again due to the width changes?
> 
> I also added the screen shots as you requested :)

@ErikSin I checked through the 7 screens and all looks (I think) the way it is supposed to. It's annoying now we have to center all the buttons, but I don't think there is any easy way around it, right? We will have to change A LOT of buttons! Good thing you checked. It (the button) looks quite different to me, but it is as it should be as far as I can tell

PR #1060, opened 2025-03-24
refactor: update state management setup for coordinate format and manual entry coordinate format settings

achou11:
Towards #1019 

Notes:

- Separates the storage locations of each setting. In this case, the coordinate format and the manual entry coordinate format are affected. 

    - Based on subsequent discussions from attempting #1018 , we realized that it's easier when only _related_ settings are stored in the same storage entry. In this case, despite similar naming, the two settings related to coordinate formats are _unrelated_ to each other, so this PR explicitly separates them.
    - There is NO migration of the previous settings to this new storage approach, so any relevant settings done by the user prior to this will lose that change.

- Some of the names are a bit verbose. Happy to consider suggestions for shortening if desired.

- Given that we're separating the states, I was hoping that using a scalar value to represent the zustand state (as opposed to an object) would be trivial. Unfortunately, Zustand's persistence plugin and how it serializes state make this tedious to achieve. Instead, this PR represents each state as an object but the hooks to consume that state return scalar values.


src/frontend/lib/coordinateFormat.ts
```diff
@@ -0,0 +1,101 @@
...
+  const longitudeCardinal = lon >= 0 ? 'E' : 'W';
+  return `${latitude} ${latitudeCardinal}, ${longitude} ${longitudeCardinal}`;
+}
+
+export function convertToUTM({lat, lon}: {lat: number; lon: number}) {
+  try {
+    const {easting, northing, zoneNum, zoneLetter} = fromLatLon(lat, lon);
+    return `UTM ${zoneNum}${zoneLetter} ${easting.toFixed()} ${northing.toFixed()}`;
+  } catch {
+    // Some coordinates (e.g. < 80S or 84N) cannot be formatted as UTM
```
ErikSin, line 74: I think we should write tests for this

src/frontend/lib/coordinateFormat.ts
ErikSin, line 1: I see why alot of these function don't have test as they are relatively straightforward. So perhaps this is a bigger question of what does our testing coverage look like, but do you think it would not be worth just writing 1 or 2 tests for these function even if they are pretty obvious? Open question, because I don't have a fully formed opinion on this.

On one hand, i think total coverage is unnecessary and doesn't actually give us the full picture. On the other hand, total coverage has the easiest mental load in terms of knowing when to write tests (aka we know to write tests for everything so we dont have to think about it).

What do you think?

ErikSin (changes requested):
1 small change, i think we should write a test for that edge case you mentioned in the comment.

Also 1 question about test coverage that perhaps we should discuss as a larger group, but would love to get your thought on it.

src/frontend/lib/coordinateFormat.ts
ErikSin: Just for clarity, I am referring to non react components with this comment (aka pure functions)? I know we already discussed our testing philosophy for react components.

src/frontend/lib/coordinateFormat.ts
achou11: can this be done as a follow-up? feels out of scope as this function existed prior to this PR.

src/frontend/lib/coordinateFormat.ts
achou11: I agree that having tests for these is actually preferable, but I think adding them to this PR is out of scope given that they existed prior to this PR (this PR just moves them to a different file). Mind if they're added as a follow-up?

achou11:
Something that I just thought about: any thoughts on updating the hooks to be more like React's `useState()`? for example, instead of:

```ts
function Example() {
  const coordinateFormat = useCoordinateFormat()
  const {setCoordinateFormat} = usCoordinateFormatActions()
}
```

Do this:

```ts
function Example() {
  const [coordinateFormat, setCoordinateFormat] = useCoordinateFormat()
}
```

Kind of diverges from our recommended approach for zustand-based state, but given that the values are scalar (i.e. not more complex objects), this seems like less verbose way of using the hooks in app code, without any of the tradeoffs when trying to do this for more complex state (with more actions)

ErikSin:
> Something that I just thought about: any thoughts on updating the hooks to be more like React's `useState()`? for example, instead of:
> 
> ```ts
> function Example() {
>   const coordinateFormat = useCoordinateFormat()
>   const {setCoordinateFormat} = usCoordinateFormatActions()
> }
> ```
> 
> Do this:
> 
> ```ts
> function Example() {
>   const [coordinateFormat, setCoordinateFormat] = useCoordinateFormat()
> }
> ```
> 
> Kind of diverges from our recommended approach for zustand-based state, but given that the values are scalar (i.e. not more complex objects), this seems like less verbose way of using the hooks in app code, without any of the tradeoffs when trying to do this for more complex state (with more actions)

Personally, i like the explicit naming, but I don't feel strongly about it. So up to you, as I agree there aren't any real tradeoffs, so just a style question at this point

achou11:
Yeah i think i'll keep things as is for now. can revisit as a follow up if needed

src/frontend/lib/coordinateFormat.ts
ErikSin: oh i see, you moved the prewritten into a new file! Ok, yeah lets do it as a follow up

ErikSin (approved):

PR #1061, opened 2025-03-24
refactor: update state setup for active project id

achou11:
Towards #1019 

Notes:

- Wasn't sure about whether we should use the existing storage location or just do a hard switch. ~Currently tries to migrate the existing storage if possible.~ EDIT: No longer tries to migrate. Instead we try to reuse the previously persisted value by maintaining the same storage version, since the shape of the relevant data is the same in this PR. See [thread](https://github.com/digidem/comapeo-mobile/pull/1061#discussion_r2012498955)


src/frontend/contexts/ActiveProjectIdStoreContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+              return {
+                projectId: persistedState.projectId,
+              };
+            }
+          }
+
+          return newState;
+        },
+      }),
+    );
```
ErikSin: overall it looks good, but before i submit an official review I just want to clarify what this migration is for? Is it because previously the "action" was in the store?

achou11: it's to make sure that the previously persisted active project id is actually ported over to the new storage version since we're bumping the version. without doing this, zustand will basically act as though no previous value was persisted and fall back to the initial value we set (i.e. no active project id)

as noted in my PR description, not sure how imperative doing this is in terms of user impact. if we don't do it, we just have to rely on the logic that's done [here](https://github.com/digidem/comapeo-mobile/blob/1861dab49b42398a3143d5cd9dc67373a38f8e4f/src/frontend/contexts/ActiveProjectContext.tsx#L35-L62), which works okay for now but is a bit fragile

achou11: alternatively, we can use the same storage but not bump the version. I think the shape of what's stored doesn't change with this PR, as I believe functions within an object (like the original `setProjectId()`) don't get persisted due to serialization via JSON stringify.

It's a bit riskier but might be okay. The main question is do we want to start with a clean slate or try to use the existing storage

ErikSin: Yeah, sorry, what I was trying to get at is understanding why we need to bump the version. 

Even if the shape of the store changes (but I think your right, setProjectId doesn't get persisted), both stores would still have the value `projectId`. so I *think* it would be fine. BUT even if it isn't fine we do have that `useEffect` check.

The complexity of this migration feels a little unnecessary and it ill become less of a problem when we have multi projects down the line

achou11: Yeah makes sense. can update to omit the migration and test that it works!

achou11: Addressed via https://github.com/digidem/comapeo-mobile/pull/1061/commits/37b12e7a4e6c36d5c04847d647a79c9373e65e10. Looks like the reuse of the old storage worked based on my local testing 👍 

Just to clarify: this keeps the same persistence key but no longer bumps the storage version. is that sufficient, or should we just update the persistence key and start fresh?

ErikSin: > Just to clarify: this keeps the same persistence key but no longer bumps the storage version. is that sufficient, or should we just update the persistence key and start fresh?

I think keeping the same persistence key is the easiest thing to do, and i don't think it will cause any problems

ErikSin (approved):

PR #1063, opened 2025-03-25
feat: add option to use system preferences for language

### LINKED ISSUE #978: App does not use system language preference when expected

### LINKED ISSUE #1019: Port zustand-based states to use preferred setup

### LINKED ISSUE #1028: Language option for following system preference
There's currently no way for the user to explicitly tell the app that it should follow the system preference(s) for language.

Based on discussions with design that originated from #978, we should add an option in the language settings screen that says `Follow System Preference` as the first option. It should be selected as the default if no language within that screen has been explicitly selected by the user.

achou11:
Closes #978 
Closes #1028 
Closes #1019

A combination of changes from #1011 and #1027 with two goals:

- update the state setup to follow other zustand-based states
- fix a bug with language resolution by introducing a new UI option for language selection, which makes language resolution much more predictable

---

image


src/frontend/hooks/useAppLanguageTag.ts
```diff
@@ -0,0 +1,17 @@
+import {useLocales} from 'expo-localization';
+import {useMemo} from 'react';
+
+import {useLocaleState} from '../contexts/LocaleStoreContext';
+import {getAppLanguageTag} from '../lib/intl';
+
+export function useAppLanguageTag() {
```
achou11, line 7: This currently returns an object that has this representation:

```ts
type ResolvedLanguageTag = {
  source: 'fallback' | 'system' | 'selected';
  value: SupportedLanguageTag;
};
```

The `source` field is not really used in the app, but I find it potentially helpful for debugging. Open to just returning the `value` field when it comes to this hook specifically though.

achou11:
Reminder to myself that if there's an e2e test related to the settings screen, I should update it to account for this new UI. ~Based on the fact that the CI passed, my guess is that there isn't. If that's the case, wondering if it'd be worth doing as part of this PR or as a follow-up~ EDIT: ah looks like there are tests. will update them to confirm that the initially selected option is this new option.

https://github.com/digidem/comapeo-mobile/blob/629129f7a45b9d447dcb03c3436b35ebdc4057ce/e2e/settings/language.yaml

https://github.com/digidem/comapeo-mobile/blob/629129f7a45b9d447dcb03c3436b35ebdc4057ce/tests/e2e/specs/settings/language.test.ts

achou11:
> Reminder to myself that if there's an e2e test related to the settings screen, I should update it to account for this new UI.

I have a clear idea of the test I want to add (basically adding an assertion [here](https://github.com/digidem/comapeo-mobile/blob/629129f7a45b9d447dcb03c3436b35ebdc4057ce/tests/e2e/specs/settings/language.test.ts#L6-L15) but I think it'll require a decent amount of changes to do this properly. From my perspective, the general steps would be:

1. Add the appropriate [accessibility attributes](https://reactnative.dev/docs/0.76/accessibility#accessibilitystate) to the relevant element(s) within the `SelectOne` component (looking specifically at [`accessibilityState`](https://reactnative.dev/docs/0.76/accessibility#accessibilitystate) and [`aria-checked`](https://reactnative.dev/docs/0.76/accessibility#aria-checked))

2. Create a helper wdio selector to query `checked` elements ([android docs](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector#checked(boolean)))

3. Update the relevant test to make use of the custom selector and write the assertion

I think (1) will require touching a decent number of things that I'm currently a little unfamiliar with, so I'm leaning towards doing this work as a follow-up as opposed to this PR. Open to other thoughts though

src/frontend/lib/intl.ts
```diff
@@ -47,24 +68,85 @@ function deriveSupportedLanguages(
...
 
-  const nonRegionalLocale = locale.split('-')[0];
+    if (languageCode === supported.languageTag) {
+      return supported.languageTag;
+    }
+  }
+}
+
+type ResolvedLanguageTag = {
+  source: 'fallback' | 'system' | 'selected';
```
cimigree, line 106: Should there be a comment to explain the source part and when/ where it will be used?

cimigree (approved):
Great idea to have Follow system preferences. That makes things so, so much clearer.
Look what happens with the app as soon as I change the system preference?
<image src="https://github.com/user-attachments/assets/d0d66aeb-5367-4547-890a-10d9deff9076" width="250" />
Very cool!
I switched back languages back and forth from English to French to Spanish in and out of the app, and it was quite responsive -- no race conditions that I could see
Loving those tests! They seem very comprehensive.
Since we are not using the source in the rest of the app, it might be better to simplify the API and just return the language value?
What is the migration plan? Sorry if I missed that. Or are we not worried about that?

achou11:
> Since we are not using the source in the rest of the app, it might be better to simplify the API and just return the language value?

Yeah I can update the `useLanguageTag` hook to just return the language tag string. I'll keep the helper the same because I do find it easier for understanding the implementation and debugging.

>  What is the migration plan? Sorry if I missed that. Or are we not worried about that?

There is no migration plan. Based on discussions we've had, we're okay with everyone being "reverted" back to `Follow system preferences` option and requiring them to go through the process of choosing a different language if needed

src/frontend/lib/intl.ts
achou11: Yeah I'll leave a comment mentioning that it's mostly helpful for understanding the implementation and debugging

achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/1063/commits/877e36770be6cf58676dbe30ed88cb3eadcb52a6

PR #1064, opened 2025-03-25
refactor: update state setup for metrics diagnostics permissions

achou11:
Towards #1019

Notes:

- reuses the same storage as the previous setup, but does not change the version number. the shape of the store is the same so previously persisted values will be used.


src/frontend/App.tsx
```diff
@@ -83,6 +84,17 @@ const persistedActiveProjectIdStore = createActiveProjectIdStore({
...
+const persistedMetricsDiagnosticsStore = createMetricsDiagnosticsStore({
+  persist: true,
+});
+
+persistedMetricsDiagnosticsStore.instance.subscribe((current, previous) => {
+  if (previous.isEnabled !== current.isEnabled) {
+    appDiagnosticMetrics.setEnabled(current.isEnabled);
+    deviceDiagnosticMetrics.setEnabled(current.isEnabled);
+  }
+});
```
achou11, line 103: got rid of the `MetricsContext` in favor of doing this. Not sure if that's preferable, but the context seemed unnecessary and we weren't using it for anything else.

ErikSin: yeah this makes sense. but, there is a race like condition (that was also happening before so you didn't introduce it, just noticing it now). We initialize `AppDiagnosticMetrics` with a default value of `false`. Then we sync the state, where it could be set to true. 

Its not a problem because when it is set to false, nothing happens, and once it is set to true, then we send metrics right away. So I suppose it is not a race condition per se. But it is not the clearest logic. I wonder if we should include the initial value in the constructor of `appDiagnosticMetrics` an `deviceDiagnosticMetrics`

ErikSin (approved):
1 small comment about how we initialize the metrics, but not blocking (just weird but functional logic)

src/frontend/App.tsx
achou11: hm yeah good point. i actually think that how i have this implemented actually introduces a bug, partially contributed to by what you've described. The code i have is only updating metrics instances when the relevant zustand state has _changed_ (e.g. previous !== current). Otherwise, on initial startup the current state we have is:

- zustand state says that metrics is _enabled_
- appDiagnosticsMetrics is _disabled_
- deviceDiagnosticsMetrics is _disabled_

probably need to update the code i have above to make it so that it doesn't try to be "efficient" i.e. remove the `if (previous.isEnabled !== current.isEnabled)` condition.

Additionally can do something similar to what you mentioned where we pass a constructor opt for those metrics instances so that we make sure that they're initialized with the same default as what we're initializing the zustand state with.

achou11: Addressed via [c16f942](https://github.com/digidem/comapeo-mobile/pull/1064/commits/c16f9425b26fcdc524e57805f2128740afd1b026). Opted to avoid updating the relevant constructors in favor of calling the relevant methods when initializing the app.

Also learned that the zustand store `subcribe()` method only fires on subsequent state changes and does not emit the initial state, which makes sense.

Issue #1062, opened 2025-03-25
add tests for coordinate format helper functions


achou11:
Extracted from discussion in https://github.com/digidem/comapeo-mobile/pull/1060. We have a few functions in [`lib/coordinateFormat.ts`](https://github.com/digidem/comapeo-mobile/blob/c65dda3be93ddd8eb79bd1600870917f8e55919a/src/frontend/lib/coordinateFormat.ts) that don't yet have tests.

PR #1065, opened 2025-03-26
chore: use coordinate format functions from `lib/coordinateFormat` instead of `lib/utils`

achou11:
Meant to do this in #1060 but forgot to port over the relevant changes before merging (dangers of copy+paste from previous attempts 😅 )

This PR doesn't add tests for the relevant functions as described in #1062 since I want this to be a quick follow-up

### DIFF
--- a/src/frontend/lib/utils.ts
+++ b/src/frontend/lib/utils.ts
@@ -1,10 +1,8 @@
 // import { Alert } from "react-native";
-import {fromLatLon} from 'utm';
 import {Preset, Observation, Track} from '@comapeo/schema';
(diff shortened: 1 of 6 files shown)


src/frontend/lib/utils.ts
achou11, line 1: for ease of comparison, the functions removed have already been copied over to [`lib/coordinateFormat.ts`](https://github.com/digidem/comapeo-mobile/blob/a0cd4f08b9a7fc84be81b96bc664f94cba1aa082/src/frontend/lib/coordinateFormat.ts)

PR #1066, opened 2025-03-26
Feature: adds obscure passcode 

### LINKED ISSUE #988: Integrate Obscure Passcode to CoMapeo
Obscure Passcode was a Security Feature in Mapeo Legacy. It was also called "Kill passcode" on its first iteration.
In CoMapeo, it is no longer available.
Goal : re-introduce the feature.

# To Do
- [ ] Set 00000 as obscure code
- [ ] Hide all observation
- [ ] Intecerpt creation of observations
- [x] Hide Screen when in background 

## Design
https://www.figma.com/design/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund

cimigree:
closes #988 
# Re-introduce Obscure Passcode Feature and Add E2E Tests

## Overview

This PR re-introduces the **Obscure Passcode** feature (previously known as "Kill Passcode" in Mapeo) into CoMapeo. The feature uses a default code of `00000` that, when entered, puts the app into a mode where:
- The observation list appears blank.
- New observations not permanently saved.
- The Security section is hidden from settings.

Additionally, this PR updates UI text components to use our new `HeaderText` and `BodyText` and adds the visual of the  text turning red on the obscure passcode menu item when a passcode is not set.

## What Changed

- **Native Module Integration**  
  - Added a custom `FlagSecureModule` (and corresponding package) via a config plugin in `expo-config-plugins/flagSecure/` that automatically copies the Java files into the Android project and patches `MainApplication` to register the module.
  
- **Security Context Updates**  
  - In the `AuthContext`, calls to `FlagSecureModule.activate()`/`deactivate()` are now triggered based on the app state and the passcode setting.
  - If the obscure passcode (`00000`) is entered, the app sets `authState` to `'obscured'`, which causes the UI to prevent access to certain screens.

- **E2E Tests**  
  - New tests have been added under `tests/e2e/specs/passcode/`:
    - **Obscure Passcode Visibility Test:** Verifies that the obscure passcode menu item displays the correct text based on whether the passcode is set.
    - **Obscure Passcode Mode Tests:**  
      - Ensure that when the obscure passcode (`00000`) is entered, the observation list shows an empty state.  
      - Verify that if you add and observation in obscure mode it is not saved.  
      - Confirm that the "Security" option is not visible in settings.
      - Verify that entering the regular passcode restores normal observations (without including any obscure-mode observations).
    - An additional test ensures that tapping on the obscure passcode when it is not allowed does not navigate away

## How to Test

1. **Manual Testing:**  
 Aside from the E2E tests, check that when the app is in Obscure Mode, and you background the app, that the "card" for it is blank (either a black or a white screen).


src/frontend/contexts/AuthContext.tsx
```diff
@@ -34,6 +35,15 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
   const shareDialogIsOpen = useIsShareDialogOpen();
+  React.useEffect(() => {
+    if (!process.env.EXPO_PUBLIC_E2E_TEST) {
+      if (passcode !== null) {
+        FlagSecureModule.activate();
+      } else {
+        FlagSecureModule.deactivate();
+      }
+    }
+  }, [passcode]);
```
ErikSin: What is the reasoning for this?

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -151,6 +153,11 @@ export const ObservationCreate = ({
 
   const createObservation = React.useCallback(async () => {
     if (!value) throw new Error('no observation saved in persisted state ');
+    if (authState === 'obscured') {
+      clearDraft();
+      navigation.popTo('Home', {screen: 'Map'});
+      return;
+    }
```
ErikSin, line 160: clever way of doing this!

src/frontend/contexts/AuthContext.tsx
```diff
@@ -43,13 +53,18 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
+          }
           if (
             nextAppState === 'active' ||
             nextAppState === 'background' ||
             nextAppState === 'inactive'
           ) {
             setAuthState('unauthenticated');
           }
+        } else if (!process.env.EXPO_PUBLIC_E2E_TEST) {
+          FlagSecureModule.deactivate();
```
ErikSin: this logic is quite confusing.

Also the FlagSecureModule should only be activated once, right? Like when the user turn on obsecure mode, we activate it, and when they turn off obscure mode they deactivate it? Im pretty sure it is designed so that it does not need to be manually turned on an off every time the user exits the screen

src/frontend/screens/ObscurePasscode.tsx
```diff
@@ -52,36 +53,43 @@ export const ObscurePasscode: NativeNavigationComponent<'ObscurePasscode'> = ({
 
   React.useEffect(() => {
     if (authState === 'obscured') {
-      navigation.popTo('Settings');
+      navigation.popTo('AppSettings');
     }
   }, [navigation, authState]);
```
ErikSin: this should be a `useLayoutEffect` so that the screen doesn't get drawn

src/frontend/screens/Security/index.tsx
```diff
@@ -36,13 +50,21 @@ export const Security: NativeNavigationComponent<'Security'> = ({
...
   const passcodeSet = useSecurityState(state => state.passcode !== null);
   const {authState} = useAuthContext();
+  const [highlight, setHighlight] = React.useState(false);
 
   React.useEffect(() => {
     if (authState === 'obscured') {
-      navigation.popTo('Settings');
+      navigation.popTo('AppSettings');
     }
   }, [navigation, authState]);
```
ErikSin: same comment as above, should be a `useLayoutEffect`

tests/e2e/specs/passcode/obscure-mode.test.ts
```diff
@@ -0,0 +1,82 @@
...
+import {byText, byTextMatches, byResourceId} from '../../utils/selectors';
+import {output} from '../../utils/naming';
+
+describe('Passcode - Obscure Passcode Mode', () => {
+  it('should show a blank Observations screen after entering obscure passcode', async () => {
+    const obsListTab = await $('~Go to ObservationsList');
+    await obsListTab.click();
+    await expect($(byTextMatches('Lake'))).toBeDisplayed();
+    await driver.terminateApp('com.comapeo.rc');
+    await driver.activateApp('com.comapeo.rc');
```
ErikSin, line 12: I don't know if we can change this, but this means these test will only work on RC builds, so we wont be able to run these tests on dev builds like andrew mentioned. I think it is fine for now, since we don't know if we can run e2e on dev build but just noting so we know @achou11

ErikSin (changes requested):
Overall the functionality is good and all my comments are minor.

I am not fully approving it because i think we don't need to turn flag secure on and off everytime the app goes into the background/foreground. I think we need to turn it on once, and it will determine what to do itself when the app goes into the background. I also think that we need to clean up the logic in the auth context. Its just a little convoluted.

On a side note (I think i might have introduced this), but can we fix these buttons, they should be full size:

image

src/frontend/contexts/AuthContext.tsx
cimigree: I will add a comment. I thought I did. My mistake... But basically the screen recording doesn't work and then you can't see a video of the tests on Browserstack (or if you did it locally).

cimigree:
I also updated the text and the button in the InputPasscode screen. Hope that works. The buttons are not full width because it is the screen content with dock screen which has a padding of 20:
https://github.com/digidem/comapeo-mobile/pull/1066/commits/3fc2a38f7435b906d2acec63b1b0768ebd10363a
<image src="https://github.com/user-attachments/assets/f0159651-36be-4f39-834a-16b89471e0f3" width="250" />

src/frontend/screens/Security/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1066/commits/3fc2a38f7435b906d2acec63b1b0768ebd10363a

src/frontend/screens/ObscurePasscode.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1066/commits/3fc2a38f7435b906d2acec63b1b0768ebd10363a

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1066/commits/3fc2a38f7435b906d2acec63b1b0768ebd10363a

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1066/commits/3fc2a38f7435b906d2acec63b1b0768ebd10363a

src/frontend/contexts/AuthContext.tsx
```diff
@@ -32,9 +33,19 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
+  React.useEffect(() => {
+    if (!isE2E) {
+      if (passcode !== null) {
+        FlagSecureModule.activate();
+      } else {
+        FlagSecureModule.deactivate();
+      }
+    }
+  }, [passcode, isE2E]);
+
```
ErikSin: I think this would be clearer and catch any cases where flagsecure was activated during e2e if you don't nest the statement:

```ts

if(passcode !== null && !isE2E){
   FlagSecureModule.activate();
} else{
    // now if flag secure was activated somewhere else but we are in e2e, then it will be deactivated
    FlagSecureModule.deactivate();
}

```

ErikSin (changes requested):
So the useLayoutEffects are actually not doing what I thought they would do...If the screen is already open, then it still shows, which is obviously a problem. Also, if the user is viewing an observation we are opening to the observation screen even on obscure mode.

SOOO radical idea, lets just ALWAYS `popTo` to the home(map) screen when obscure mode is set. then we do not have to worry about these useEffects/useLayout Effects, and there will be less code to maintain.

cimigree:
We now navigate to home/ map screen when coming in from the auth screen, but I have to check for obscured passcode in the mapscreen useFocusEffect or you still get navigated to an observation in progress.
Removed three other instances of useLayoutEffect where we were checking for obscured passcode in those cases.
The passcode screen became just a renderer of the intro screen so I sort of merged them (and updated the button and text). Hope that is ok. I can undo that if that is not semantic/ right.

ErikSin (approved):
Yeah, the navigating to home just feels much cleaner and less developer overhead

PR #1067, opened 2025-03-27
feat: observation metadata screen

### LINKED ISSUE #1044: Exposing Observation Metadata
**Description**
To introduce the Observation Metadata, we are introducing the ability by tapping the grey bar that has the date and shield (the entire bar is tappable) or And by tapping the gps coordinates bubble (the entire bubble is tappable)
We are not displaying the **How we validate** icon

**Feature Spec**
https://www.notion.so/digidem/Exposing-Observation-Data-GPS-Photos-6f9cd1cf9f0e40ed8eda96b6a70f3aa3?pvs=4

**User Requirements**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1c81b08162d58099b1bd000c34450044&pvs=4

**To Do**

Image

**Validated state**

Image

- [ ] On the Observation, allow to tap the grey bar 
- [ ] Validated state view - items listed
- Date and time
- GPS coordinates
- Latitude
- Longitude
- Accuracy
- Altitude
- Altitude accuracy
- Speed
- ~(Share Button)~ to be done https://github.com/digidem/comapeo-mobile/issues/1053

**Un-validated state**
Image

- [ ] On the Observation, allow to tap gps coordinates bubble
- [ ] (Un-validated) state - items listed
- User sees the metadata is un-validated by reading “This data was manually entered”
- Date and time
- GPS coordinates
- Latitude
- Longitude
- ~(Share Button)~ to be done https://github.com/digidem/comapeo-mobile/issues/1053

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=4462-3153&t=0XLf5Fh7LsLDmqGy-0

ErikSin:
closes #1044 

image

image


src/frontend/screens/Observation/index.tsx
```diff
@@ -90,16 +91,25 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
       style={styles.root}
       contentContainerStyle={styles.scrollContent}>
       <>
-        {/* check lat and lon are not null or undefined */}
-        {lat != null && lon != null && <InsetMapView lat={lat} lon={lon} />}
-        <View>
-          <BodyText variant="smallMeta" style={styles.time}>
+        {/* check lat and lon are undefined. We cannot do `!lat && !lon` as 0 is a valid number but a falsy value */}
```
cimigree, line 94: I don't think this is supposed to have the shield if it is not verified, no?

src/frontend/screens/Observation/index.tsx
```diff
@@ -90,16 +91,25 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
+          }}
+          style={styles.time}>
+          <BodyText variant="smallMeta">
             <FormattedObservationDate
               createdDate={observation.createdAt}
               variant="long"
             />
           </BodyText>
-        </View>
+          <VerifiedBadge style={{marginLeft: 10}} />
```
cimigree, line 112: I don't think this is supposed to have the shield if it is not verified, no?

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -0,0 +1,346 @@
...
+            const key = Object.keys(item)[0] as string;
+            return (
+              <View
+                style={[
+                  styles.listItem,
+                  {backgroundColor: index % 2 === 0 ? VERY_LIGHT_GREY : WHITE},
+                ]}>
+                <View style={styles.flexRow}>
+                  {item[key]!.icon}
+                  <BodyText
```
cimigree, line 267: This looks like it is supposed to be bold in the Figma mocks

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -0,0 +1,346 @@
...
+                    style={{textAlign: 'left', marginLeft: 5}}
+                    variant="smallMeta">
+                    {item[key]!.label}:{' '}
```
cimigree: The colon isn't in the figma...

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -0,0 +1,346 @@
...
+};
+
+export const VerifiedHeader = () => {
+  const {formatMessage} = useIntl();
+
+  return (
+    <>
+      <BadgeWithCheck />
+      <View style={[styles.flexRow, {marginBottom: 10}]}>
+        <VerifiedBadge stroke={COMAPEO_BLUE} style={{marginRight: 10}} />
```
cimigree: These seem a smidge too close together. The badge and the How we validate
<image src="https://github.com/user-attachments/assets/7c1eb1b2-59ea-40fd-8c11-735066efcbe4" width="250" />

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -0,0 +1,346 @@
...
+  'ObservationMetadata'
+> = ({route}) => {
+  const {formatMessage} = useIntl();
+  const {
+    data: {createdAt, lat, lon, metadata},
+  } = useObservation(route.params.observationId);
+  const coordinateFormat = useCoordinateFormat();
+  const manualLocation = metadata?.manualLocation;
+
+  const listData: {
```
cimigree, line 91: Would it make sense to extract the metadata item formatting (from listData through filteredListData) into a helper function/ separate file? Not sure how testing will work, but I am wondering if it would be easier to test this logic without touching UI later.

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -0,0 +1,346 @@
...
+        unit: 'm/s',
+        icon: (
+          <MaterialIcons name="speed" color={NEW_DARK_GREY} size={ICON_SIZE} />
+        ),
+      },
+    },
+  ];
+
+  // takes out any undefined values
+  const filteredListData = listData.filter(
```
cimigree, line 186: So it looks like you set it up to skip metadata fields with undefined values (like speed or accuracy). Should we instead include them with a "Not available" placeholder instead? I could not see what/ where design said what to do in that situation.

src/frontend/screens/Observation/InsetMapView.tsx
```diff
@@ -1,54 +1,68 @@
...
+          source={require('../../images/observation-icon.png')}
         />
-      </MapboxGL.MapView>
-    </View>
-  );
-});
+        <TouchableOpacity
+          onPress={() => navigate('ObservationMetadata', {observationId})}
+          style={styles.coords}>
+          <View style={styles.coordsPointer} />
```
cimigree: My map looks different than the figma? Looks like the orange pointy thing (the observation-icon) is supposed to be in the box but it is instead in the map.
<image src="https://github.com/user-attachments/assets/13138a27-0054-41ce-bc45-369fd0e233ae" width="250" />

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/(Mobile)-CoMapeo?node-id=4461-2835&t=7edLLMm4st67EbMm-4

cimigree (changes requested):
Mostly just comments and questions, style things mostly... but one thing I _think_ should be fixed, though I am not certain...
When the data is unverified, that shield is still in the date view on the observation. (that hunting site observation from me is manually entered). It seems like it should not be there in that case.
And the pointer being in the wrong place seems like it should be fixed unless I am missing something.
The rest is up to you as you see fit.

src/frontend/screens/Observation/index.tsx
ErikSin: That was not what I was told. I will confirm though

src/frontend/screens/ObservationMetadata.tsx
ErikSin: I'm not sure what you mean with this comment. Can you elaborate?

src/frontend/screens/Observation/index.tsx
ErikSin: I spoke to Sabella and this is suppose to always show. Its not suppose to be an indicator that is is verified, just an icon to check whether or not is verified.

src/frontend/screens/ObservationMetadata.tsx
ErikSin: The logic is simpler to just skip it, and i don't want to inject UI that has not been approved

src/frontend/screens/Observation/index.tsx
cimigree: Ok

src/frontend/screens/ObservationMetadata.tsx
cimigree: What I meant is the logic that builds the array of metadata items (listData) and filters out the undefined ones (filteredListData) could be put into a helper function in a separate file — something like `getObservationMetadataItems({lat, lon, metadata, formatMessage})`
And if you did that, you could just test the logic...
For example, you could write a unit test that confirms which metadata fields appear depending on what data is passed in.
Right now this logic lives inline in the component, so to test it you would need to render the whole screen and do assertions based on what ends up on the screen. But if it were extracted, you could just check what inputs give what outputs without needing to render anything.

cimigree (approved):
Ok looks good! One color change and then good to go I think.

ErikSin:
> Ok looks good! One color change and then good to go I think.

sorry, which color change?

src/frontend/screens/Observation/index.tsx
```diff
@@ -159,6 +174,8 @@ const styles = StyleSheet.create({
   time: {
     backgroundColor: LIGHT_GREY,
```
cimigree: I think This LIGHT_GREY is supposed to be VERY_LIGHT_GREY

cimigree:
Sorry! I don't know why my comment didn't come through. Should be there now.

src/frontend/screens/ObservationMetadata.tsx
ErikSin: Right, I agree and attempted that initially. The problem is that 2 of the properties are react dependent. The `label` is a text, but it has to be translated, so we need to use `formatMessage` which requires the `useIntl` hook. Hooks have to be used inside a react component. The second one is the icon is a `ReactNode`.

I attempted to seperate those things,  so the object would be 
`{label: MessageDescription;
      value: number | undefined | string;
      unit: string;}`
 
 Then the REact component would look like this:
 ```ts
{filteredListData.length > 0 && (
        <FlatList
          data={filteredListData}
          renderItem={({item, index}) => {
            return (
              <View>
                  {determineIcon(item.label)}
                  {formatMEssage(item.label)}
                  {value}
              </View>
            );
          }}
        />
      )}
    </View>
  );
};

function determingIcon(label){
    //some switch logic to determine icon
}

```

We could do it this way, but now we are not co-locating the label with the icon, so it feels a little more convoluted...I suppose its more testable, but less clear?...im not sure

cimigree: That makes sense — I see your point now about formatMessage and the icon being tied to the component, so thanks for explaining.

I can see how separating the label/icon from the value/unit could make it more testable, but also harder to trace visually when reading the code and would be more "convoluted." So I get what you mean about clarity vs testability. I generally like things less convoluted and if we won't gain much from testing, how you did it makes a lot of sense.

So I am happy, as always, to go with what feels right to you. This was never a blocking thing in any way.

ErikSin: I appreciate your opinion. i know it wasnt blocking, but i just wanted to make sure that we agreed before i moved forward! Im going to go ahead and merge it. I also realize that we should write e2e test for this, but i will do that in a follow up pr

PR #1072, opened 2025-04-01
chore: clean up createPersistedState file

achou11:
Ideally this whole file ceases to exist given our shift in how we set up persisted zustand state. Can't do that just yet since `usePersistedDraftObservation` relies on this file too much to make it trivial to remove.

This PR does the following:

- Removes the exported `createPersistedStore()` function. Nothing's using this anymore.

- Adds a deprecation notice for the `createPersistedState()` function. Only the old draft observation hook is using this and I don't expect anything to accidentally use this, but doing just for clarity.

### DIFF
--- a/src/frontend/hooks/persistedState/createPersistedState.ts
+++ b/src/frontend/hooks/persistedState/createPersistedState.ts
@@ -1,4 +1,4 @@
-import {StateCreator, create, createStore} from 'zustand';
+import {StateCreator, create} from 'zustand';
 import {
   StateStorage,
   persist,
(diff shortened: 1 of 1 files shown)


src/frontend/hooks/persistedState/createPersistedState.ts
```diff
@@ -9,14 +9,8 @@ import {MMKV} from 'react-native-mmkv';
...
 export const storage = new MMKV();
 
-type PersistedStoreKey =
-  | 'MapeoLocale'
-  | '@MapeoDraft'
-  | 'MapeoTrack'
-  | 'Passcode'
-  | 'ActiveProjectId'
-  | 'Settings'
-  | 'MetricDiagnosticsPermission';
```
achou11, line 19: decided to remove the keys that are no longer relevant here and only keep the one that's actually used in the app (`@MapeoDraft`)

ErikSin (approved):

PR #1073, opened 2025-04-01
chore: rely less on direct access to singleton mmkv storage instance

achou11:
I identified various places where there's an "impure" dependency on being able to access a globally defined singleton of the mmkv storage instance. Most notably:

1. the zustand store creators
2. the metrics-related classes
3. the essentially deprecated `createPersistedState` function in `hooks/persistedState/createPersistedState.ts`

This originally started as an endeavor to address (1), but I've also added some changes that affect (2). (3) is tougher, so there are no major changes for how that's set up currently (aside from updating variable names and import locations).

This PR introduces the following changes:

- Moves the instantiation of the mmkv instance and the zustand state storage interface to the constants file. Ideally these would be instantiated in `App.tsx`, but how the `usePersistedDraftObservation()` hook is set up is impure and relies on direct access to a singleton instance of the zustand state storage. I've added todos to indicate that these variables should eventually be moved to `App.tsx` once we start using the new draft observation hook implementation.

- Updates the zustand store creators to require the `persist` option. When `persist` is `true`, a `storage` option is also required. This allows us to test the persistence behavior more easily, as we can supply a conforming in-memory storage instance instead of having to mock/observe external modules.

  - adds basic tests related to persistence for each

- Updates the metrics-related classes to accept a `storage` constructor option and converts several standalone functions as private methods in order to access the `storage` when it's not a global singleton.


src/frontend/metrics/AppDiagnosticMetrics.ts
achou11, line 1: admittedly not the most confident about these changes, mostly because I don't really know how to verify that I didn't break anything 😅 mostly just moving functions around but would be nice to have some confirmation...

src/frontend/metrics/DeviceDiagnosticMetrics.ts
achou11, line 1: same comment about not being able to easily confirm that the changes here don't affect behavior

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
```diff
@@ -43,13 +46,17 @@ export function convertPosition(
...
 
-  if (persist) {
+  if (opts.persist) {
     instance = createStore(
-      createPersistedState(createEmptyStoreState, {name: '@MapeoDraft'}),
+      createPersistedState(createEmptyStoreState, {
+        name: '@MapeoDraft',
+        storage: createJSONStorage(() => opts.storage),
+        version: 0,
+      }),
```
achou11, line 59: this isn't being used by anything in the app yet, but if it were, it would've been not working properly prior to this PR 😅

ErikSin:
To me this feels like we are adding complication and extra boilerplate to what we had previously agreed upon. We’re expanding the API surface to accommodate injecting a store, even though we know the store should always be the same. While I understand this allows us to test whether persistence is actually working, injecting a store just to test persistence feels like we’re testing the internals of zustand’s persist middleware — not our own logic.

Our test are to ensure that our application's (React) state management behaves correctly in the context of our app. We're not necessarily testing how the persistence mechanism works internally. If our persistance logic was more complicated I could see an argument to add these test. But the logic simply just usea the persist middleware directly as is. So I dont think adding extra complication to test that is necessary at this time.

I also don't think the use of a singleton is inherhently bad and necessarily means leaky implementation. Mainly because that is how zustand does things under the hood. So i don't think we need to avoid using a singleton if zustand is able to (and designed) to rectify the singleton and react state. were adding unnecessary complication to deal with a hypotethical that I don't think will cause a problem.

achou11:
yeah that's a reasonable evaluation. I did this under the pretense that it was relatively straightforward to implement and would make testing easier (which it does!), but if we don't think the tests it enables are useful then agree that it's not super necessary.

achou11:
Ah actually I just remembered that another reason for doing this was to work towards ensuring test isolation, which I think is going to be relevant for introducing integration tests via #910 .

Will re-open this if that's the case, haven't gotten far enough to know if it's actually necessary

achou11:
Yeah given that MMKV [injects](https://github.com/mrousavy/react-native-mmkv/blob/v2.12.2/src/MMKV.ts#L150-L152) a proper [mock](https://github.com/mrousavy/react-native-mmkv/blob/v2.12.2/src/createMMKV.mock.ts), I'm pretty sure we're going to run into test isolation issues. As you said, the existence of a singleton isn't a problem, but the implicit access to it by a variety of things makes it harder to mock out and test because there's no way to reset the instance.

Happy to update the PR to elaborate on this reasoning. 

EDIT: there might be a way of stubbing out the existing global and doing that each time between tests, which would address the test isolation issue without needing the changes in this PR (which doesn't currently address all necessary areas of access anyways). can look into that as I get deeper into the integration test work.

ErikSin:
> Yeah given that MMKV [injects](https://github.com/mrousavy/react-native-mmkv/blob/v2.12.2/src/MMKV.ts#L150-L152) a proper [mock](https://github.com/mrousavy/react-native-mmkv/blob/v2.12.2/src/createMMKV.mock.ts), I'm pretty sure we're going to run into test isolation issues. As you said, the existence of a singleton isn't a problem, but the implicit access to it by a variety of things makes it harder to mock out and test because there's no way to reset the instance.
> 
> Happy to update the PR to elaborate on this reasoning.
> 

No, I would be curious to understand more. Im open to these changes if its going to make testing easier. I think i just dont fully understand the architecture of mmkv and how this would help!

PR #1076, opened 2025-04-02
refactor: invites

ErikSin:
Creating as a draft so we can review together.

The API did not change much. SO im wondering if there are certain things I am missing that would help with error handling. As of now, I refactored the bottom sheet to use our new architecture (this is unrelated to the Api change). The only significant change so far is that the user is navigated away from the invite screen when the `invite.state` is updated.


src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    mapeoApiInvite.addListener('invite-updated', navigateBasedOnInviteState);
+
+    return () => {
+      mapeoApiInvite.removeListener(
+        'invite-updated',
+        navigateBasedOnInviteState,
+      );
+    };
+  }, [mapeoApiInvite, inviteId, navigation]);
```
ErikSin: this is the most significant change. Now we are listening for updates to the invite state, and navigating accordingly. More specifically, if the invite gets cancelled, the user is navigated to the cancel screen. Any other state that is not `pending` the user is navigated away from this screen (in order to avoid the user getting stuck on this screen)

ErikSin:
Also, just want to confirm, when a receiver accepts an invites all the other invites that the receiver had in the queue get cleared, correct?

RangerMauve:
Affirmative, pending invites are deleted on accept. https://github.com/digidem/comapeo-core/blob/main/src/invite-api.js#L380

src/frontend/screens/Invites/InviteReceived.tsx
RangerMauve: Does this also show error messages?

ErikSin: I don't know if I am understanding your question fully. But the invite itself doesn't actually expose an error. It has an `error` state, but it doesn't have a message. We catch errors (and send them to sentry) when there is an error accepting or rejecting an invite.

achou11: an invite with `state: 'error'` also has an `error: Error` field (see https://github.com/digidem/comapeo-core/blob/6d88d68cf2d505fb6c6d8af2d4b73a45a2a10682/src/invite/invite-api.js#L35), so the actual error could be exposed. don't remember what the designs specified though.

ErikSin: oh interesting! We dont ever expose errors to the user. I will send the error to sentry though

RangerMauve: Reporting to sentry could suffice then. If we get errors during the next training would it be possible to get the sentry logs for those devices?

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1076/commits/3cbd895e41b2c86f119f2ebf10f0bc7d9759f869 and https://github.com/digidem/comapeo-mobile/pull/1076/commits/9feabb133f647644317070546a8b535673b2dfb4
So now we are sending that error to sentry and popping up the error bottom sheet. Im not 100% sure if we need to show the error bottom sheet. Like we don't ever expose the errors to user, so it might just be the same to clear the invite. Perhaps that is a question for design

ErikSin: > Reporting to sentry could suffice then. If we get errors during the next training would it be possible to get the sentry logs for those devices?

yes sentry logs all the device info along with the errors

ErikSin:
@RangerMauve and @achou11 Are there any other big picture/architectural things that we should be considering? Or does this suffice for now?

RangerMauve:
IMO this should be good 👍

achou11:
> IMO this should be good 👍

Agreed! Can't see any other notable changes to consider right now

ErikSin:
Thanks! I have tested this locally (with 2 devices), and it seems to be working. I will create dev builds so other people can confirm. In the mean time, should go ahead an release a new version of core with this new API?

Issue #1047, opened 2025-03-18
Map/ Camera/Tracks UI Update


ximenabb:
**Description**
User story : As a mobile user, I want to be able to start collecting observations on my own, and add collaborators later, without losing observations I have already collected.

**Feature spec**
https://www.notion.so/digidem/Invite-devices-to-default-project-1791b08162d580889f78f08b034b710d?pvs=4

**User requirements**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1881b08162d580e58a62000c7772d608&pvs=4

3 possible states:
Solo UX = User mapping on their own without a named project, no collaborators invited
Coordinator UX = User decided to name a project and invite collaborators. OR user was invited to a project and the were selected as Coordinator
Participant UX = User invited by a Coordinator (set this user as Parcipitant) and is mapping in a team on a named project

See how this affects the Map/ Camera/Tracks views

**Map View**

<img width="225" alt="Image" src="https://github.com/user-attachments/assets/5c0b8fe5-ec99-463d-8bd8-2cb0f3f3f535" />

<img width="228" alt="Image" src="https://github.com/user-attachments/assets/7c5132d4-4b17-4865-a40e-700c3cc65f3d" />

- [ ] Remove Exchange button; do not display
- [ ] Solo UX- Introduce title of default project: “My Solo Project”
- [ ] Coordinator & Participant UX-  Introduce title of default project: “[Current Project Name]”
- [ ] Design titles as shown on Figma
- [ ] Updates GPS Accuracy UI and introduces Device Icon
- [ ] GPS opens current GPS screen
- [ ] Device Icon opens updated Menu screen
- [ ] All other current UI remains the same.

**Camera View**

<img width="227" alt="Image" src="https://github.com/user-attachments/assets/9c4851f9-55a8-4358-a76e-9f424c65ceae" />

<img width="230" alt="Image" src="https://github.com/user-attachments/assets/2d7ff151-af15-456b-a201-b9f1ab4e74be" />

- [ ] Same as Map View

**Tracks View**

<img width="226" alt="Image" src="https://github.com/user-attachments/assets/7fd6123e-3173-495c-854c-2464a947f05d" />

<img width="231" alt="Image" src="https://github.com/user-attachments/assets/5d8f71f2-799b-4639-8dd4-4ed4750e0027" />

- [ ] Same as Map View

**Design**
[...]

PR #1078, opened 2025-04-02
feature: default project observations list

### LINKED ISSUE #1045: Observation List Update
**Description**
User story : As a mobile user, I want to be able to start collecting observations on my own, and add collaborators later, without losing observations I have already collected.

**Feature spec**

**User requirements**

3 possible states:
Solo UX = User mapping on their own without a named project, no collaborators invited
Coordinator UX = User decided to name a project and invite collaborators. OR user was invited to a project and the were selected as Coordinator
Participant UX = User invited by a Coordinator (set this user as Parcipitant) and is mapping in a team on a named project

**To Do**
### Solo UX
Image

- [x] Remove all previous UX copy
- [x] Remove Back arrow button
- [x] Header displays default name for all default projects: “My Solo Project”
- [x] Introduce Device Icon which opens updated Menu screen
- [x] Introduce copy indicating this device’s role in the current project. For default projects, the UI displays: “You’re mapping on your own.”
- [x] Design “You’re mapping on your own.” as shown on figma

### Coordinator UX

Image

- [ ] Remove all previous UX copy
- [ ] Remove Back arrow button
- [ ] Header displays default name for all default projects: “[Current Project Name]”
- [ ] Introduce Device Icon which opens updated Menu screen
- [ ] Introduce  copy indicating this device’s role in the current project. For new projects created by this device, the UI displays: “You’re a coordinator on this project.”
- [ ] Design “You’re a coordinator on this project.” as shown on figma

### Participant UX

Image

- [ ] Remove all previous UX copy
- [ ] Remove Back arrow button
- [ ] Header displays default name for all default projects: “[Current Project Name]”
- [ ] Introduce Device Icon which opens updated Menu screen
- [ ] Introduce  copy indicating this device’s role in the current project. For new projects created by this device, the UI displays: “You’re a participant on this project.”
- [ ] Design “participant on this project.” as shown on figma

**Design**

cimigree:
closes #1045 
## Changes
- Updated Map, Camera and ObservationsList headers to show "My Solo Project" or named project and a device icon.
- Added a ProjectCard row that shows user role (“You’re mapping on your own.”) in ObservationsList and empty observation list.
- Had to add a suspense wrapper to the HomeHeader in the Map project tab so the map would reload when the app goes from background to foreground
- Adjusted E2E tests for new headers, icons, and card text.

**Note:** This contains some pieces from[ 1047](https://github.com/digidem/comapeo-mobile/issues/1047), but does not include the new GPS button yet

---
## Screenshots
<image src="https://github.com/user-attachments/assets/0bbb23bc-1af7-43e0-9e2a-e9a9ae42e8ef" width="250" />
<image src="https://github.com/user-attachments/assets/2ae4c3f4-cd7e-4e0b-aa30-820001959db9"  width="250"/>
<image src="https://github.com/user-attachments/assets/97e5d160-e6e5-47c1-a798-f02c7858a270" width="250" />
<image src="https://github.com/user-attachments/assets/faa5bd0a-b4ce-4cb2-b1f0-c24e34356cbb" width="250" />
<image src="https://github.com/user-attachments/assets/f162f4e8-6a5b-4a49-b0ea-4d2792ad4f92" width="250" />
<image src="https://github.com/user-attachments/assets/e31297d0-864f-41fa-ae6d-2230d224bb36" width="250" />
<image src="https://github.com/user-attachments/assets/52a504ab-55dd-4f75-9318-5789fd844ff2" width="250" />


src/frontend/Navigation/Tab/index.tsx
```diff
@@ -59,13 +55,22 @@ export const HomeTabs = () => {
...
         <Tab.Screen
           name="Map"
           component={MapScreen}
           options={{
-            header: props => <HomeHeader {...props} openDrawer={openDrawer} />,
+            header: props => (
+              <React.Suspense fallback={<Loading />}>
+                <HomeHeader {...props} openDrawer={openDrawer} />
+              </React.Suspense>
+            ),
```
ErikSin: Just looking at figma, these header should all be the same (with a different background color). Also technically they should all be wrapped with a `suspense`. Its working now because they map screen loads the data into the cache, so when navigating to the other tabs, the suspense is not called again. But this could lead to potential bugs if the cache ever gets cleared.

src/frontend/screens/ObservationsList/index.tsx
```diff
@@ -34,6 +30,18 @@ const m = defineMessages({
...
+    defaultMessage: 'You’re mapping on your own.',
+  },
+  coordinator: {
+    id: 'observationsList.ObservationListHeader.coordinator',
+    defaultMessage: 'You’re a coordinator on this project.',
+  },
+  participant: {
+    id: 'observationsList.ObservationListHeader.participant',
+    defaultMessage: 'You’re a participant on this project.',
+  },
```
ErikSin: these are not being used.

src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
```diff
@@ -50,6 +52,13 @@ export const ObservationEmptyView = ({
           text={t(m.backButton)}
         />
       }>
+      <View
+        style={{
+          width: '110%',
```
ErikSin, line 57: out of curiosity, why 110?

src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -1,61 +1,89 @@
...
-  navigation,
-  openDrawer,
-}: BottomTabHeaderProps & {openDrawer: () => void}) => {
+export function HomeHeader({openDrawer}: HomeHeaderProps) {
   const insets = useSafeAreaInsets();
+  const {formatMessage} = useIntl();
+  const {data, isRefetching} = useProjectSettings();
+
+  const projectName =
+    !isRefetching && data?.name ? data.name : formatMessage(m.mySoloProject);
```
ErikSin: I don't think we should use `isRefetching` here. It refetches everytime the screen mounts, so its going to cause solo project to flicker if they are not on a solo project

ErikSin (changes requested):
You are checking whether it is a "solo" project in 2 different ways:
1. in the `projectCard` you are determining whether it is a solo project based on the number of projects
2. in the headers, you are basing it off whether there is a project name.

In our conversation we discussed that we would use method 2.

Also, this can be done in a follow up PR, but we also discussed moving the menu away from the animated view, and just making it part of the stack navigation (with a layout navigation to slide across the screen from the right)

src/frontend/screens/ObservationsList/index.tsx
cimigree: I am confused. I think they are used in lines 55 - 66?

src/frontend/sharedComponents/HomeHeader.tsx
cimigree: ok. I wasn't sure. I will remove.

cimigree:
> You are checking whether it is a "solo" project in 2 different ways:
> 
> 1. in the `projectCard` you are determining whether it is a solo project based on the number of projects
> 2. in the headers, you are basing it off whether there is a project name.
> 
> In our conversation we discussed that we would use method 2.
> 
> Also, this can be done in a follow up PR, but we also discussed moving the menu away from the animated view, and just making it part of the stack navigation (with a layout navigation to slide across the screen from the right)

Ok sure. Will change.
Yes the menu is going to be done in a follow up PR.

src/frontend/screens/ObservationsList/ObservationsEmptyView.tsx
cimigree: I wanted it to be as wide as it is in the ObservationList, but the Empty View uses the ScreenContentWithDock which has a padding of 20 that I can't remove. So the 110% balances that out.

src/frontend/Navigation/Tab/index.tsx
cimigree: I swear this changed since last week?! I see that it is the same now, but I swear it didn't used to be. Ok. I will work on this but it will take a bit.

src/frontend/screens/ObservationsList/index.tsx
ErikSin: I think this is a duplicate. They are being using in line 55-66 of 'projectCard' not in this file

src/frontend/Navigation/Tab/index.tsx
ErikSin, line 1: it would just be slightly more streamline if you set the `Tab.Navigator` Header Option to be the Home Header with a transparent background and no bottom line, and then only overwrote it in the observation list. That way you only have to write it twice instead of 3 times

src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -1,61 +1,102 @@
...
-        style={{marginRight: 20}}
-        onPress={openDrawer}
-        testID="drawer-icon-home"
-        accessibilityLabel="Open Navigation Drawer"
-      />
+    <View
+      style={[
+        styles.container,
+        {
+          paddingTop: insets.top,
```
ErikSin: The padding looks a little uneven, looks like there is more space below. perhaps `justifyContent:'center'` (that is just a guess)

src/frontend/screens/ObservationsList/ProjectCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+      RoleIcon = NoProjectIcon;
+      break;
+  }
+
+  return (
+    <View style={styles.card}>
+      <RoleIcon width={20} height={20} />
+      <BodyText testID="OBS.card-text" variant="smallMeta">
+        {userRoleText}
+      </BodyText>
```
ErikSin: the text is overflowing a bit (just ever so slightly).:
image

ErikSin (approved):
some slight styling changes, which are blocking, but I don't think require a re-review when they are fixed so approving!

src/frontend/sharedComponents/HomeHeader.tsx
ErikSin: image

PR #1080, opened 2025-04-02
refactor: invites

ErikSin:
- refactored invites bottom sheet to use the new bottom sheet architecture.
- Created `InvitesListener` component. This component holds all the invites from `useManyInvites`. `useManyInvites` is updated when invites are updated or invites are received. These updates to `useManyInvites` happen in `useInvitesListeners()`, which listens for these events and invalidate the tanstack query cache. Two major architectural decision were made explicitly:
1.  I opted for a component, which returns null, instead of a hook, to avoid unncessary re-renders. the changes in invite state cause navigation events so it is not necessary to re-render the react state tree on state updates. As well, this listener is used in the Navigation Layout which does not accept hooks.
2. We are using `useManyInvites` instead of just using the `invite-updated` and `invite-received` event listeners because we want to keep track of ALL the invites so we can queue the invites. In other words if a user receives an invite and declines it, if there is another invite in the queue, we then want to show that invite. useManyInvites allows us to keep track of ALL the invites, not just the most recently received/updated.
- Created a `useListenToInviteStateUpdate` hook. This listens to any updates to the current invite the user is interacting with. Therefore we can reactively listen to when the invite gets cancelled, or listen for any errors. This also solves the problem of the user getting stuck on the invite screen when there is an error with an invite (as was happening before).

*** to note:
- 10 of the files changed are deleted file, and 7 of the files are associated with using a tarballs for the dependencies, so there are only 16 "files changed" that need to be reviewed


src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -441,11 +446,19 @@ export const createDefaultScreenGroup = ({
         name="AudioSavedBottomSheet"
         component={AudioSavedBottomSheet}
       />
+      <RootStack.Screen name="InviteReceived" component={InviteReceived} />
+      <RootStack.Screen
+        name="InviteSuccessfullyJoined"
+        component={InviteSuccessfullyJoined}
+      />
+      <RootStack.Screen name="InviteCancelled" component={InviteCancelled} />
```
achou11: an absolute nit but i believe core spells it using one `l` i.e. `canceled`.

```suggestion
      <RootStack.Screen name="InviteCanceled" component={InviteCanceled} />
```

the practical argument i'll pose is that the consistency helps for grepping through code 😄

src/frontend/screens/Invites/AlreadyOnProject.tsx
```diff
@@ -0,0 +1,97 @@
...
+});
+export const AlreadyOnProject = ({
+  route,
+  navigation,
+}: NativeRootNavigationProps<'AlreadyOnProject'>) => {
+  const {formatMessage} = useIntl();
+  const projectId = useActiveProjectId();
+  const {
+    data: {name: currentProjectName},
+  } = useProjectSettings({projectId: projectId!});
```
achou11: this should be using `useActiveProject()` which guarantees the existence of the `projectId` i.e.

```suggestion
  const {projectId} = useActiveProject();
  const {
    data: {name: currentProjectName},
  } = useProjectSettings({projectId});
```

a bit confusing for now but one of mentioned items to clean up in #1081

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -0,0 +1,149 @@
...
+  const acceptInvite = useAcceptInvite();
+  const rejectInvite = useRejectInvite();
+  const {setActiveProjectId} = useActiveProjectIdActions();
+  const projects = useManyProjects();
+
+  useListenToInviteStateUpdate(inviteId);
+
+  function accept() {
+    if (projects.data.length > 1) {
+      navigation.replace('AlreadyOnProject', {
```
achou11: based on the screen name alone, took me a moment to realize that this is for when you're on an existing project and attempting to join a new project. i initially thought this is referring to some UI that is explaining that you're already on the project that you're being invited to.

think it would be helpful to rename that screen so that difference is clearer.

src/frontend/screens/Invites/InviteSuccessfullyJoined.tsx
```diff
@@ -0,0 +1,80 @@
...
+    id: 'screens.InviteSuccess.youHaveJoined',
+    defaultMessage: 'You have joined {projectName}',
+  },
+  goToSync: {
+    id: 'screens.InviteSuccess.goToSync',
+    defaultMessage: 'Go To Sync',
+  },
+});
+
+export const InviteSuccessfullyJoined = ({
```
achou11: sorry for another nit but `InviteSuccessfullyAccepted` seems like a more accurate naming of what this is reflecting. maybe it's just me but I read this as someone tried to join an invite, which doesn't make sense to me 😅

src/frontend/screens/Invites/LeaveProject.tsx
```diff
@@ -0,0 +1,205 @@
...
+              onError: err => {
+                Sentry.captureException(err);
+                navigation.replace('ErrorBottomSheet');
+              },
+            },
+          );
+        },
+        onError: err => {
+          Sentry.captureException(err);
+          navigation.replace('ErrorBottomSheet');
```
achou11, line 108: out of curiosity, can you explain why this is using `replace`? maybe i'm missing something about how this is set up in the navigation...

src/frontend/sharedComponents/InvitesListener.tsx
```diff
@@ -0,0 +1,26 @@
+import {useManyInvites} from '@comapeo/core-react';
+import {useEffect} from 'react';
+import {isEditingScreen, isInviteScreen} from '../lib/screenNameChecks';
+
+export const InvitesListener = ({
```
achou11: is there any way to make this more descriptive?

src/frontend/lib/screenNameChecks.test.ts
achou11, line 1: worth adding some similar tests for `isInviteScreen`?

src/frontend/hooks/useListenToInviteStateUpdate.ts
```diff
@@ -0,0 +1,37 @@
...
+          projectName: invite.projectName,
+        });
+        return;
+      }
+
+      if (invite.state === 'error') {
+        Sentry.captureException(invite.error);
+        navigation.replace('ErrorBottomSheet');
+        return;
+      }
```
achou11: does this conflict with cases where where responding to an invite (e.g. accepting or rejecting) fails. seems like this is also being called in those cases

achou11 (changes requested):
haven't tested this on device just yet but overall these changes are looking great!

src/frontend/sharedComponents/InvitesListener.tsx
ErikSin: im not sure, do you have any suggestions?

src/frontend/hooks/useListenToInviteStateUpdate.ts
ErikSin: oh yeah, your right...hmm im not sure which one I should take out. What do you think?

ErikSin: i guess this listener listens for all errors, while responding to an invite only deals with the errors related to those mutations. So we should take out the accepting and rejecting navigation events and let. the listener deal with it

achou11: i think it makes sense for this hook to handle states that can happen "in the background" due to the something from the invitor's side, which in this case I think is only the `canceled` state? if that's the case, you rename this hook to be more specifically tied to handling that explicit state.

so i would say remove this check here and leave the check in the other places where responding to the invite happens

achou11: > i guess this listener listens for all errors, while responding to an invite only deals with the errors related to those mutations. So we should take out the accepting and rejecting navigation events and let. the listener deal with it

hm i guess you have a point there. although i think that errors that occur while attempting to reject or accept the invitation are reflected in the invitation state as `error`, but not entirely sure about that i guess.

still, i think the "error" state for an invite can only be reached based on the invitee's actions i.e. when responding to the invite (based on a brief glance at the [state chart diagram](https://github.com/digidem/comapeo-core/blob/main/src/invite/StateDiagram.md)). feels like it'd be more appropriate to handle that where the interaction happens, as opposed to an effect here

ErikSin: hmm, but are there any other errors that we should be worried about? Or do errors only happen during `accept` or `reject`. I just don't want the user to get stuck on the invites screen if an error happens outside of those mutations

ErikSin: github feature request: turn comments into live chats

src/frontend/sharedComponents/InvitesListener.tsx
achou11: maybe something related to handling pending invites? something like `PendingInvitesListener` would be a slight improvement at least

src/frontend/hooks/useListenToInviteStateUpdate.ts
achou11: > hmm, but are there any other errors that we should be worried about? Or do errors only happen during accept or reject. I just don't want the user to get stuck on the invites screen if an error happens outside of those mutations

based on the implementation and the state chart in core, the invite enters the error state under the following scenarios:

- [an error occurs when the invitee tries to accept an invitation](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L97)

- [an error occurs when the invitee tries to reject an invitation](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L115)

- [an error occurs on the invitees side due to realizing that they are already in the project](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L67), [which can only happen when the invitee accepts the invite](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L96)

- [an error occurs when the invitee is waiting for project details but it times out](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L141), [which can only enter this state after accepting the invite](https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L96)

I feel decently confident about my reading of this and saying that yes, the error state for an invite can only occur based on the invitee's action and all of these errors are covered in the `onError` callbacks for the relevant actions from core-react (although we probably can't match/narrow the kind of error in those callbacks, i think)

src/frontend/screens/Invites/LeaveProject.tsx
ErikSin: Yeah, the decision to use replace everwhere was because I didn't want the user to get stuck on an invite if it was no longer `pending`. If I used `navigate` instead, the go back function would have allowed the user to navigate back to the invite, even in cases where invite is no longer pending. And then they would get stuck on the invite screen. So by using replace, it guarantees that the invite screen is no longer open and going back would just take them to the screens outside of navigation. AND if there was still a invitation in pending, it would just open again via the invites listener.

achou11:
Alright did some device-based testing. Ran into a couple of major issues. ~My guess is that they're purely client-side issues.~ EDIT: maybe I shouldn't jump to conclusions 😄 

Hopefully the way i've written the things I've tried is clear enough

✅ = outcome is expected
‼️ = outcome is unexpected
❓= not sure if outcome is expected or not

---

1. X receives invite from Y -> Y cancels -> canceled sheet shows ✅

2. X receives invite from Y -> Y closes app -> X declines -> error screen shows for X ❓

   - attempting to go back with X doesn't work: `ERROR  The action 'GO_BACK' was not handled by any navigator.` can't leave error screen ‼️

3. X receives invite from Y -> Y closes app -> X accepts invite -> error screen shows ✅

4. X receives invite from Y -> X accepts invite -> Y sees that X accepted invite ✅ , X enters pending state ✅ -> error screen shows for X ‼️ ->

    - From Y's point of view, X is now on project ✅
    - From X's point of view, X is not on Y's project ‼️

5. X receives invite from Y -> X rejects invite -> Y sees that X rejected invite ✅.

6. X create project -> X receives invite from Y -> X sees "already on project" screen ✅ -> X attempts to leave own project -> Y sees that X has accepted invite ✅ , X enters state of leaving own project to join Y's project ✅ -> error screen shows for X ‼️

    - From Y's point of view, X is now on project ✅
    - From X's point of view, X is not on Y's project ‼️

---

let me know if you're getting different results for any of these. can try again to confirm that what i've initially reported is reproducible.

src/frontend/screens/Invites/LeaveProject.tsx
achou11: haven't fully wrapped my head around it, but could this be contributing to the issue I'm seeing in https://github.com/digidem/comapeo-mobile/pull/1080#issuecomment-2776860075 related to the go back button not working?

my initial thought is that this seems fragile and relies on implicit knowledge that isn't entirely clear to me, but i guess i can't really come up with a much stronger suggestion at the moment

ErikSin:
>     2. X receives invite from Y -> Y closes app -> X declines -> error screen shows for X ❓

I think this is happening because X cannot properly decline the invite because the connection was lost. So technically it is an error? Its just nor meaningful for X. But we can solve this be having Y cancel the invite on close.

>        * attempting to go back with X doesn't work: `ERROR  The action 'GO_BACK' was not handled by any navigator.` can't leave error screen ‼️

This should be fixed now. It was caused by 2 error event happening (by the listener and the mutation as discussed above) Since there were 2 `replace` event happening, it replaces the bottoms sheet and the map


>     4. X receives invite from Y -> X accepts invite -> Y sees that X accepted invite ✅ , X enters pending state ✅ -> error screen shows for X ‼️ ->
>        
>        * From Y's point of view, X is now on project ✅
>        * From X's point of view, X is not on Y's project ‼️
>     6. X create project -> X receives invite from Y -> X sees "already on project" screen ✅ -> X attempts to leave own project -> Y sees that X has accepted invite ✅ , X enters state of leaving own project to join Y's project ✅ -> error screen shows for X ‼️
> 

For these 2 events, is it happening every time? If so can you console.log the error?

achou11:
> I think this is happening because X cannot properly decline the invite because the connection was lost. So technically it is an error? Its just nor meaningful for X. But we can solve this be having Y cancel the invite on close.

seems like the error screen is no longer showing in this case now, which i guess is reasonable?

> This should be fixed now. It was caused by 2 error event happening (by the listener and the mutation as discussed above) Since there were 2 replace event happening, it replaces the bottoms sheet and the map

yep seems like this is no longer an issue

> For these 2 events, is it happening every time? If so can you console.log the error?

So for (4), i got the following error from the invitee's side: `[TimeoutError: Timed out adding project]`, which i guess is a server side issue. the timeout defined in core is 10 seconds, which seems aligned with how long i was waiting until that error surfaced:

https://github.com/digidem/comapeo-core/blob/5b1e22720a2c4c41ba1c0ed18836b5bf60a00f6a/src/invite/invite-state-machine.js#L8

So I guess this could either be device-specific or some other issue in core? The invitee in my case is a relatively old Pixel 2 device. Can try with my newer beefier device to see if I can avoid the timeout error.

For (6), I'm now running in an issue where after invitee presses `Accept Invite`, the invite received sheet disappears and then reappears. i'm never navigated to the `Already on project` screen

achou11:
> For (6), I'm now running in an issue where after invitee presses Accept Invite, the invite received sheet disappears and then reappears. i'm never navigated to the Already on project screen

I cancelled the invite and I'm actually now in a rather confusing navigation loop on the invitee's side. Once the invitor cancelled:

1. the invitee's `invite received` sheet disappeared but I noticed that the backdrop shadow associated with sheet screen was coming in and out a few times (but no actual sheet)

2. then the `already on project` screen showed up!

3. then pressing `go back` brought me to the error sheet screen.

4. then i frantically pressed more buttons and eventually was able to get rid of the invite screens as a whole 😅

ErikSin:
> > For (6), I'm now running in an issue where after invitee presses Accept Invite, the invite received sheet disappears and then reappears. i'm never navigated to the Already on project screen
> 
> I cancelled the invite and I'm actually now in a rather confusing navigation loop on the invitee's side. Once the invitor cancelled:
> 
>     1. the invitee's `invite received` sheet disappeared but I noticed that the backdrop shadow associated with sheet screen was coming in and out a few times (but no actual sheet)
> 
>     2. then the `already on project` screen showed up!
> 
>     3. then pressing `go back` brought me to the error sheet screen.
> 
>     4. then i frantically pressed more buttons and eventually was able to get rid of the invite screens as a whole 😅

This is happening because i updated the screen names, but didn't update the "INVITE_SCREENS" constant

ErikSin:
> > > For (6), I'm now running in an issue where after invitee presses Accept Invite, the invite received sheet disappears and then reappears. i'm never navigated to the Already on project screen
> > 
> > 
> > I cancelled the invite and I'm actually now in a rather confusing navigation loop on the invitee's side. Once the invitor cancelled:
> > ```
> > 1. the invitee's `invite received` sheet disappeared but I noticed that the backdrop shadow associated with sheet screen was coming in and out a few times (but no actual sheet)
> > 
> > 2. then the `already on project` screen showed up!
> > 
> > 3. then pressing `go back` brought me to the error sheet screen.
> > 
> > 4. then i frantically pressed more buttons and eventually was able to get rid of the invite screens as a whole 😅
> > ```
> 
> This is happening because i updated the screen names, but didn't update the "INVITE_SCREENS" constant

I updated the typing so typescript should catch this now. I was able to recreate this issue, and now it is no longer happening on my devices

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/WaitingForInviteAccept.tsx
```diff
@@ -36,16 +36,18 @@ export const WaitingForInviteAccept = ({
...
+  React.useEffect(() => {
+    // cancels invite if app is closed or goes into background
+    const subscription = AppState.addEventListener('change', nextState => {
+      if (nextState === 'background' || nextState === 'inactive') {
+        cancelInvite();
+      }
+    });
+
+    return () => subscription.remove();
+  }, [cancelInvite]);
```
achou11: not sure this should happen when the app is backgrounded. doing this when  the app is closed makes sense to me though

achou11 (commented):
I've only been able to successfully accept an invite once out of the various times I've tried to so far 😬 (from either device I'm testing). Think this is could be more of a backend issue though (see https://github.com/digidem/comapeo-core/issues/1010).

However, in the cases where I do run into the error, the invitee's side sometimes ends up in a bad state that's surfaced in the UI:

https://github.com/user-attachments/assets/e9a11264-37c2-49f1-98ce-97e9b1e0152e

Some parts of the UI suggest that the invitee is actually on the project (like the sync screen), although we're missing key project info like the name. Other parts indicate that it's not on the project it was invited to (like the sidebar or the Your Team screen). Again, this might be more of a backend issue, although some of the issues might be related to fragile checks in the frontend.

---


Also a couple of minor UX issues I've noticed:

- awkward loading indicator appearance in the backdrop as invite sheets appear

    https://github.com/user-attachments/assets/f001b679-89c6-4489-888f-81f0900c9831

- in the following sequence:

  1. `X` invites `Y`.
  2. Y closes app
  3. Y reopens app.
  
  `X` is still on the screen about waiting for a response to the invite that was sent. However, `Y` no longer has access to this invite after (3), so `X` has to cancel and resend the invite for `Y` to receive it.
  
  Very minor, but could be place where `Y` automatically declines the invite when it closes the app. Not sure if that's actually desired UX but avoids this case of a "dangling" invite on `X`'s side.
  
  Again, very minor.

src/frontend/screens/Invites/LeaveProject.tsx
ErikSin: The alternative is to use `navigate`. That would stack the error screen on top of the invitations screen. But the error has a good chance that it means the invitation is no longer valid. And if that is the case, the user would go back to the invitation screen, and be stuck on it.

The way it is now guarantees that the invitation is only open if it is pending.

ErikSin:
>       Very minor, but could be place where `Y` automatically declines the invite when it closes the app. Not sure if that's actually desired UX but avoids this case of a "dangling" invite on `X`'s side.
>       Again, very minor.

Im not too worried because it seems like its not causing any weird states or error states. Yes, the invitor is stuck thinking they are waiting on a response. But it is really easy for them to cancel and try again. Like you said, its pretty minor, so I don't think its worth the complication of adding a listener and trying to decline when the app goes in the background

src/frontend/screens/Settings/ProjectSettings/YourTeam/ReviewAndInvite/WaitingForInviteAccept.tsx
ErikSin: yeah, you know what there actually is no way to detemine if the app is about to close (in android). `inactive` is only for ios, and android only has `background` or `active`. So i think if we do want to use this, we can only actually cancel it if it goes into the background

achou11: ah right. without adding this listener, what happens in the following sequence?

1. `X` invites `Y`
2. `Y` sees invite
3. `X` closes app

Does `Y` get notified that the invite has been canceled? Basically I'm wondering if the backend automatically cancels the invite this such that this extra listener isn't necessary. If it doesn't, I guess leaving this as is makes sense.

i'm working from elsewhere and forgot to bring to a second device to test this 🙃

ErikSin: I will test it right now, but I don't think the backend handles it!

ErikSin: yeah, it doesnt deal with it. So I think we should cancel it if the app goes into the background

src/backend/patches/@comapeo+core+3.0.0.patch
achou11, line 1: non blocking but I generally prefer to use the `--append` flag when using patch package, which makes it easier to isolate patches and remove specific ones when no longer applicable.

https://github.com/ds300/patch-package/?tab=readme-ov-file#creating-multiple-patches-for-the-same-package

achou11 (approved):
Overall the code changes look good! Unfortunately haven't done thorough testing of the success flow due to lacking a second device today, but I trust that you've gone through that various times at this point 😄

ErikSin:
> Overall the code changes look good! Unfortunately haven't done thorough testing of the success flow due to lacking a second device today, but I trust that you've gone through that various times at this point 😄

I don't think im going to merge yet. I created an internal build, so Ill have people test that out first beforehand

achou11:
also would you mind doing the following to address patch-package warning about a version mismatch with `@comapeo/ipc`? (which this PR updates):

1. Run `npx patch-package @comapeo/ipc` (optionally using the `--append` flag)
2. Commit the new patch file
3. Remove the old patch file
4. Fix the relevant link to the patch file in the patch readme.

ErikSin:
Merging. The problem hasn't been fully solved yet, but it is an improvement so we are just going to build an RC and add more invite improvements later

Issue #1081, opened 2025-04-03
changes to consider or make now that `@comapeo/core-react` is more fully integrated


achou11:
Basically extracting a variety of potential tasks based on feedback in https://github.com/digidem/comapeo-mobile/pull/1038. Just going to consolidate these here for now and we can create separate items later if needed.

### 1. Avoid usage of hardcoded/vendored query keys.

Due to missing functionality in `@comapeo/core-react`, we still have some reliance on knowing the query keys used by it in order to do proper invalidation of relevant read hooks. Basically anything that references [`ROOT_QUERY_KEY`](https://github.com/digidem/comapeo-mobile/blob/0af0bfab515463d66caa67d4eb2bc4d390285033/src/frontend/constants.ts#L23). As of this writing, some hooks related to maps, invites, and sync are affected by this.

The ideal solution is to update `@comapeo/core` and/or `@comapeo/core-react` to provide functionality such that manual invalidation in the mobile app isn't necessary. 

### 2. Simplify [`ActiveProjectContext.tsx`](https://github.com/digidem/comapeo-mobile/blob/0af0bfab515463d66caa67d4eb2bc4d390285033/src/frontend/contexts/ActiveProjectContext.tsx).

Before the core react integration, the implemented provider ensured that there was an active project id and project api instance for the rest of the app to work with. the usage of this generally looked like this:

```tsx
const {projectId, projectApi} = useActiveProject();
```

After the major integration work, it's still technically doing the same thing but there's no need for it to hold+pass a project api instance, as that should is handled by using `useSingleProject()` throughout the app. now the consumption of that context looks like this:

```tsx
// 🟠 no more need for the projectApi field
const {projectId} = useActiveProject()

const {data: projectApi} = useSingleProject({projectId})
```

We still need the context for guaranteeing an active project id and making that available for the rest of the app. I'm proposing that the context be reduced to the following responsibilities:

  1. conditionally rendering until an active project id has been identified/selected
  2. pass the value from (1) down to the rest of the app.

The ideal usage of this context looks like the following for any component that deals with project-specific APIs:

```ts
const activeProjectId = useActiveProjectId()

const {data: projectApi} = useSingleProject({projectId: activeProjectId})
```

### 3. Reducing the number of "wrapper" hooks where appropriate

_Note: this one could be viewed as more stylistic so could be less urgent to address_

[...]

PR #1091, opened 2025-04-09
chore: use available sync hooks from @comapeo/core-react

achou11:
Towards #1081 (specifically simplifying the `useActiveProject()` hook).

There should be no user-facing changes with this, as the [implementation from core-react](https://github.com/digidem/comapeo-core-react/commit/d181ea7d280e185d06f4052a13ce109a56da3e24) is directly ported from the implementation that was here.

There are some methods related to sync that are used in the app that haven't been ported to core-react just yet, but wouldn't consider this a blocker since they don't rely on query invalidations right now.

Also adds a unit test for a helper function related to sync state.


cimigree (approved):
Everything from the old useSyncState seems to be removed, and I see the new useSyncState({ projectId }) from @comapeo/core-react is replacing it. The new test for getConnectedPeersCount covers the logic well as far as I can tell. I don’t see any stray references to the old sync code or manual query invalidations, so it all looks clean. I can’t actually test the Sync UI right now, but from reading through these diffs, the code looks straightforward and consistent.

PR #1093, opened 2025-04-10
feat:default project menu update

### LINKED ISSUE #1048: Menu Update
**Description**
User story : As a mobile user, I want to be able to start collecting observations on my own, and add collaborators later, without losing observations I have already collected.

**Feature spec**

**User requirements**

3 possible states:
Solo UX = User mapping on their own without a named project, no collaborators invited
Coordinator UX = User decided to name a project and invite collaborators. OR user was invited to a project and the were selected as Coordinator
Participant UX = User invited by a Coordinator (set this user as Parcipitant) and is mapping in a team on a named project

**To Do**

### Overall

- [ ] Rename **App Settings** to **Settings**
- [ ] Tapping the featured card also goes to Project Settings

### Solo UX

Image

- [ ] The menu now spans the width of the screen.
- [ ] Display Device Icon with [Device Name]
- [ ] Introduce “Current Project” featured card.
- [ ] Give card a background color : #e5f0ff
- [ ] Display default project name: “My Solo Project”
- [ ] Display current role in project: “You are mapping on your own.”
- [ ] Display View and Invite buttons
- [ ] Tapping on View opens
- [ ] Tapping on Invite opens
- [ ] Anchor list of additional menu option to the bottom of the screen
- [ ] Display Exchange, Settings, Data & Privacy, and About CoMapeo - Tapping on them opens to the usual CoMapeo screens

### Coordinator UX

Image

- [ ] The menu now spans the width of the screen.
- [ ] Display Device Icon with [Device Name]
- [ ] Introduce “Current Project” featured card.
- [ ] Give card a background color (different than Solo UX) : #fff5eb
- [ ] Display current Project Name: “[Current Project Name]”
- [ ] Display current role in project: “You are a coordinator on this project.”
- [ ] Display View and Invite buttons
- [ ] Tapping on View opens
- [ ] Tapping on Invite opens
- [ ] Anchor list of additional menu option to the bottom of the screen
- [ ] Display Exchange, Settings, Data & Privacy, and About CoMapeo - Tapping on them opens to the usual CoMapeo screens

### Participant UX

Image

- [ ]  The menu now spans the width of the screen.
- [ ]  Display Device Icon with [Device Name]
- [ ] Introduce “Current Project” featured card.
- [ ] Give card a background color (different than Solo UX, same as Coordinator) : #fff5eb
- [ ] Display current Project Name: “[Current Project Name]”
- [ ] Display current role in project: “You are a participant on this project.”
- [ ] **Only** display View button
- [ ] Tapping on View opens
- [ ] Anchor list of additional menu option to the bottom of the screen
- [ ] Display Exchange, Settings, Data & Privacy, and About CoMapeo - Tapping on them opens to the usual CoMapeo screens

**Design**

cimigree:
closes #1048 

Move Menu to a Stack Screen & Make Updates to It Per Design Specifications

**Reference**: [Invite devices to default project](https://www.notion.so/digidem/Invite-devices-to-default-project-1791b08162d580889f78f08b034b710d?pvs=4)

## Summary

- Replaces the old custom “drawer” overlay with a dedicated **Menu** stack screen that slides in from the right.
- Implements **Solo, Coordinator, and Participant** project role UX:
  - **Solo** or **Coordinator**: Shows “View” and “Invite” buttons in the project card.
  - **Participant**: Only shows “View” button.
  - changes color depending on if it is a named project
- Reorganizes the **Menu** code to rely on a single `MenuContent` component plus a short `MenuScreen.tsx`.
- Removes the old `useCloseDrawerOnBackPress` hook (no longer needed for stack navigation).
- I updated the text color to Dark grey because that is what it always seems to be... but that was just a guess!

## Changes

1. **New Menu Stack Screen**  
   - Added `MenuScreen.tsx` to our RootStack, using `animation: 'slide_from_right'`.
   - Pressing the device icon in the header navigates to `'Menu'` instead of opening the drawer.

2. **Project Card**  
   - Card showing the current project name and role that when clicked goes to Project Settings.
   - Role determined by added `useProjectRole` hook.

3. **Buttons Layout**  
   - When user is `'participant'`, only one button is shown, occupying half of the row (the second half is empty).
   - `'solo'` or `'coordinator'` sees both “View” and “Invite” side by side.

4. **Menu Items**  
   - Updated to use a local `<MenuItem>` helper that wraps `MenuListItem` with consistent padding/gap.
   - Anchored bottom items in the scroll view.
  
5. **Test changes**
   - I adjusted the Menu test and documentation
   - I commented out some things to make them work for now as Project naming is tricky at the moment. Left todos and notes to be able to put everything back. 
   - Will be able to add a bit more testing after project naming is easy
   - Changed the accessibility label for the menu to reference menu and not drawer (that is a bunch of files!)

**Trying to get this UI in so Exchange can be gotten to...**

## How to Test

1. **Open the app** and navigate to the main map or observation list.  
2. **Tap the device icon** in the header. Confirm the new menu slides in from the right.  
3. **Check the top project card**: 
   - If your device is on a “solo” project (unnamed), see two buttons (“View,” “Invite”). 
4. **Name your project**:
   - You can do this by clicking on Invite and then Create Or Join Project then Create a Project.
   - If you name the project, you remain “coordinator,” still see two buttons.  
7. **Invite someone to your project through usual flow**:
   - Participant will only see the “View” button half-width.  
8. **Press the back button**. Confirm the menu dismisses properly.
9. **Try the menu items**: Exchange, Settings, Data & Privacy, About. They should navigate as expected.

## Screenshots

<image src="https://github.com/user-attachments/assets/61ae6aca-f8e7-4847-93d6-3d95094497d5" width="250" />
<image src="https://github.com/user-attachments/assets/4b0683ef-522d-41a4-93dd-b862f66e28a9" width="250" />
<image src="https://github.com/user-attachments/assets/8010be7f-21bb-4b90-8924-f68d91c95ea3" width="250" />


src/frontend/screens/MenuScreen.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const navigation = useNavigation();
+
+  function closeMenu() {
+    navigation.goBack();
+  }
+
+  return (
+    <View style={styles.container}>
+      <MenuContent closeMenu={closeMenu} />
+    </View>
```
ErikSin: Your menu screen is just "forwarding" the menu content. So lets just get rid of menu content and put everyting on this screen. We previously had to do it that way because of the animation, but I don't think that is necessary anymore

src/frontend/hooks/useProjectRole.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (!role?.roleId) return 'solo';
+
+  if (role.roleId === COORDINATOR_ROLE_ID || role.roleId === CREATOR_ROLE_ID) {
+    return 'coordinator';
+  } else if (role.roleId === MEMBER_ROLE_ID) {
+    return 'participant';
+  } else {
+    return 'solo';
+  }
+}
```
ErikSin: this has 2 possible sources of truth. Technically `ProjectData` can come from one project, while the projectId and role can come from another project. I think it would be better to take projectId as the function param. And then based on that use all the relevant hooks directly inside this hook. The logic you have is sound, just where your getting the information can get a little messy

src/frontend/sharedComponents/Text/BodyText.tsx
```diff
@@ -31,7 +31,7 @@ export const BodyText = ({
   return (
     <RNText
       style={[
-        {color: BLACK, fontSize: fontSizeMap[variant || 'regular']},
+        {color: DARK_GREY, fontSize: fontSizeMap[variant || 'regular']},
```
ErikSin: This is changing ALL the `BodyText` in the entire app to dark grey...is that intentional?

src/frontend/sharedComponents/MenuContent.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+export function MenuContent({closeMenu}: MenuContentProps) {
+  const {data: deviceData} = useOwnDeviceInfo();
+  const {data: projectData} = useProjectSettings();
+  const role = useProjectRole(projectData);
+  const {formatMessage} = useIntl();
+  const {navigate} = useNavigationFromRoot();
+
+  const deviceName = deviceData?.name;
+  const projectName = projectData?.name || 'My Solo Project';
```
ErikSin: A semi similar concern to what I was saying about the hook. Here we are fetching the project name, but then the project name is also fetched in the hook. I know they are the both using the same hook to get the projectName, but it feels like were making calls that are so tied together, but abstracting them away so we need to call them twice.

I think we should update the hook to be `useProjectRoleAndName`. The 2 are used together often and it would make it more explicit that the role is based on the name. In the future when and if the role and name are no longer attached (which tbh i don't think they should be in the first place...), then we can explicitly seperate them.

src/frontend/sharedComponents/MenuContent.tsx
```diff
@@ -0,0 +1,287 @@
...
+    <View style={styles.container}>
+      <View style={styles.topBar}>
+        <View style={styles.deviceRow}>
+          <DeviceIcon width={32} height={32} />
+          <HeaderText variant="header4">{deviceName}</HeaderText>
+        </View>
+        <TouchableOpacity onPress={closeMenu} accessibilityLabel="Close Menu">
+          <CloseIcon size={32} />
+        </TouchableOpacity>
+      </View>
```
ErikSin: I think we should be leveraging React navigations header to be consistent. So creating this header as its own component and setting it at the `Stack/AppScreens.tsx` level

src/frontend/sharedComponents/Text/HeaderText.tsx
```diff
@@ -44,7 +44,7 @@ export const HeaderText = ({
       style={[
         {
           fontFamily: 'Rubik_500Medium',
-          color: BLACK,
+          color: DARK_GREY,
```
ErikSin, line 47: Same comment as the Body Text above

src/frontend/Navigation/Tab/index.tsx
```diff
@@ -16,84 +13,47 @@ import {WHITE} from '../../lib/styles';
...
-        <View style={styles.backdrop}>
-          <DrawerContent closeDrawer={closeDrawer} />
-        </View>
+    <Tab.Navigator
+      tabBar={TabBar}
+      screenOptions={{
+        tabBarShowLabel: false,
+        headerTransparent: true,
+        header: props => (
+          <React.Suspense fallback={<Loading />}>
```
ErikSin: Can we change the fallback to `null` instead of `Loading` here? Its causing 2 loaders when we first open the app which doesn't look that great

src/frontend/sharedComponents/MenuContent.tsx
```diff
@@ -0,0 +1,287 @@
...
+  return (
+    <MenuListItem
+      {...props}
+      paddingLeft={0}
+      paddingRight={15}
+      columnGap={15}
+      style={{paddingVertical: 10}}
+    />
+  );
+}
```
ErikSin: I don't think we should use this. I created the `MenuListItem` before we agreed on the architecture. It also is designed for a flat list.

I think to best follow our architecture guidelines I would create a share wrapper component:

```ts
const MenuWrapper = ({child, onPress})=>(
     <TouchableOpacity onPress={onPress} style={...someSharedStyle} >
           {child}
     </TouchableOpacity>
)

```
The focus here is to reduce the prop surface area, while still having consistent styling

ErikSin (changes requested):
Good first pass. I have a minor, but blocking concerns. The overall architecture is good, but I think we just need to refine the hooks.

Also a couple comments regarding component/prop forwarding. Just a reminder, we should try avoiding creating a component that just returns another component!

src/frontend/sharedComponents/Text/BodyText.tsx
cimigree: Yes, as I said in the summary. 90% of the time when I look at the text color, body or header, in FIGMA the text is Dark Grey (333333). Should we ask Aldo and/ or Sabella to confirm?

> I updated the text color to Dark grey because that is what it always seems to be... but that was just a guess!

src/frontend/sharedComponents/Text/HeaderText.tsx
cimigree: See above.

src/frontend/Navigation/Tab/index.tsx
cimigree: Ok I did it for the default and for the ObservationList. Hope that is what you meant. I never see it in either place so I wasn't sure.

src/frontend/sharedComponents/MenuContent.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1093/commits/847c6a554d7be919f14bed4ab7e64db2b35b2885

src/frontend/screens/MenuScreen.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1093/commits/847c6a554d7be919f14bed4ab7e64db2b35b2885

src/frontend/sharedComponents/MenuContent.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1093/commits/847c6a554d7be919f14bed4ab7e64db2b35b2885

src/frontend/hooks/useProjectRole.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1093/commits/847c6a554d7be919f14bed4ab7e64db2b35b2885

src/frontend/Navigation/Tab/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1093/commits/847c6a554d7be919f14bed4ab7e64db2b35b2885

cimigree:
I added a fontsize of 18 to the body text because that is what is in the specs for the FIGMA for the menu (and not bold and there is no way to override the bold in the HeaderText) as a medium variant. Hope that is ok.

src/frontend/sharedComponents/Text/BodyText.tsx
ErikSin: Yes, we should confirm with someone before we make a change like this!

cimigree: Ok I sent a message in Slack... Or is there a better way to suss this out?

src/frontend/sharedComponents/MenuList/MenuWrapper.tsx
```diff
@@ -0,0 +1,45 @@
...
+  StyleProp,
+  TouchableOpacityProps,
+} from 'react-native';
+
+type MenuWrapperProps = {
+  onPress?: TouchableOpacityProps['onPress'];
+  children: React.ReactNode;
+  style?: StyleProp<ViewStyle>;
+  testID?: string;
+  accessibilityLabel?: string;
```
ErikSin: `style` and `testID` are not being used, so I think we should take them out. (if we use them later it would be easy to add). `onPress` and `accessibilityLabel` should NOT be optional as they are always being used

src/frontend/sharedComponents/MenuList/MenuWrapper.tsx
ErikSin, line 1: Just for clarity I think we should not co-locate this with `sharedComponents/MenuList` as it is being used for something completely different.

Perhaps Call it `MainMenuItemWrapper` have it as a stand alone file in sharedComponents?

src/frontend/screens/MenuScreen.tsx
```diff
@@ -0,0 +1,221 @@
...
+    defaultMessage: 'View',
+  },
+});
+
+export function MenuScreen() {
+  const {formatMessage} = useIntl();
+  const navigation = useNavigationFromRoot();
+
+  const {projectId} = useActiveProject();
+  const {role, projectName} = useProjectRoleAndName(projectId);
```
ErikSin: yeah, this feels much cleaner and its clear that the role and project name are linked to the same project id

src/frontend/hooks/useProjectRoleAndName.ts
```diff
@@ -0,0 +1,31 @@
...
+  COORDINATOR_ROLE_ID,
+  CREATOR_ROLE_ID,
+  MEMBER_ROLE_ID,
+} from '../sharedTypes';
+
+type ProjectRole = 'solo' | 'coordinator' | 'participant';
+
+export function useProjectRoleAndName(projectId: string): {
+  role: ProjectRole;
+  projectName: string;
```
ErikSin: Perhaps a slightly better thing to do would be to return a "projectHeader" instead. This just implies that the `projectName` is "My Solo Project" when it is actually undefined. So it might be confusing to use.

Like I can imagine a developer trying to get the name of a project via this hook, and it would be confusing as its not actually always returning the name

so maybe the return type is 
```ts
{role:'solo', projectHeader: 'My Solo Project'} | {role:'coordinator' | 'participant', projectName: string}

```
and then perhaps rename it: `useProjectRoleAndDetails`

And then in the of the menu where you are using the `projectName` your could just do:
```ts
const projectInfo = useProjectRoleAndDetails(projectId)
<HeaderText>{projectInfo.projectName || projectInfo.ProjectHeader}</HeaderText>
```

ErikSin (approved):
Im approving, but i do think we should update the hook via my suggestion. Otherwise there are a couple of relatively small comments

src/frontend/Navigation/Tab/index.tsx
ErikSin: Yeah, i only noticed it on my really slow phone briefly when opening the app. But changing it to null should get rid of that problem!

src/frontend/hooks/useProjectRoleAndName.ts
ErikSin: or even better ALWAYS return the name... just have the name as undefined when role is "solo"

ErikSin: So it would be 
```ts
{
    role:'solo', 
    projectHeader: 'My Solo Project', 
    projectName:undefined
}
 | 
{
     role:'coordinator' | 'participant',
     projectName: string
}
````

PR #1104, opened 2025-04-23
chore: update @comapeo/ipc patch file name

achou11:
Forgot to mention doing this as part of #1080, which updated `@comapeo/ipc`


ErikSin (approved):
I didn't check it on my device but since all the CI is passing it seems like it working

PR #1105, opened 2025-04-23
chore(test): get basic react-testing-library setup working

achou11:
Towards #910 

This does not get us to a point where we can do "integration" tests (i.e. using a mocked backend) but brings us closer to being able to run DOM-based react-testing-library tests more consistently. ~For example, this PR introduces a rather incomplete (and currently failing) "component unit" test for the `GPSPill`, which I identified as a straightforward candidate that would benefit from having this kind of test (with some refactoring). See https://github.com/digidem/comapeo-mobile/pull/1105/commits/0417640c5dcc08b57d7f39b5b6f014dc5733f096 for the relevant changes.~ EDIT: moving these changes to a stacked PR for better isolation.

There are potentially some mocks that are missing but it's hard to tell without having tests that try to make use of the things that need to be mocked. Some of the current changes related to mocks were artifacts of when my goal was to try to get full integration tests set up, and I figured it would be good to include as part of the PR anyways.


src/frontend/lib/intl.ts
```diff
@@ -44,7 +44,10 @@ function getUsableLanguages(
     if (!hasAtLeastOneTranslatedString) continue;
 
     if (!isSupportedLanguageTag(languageTag)) {
-      if (process.env.APP_VARIANT === 'development') {
+      if (
+        process.env.APP_VARIANT === 'development' &&
+        process.env.NODE_ENV !== 'test'
+      ) {
```
achou11, line 50: got annoyed with this warning being logged when running the jest tests. This updates the logging condition such that it only logs when running the actual app (and not tests).

src/frontend/sharedComponents/GPSPill.test.tsx
achou11, line 1: this was mostly included to provide an idea of the kinds of tests that are enabled by the changes in this PR. can commit to writing out these tests more thoroughly if desired

src/frontend/sharedComponents/GPSPill.tsx
```diff
@@ -1,33 +1,19 @@
...
+type GPSPillProps = LocationStatusResult & {
   onPress?: () => void;
-  locationProviderStatus?: LocationProviderStatus;
 };
 
-export const GPSPill: FC<GPSPillProps> = ({
-  onPress,
-  locationProviderStatus,
-}) => {
-  const {locationState, fgPermissions} = useSharedLocationContext();
```
achou11: i identified this usage of the context as unnecessary and refactored this component to take the relevant information as props, which makes testing the component much more straightforward.

cimigree (changes requested):
I reviewed the code changes related to jest the best that I could. I am not sure at all what they do or how they work and what consequences they have, but I assume your research lead you to this configuration, and it doesn't break any of our current tests or our app (as far as I can tell!)
I tested it out and the GPS pill seemed to work as before.
As far as the example test goes, it is conceptually helpful, but its failing state makes it harder to evaluate the usefulness of the new setup in isolation.

Suggestion:

Either:
1. Remove the test and push it in a stacked PR, as you suggested.
Or: 
2. fix the test so it passes (even minimally), just to demonstrate the setup is working.

Your call — happy with either.

So, in sum, it’s hard to assess the “value” of this setup without a green test, since part of the benefit is psychological ("tests are working now!") as well as functional.

Still, I think it's worth merging with or without the GPSPill test, as long as that test doesn't block others or pollute CI (e.g., it's skipped or deleted for now).

Recommendation
I can approve it with one of the following...
Either remove the non-working test
Or mark it as test.skip(...)
Or fix the component or expectation to demonstrate green test

Once that’s done, this is ready to merge from my perspective.

achou11:
Yeah I think I'll opt for extracting the GPSPill-related changes and the test for it to a separate PR that's stacked on top of this.

achou11:
created a stacked PR for adding the GPS pill test here: https://github.com/digidem/comapeo-mobile/pull/1115

cimigree (approved):

PR #1113, opened 2025-04-24
chore: add menu as modal

ErikSin:
fixes a visual problem with animation of the menu. When it is under the `Rootstack.group` where `{presentation:'card'}`, it animates by moving the entire previous screen away, which is not the desired behaviour.

When it is under the `Rootstack.group` where `{presentation:'transparentModal'}`, it animates on top the the previous screen, which is the desired behaviour.
 
presentation:transparentModal (Correct)

https://github.com/user-attachments/assets/b80b334d-d02e-48d0-8594-d79f707f721e

presentation:card (incorrect)

https://github.com/user-attachments/assets/c60777fb-acfc-4e6b-9efa-499860bcd5c3

### DIFF
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -424,15 +424,7 @@ export const createDefaultScreenGroup = ({
         component={ObservationMetadata}
         options={{headerTitle: intl(ObservationMetadata.navTitle)}}
       />
-      <RootStack.Screen
-        name="Menu"
-        component={MenuScreen}
-        options={{
-          headerShown: true,
-          animation: 'slide_from_right',
-          header: () => <MenuHeader />,
-        }}
-      />
+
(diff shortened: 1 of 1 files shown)


cimigree (commented):
The reason I moved it had to do with the behavior after opening the modal from within it...
Do you notice the kind of jerking that happens when opening the submenus?

See screen recording of having the menu as a modal:

https://github.com/user-attachments/assets/31fd1fe9-40c2-40f5-b3ed-a2d6d83c66c0

And here it is Not as a modal -- much smoother.

https://github.com/user-attachments/assets/b252c8bb-ec7e-417c-aa01-497d24207fff


Is that jerking something that can be addressed some other way?
At this point, it is deciding which UI is preferable and that is up to you. I just wanted make sure you are aware of the consequences downstream.

ErikSin:
Closing as it seems like we are good with the current menu!

PR #1115, opened 2025-04-24
test: add basic component test for GPS pill component

achou11:
Stacked on #1105 

Provides an example of a isolated component unit test using React Testing Library.

Notable changes:

- Updates the `GPSPill` component to make it more testable:

  - removes the internal usage of a context in favor defining props for the relevant information
  - exposes the test ID and accessibility label as props instead of defining them internally.
      - test IDs should always be defined as props for shared components because otherwise, they're not shareable (as the IDs are supposed be unique).
      - accessibility labels are contextual so a shared component should not define it internally and exposing it as a prop allows for more flexible usage.
      
      
Still have some questions about the desired behavior of the component, mostly in the context of how it displays accuracy values.

### DIFF
--- a/src/frontend/sharedComponents/GPSPill.test.tsx
+++ b/src/frontend/sharedComponents/GPSPill.test.tsx
@@ -1,0 +1,45 @@
+import {render, screen} from '@testing-library/react-native';
+
+import {GPSPill} from './GPSPill';
+
+test.skip('searching status', async () => {
+  render(<GPSPill status="searching" />);
+
+  // TODO: Check icon color
+});
+
+test('error status', async () => {
+  render(<GPSPill status="error" />);
+
+  expect(screen.getByText('--')).toBeOnTheScreen();
+
+  // TODO: Check icon color
+});
+
+test('good status', async () => {
+  render(<GPSPill status="good" accuracy={1} />);
+
+  expect(screen.getByText('± 1 m')).toBeOnTheScreen();
+
+  // TODO: Check icon color
+});
+
+test('displayed accuracy', async () => {
+  // Handles integers
+  render(<GPSPill status="good" accuracy={10} />);
+  expect(screen.getByText('± 10 m')).toBeOnTheScreen();
+
+  // Handles negative accuracy elegantly
+  render(<GPSPill status="good" accuracy={-1} />);
+  expect(screen.getByText('± 1 m')).toBeOnTheScreen();
+
+  // Handles floats
+  render(<GPSPill status="good" accuracy={0.5} />);
+  expect(screen.getByText('± 1 m')).toBeOnTheScreen();
+
+  render(<GPSPill status="good" accuracy={5.25} />);
+  expect(screen.getByText('± 5 m')).toBeOnTheScreen();
+
+  render(<GPSPill status="good" accuracy={10.75} />);
+  expect(screen.getByText('± 11 m')).toBeOnTheScreen();
+});
--- a/src/frontend/sharedComponents/GPSPill.tsx
+++ b/src/frontend/sharedComponents/GPSPill.tsx
@@ -1,33 +1,29 @@
-import React, {FC} from 'react';
 import {StyleSheet, TouchableOpacity} from 'react-native';
-import {DARK_GREY, WHITE} from '../lib/styles';
-import {GpsErrorIcon, GpsSearchingIcon, GpsGoodIcon} from './icons';
-import {useSharedLocationContext} from '../contexts/SharedLocationContext';
-import {getLocationStatus} from '../lib/utils';
-import {BodyText} from './Text/BodyText';
 import {UIActivityIndicator} from 'react-native-indicators';
-import type {LocationProviderStatus} from 'expo-location';
+
+import {DARK_GREY, WHITE} from '../lib/styles';
+import {GpsErrorIcon, GpsGoodIcon, GpsSearchingIcon} from './icons';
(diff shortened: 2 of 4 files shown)


src/frontend/sharedComponents/GPSPill.test.tsx
```diff
@@ -0,0 +1,45 @@
+import {render, screen} from '@testing-library/react-native';
+
+import {GPSPill} from './GPSPill';
+
+test.skip('searching status', async () => {
+  render(<GPSPill status="searching" />);
+
+  // TODO: Check icon color
+});
```
achou11, line 9: there's no text that's rendered from what I can tell (instead it's the activity indicator). not really sure how to query that as of right now but open to adding an assertion if there's a clear way.

src/frontend/sharedComponents/GPSPill.test.tsx
```diff
@@ -0,0 +1,45 @@
+import {render, screen} from '@testing-library/react-native';
+
+import {GPSPill} from './GPSPill';
+
+test.skip('searching status', async () => {
+  render(<GPSPill status="searching" />);
+
+  // TODO: Check icon color
```
achou11, line 8: checking icon color is a bit tricky, mostly because of how the existing implementation decides to render the icon portion of the pill. first thought is to assign an appropriate accessibility role to the icons used, query for that, and then make an assertion on the styling. not sure if that will work given that one of the icons used is based on the vector icons library.

src/frontend/sharedComponents/GPSPill.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  // Handles floats
+  render(<GPSPill status="good" accuracy={0.5} />);
+  expect(screen.getByText('± 1 m')).toBeOnTheScreen();
+
+  render(<GPSPill status="good" accuracy={5.25} />);
+  expect(screen.getByText('± 5 m')).toBeOnTheScreen();
+
+  render(<GPSPill status="good" accuracy={10.75} />);
+  expect(screen.getByText('± 11 m')).toBeOnTheScreen();
```
achou11, line 44: it'd be good to get a clearer idea of what the desired behavior is. I wrote these assertions based on what was already implemented (with a sprinkle of what I think is reasonable), but that could be different from what's desired 😄

src/frontend/sharedComponents/GPSPill.tsx
```diff
@@ -49,17 +37,13 @@ export const GPSPill: FC<GPSPillProps> = ({
...
-      accessibilityLabel="Open GPS Modal">
+      testID={props.testID}
+      accessibilityLabel={props.accessibilityLabel}>
       <IconToRender />
 
-      <BodyText
-        variant="smallMeta"
-        style={styles.text}
-        numberOfLines={1}
-        testID="MAP.gps-pill-text">
```
achou11, line 62: removing this test ID has a small effect on one of the e2e tests. see other comment in relevant test file.

tests/e2e/specs/main/gps.test.ts
```diff
@@ -7,8 +7,8 @@ describe('MAIN - GPS Component Test', () => {
     const mapView = await $(byResourceId('MAIN.mapbox-map-view'));
     await expect(mapView).toBeDisplayed();
 
-    const gpsText = await $(byResourceId('MAP.gps-pill-text'));
-    await expect(gpsText).toBeDisplayed();
+    const gpsPill = await $(byResourceId('MAP.gps-pill'));
+    await expect(gpsPill).toBeDisplayed();
```
achou11, line 11: I removed the test ID that was used internally by this component and updated this test to confirm that the pill as a whole is displayed, not the text component within the pill. i think it's a minor change but maybe that's not desired.

src/frontend/sharedComponents/GPSPill.tsx
```diff
@@ -37,7 +33,7 @@ export const GPSPill: FC<GPSPillProps> = ({
       IconToRender = GpsSearchingIcon;
       break;
     case 'good': {
-      textValue = `± ${Math.round(locationStatus.accuracy)} m`;
+      textValue = `± ${Math.abs(Math.round(props.accuracy))} m`;
```
achou11, line 36: although `accuracy` isn't supposed to be negative, i figured it made more sense to take the absolute value than to display something like `± -10 m`. Open to an alternative fix (e.g. throwing an error if negative?)

src/frontend/sharedComponents/GPSPill.test.tsx
cimigree: And I happen to know there are plans to change the colors slightly, so best not to put too much into it right now.

src/frontend/sharedComponents/GPSPill.test.tsx
cimigree: Yeah I am not sure either.

src/frontend/sharedComponents/GPSPill.tsx
cimigree: I agree.

src/frontend/sharedComponents/GPSPill.test.tsx
achou11: Guess the safest thing to do is to keep the existing behavior, and then update things when someone points out an issue 😄

cimigree (approved):
This is an exciting start!

PR #1116, opened 2025-04-24
feat: updates invite received sheet

### LINKED ISSUE #1050: Receiving Invite
**Description**
User story : As a mobile user, I want to be able to start collecting observations on my own, and add collaborators later, without losing observations I have already collected. We are taking the oportunty to update the design of the invite to project bottom sheet

**Feature spec**

**User requirements**

**To Do**

Image

- [ ] Update bottom sheet as designed
- [ ] Use the "Current Project" feature card as seen https://github.com/orgs/digidem/projects/39/views/13?pane=issue&itemId=102533321&issue=digidem%7Ccomapeo-mobile%7C1048
- [ ] Tapping on Decline Invite or Join Project directs to the usual CoMapeo flows

**Design**

cimigree:
closes #1050 
Adds new styling and UI for the invite received sheet.
**Really large font and display:**
<image src="https://github.com/user-attachments/assets/fda687a0-9e3e-4feb-b014-01602bb242e3" width="250" />

**Really long project name:**
<image src="https://github.com/user-attachments/assets/4979342e-ec12-47f9-adae-37caaaa65596" width="250" />

**Sort of regular project name and size**
<image src="https://github.com/user-attachments/assets/a08c5544-74af-4cd6-b6a3-b56344282dfc" width="250" />

### DIFF
--- a/src/frontend/screens/Invites/InviteReceived.tsx
+++ b/src/frontend/screens/Invites/InviteReceived.tsx
@@ -18,9 +18,9 @@ import {useListenToInviteCancel} from '../../hooks/useListenToInviteCancel';
 import {BLACK, NEW_DARK_GREY, VERY_LIGHT_GREY} from '../../lib/styles';
 
 const m = defineMessages({
-  acceptInvite: {
-    id: 'screens.InviteReceived.acceptInvite',
-    defaultMessage: 'Accept Invite',
+  joinProject: {
+    id: 'screens.InviteReceived.joinProject',
+    defaultMessage: 'Join Project',
   },
   declineInvite: {
     id: 'screens.InviteReceived.declineInvite',
@@ -28,11 +28,19 @@ const m = defineMessages({
   },
   invitedToJoin: {
     id: 'screens.InviteReceived.invitedToJoin',
-    defaultMessage: "You've been invited to join:",
+    defaultMessage: "You've been invited to...",
   },
   joinAsRole: {
     id: 'screens.InviteReceived.joinAsRole',
     defaultMessage: 'Join as a {role}?',
+  },
+  coordinatorRole: {
+    id: 'screens.InviteReceived.coordinatorRole',
+    defaultMessage: 'coordinator',
+  },
+  participantRole: {
+    id: 'screens.InviteReceived.participantRole',
+    defaultMessage: 'participant',
   },
 });
 
@@ -43,8 +51,10 @@ export const InviteReceived = ({
   const {formatMessage} = useIntl();
   const inviteId = route.params.inviteId;
   const {data: invite} = useSingleInvite({inviteId});
-  const role =
-    invite?.roleName === 'Coordinator' ? 'coordinator' : 'participant';
+  const translatedRole =
+    invite?.roleName === 'Coordinator'
+      ? formatMessage(m.coordinatorRole)
+      : formatMessage(m.participantRole);
 
   const acceptInvite = useAcceptInvite();
   const rejectInvite = useRejectInvite();
@@ -101,10 +111,10 @@ export const InviteReceived = ({
 
         <View style={styles.cardContainer}>
           <HeaderText variant="header2" style={styles.projectName}>
-            {invite?.projectName}
+            {invite.projectName}
           </HeaderText>
           <BodyText variant="smallMeta" style={styles.rolePrompt}>
-            {formatMessage(m.joinAsRole, {role})}
+            {formatMessage(m.joinAsRole, {role: translatedRole})}
           </BodyText>
         </View>
 
@@ -122,7 +132,7 @@ export const InviteReceived = ({
               <PrimaryButton
                 fullSize
                 onPress={accept}
-                text={formatMessage(m.acceptInvite)}
+                text={formatMessage(m.joinProject)}
               />
             </>
           )}
--- a/src/frontend/screens/Invites/InviteReceived.tsx
+++ b/src/frontend/screens/Invites/InviteReceived.tsx
@@ -1,7 +1,5 @@
 import {BottomSheetWrapper} from '../../sharedComponents/BottomSheetWrapper';
-import InviteIcon from '../../images/AddPersonCircle.svg';
 import {StyleSheet, View} from 'react-native';
-import {LIGHT_GREY} from '../../lib/styles';
 import {PrimaryButton, SecondaryButton} from '../../sharedComponents/Buttons';
 import {defineMessages, useIntl} from 'react-intl';
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -27,13 +26,13 @@ const m = defineMessages({
...
     defaultMessage: 'Decline Invite',
   },
-  joinProject: {
-    id: 'screens.InviteReceived.joinProject',
-    defaultMessage: 'Join {projectName}',
-  },
   invitedToJoin: {
     id: 'screens.InviteReceived.invitedToJoin',
-    defaultMessage: "You've been invited to join {projectName}",
+    defaultMessage: "You've been invited to join:",
```
achou11: the designs have slightly different wording

```suggestion
    defaultMessage: "You've been invited to…",
```

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -92,52 +94,81 @@ export const InviteReceived = ({
...
+            <>
+              <SecondaryButton
+                fullSize
+                onPress={reject}
+                text={formatMessage(m.declineInvite)}
+              />
+              <PrimaryButton
+                fullSize
+                onPress={accept}
+                text={formatMessage(m.acceptInvite)}
```
achou11: the designs have different copy. should say `Join Project`

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -92,52 +94,81 @@ export const InviteReceived = ({
...
+  cardContainer: {
+    backgroundColor: '#fff5eb',
+    borderColor: VERY_LIGHT_GREY,
     borderWidth: 1,
-    borderRadius: 100,
-    alignItems: 'center',
-    alignSelf: 'center',
-    width: 80,
+    borderRadius: 6,
+    padding: 20,
```
achou11, line 160: very minor but the spacing between the top of the sheet and the text containing `You've been invited to` seems to be notably larger than the designs. if it's not a trivial fix, i wouldn't worry about fixing but my guess is that adjusting this padding value would be sufficient.

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -92,52 +94,81 @@ export const InviteReceived = ({
...
-          </>
-        )}
+      <View style={styles.container}>
+        <BodyText variant="tinyMeta" style={styles.invitedLabel}>
+          {formatMessage(m.invitedToJoin)}
+        </BodyText>
+
+        <View style={styles.cardContainer}>
+          <HeaderText variant="header2" style={styles.projectName}>
+            {invite?.projectName}
```
ErikSin: i don't think this '?' is needed

ErikSin (approved):
Looks good!

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -27,13 +26,13 @@ const m = defineMessages({
...
-    defaultMessage: 'Join {projectName}',
-  },
   invitedToJoin: {
     id: 'screens.InviteReceived.invitedToJoin',
-    defaultMessage: "You've been invited to join {projectName}",
+    defaultMessage: "You've been invited to join:",
+  },
+  joinAsRole: {
+    id: 'screens.InviteReceived.joinAsRole',
+    defaultMessage: 'Join as a {role}?',
```
achou11, line 35: note that whatever value that's passed as `role` does NOT get exposed for translation and will be presented as-is. I'm assuming we want to surface the role that you're invited to as a translated string, which means you either you have to:

1. Define a translation message for each role so that the usage is basically `t(m.invitedToJoin, { role: role === 'coordinator' ? t(m.coordinator) : ... })`

2. Use the "select" ICU message format: https://formatjs.github.io/docs/core-concepts/icu-syntax#select-format

I personally prefer (2) when it makes sense to, but most of the examples in the code base use (1) so maybe it's preferable to stick with that approach.

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -44,6 +43,9 @@ export const InviteReceived = ({
   const {formatMessage} = useIntl();
   const inviteId = route.params.inviteId;
   const {data: invite} = useSingleInvite({inviteId});
+  const role =
+    invite?.roleName === 'Coordinator' ? 'coordinator' : 'participant';
```
achou11: related to other comment about translation, ideally these would be translatable strings.

achou11 (changes requested):
Mostly related to fixes in copy and one notable one about translations

src/frontend/screens/Invites/InviteReceived.tsx
cimigree: I think that comes from the paddingTop in the BottomSheetWrapper -- 
```
{displayContent && (
        <Animated.View
          style={{
            backgroundColor: WHITE,
            padding: 20,
            paddingTop: 40,
            borderTopLeftRadius: 10,
            borderTopRightRadius: 10,
          }}
```
Changing the padding in the InviteReceived component only affects the stuff below the You've Been Invited To... text. I don't want to mess with that look (because it is according to spec), or the BottomSheetWrapper, so I am leaving as is.

src/frontend/screens/Invites/InviteReceived.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1116/commits/c4da8c716de70da7005a8bfe28077412d818c806

src/frontend/screens/Invites/InviteReceived.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1116/commits/c4da8c716de70da7005a8bfe28077412d818c806

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -44,6 +51,11 @@ export const InviteReceived = ({
   const {formatMessage} = useIntl();
   const inviteId = route.params.inviteId;
   const {data: invite} = useSingleInvite({inviteId});
+  const translatedRole =
+    invite?.roleName === 'Coordinator'
```
achou11: optional chain isn't needed here
```suggestion
    invite.roleName === 'Coordinator'
```

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -44,6 +51,11 @@ export const InviteReceived = ({
...
+      ? formatMessage(m.coordinatorRole)
+      : formatMessage(m.participantRole);
```
achou11: we can only invite someone as a coordinator or participant at the moment so maybe not something to worry about for now, but this would be prone to minor issues if there are other roles that one could be invited for in the future.

This would be more thorough but then the UI would have to account for when the role name is undefined.

```ts
const translatedRole =
  invite.roleName === 'Coordinator'
    ? formatMessage(m.coordinatorRole)
    : invite.roleName === 'Member'
      ? formatMessage(m.participantRole)
      : invite.roleName;
``` 

Think what you have here is okay for now though, since "participant" is a suitably vague generalization in the case I'm considering 😄

achou11 (approved):
Couple non-blocking comments but otherwise lgtm!

src/frontend/screens/Invites/InviteReceived.tsx
cimigree: Yes I realize that things might change and a map or object might be more suitable _in the future,_ but I did not want to write code that is for a future time that might not be needed.

PR #1118, opened 2025-04-25
Get first component test running

EvanHahn:
_To be merged into [this branch][0], which scaffolds component tests._

My goal: get this running:

```sh
npx dotenv -- jest src/frontend/screens/Exchange.test.tsx
```

Here's what I did.

## Get the tests running

1. _not shown here_: **make `styled-map-package` compatible with CommonJS.** Did this by manually editing `node_modules/styled-map-package/package.json`.

   Best I understand, Jest uses CommonJS for everything, not ECMAScript modules. Though it has [experimental ESM support][1], I could not make this work.[^1] Instead, my solution was to edit the `"exports"` field of `styled-map-package`'s `package.json`, doing something like this for all exports:

   ```diff
    "exports": {
      ".": {
        "types": "./dist/index.d.ts",
   +    "require": "./lib/index.js",
        "import": "./lib/index.js"
      },
   ```

   I don't think this is the perfect solution because, well, these aren't CommonJS modules! But Jest is able to transform them correctly here.

   I think the best _real_ solution would be to add CommonJS support to `styled-map-package`. Inconvenient, but hopefully not too tricky.

   Doing this fixes the `Cannot find module 'styled-map-package' from 'node_modules/@comapeo/core/src/fastify-plugins/maps.js'` error, but exposed new problems...

1. **Transform `import.meta.url` into something else.** See `jest.config.js`.

   `@comapeo/core` uses `import.meta.url`, which doesn't work with CommonJS and gives a syntax error.

   [This Babel plugin][2] addresses that problem. (I set its `module` option. Without it, things like `createRequire` break.)

1. **Remove async `describe` from the exchange test.**

   All I did was remove the word `async` from the `describe` block, because Jest was giving an error. I guess these have to be synchronous.

1. **Mock `NetInfo`.** See `jest.setup.js`.

   After this I was seeing errors from `@react-native-community/netinfo`, but good news: [they tell you just how to fix it][3]. I basically just copy-pasted their solution.

At this point, tests were running!

## Stop tests from hanging

Unfortunately, things weren't getting cleaned up so the test would hang after passing. We want the test process to actually finish!

To fix that problem:

1. **Clean up the interval in `AppDiagnosticMetrics`.** See #1117.

1. **Clean up app provider wrappers.** See `tests/integration/helpers/react.tsx`.

   The provider helpers set up a bunch of stuff. We need to tear a couple of those things down in tests, so I exposed and used that.

And now the test runs! Success!

[^1]: I tried following the instructions. I tried mocking the module. I tried `patch-package`, which can't edit `package.json` I don't think.

[0]: https://github.com/digidem/comapeo-mobile/tree/910/setup-integration-testing
[1]: https://jestjs.io/docs/ecmascript-modules
[2]: https://www.npmjs.com/package/babel-plugin-transform-import-meta
[3]: https://www.npmjs.com/package/@react-native-community/netinfo#user-content-errors-while-running-jest-tests


achou11:
Just an end of week update on this:

I have an open PR in `styled-map-package` to enable dual publishing (see https://github.com/digidem/styled-map-package/pull/45). It's been approved but wanted to mull over the solution a bit before merging and - if possible - wait for an upstream fix to one of the dev deps that it introduces.

I tested a packed version of that branch in core and tests seemed to pass. 

One last thing to do to confirm is defining an override of `styled-map-package` in this branch and attempting to get (1) fixed!

EvanHahn:
Love it. Let me know if I can help with anything.

achou11:
Packed that PR's branch as of https://github.com/digidem/styled-map-package/pull/45/commits/c60491e7f0b661f7b0e2a21274726cf464ce98a9 and added this to the `package.json` here:

```json
  "overrides": {
    "@comapeo/core": {
      "styled-map-package": "file:styled-map-package-2.2.1.tgz"
    }
  }
```

Got the following error when running the test:

```console
TypeError: _styledMapPackage.ReaderWatch is not a constructor

      at node_modules/@comapeo/core/src/fastify-plugins/maps.js:90:29
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:17)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:17:9)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:7)
      at node_modules/@babel/runtime/helpers/asyncToGenerator.js:14:12
      at apply (node_modules/@comapeo/core/src/fastify-plugins/maps.js:132:2)
      at apply (node_modules/@comapeo/core/src/fastify-plugins/maps.js:25:29)
      at Plugin.func (node_modules/avvio/lib/plugin.js:125:28)
      at Boot.exec [as _loadPlugin] (node_modules/avvio/boot.js:432:10)
```

Probably something to do with default exports being used. Going to update that branch to avoid default exports and see if it fixes it.

achou11:
> Probably something to do with default exports being used. Going to update that branch to avoid default exports and see if it fixes it.

Yep, looks like removing default exports+imports in favor of named ones did the trick. Was able to get the test passing using the packed version as of https://github.com/digidem/styled-map-package/pull/45/commits/1cfbcd6dc16a86f261b18d7c432384fc35bff218

achou11:
Updated `@comapeo/core` to pull in the updated `styled-map-package` dep via https://github.com/digidem/comapeo-mobile/pull/1118/commits/f56d4a844cb69b88e31023d9c2e3ec7c08a8da35. The test of interest ran successfully for me locally, although it took almost 10 seconds 😅

achou11:
couple of type errors showing up but they're introduced via changes in the `910/setup-integration-testing` branch. probably a better use of time to address them on that branch than in this PR

PR #1119, opened 2025-04-27
chore: update `@rollup/plugin-esm-shim`

EvanHahn:
This Rollup plugin was slightly out of date. It had [a bug][0] which [seems to have been fixed][1], so we can remove our patch workaround.

I'm doing this because [the bug issue][0] was closed and I figured I'd follow up. This change is not urgent.

[0]: https://github.com/rollup/plugins/issues/1709#event-17420158881
[1]: https://github.com/rollup/plugins/commit/f23312a72cb7e6c0219da9f09051421f3e2d86a0


achou11 (approved):
awesome, thanks for following up with this!

PR #1126, opened 2025-04-28
fix: show previews

### LINKED ISSUE #934: Media Manager/ preview only/ photo issues
https://www.notion.so/digidem/Media-Manager-preview-only-photo-issues-18f1b08162d5806cb6afc5e2934aaef6

Both devices choose to exchange previews only

exchange data

The exchanged data appears correctly, but when clicking on an image…the detail view is black

ErikSin:
closes #934 

Previously we were always trying to show the 'Original' photo, even if it was not available. Technically if the photo is not available a url is still being returned, im not sure if that is the intended behaviour. I would argue that is is not desired behaviour and that it should throw an error instead, but if its not a trivial fix the front end can deal with it.

I could see a world in which we check whether "sync previews only" is set, and return the "original" vs the "preview". I don't know if that is necessary as it complicates the code, and we cannot do anything with the current photo. For example we cannot zoom into the photo, so showing the preview vs the original has no real functionality benefit

### DIFF
--- a/src/frontend/screens/PhotoPreviewModal.tsx
+++ b/src/frontend/screens/PhotoPreviewModal.tsx
@@ -79,7 +79,8 @@ export const PhotoPreviewModal: FC<
         <PhotoUnpreparedView
           onPress={() => setShowHeader(!showHeader)}
           photo={photo}
-          variant={'original'}
+          // This needs to be "preview" otherwise the photo will not be shown when the user is only syncing previews
+          variant={'preview'}
         />
       ) : (
         <PhotoPreparedView


achou11 (approved):
Think this fix is sufficient but would prefer to see a more resilient image loading solution at some point. I quickly implemented an example of attempting to load the original first and then resorting to the preview if necessary, which wasn't too complicated to get working: https://github.com/digidem/comapeo-mobile/commit/b845b4f65efbdcf5624ea45ad6495f27c8797eba

If our needs become more complex, can consider using something like https://github.com/mthahzan/react-native-image-fallback to handle a variety of fallbacks (or implement our own thing internally).

ErikSin:
> Think this fix is sufficient but would prefer to see a more resilient image loading solution at some point. I quickly implemented an example of attempting to load the original first and then resorting to the preview if necessary, which wasn't too complicated to get working: [52ba18c](https://github.com/digidem/comapeo-mobile/commit/52ba18c02baf52ee398eed1e6d828272cc7513d8)

This makes sense and is not super complicated which is great. The reason that I don't think this is necessary right now is that viewing the original vs the preview does not effect the user experience. The user can't interact with the photo. The only thing they can do is open the photo and view it. The difference between the original and preview is imperceptible on a mobile device. On a tablet, there may be a difference, and technically we can run this on certain tablets, but we are not really designing for that.

All that being said, this does make me realize when a user shares a photo (eg, via whats app). We probably should try to share the orginal if possible. But doing that check without trying to open the photo first is a little more difficult. So if you have any ideas on how to do that...

PR #1133, opened 2025-04-30
feat: block invites if user taking tracks

### LINKED ISSUE #1089: Tracks / Error States
**Description**
User story : User can store data from different projects in CoMapeo, keep it separate, and switch between projects
Here are the possible Errors that could occur with this new feature

**Feature spec**
https://www.notion.so/digidem/Multiple-projects-ce7d9913220342ed968311ac9e8a523b?pvs=4

**User requirements**
See **Must Haves**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1981b08162d58065a598000c95d41deb&pvs=4

**To Do**

### Project Invite while Recording Tracks

Image

When a person receives a Project Invite and accepts, then they are taken to that project right away. They can return to their previous project through the All Projects screen.
If a person receives a Project Invite while Recording Tracks and they attempt to accept : 

- [ ] Display the bottom sheet error letting user know they need to Stop Recording Tracks. 
- [ ] Display copy :  **Recording Tracks!**
- [ ] Display copy : **You’re currently recording a track.  To join project, stop recording now.**
- [ ] Display button **Stop Tracks**
- [ ] Display button **Cancel**
- [ ] If tapped on Stop Tracks , take user to complete the New Track flow. 
- [ ] When track saved, user returns to the Project Invite and they can attempt to accept again.

### Switching Projects while Recording Tracks

Image

From the All Projects screen, when a person attempts to Switch Projects while Recording Tracks : 

- [ ] Display the bottom sheet error letting user know they need to Stop Recording Tracks.
- [ ] Display copy :  **Recording Tracks!**
- [ ] Display copy : **You’re currently recording a track.  To join project, stop recording now.**
- [ ] Display button **Stop Tracks**
- [ ] Display button **Cancel**
- [ ] If tapped on Stop Tracks , take user to complete the New Track flow. 
- [ ] When track saved, user returns to the Project they were switching to. 

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4741-2651&t=bQDpwjZbr3d6dKAq-0

ErikSin:
closes #1089 

https://github.com/user-attachments/assets/e76dbd89-ab53-4b89-ad06-7cc88435a42d

I simplified the useTracks hook in the process:
1. There was some leaky implementation. We only save a track if there is more than 1 location saved. Previously the hook was not responsible for this check, and would only save or cancel a track. This meant that the consuming component was checking for the location via the `useTrackState`. The useTracks() is partly an abstraction of useTrackState. So we were using 2 hook unnecessarily. I updated the hook so that useTracks does check for the location. Now when tracks is stopped, it simply return a boolean whether there is more than 1 location in persisted state.
2. Removed the loading state from the hook. The loading state is unncessary as we don't need to block any interactions while we are turning on tracks. It is turned on and runs in the background. If there is an error, we show an error sheet but a loading state is unnecessary for that. Also when turning off the tracking, the user does not need to wait for track to turn off. It can happen in the background while they are saving the track.

### DIFF
--- a/src/frontend/hooks/useTracking.ts
+++ b/src/frontend/hooks/useTracking.ts
@@ -6,6 +6,7 @@ import {
 } from '../contexts/TrackStoreContext.tsx';
 import {LOCATION_TASK_NAME} from '../sharedTypes/location.ts';
 import {useNavigation} from '@react-navigation/native';
+import * as Sentry from '@sentry/react-native';
 
 export function useTracking() {
   const {setTracking, clearCurrentTrack} = useTrackActions();
@@ -22,9 +23,10 @@ export function useTracking() {
     Location.startLocationUpdatesAsync(LOCATION_TASK_NAME, {
       accuracy: Location.Accuracy.Highest,
       activityType: Location.LocationActivityType.Fitness,
-    }).catch(() => {
+    }).catch(err => {
+      Sentry.captureException(err);
       setTracking(false);
-      // @ts-expect-error - this is typing issue, we are using the not strongly typed hook as this can technically be used in any screen. But regardless of the screen, we want to show the error bottom sheet
+      // @ts-expect-error - this is a typing issue, we are using the non-strongly typed hook as this can technically be used in any screen. But regardless of the screen, we want to show the error bottom sheet.
       navigation.navigate('ErrorBottomSheet');
     });
   }, [isTracking, setTracking, navigation]);
@@ -32,9 +34,11 @@ export function useTracking() {
    * Cancels location tracking and stops background updates.
    * @returns {boolean} True if location history has more than one entry (track should be kept), false otherwise.
    */
-  const cancelTrackingAndReturnIfTracksSaved = useCallback(() => {
+  const cancelTracking = useCallback(() => {
     setTracking(false);
-    Location.stopLocationUpdatesAsync(LOCATION_TASK_NAME);
+    Location.stopLocationUpdatesAsync(LOCATION_TASK_NAME).catch(err => {
+      Sentry.captureException(err);
+    });
     const hasTrackWithMoreThan1Point = locationHistory.length > 1;
 
     if (!hasTrackWithMoreThan1Point) {
@@ -44,5 +48,9 @@ export function useTracking() {
     return hasTrackWithMoreThan1Point;
   }, [setTracking, clearCurrentTrack, locationHistory]);
 
-  return {isTracking, startTracking, cancelTrackingAndReturnIfTracksSaved};
+  return {
+    isTracking,
+    startTracking,
+    cancelTracking,
+  };
 }
--- a/src/frontend/hooks/useTracking.ts
+++ b/src/frontend/hooks/useTracking.ts
@@ -5,34 +5,44 @@ import {
   useTrackState,
 } from '../contexts/TrackStoreContext.tsx';
 import {LOCATION_TASK_NAME} from '../sharedTypes/location.ts';
+import {useNavigation} from '@react-navigation/native';
 
 export function useTracking() {
-  const [loading, setLoading] = useState(false);
-  const {setTracking} = useTrackActions();
+  const {setTracking, clearCurrentTrack} = useTrackActions();
+  const navigation = useNavigation();
   const isTracking = useTrackState(state => state.isTracking);
+  const locationHistory = useTrackState(state => state.locationHistory);
 
-  const startTracking = useCallback(async () => {
+  const startTracking = useCallback(() => {
     if (isTracking) {
       console.warn('Start tracking attempt while tracking already enabled');
-      setLoading(false);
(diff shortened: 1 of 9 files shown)


src/frontend/hooks/useTracking.ts
```diff
@@ -5,34 +5,44 @@ import {
   useTrackState,
 } from '../contexts/TrackStoreContext.tsx';
 import {LOCATION_TASK_NAME} from '../sharedTypes/location.ts';
+import {useNavigation} from '@react-navigation/native';
 
 export function useTracking() {
-  const [loading, setLoading] = useState(false);
-  const {setTracking} = useTrackActions();
+  const {setTracking, clearCurrentTrack} = useTrackActions();
```
ErikSin, line 12: IMO, i think we shouldn't manually be setting 'setTracking'. `isTracking` should be a derived value based on whether the `Location.startLocationUpdatesAsync` is running..but that is a little out of scopr

ErikSin:
oh, my ESLINT was disabled on vs code. Fixing the eslint issues now

src/frontend/hooks/useTracking.ts
```diff
@@ -1,38 +1,48 @@
...
+      navigation.navigate('ErrorBottomSheet');
     });
+  }, [isTracking, setTracking, navigation]);
+  /**
+   * Cancels location tracking and stops background updates.
+   * @returns {boolean} True if location history has more than one entry (track should be kept), false otherwise.
+   */
+  const cancelTrackingAndReturnIfTracksSaved = useCallback(() => {
+    setTracking(false);
+    Location.stopLocationUpdatesAsync(LOCATION_TASK_NAME);
```
achou11: should probably add a catch handler here that at least logs to Sentry

src/frontend/hooks/useTracking.ts
```diff
@@ -1,38 +1,48 @@
...
+      setTracking(false);
+      // @ts-expect-error - this is typing issue, we are using the not strongly typed hook as this can technically be used in any screen. But regardless of the screen, we want to show the error bottom sheet
+      navigation.navigate('ErrorBottomSheet');
     });
+  }, [isTracking, setTracking, navigation]);
+  /**
+   * Cancels location tracking and stops background updates.
+   * @returns {boolean} True if location history has more than one entry (track should be kept), false otherwise.
+   */
+  const cancelTrackingAndReturnIfTracksSaved = useCallback(() => {
```
achou11: I'm all for long variable names, but wondering if keeping this to `cancelTracking` is sufficient given the documentation you added 😄

src/frontend/hooks/useTracking.ts
```diff
@@ -1,38 +1,48 @@
...
     }
-
-    setLoading(true);
-
-    await Location.startLocationUpdatesAsync(LOCATION_TASK_NAME, {
+    setTracking(true);
+    Location.startLocationUpdatesAsync(LOCATION_TASK_NAME, {
       accuracy: Location.Accuracy.Highest,
       activityType: Location.LocationActivityType.Fitness,
+    }).catch(() => {
```
achou11: should probably log the error to Sentry in this case.

src/frontend/screens/TrackRecordingActive.tsx
```diff
@@ -0,0 +1,75 @@
...
+  navigation,
+}: NativeRootNavigationProps<'TrackRecordingActive'>) => {
+  const {formatMessage} = useIntl();
+  const {cancelTrackingAndReturnIfTracksSaved} = useTracking();
+
+  function handleStopTracks() {
+    const hasTracksSaved = cancelTrackingAndReturnIfTracksSaved();
+
+    if (hasTracksSaved) {
+      navigation.replace('SaveTrack');
```
achou11, line 41: maybe more of a UX question, but even if the track is saved, the user loses access to it if they eventually accept the invite, right? giving the option of saving something that they won't be able to access seems like it could lead to confusion.

achou11 (commented):
Made a couple of small suggestions but need some clarification on a UX detail.

src/frontend/screens/TrackRecordingActive.tsx
achou11: Decided it was more suitable to ask this in the original issue: https://github.com/digidem/comapeo-mobile/issues/1089#issuecomment-2852259110

achou11 (approved):
aside from my suggestions, changes seem to work as intended!

src/frontend/hooks/useTracking.ts
ErikSin: haha i named it like this for you!

ErikSin: Yeah, ill change it to `cancelTracking`!

PR #1142, opened 2025-05-05
fix: allow "failed" members to be invitable

achou11:
`@comapeo/core@3.1.0` introduced a new "failed" role that indicates a member's failure to properly join a project under valid circumstances, usually caused by external conditions such as device-related issues or interruptions to the invite flow. This new role was added to indicate the ability to re-invite the device with the intent of properly adding it to the project.

This PR updates the UI for device selection step when inviting a device to account for this new role such that project members with this role can be invited.

### DIFF
--- a/src/frontend/sharedTypes/index.ts
+++ b/src/frontend/sharedTypes/index.ts
@@ -39,4 +39,5 @@ export const COORDINATOR_ROLE_ID = 'f7c150f5a3a9a855';
 export const MEMBER_ROLE_ID = '012fd2d431c0bf60';
 export const BLOCKED_ROLE_ID = '9e6d29263cba36c9';
 export const LEFT_ROLE_ID = '8ced989b1904606b';
+export const FAILED_ROLE_ID = 'a24eaca65ab5d5d0';
 export const NO_ROLE_ID = '08e4251e36f6e7ed';
--- a/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
@@ -11,6 +11,7 @@ import {DeviceCard} from '../../../../sharedComponents/DeviceCard';
 import {Text} from '../../../../sharedComponents/Text';
 import {NativeNavigationComponent} from '../../../../sharedTypes/navigation';
 import {useActiveProject} from '../../../../contexts/ActiveProjectContext';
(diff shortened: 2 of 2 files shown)


src/frontend/sharedTypes/index.ts
```diff
@@ -39,4 +39,5 @@ export const COORDINATOR_ROLE_ID = 'f7c150f5a3a9a855';
 export const MEMBER_ROLE_ID = '012fd2d431c0bf60';
 export const BLOCKED_ROLE_ID = '9e6d29263cba36c9';
 export const LEFT_ROLE_ID = '8ced989b1904606b';
+export const FAILED_ROLE_ID = 'a24eaca65ab5d5d0';
```
achou11, line 42: Copied from https://github.com/digidem/comapeo-core/blob/ffe5198d8a76b1de25a30bb922b19ec06eb82360/src/roles.js#L15

RangerMauve (approved):
Looks good. Is it okay that members with role `BLOCKED` and `LEFT` also get shown in this list? It feels like it'd be good to allow `LEFT` members to be re-added too.

achou11:
> Looks good. Is it okay that members with role `BLOCKED` and `LEFT` also get shown in this list? It feels like it'd be good to allow `LEFT` members to be re-added too.

Pretty sure BLOCKED members shouldn't be able to be re-invited. Not overly sure about LEFT either - maybe there's an e2e test in core that clarifies that? From a UX perspective it would make sense, just not sure if core actually allows it.

RangerMauve:
Just checked and we should be able to re-add BLOCKED or LEFT members from the perspective of core. Maybe we can bring this up with the other teams for advice on what they prefer? I don't think this is blocking the PR but would be good not to lose it.

cc @ximenabb

achou11:
~Hm actually upon reflection, are the changes I'm introducing actually doing what I claimed? For some reason I thought that the existing code was somehow filtering out the "failed" members, but I guess technically it's not. It just filters out members whose member ids aren't found from the discovery layer.~ EDIT: nevermind, my change is actually doing what I described 😄

achou11:
> I don't think this is blocking the PR but would be good not to lose it.

Just to clarify, I don't think this change causes a regression, if that's what you're implying. Not sure if that's what you meant but just in case!

Before: any discovered device id that is a member (whether blocked, failed, left, etc) is not inviteable
After: any discovered device id that is a member must be failed in order to be inviteable

PR #1146, opened 2025-05-05
fix: properly save audio on observation edit

### LINKED ISSUE #1127: Edit Obsv/ cant save when Audio added
**Describe the bug**
**Problem:** 

- Create an observation with a photo and Save it
- Go back to the observation and tap on Edit
- Create an audio recording
- Try to save the observation with the new recording
- Screen will open saying “something went wrong”
- Can’t save the observation

- If i try to delete the audio recording on the unsaved Observation
- And then try to save again
- Screen will open saying “something went wrong”

https://www.notion.so/digidem/Edit-Obsv-cant-save-when-Audio-added-1e41b08162d5803194bbc1f15603e382?pvs=4

Also noted here: https://github.com/digidem/comapeo-mobile/issues/1022

ErikSin:
closes: #1127 

This error was being caused by `AudioSavedBottomSheet` navigating to `CreateObservation` page always, even when it should have been navigating back to the `EditObservation` page. I solved this by simply using `navigation.goBack()`

### DIFF
--- a/src/frontend/screens/Audio/AudioSavedBottomSheet.tsx
+++ b/src/frontend/screens/Audio/AudioSavedBottomSheet.tsx
@@ -5,8 +5,8 @@ import {defineMessages, useIntl} from 'react-intl';
 import GreenCheck from '../../images/GreenCheck.svg';
 import {HeaderText} from '../../sharedComponents/Text/HeaderText';
 import {BodyText} from '../../sharedComponents/Text/BodyText';
-import {Button} from '../../sharedComponents/Button';
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
+import {PrimaryButton, SecondaryButton} from '../../sharedComponents/Buttons';
 
 const m = defineMessages({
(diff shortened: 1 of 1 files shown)


achou11:
I'm still encountering the issue based on the steps provided in the original issue. The `Something went wrong` screen appears after I try to save the observation (in the edit flow). Also got the following log: `Error: Invalid value [object Object]`. not very helpful as is 😅

achou11:
hm now i'm not able to re-create it...maybe a local dev issue on my end 🤔

achou11 (approved):
since i can't reliably repro the issue anymore, going to assume I ran into it due to some dumb local dev mistake.

ErikSin:
yeah, on my attempt to figure out what was going on, I re-ran it. I actually did run into the original issues, but i think you are right and it is related to our dev environment. Because when i just rebuilt everything and tried again it was fixed. Im pretty sure it is related to our hot reloading just not working properly.

PR #1147, opened 2025-05-06
fix: audio error - does not throw when time runs out

### LINKED ISSUE #1137: Audio recording max 5mins/ error message
When the recording reaches 5mins - the “Something went wrong” screen appears
When tapping “Go back” user goes back to the playback screen and can save the observation
*Correction- can save the file but cannot save the observation until you delete the audio recording

https://www.notion.so/digidem/Audio-recording-max-5mins-error-message-1e61b08162d580699e04cb69cce9c0f0?pvs=4

ErikSin:
closes #1137 

The issue was being caused by useEffect which stopped the recording at 5 mins. The useEffect would check to see if the timeElapsed was equal to or greater than 5 mins. If it was 5 mins, it would call `finishRecording`, which would cause another re-render. This would cause the useEffect to be run again, and since the `timeElapsed` was greater that 5 mins, it would call `finishRecording()` again. But since there was no active recording, as finish recording had been already called once before, it would throw an error.

This fix adds a check to only call `finishRecording` if the recording is active.

To test, you can change [this number](https://github.com/digidem/comapeo-mobile/blob/489519d907500918678618e892552c53cef1ff09/src/frontend/screens/Audio/AudioRecording/index.tsx#L18) to a smaller number so you do not need to wait 5 mins.

### DIFF
--- a/src/frontend/screens/Audio/AudioRecording/index.tsx
+++ b/src/frontend/screens/Audio/AudioRecording/index.tsx
@@ -15,7 +15,7 @@ import {UIActivityIndicator} from 'react-native-indicators';
 import {millisecondsToMMSS} from '../../../lib/millisecondsToFormattedTime';
 
 // 5 minutes
-const MAX_RECORDING_DURATION_MS = 5000;
+const MAX_RECORDING_DURATION_MS = 30000;
 const PRIMARY_CONTROL_DIAMETER = 96;
 
 const m = defineMessages({
--- a/src/frontend/screens/Audio/AudioRecording/index.tsx
+++ b/src/frontend/screens/Audio/AudioRecording/index.tsx
@@ -15,7 +15,7 @@ import {UIActivityIndicator} from 'react-native-indicators';
 import {millisecondsToMMSS} from '../../../lib/millisecondsToFormattedTime';
 
 // 5 minutes
-const MAX_RECORDING_DURATION_MS = 300000;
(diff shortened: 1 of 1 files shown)


src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -15,7 +15,7 @@ import {UIActivityIndicator} from 'react-native-indicators';
 import {millisecondsToMMSS} from '../../../lib/millisecondsToFormattedTime';
 
 // 5 minutes
-const MAX_RECORDING_DURATION_MS = 300000;
+const MAX_RECORDING_DURATION_MS = 5000;
```
achou11: need to restore this value

achou11 (changes requested):
Fix makes sense but need to restore the duration limit constant 😄

src/frontend/screens/Audio/AudioRecording/index.tsx
ErikSin: whoops! good catch

ErikSin: ~https://github.com/digidem/comapeo-mobile/pull/1147/commits/fe7437ddff6c90431abccb59b133d3e66ef226e7~

https://github.com/digidem/comapeo-mobile/pull/1147/commits/18f7145594d6bd41c701b0db22916a0f98ce2e79

achou11 (approved):

PR #1148, opened 2025-05-06
feat: indicate track from other users

### LINKED ISSUE #1130: Exchange Tracks/ No blue line in obvs list pt2
tested on CoMapeoV1.1.0-RC.3

not fixed https://github.com/digidem/comapeo-mobile/issues/936

ErikSin:
closes #1130

image


src/frontend/hooks/useIsMyDocument.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function useIsMyDocument(originalVersionId: string) {
+  const {projectId} = useActiveProject();
+  const {data: deviceInfo} = useOwnDeviceInfo();
+  const {data: createdByDeviceId} = useDocumentCreatedBy({
+    projectId,
+    originalVersionId: originalVersionId,
+  });
+
+  return createdByDeviceId === deviceInfo?.deviceId;
```
cimigree: Same question as last time, should we fallback to true in useIsMyDocument if the query results haven’t loaded yet, to avoid mistakenly showing the blue line?

src/frontend/hooks/useIsMyDocument.ts
```diff
@@ -0,0 +1,13 @@
+import {useOwnDeviceInfo, useDocumentCreatedBy} from '@comapeo/core-react';
```
cimigree: Should this be in the server folder since it uses server queries? Not sure at all how to decide... Just curious what your thoughts are on that.

cimigree (approved):
Works well. I like the useIsMyDocument hook.
Thanks for updating the list items with the new text components and no more touchable highlight.
Two non-blocking comments.

src/frontend/hooks/useIsMyDocument.ts
ErikSin: now that were using suspense queries, we don't need to worry about loading states. If its loading, the screen won't render

src/frontend/hooks/useIsMyDocument.ts
ErikSin: i also wasn't sure. But I think if you are thinking the same thing, lets move it to the server folder!

PR #1149, opened 2025-05-06
fix: after invite navigation

### LINKED ISSUE #1124: Inviter/ Accepted invite, goes back to inviting device flow
**Describe the bug**
Once the inviter gets the confirmation that the invitee has accepted, they are taken to the Your Team screen and then back to the accepting invite flow. See below

**To Reproduce**
- Inviter sends a project invite
- Invite gets accepted
- Inviter sees the “accepted invite” screen, clicks on Close
- Inviter is taken to Your Team
- Inviter clicks on the Back arrow
- Inviter sees the “waiting for device to accept”
- Have to click “Cancel” and is taken back to Your Team screen
- Inviter clicks on the Back arrow again
- Have to select the role and inviting new device
- Inviter clicks on the Back arrow again and back to Your Team
- Inviter clicks on the Back arrow again, goes to map

**Expected behavior**
- When inviter is in the Your team screen
- when clicking on Back Arrow, go to Map
- 
**Video*

https://www.notion.so/digidem/Inviter-Accepted-invite-goes-back-to-inviting-device-flow-1e01b08162d5801594befd79e78db756

cimigree:
closes #1124 

**The problem**
After accepting or declining a device invite, tapping "Close" would return to the YourTeam screen, but the invite flow was still in the navigation stack. Pressing the back button would take the user back into the invite flow, which was confusing.

**The fix**
Added a shared resetToYourTeam function to replace the navigation stack with:
Home → ProjectSettings → YourTeam
Passed this function into InviteAccepted, InviteDeclined, and UnableToCancelInvite as an onClose param.
Used it in the cancelInvite flow as well.

**Result**
Now when the user taps "Close," they return to YourTeam, and the back button works as expected without going back into the invite screens.

### DIFF
--- a/src/frontend/lib/resetToYourTeam.ts
+++ b/src/frontend/lib/resetToYourTeam.ts
@@ -1,0 +1,13 @@
+import {CommonActions, NavigationProp} from '@react-navigation/native';
+import type {AppStackParamsList} from '../sharedTypes/navigation';
+
+export function resetToYourTeam(
+  navigation: NavigationProp<AppStackParamsList>,
+) {
+  navigation.dispatch(
+    CommonActions.reset({
+      index: 2,
+      routes: [{name: 'Home'}, {name: 'ProjectSettings'}, {name: 'YourTeam'}],
+    }),
+  );
+}
--- a/src/frontend/lib/resetToYourTeam.ts
+++ b/src/frontend/lib/resetToYourTeam.ts
@@ -1,10 +1,11 @@
 import {CommonActions, NavigationProp} from '@react-navigation/native';
 import type {AppStackParamsList} from '../sharedTypes/navigation';
 
-export function resetToYourTeam(
-  navigation: NavigationProp<AppStackParamsList>,
-) {
-  navigation.dispatch(
+type Navigation = NavigationProp<AppStackParamsList>;
+type Dispatch = Navigation['dispatch'];
+
+export function resetToYourTeam(dispatch: Dispatch) {
+  dispatch(
     CommonActions.reset({
       index: 2,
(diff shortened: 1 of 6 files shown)


ErikSin:
A simpler alternative is to export `'resetToYourTeam`, and just have each of the pages import it. Then we wouldn't have to pass it around via navigation. Unless there is a reason for passing it via navigation that i am missing.

cimigree:
I don't know why I had it in my head to use navigation params... I made an exportable file instead. Sorry about that.
I also replaced the buttons.

src/frontend/lib/resetToYourTeam.ts
```diff
@@ -0,0 +1,13 @@
...
+export function resetToYourTeam(
+  navigation: NavigationProp<AppStackParamsList>,
+) {
+  navigation.dispatch(
+    CommonActions.reset({
+      index: 2,
+      routes: [{name: 'Home'}, {name: 'ProjectSettings'}, {name: 'YourTeam'}],
+    }),
+  );
+}
```
ErikSin: i dont thing we should pass the entire navigation object, since were only using the dipatch. You can do this for the typing:

```ts
type Navigation = NavigationProp<AppStackParamsList>;
type Dispatch = Navigation['dispatch'];

export function resetToYourTeam(dispatch: Dispatch) {
  dispatch(
    CommonActions.reset({
      index: 2,
      routes: [{name: 'Home'}, {name: 'ProjectSettings'}, {name: 'YourTeam'}],
    }),
  );
}
```

ErikSin (approved):
1 comment, but can merge once that is fixed

PR #333, opened 2024-05-09
chore: downgrade expo-updates to expected version

achou11:
Noticed this build issue when running `npx expo run:android`:

```sh
Could not find method useDefaultAndroidSdkVersions() for arguments [] on project ':expo-eas-client' of type org.gradle.api.Project.
```

Running `npx expo install --fix` downgraded `expo-updates`, which seems to have resolved the issue for me.

PR #1152, opened 2025-05-06
feat: header indicates project color

### LINKED ISSUE #1085: Update all Tab header
**Description**
User story : User can store data from different projects in CoMapeo, keep it separate, and switch between projects
Goal is to display the Project Name in the Observation List, Map, Camera and Tracks screens.

**Feature spec**
https://www.notion.so/digidem/Multiple-projects-ce7d9913220342ed968311ac9e8a523b?pvs=4

**User requirements**
See **Must Haves**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1981b08162d58065a598000c95d41deb&pvs=4

**To Do**

### Observation List

Image

- [ ] Integrate Project Color into the Tabs header UI. 

### Map

Image

- [ ] Integrate Project Color into the Tabs header UI. 

### Camera

Image

- [ ] Integrate Project Color into the Tabs header UI. 

### Tracks

Image

- [ ] Integrate Project Color into the Tabs header UI. 

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4741-2651&t=bQDpwjZbr3d6dKAq-0

ErikSin:
closes #1085

image

image


src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -89,8 +89,6 @@ const styles = StyleSheet.create({
     backgroundColor: '#333333E6',
   },
   text: {
-    color: WHITE,
-    fontFamily: 'Rubik_600SemiBold',
```
cimigree, line 93: Why did you remove the bolder text? It looks like that is what the font still is in figma? And it also looks like it is #333 black instead of Dark Grey...

cimigree (approved):
That's great that we can use the context in the header now (I think that is what that change is doing?). I do think there is a little different style of the text in the header based on Figma, but I will leave that up to you to confirm.

ErikSin:
> That's great that we can use the context in the header now (I think that is what that change is doing?).

Actually i did the opposite. So `ScreenLayout` wraps the individual screen (but not the header - because of how headers are separate to the screen). `Layout` wraps the entire navigator, which includes the header. Since the GPS pill was moved out of the header, the header does not need access to the location context anymore.

src/frontend/sharedComponents/HomeHeader.tsx
ErikSin: both `BodyText` and `HeaderText` have opinionated styling when it comes to `fontSize`, `fontWeight`, and `fontFamily`. 

Im a little confused what design wants. I was told that they wanted the opinionated text components, but they do this sometimes...so for simplicity, im just using the opinionated text components...

PR #1100, opened 2025-04-17
feat: default project project settings

cimigree:
closes #1052 

To be considered after #1094 

## Refactor Project Settings Screen + Add E2E Tests 

**Reference**: [Notion - Invite Devices to Default Project](https://www.notion.so/digidem/Invite-devices-to-default-project-1791b08162d580889f78f08b034b710d)

### Summary

This PR updates the **Project Settings** screen to support and visually distinguish between three user roles:

- **Solo**: A user mapping alone
- **Coordinator**: A user who named a project or was invited as a coordinator
- **Participant**: A user invited to a project (cannot be tested in E2E due to Browserstack limitations)

Also adds new E2E test coverage for the Project Settings Screen for both **solo** and **coordinator** roles.

---

#### 1.  **Refactored Project Settings Screen**
- Each role sees a different combination of cards based on role:
  - **Solo**: "My Solo Project" + Invite button, Project Categories
  - **Coordinator**: Project Info, Project Collaborators, Project Categories
  - **Participant**: Project Info, Project Collaborators (not testable currently)

#### 2.  **New E2E Tests**
- **`project-settings-no-proj.test.ts`** (solo)
- **`project-settings-proj.test.ts`** (named project / coordinator)
- Both are added in `flow/index.ts` in correct order and verify content without navigating deeper into sub-screens.

#### 3.  **Updated docs**
- For **`project-settings-no-proj.test.ts`** (solo)
- For **`project-settings-proj.test.ts`** (named project / coordinator)

### How to Test

1. Launch app as **Solo User**
   - Open side drawer, click **View**
   - See "My Solo Project" card with "You’re mapping on your own"
   - See **Invite Collaborators** and **Project Categories** cards

2. Create or switch to **Named Project / Coordinator**
   - Open drawer, click **View**
   - See:
     - Project Info card with "test project"
     - Project Collaborators card with "This device is a coordinator"
     - Project Categories card with config filename (e.g., @mapeo/default-config)

3. Participant flow currently **not testable in Browserstack**

---

### Open Questions

1. **How does someone edit a project name?**
   - Currently I had the “Edit Info” button links to `ProjectSettings`, but there is no dedicated screen to edit a project name, right? Is this planned? Should I create one?

2. **Can someone still edit a device name?**
   - What’s the path for this now? The test `edit-device-name.test.ts` is commented out until this is figured out.

---
####  **Solo**

---

####  **Coordinator on a Named Project**

---

PR #1154, opened 2025-05-07
fix: add back settings

### LINKED ISSUE #1131: New Menu / Add back Device Name
**Description**
https://www.notion.so/digidem/Add-back-Device-Name-1e31b08162d58068892dfb3275c54192
https://www.notion.so/digidem/Address-Device-Name-across-Multiple-Projects-1e31b08162d580fa9c25e243cd1279e1

Device Name was removed Project Settings, and now users do not have a way to edit their Device name. Also, with the introduction of Multi-project, editing the name device-wide is ideal, instead of project-wide.

**Todo**

- [ ] Add **Device Name** button to top of **Settings/App Settings** (top of Language)
- [ ] When editing the device name, this name should be edited across all projects

### LINKED ISSUE #1136: New Menu / Add back Remote Archive
https://www.notion.so/digidem/Add-back-Remote-Archive-1e61b08162d580c6a67fd1c871c399a3

**Todo**
Remote archive ON or OFF - Coordinator

Image

- [ ] Introduce the Remote Archive button as design
- [ ] When tapping on "View details" go to the existing screen

Remote archive ON  Participant
- [ ] Display the Remote Archive button as designed, set to ON
- [ ] Card isnt clickable

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4556-2052&t=iOW2jdEc1YPY67de-0

### LINKED ISSUE #1139: Add back Test Data
Add it back to App Settings

cimigree:
closes #1131
closes #1136 
closes #1139

### Changes
- Adds back the menu item to edit a device name and puts it in App Settings at the top
   - The issue asks that the device name be tied to multiple projects, but that is already the case so nothing needs to be changed for that.
- Adds back the menu item to create test data in settings
- Adds the card within the project menu to add a remote archive 
  - that card is there for any coordinator, even one on a solo project. Not sure if that is as it should be.

### Screenshots
<image src="https://github.com/user-attachments/assets/06994fe2-5043-488e-a3f4-9beaee09f644" width="250" />
<image src="https://github.com/user-attachments/assets/bbf67dc0-a397-4633-b8a2-d481996c6b39" width="250" />

### DIFF
--- a/src/frontend/screens/Settings/ProjectSettings/index.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/index.tsx
@@ -126,27 +126,27 @@ export const ProjectSettings = () => {
           onPress={() => navigate('YourTeam')}
         />
       )}
+      {isCoordinator && (
+        <SettingsCardRow
+          icon={<ExchangeIcon width={24} height={24} />}
+          title={formatMessage(
+            remoteArchiveOn ? m.remoteArchiveOn : m.remoteArchiveOff,
+          )}
+          subtitle={formatMessage(m.remoteArchiveDesc)}
+          buttonText={formatMessage(m.viewDetails)}
+          onPress={() =>
+            navigate(remoteArchiveOn ? 'RemoteArchiveOn' : 'RemoteArchiveOff')
+          }
+        />
+      )}
       {projectInfo.role !== 'participant' && (
-        <>
-          <SettingsCardRow
-            icon={<ExchangeIcon width={24} height={24} />}
-            title={formatMessage(
-              remoteArchiveOn ? m.remoteArchiveOn : m.remoteArchiveOff,
-            )}
-            subtitle={formatMessage(m.remoteArchiveDesc)}
-            buttonText={formatMessage(m.viewDetails)}
-            onPress={() =>
-              navigate(remoteArchiveOn ? 'RemoteArchiveOn' : 'RemoteArchiveOff')
-            }
-          />
-          <SettingsCardRow
-            icon={<ProjectCategoriesIcon width={24} height={24} />}
-            title={formatMessage(m.configTitle)}
-            subtitle={configData?.configMetadata?.name}
-            buttonText={formatMessage(m.updateCategories)}
-            onPress={() => navigate('Config')}
-          />
-        </>
+        <SettingsCardRow
+          icon={<ProjectCategoriesIcon width={24} height={24} />}
+          title={formatMessage(m.configTitle)}
+          subtitle={configData?.configMetadata?.name}
+          buttonText={formatMessage(m.updateCategories)}
+          onPress={() => navigate('Config')}
+        />
       )}
     </ScrollView>
   );
--- a/src/frontend/screens/Settings/ProjectSettings/index.tsx
+++ b/src/frontend/screens/Settings/ProjectSettings/index.tsx
@@ -9,6 +9,7 @@ import {BodyText} from '../../../sharedComponents/Text/BodyText';
 import NoProjectIcon from '../../../images/NoProjectIcon.svg';
 import ProjectParticipantIcon from '../../../images/ProjectParticipant.svg';
 import ProjectCategoriesIcon from '../../../images/ProjectCategories.svg';
+import ExchangeIcon from '../../../images/Exchange.svg';
 import {
   COMAPEO_BLUE,
   NEW_DARK_GREY,
@@ -18,6 +19,7 @@ import {
 } from '../../../lib/styles';
 import {useProjectSettings} from '../../../hooks/server/projects';
 import {useNavigationFromRoot} from '../../../hooks/useNavigationWithTypes';
+import {useGetRemoteArchives} from '../../../hooks/server/projects';
 
 const m = defineMessages({
   title: {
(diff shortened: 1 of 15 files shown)


achou11 (dismissed):
Thanks for fixing!

achou11:
> that card is there for any coordinator, even one on a solo project. Not sure if that is as it should be.

I _thought_ this would be fine but I ran into an error trying to add a remote server when on the solo project. my guess is that it's because the project doesn't have a name, but will confirm that in a moment. might need to hide that card in the case of the solo project.

src/frontend/screens/Settings/ProjectSettings/index.tsx
```diff
@@ -106,13 +127,26 @@ export const ProjectSettings = () => {
...
+            icon={<ExchangeIcon width={24} height={24} />}
+            title={formatMessage(
+              remoteArchiveOn ? m.remoteArchiveOn : m.remoteArchiveOff,
+            )}
+            subtitle={formatMessage(m.remoteArchiveDesc)}
+            buttonText={formatMessage(m.viewDetails)}
+            onPress={() =>
+              navigate(remoteArchiveOn ? 'RemoteArchiveOn' : 'RemoteArchiveOff')
+            }
+          />
```
achou11: > that card is there for any coordinator, even one on a solo project. Not sure if that is as it should be.

I _thought_ this would be fine but I ran into an error trying to add a remote server when on the solo project. my guess is that it's because the project doesn't have a name, but will confirm that in a moment. might need to hide that card in the case of the solo project.

achou11 (changes requested):

src/frontend/screens/Settings/ProjectSettings/index.tsx
achou11: yeah can confirm that once I named the project, adding the remote server worked. guess this needs to be conditionally rendered based on whether the current project has a name at least

cimigree:
> * that card is there for any coordinator, even one on a solo project. Not sure if that is as it should be.

Yeah, @achou11  I was wondering about that.

src/frontend/screens/Settings/ProjectSettings/index.tsx
achou11: alternatively, can update the remote archive screen to have some UI similar to the invite screen in the case where the project doesn't have a name yet i.e. prompting the user to set a project name. that probably requires more design work so i think conditionally rendering is sufficient for now.

cimigree: > alternatively, can update the remote archive screen to have some UI similar to the invite screen in the case where the project doesn't have a name yet i.e. prompting the user to set a project name

What do you think? Is there someone I should ask about that determination?

achou11: I think for now, conditionally rendering is sufficient since that's what was done in the past:

https://github.com/digidem/comapeo-mobile/pull/1100/files#diff-895678048c685b29813a9219ff1c3487e7f1ca70a230e3bc991a23522455a4e9L81

however, I think the check should be based on the project having a name, as opposed to the number of projects we have. I don't see any reason to disallow adding a remote archive on the solo project.

cimigree: > I don't see any reason to disallow adding a remote archive on the solo project.


Hmm. I find this confusing. The thing that determines if it is a "solo" project is if it has a name... not actually if there is anyone else on the project. So, yes the check I added is if it is A Solo Project (meaning it is unnamed and still has the default name as we have it -- "My Solo Project").

achou11: Yeah maybe my use of "solo" is too confusing. My main point is that you should be able to add a remote archive regardless of how many projects exist. As long as a project has a name, adding a remote archive should be allowed.

src/frontend/screens/Settings/ProjectSettings/index.tsx
```diff
@@ -105,6 +126,19 @@ export const ProjectSettings = () => {
...
+          title={formatMessage(
+            remoteArchiveOn ? m.remoteArchiveOn : m.remoteArchiveOff,
+          )}
+          subtitle={formatMessage(m.remoteArchiveDesc)}
+          buttonText={formatMessage(m.viewDetails)}
+          onPress={() =>
+            navigate(remoteArchiveOn ? 'RemoteArchiveOn' : 'RemoteArchiveOff')
+          }
+        />
+      )}
```
achou11, line 141: this doesn't properly guard against the issue i ran into. needs a check like `isCoordinator && projectInfo.name !== undefined`

achou11 (changes requested):

src/frontend/screens/Settings/ProjectSettings/index.tsx
achou11: unless the intention is to be okay with running into the error in that case? i'd be okay with that too but need some clarity on what the intended UX is

achou11 (approved):
decided that the issue i raised is non-blocking and kind of an edge case. just a little bit confusing when you do run into it but that can be addressed separately

src/frontend/screens/Settings/ProjectSettings/index.tsx
achou11: decided that this is non-blocking. being able to go through this flow and running into the error is fine, although a bit unclear since we don't surface the actual reason right now. can be addressed separately.

cimigree:
> decided that the issue i raised is non-blocking and kind of an edge case. just a little bit confusing when you do run into it but that can be addressed separately

Did you look at `useProjectRoleandDetails` because I think your point about lack of name is accounted for in there. If it isn't, I think we need to take a look at the hook again.

achou11:
> Did you look at `useProjectRoleandDetails` because I think your point about lack of name is accounted for in there. If it isn't, I think we need to take a look at the hook again.

~guess I wasn't familiar with this hook! i think it's tangential to the issue I'm raising though, since it (fortunately) doesn't try to use a default project name if it's missing (so `projectName` could still be `undefined`).~ EDIT: I see what you're referring to in that hook. The check you added in https://github.com/digidem/comapeo-mobile/pull/1154/commits/b51030f467b0e8ccb564c29f64e79f606084aec4 is narrowing based on discriminate union provided by that hook, such that the project name is defined. That's the part that wasn't apparent to me - apologies 😄

src/frontend/screens/Settings/ProjectSettings/index.tsx
achou11, line 141: Upon some additional thinking, one thing that isn't clear to me from a UX perspective is if this screen should only be able to be accessed by a coordinator.

My initial thought is that anyone should be able to view this screen, but the actions that can be taken within it should be restricted (that screen currently does that). If that's the desired UX, this conditional check here is misaligned with that.

Issue #1088, opened 2025-04-09
Edit Info Screen


ximenabb:
**Description**
User story : User can store data from different projects in CoMapeo, keep it separate, and switch between projects

**Feature spec**
https://www.notion.so/digidem/Multiple-projects-ce7d9913220342ed968311ac9e8a523b?pvs=4

**User requirements**
See **Must Haves**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1981b08162d58065a598000c95d41deb&pvs=4

**To Do**

### Coordinators Only

<img width="308" alt="Image" src="https://github.com/user-attachments/assets/d8452172-b6ae-4c6e-8150-54395ad5c0de" />

- [ ] The “My Solo Project” cannot be edited unless it is made into a project. 
- [ ] Participants cannot edit Projects they are on.
- [ ] All new projects start off as orange until a Coordinator edits the project and selects a different Project Card Color.

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4741-2651&t=bQDpwjZbr3d6dKAq-0

PR #1156, opened 2025-05-07
chore: update horizontal scroll view calculations

ErikSin:
this is towards #1088

A horizontal scroll view is needed for 1088. On working with it, I realized that it is not properly calculating the correct size. This updates it so the size is calculated based on the min item size, the gap, and the size of the scroll view (not the size of the window).

I also changed the min size of the photo thumbnails to 100 (previously it was 150). The previous implementation was shrinking the thumbnails. With the new implementation the smaller min size keeps it a similar size to the previous implementation.

I also added the `shouldShowLastItem` prop. This allows the user to see the last photo that was taken when making an observation, but it is no longer the default behaviour (we don't want it to be the default behaviour for #1088)


cimigree (approved):
I honestly didn't realize there was anything wrong with the previous way...
But glad we have a reusable scroll view now for when we need it.
I tested it out and while the gap does appear slightly big to me, that is just a visual preference and is certainly non-blocking.
I have to admit I don't totally understand the calculations, but I did the best I could to parse them and they all seem to work.
Glad there is a new test just in case.

PR #1157, opened 2025-05-07
feat: add ability to remove remote archive

### LINKED ISSUE #1140: Add ability to Remove Server
Update UI to add a Remove Server

**Todo**

Image

- [ ] Add  **Remove Server** button

Image

- [ ] When Remove Sever tapped, display bottom sheet as designed
- [ ] When deleting Remote Archive, user sees the Remote Archive screen set to OFF
- [ ] When tapping on Cancel, user goes back to  Remote Archive screen set to ON

**Design**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=2909-1931&t=5PImYATjW2jMqw32-0

achou11:
Closes #1140

Notable changes:

- Updates `@comapeo/core-react` to v4.2.1 to introduce the necessary write hook.

- Consolidates the remote archive setting screen to a single screen. This was agreed upon to reduce the technical awkwardness that having two navigation screens caused.

- UI adjustments to various remote-archive related screens/states to more closely match the designs.

---

image

image

image

https://github.com/user-attachments/assets/b29293c1-53c4-4843-b4c3-68bb47bbaad8

### DIFF
--- a/src/frontend/hooks/server/projects.ts
+++ b/src/frontend/hooks/server/projects.ts
@@ -1,4 +1,4 @@
-import {useMemo} from 'react';
+import {type MemberInfo} from '@comapeo/core/dist/member-api';
 import {
   useProjectSettings as useComapeoProjectSettings,
   useManyMembers,
@@ -7,16 +7,7 @@ import {
 import {useQuery} from '@tanstack/react-query';
 
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
-
-export const ALL_PROJECTS_KEY = 'all_projects';
-export const PROJECT_SETTINGS_KEY = 'project_settings';
-export const CREATE_PROJECT_KEY = 'create_project';
-export const PROJECT_KEY = 'project';
-export const PROJECT_MEMBERS_KEY = 'project_members';
-export const ORIGINAL_VERSION_ID_TO_DEVICE_ID_KEY =
-  'originalVersionIdToDeviceId';
-export const THIS_USERS_ROLE_KEY = 'my_role';
-export const REMOTE_ARCHIVE = 'remote_archive';
+import {MEMBER_ROLE_ID} from '../../sharedTypes';
 
 export function useProjectSettings() {
   const {projectId} = useActiveProject();
@@ -28,19 +19,29 @@ export function useGetOwnRole() {
   return useOwnRoleInProject({projectId});
 }
 
-export function useGetRemoteArchives() {
-  const {projectId} = useActiveProject();
-  const {data: members, error, isRefetching} = useManyMembers({projectId});
+// TODO: Ideally this is handled in @comapeo/core (https://github.com/digidem/comapeo-core/issues/1031)
+export type ArchiveServerMemberInfo = MemberInfo & {
+  deviceType: 'selfHostedServer';
+  selfHostedServerDetails: NonNullable<MemberInfo['selfHostedServerDetails']>;
+};
 
-  const archives = useMemo(() => {
-    return members?.filter(m => m.deviceType === 'selfHostedServer') ?? [];
-  }, [members]);
+// TODO: Ideally this is handled in @comapeo/core (https://github.com/digidem/comapeo-core/issues/1031)
+function isActiveArchiveServerMember(
+  member: MemberInfo,
+): member is ArchiveServerMemberInfo {
+  if (member.deviceType !== 'selfHostedServer') return false;
+  if (!member.selfHostedServerDetails) return false;
+  if (member.role.roleId !== MEMBER_ROLE_ID) return false;
+  return true;
+}
 
-  return {
-    data: archives,
-    error,
-    isRefetching,
-  };
+export function useActiveArchiveServer({
+  projectId,
+}: {
+  projectId: string;
+}): ArchiveServerMemberInfo | undefined {
+  const {data: members} = useManyMembers({projectId});
+  return members.find(isActiveArchiveServerMember);
 }
 
 export function useFindRemoteArchive({url}: {url?: string}) {
--- a/src/frontend/screens/Exchange/index.tsx
+++ b/src/frontend/screens/Exchange/index.tsx
@@ -3,7 +3,7 @@ import {useSyncState} from '@comapeo/core-react';
 
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
 import {
-  useGetRemoteArchives,
+  useActiveArchiveServer,
   useProjectSettings,
 } from '../../hooks/server/projects';
 import {useLocalDiscoveryState} from '../../hooks/useLocalDiscoveryState';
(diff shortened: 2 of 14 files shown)


src/frontend/screens/Exchange/index.tsx
```diff
@@ -26,18 +26,15 @@ const m = defineMessages({
...
-
-  const hasRemoteArchive = remoteArchive && remoteArchive.length > 0;
-
   const hasInternetAccess = useNetInfo().isConnected;
 
   const {projectId} = useActiveProject();
   const syncState = useSyncState({projectId});
   const projectSettingsQuery = useProjectSettings();
 
-  if (!syncState || remoteArchiveLoading) {
```
achou11, line 40: just noting the change here. i don't think this should have a major impact on this screen, but could be wrong!

src/frontend/hooks/server/projects.ts
```diff
@@ -7,16 +7,7 @@ import {
...
-
-export const ALL_PROJECTS_KEY = 'all_projects';
-export const PROJECT_SETTINGS_KEY = 'project_settings';
-export const CREATE_PROJECT_KEY = 'create_project';
-export const PROJECT_KEY = 'project';
-export const PROJECT_MEMBERS_KEY = 'project_members';
-export const ORIGINAL_VERSION_ID_TO_DEVICE_ID_KEY =
-  'originalVersionIdToDeviceId';
-export const THIS_USERS_ROLE_KEY = 'my_role';
-export const REMOTE_ARCHIVE = 'remote_archive';
```
achou11, line 19: all of these are unused since the fuller integration of core-react!

src/frontend/screens/Settings/ProjectSettings/RemoteArchive/RemoveRemoteArchive.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const removeServerPeer = useRemoveServerPeer({projectId});
+
+  return (
+    <BottomSheetWrapper>
+      <View style={styles.container}>
+        <View style={styles.infoContainer}>
+          <Error />
+          <HeaderText variant="header2" style={{textAlign: 'center'}}>
+            {t(m.title, {name: name || baseUrl})}
```
achou11, line 62: technically the `name` can be undefined so figured it would be okay to display the base url if that's the case. open to other suggestions though

src/frontend/hooks/server/projects.ts
```diff
@@ -28,19 +19,29 @@ export function useGetOwnRole() {
   return useOwnRoleInProject({projectId});
 }
 
-export function useGetRemoteArchives() {
```
achou11, line 31: noticed that this hook was being used to figure out if there's an active remote archive member. in some cases, the usages were slightly incorrect because they didn't account for the role (`useManyMembers()` currently doesn't do _any_ filtering of members).

opted for a more focused and succinct hook (`useActiveArchiveServer()`) that better addresses the needs of the existing usages.

src/frontend/hooks/server/projects.ts
```diff
@@ -28,19 +19,29 @@ export function useGetOwnRole() {
...
-    error,
-    isRefetching,
-  };
+export function useActiveArchiveServer({
+  projectId,
+}: {
+  projectId: string;
+}): ArchiveServerMemberInfo | undefined {
+  const {data: members} = useManyMembers({projectId});
+  return members.find(isActiveArchiveServerMember);
```
ErikSin, line 44: I think this is right, but I just want to make sure. Each project should only be connected to one archive server. I know that we technically can be part of many archive servers, but we don't allow that on the client side right now, correct?

ErikSin (changes requested):
the code looks good, but i am getting stuck in a state where `removeServerPeer.status` is pending. Basically the bottoms sheet for removing the archive is stuck in the loading state. I have to close the app entirely, and when i reopen, i have been removed from the project. This makes me think that it is a server issue and not a frontend issue, but i think we should make sure of that before we merge.

I am using this url: https://comapeo-cloud.fly.dev/
and the projectId: xpdbbm9dnfp531pcxjhhaq38guhxjjz6o76y1mpybjhitc3y18wy

achou11:
> the code looks good, but i am getting stuck in a state where `removeServerPeer.status` is pending. Basically the bottoms sheet for removing the archive is stuck in the loading state. I have to close the app entirely, and when i reopen, i have been removed from the project. This makes me think that it is a server issue and not a frontend issue, but i think we should make sure of that before we merge.

yeah i was running into this as well but not consistently. core immediately attempts to connect with the server to let it know that it's been removed, so my early guess is either:

1. The server was suspended by Fly (which is sometimes aggressive about it) and it needs some time to get back online.

2. Some issue in the core implementation. Could be related to (1), such as not using a timeout, or one that's too large?

In the case of (1), do you run into the issue if the relevant Fly machine is in the deployed state before attempting to remove the server? In the Fly dashboard UI, it should have this badge in the top right corner:

image

src/frontend/hooks/server/projects.ts
achou11: Yeah that's how I understand it. Core allows you to connect to multiple archive servers, but in the client side we only support connecting to a single one.

achou11:
@RangerMauve any thoughts on the issue mentioned in https://github.com/digidem/comapeo-mobile/pull/1157#pullrequestreview-2825954425? Maybe there's something in core that needs to be adjusted 🤔

RangerMauve:
Hmm, not sure why it'd be stuck waiting. I could try reproducing on Monday to see if there's something we can do at the websocket creation level. If the server is unresponsive it might cause delays in detecting a failure.

Did it stay stuck more than 30 seconds? A timeout might be a good idea. We should be listening to socket errors [here](https://github.com/digidem/comapeo-core/blob/main/src/member-api.js#L459)

achou11:
> Did it stay stuck more than 30 seconds? A timeout might be a good idea. We should be listening to socket errors [here](https://github.com/digidem/comapeo-core/blob/main/src/member-api.js#L459)

From when I ran into it, yes. I let it run for at least a minute before needing to stop the app.

achou11:
@ErikSin are you able to repro the issue consistently? I haven't been able to do so with my fresh attempts today.

> The server was suspended by Fly (which is sometimes aggressive about it) and it needs some time to get back online.

I've been testing various scenarios related to the above and no luck getting into a stuck state. Tried the following:

1. Manually suspend the Fly machine. Results in successful removal.

2. Manually stop the Fly machine. Results in successful removal.

3. Connect to wifi network without internet access. Results in error sheet appearing instantly.

(2) seemed to require a bit longer of a wait time than (1), but it wasn't significant and I didn't run into the issue.

RangerMauve:
Sry exports have been taking longer than expected. Is this blocked by me?

achou11:
> Sry exports have been taking longer than expected. Is this blocked by me?

I think it would be if we had a way of reliably reproducing the issue (from the UI side). If you have the bandwidth to attempt at reproducing it from core, that'd be great, but otherwise I think it can be something that we follow up on if it happens more consistently.

ErikSin (approved):
yeah, this seems like a backend issue. I also don't think it blocking...like its still successfully removing, the backend is just not properly communicating it with the front end.

achou11:
> yeah, this seems like a backend issue. I also don't think it blocking...like its still successfully removing, the backend is just not properly communicating it with the front end.

Yeah agreed. Will log an issue in core to track though! 

EDIT: https://github.com/digidem/comapeo-core/issues/1034

Issue #1129, opened 2025-04-29
E2E Testing


ximenabb:
**Description**

Continue to automate existing scripts as identified:

https://www.notion.so/digidem/E2E-Cross-referencing-1c01b08162d580288393d0e2194643f3

Also, try to speed up tests at points where they hang.

For this issue:
- Sped up the end to end tests from over 30 minutes to around 11 minutes (PR https://github.com/digidem/comapeo-mobile/pull/1158)
- Changes the metadata for the tests to make them easier to understand on Browserstack (https://github.com/digidem/comapeo-mobile/pull/1170)
- Adds tests and documentation, utilities, and test ids when necessary for (https://github.com/digidem/comapeo-mobile/pull/1178)
  - Observation Metadata
  - Multiple Projects (including editing)
  - Adding a remote server

These PRs

PR #1158, opened 2025-05-12
chore: speed up e2e

cimigree:
towards #1129 

These changes get the tests down from taking over 30 minutes to taking less than 12!

**Changes included:**

*  Replaces or eliminates animations for the GPS Pill and Audio because they make the tests take forever for some reason 
*  Improved passcode input speed
   * Replaced the testing framework method with an Android method, `driver.keys()`, for reliability (esp. repeated digits like `'00000'`)
   * Made a reusable helper for that
   * Lowered implicit timeout during passcode entry

* Skipped the SVG rendering on AuthScreen during E2E

  * Disabled `<CoMapeoLogoSvg />` in test mode to reduce layout/render delays

* Improved checks for when something should not be on a screen

  * Added `checkForElementGone()` helper with reduced timeout + polling
  * Disabled implicit timeout during these checks to avoid test stalls
* Cleaned up a few places where items were being looked for that weren't there


tests/e2e/utils/enterText.ts
```diff
@@ -0,0 +1,7 @@
+export async function enterPasscodeWithKeyEvents(passcode: string) {
+  await browser.setTimeout({implicit: 1000});
+
+  await driver.keys(passcode.split(''));
+
+  await browser.setTimeout({implicit: 10000});
```
achou11: what's the purpose of updating the timeouts here?

achou11 (approved):
Nice to see the improvement in running times! Will admit that I'm a bit worried about the increasing need for test-specific changes to the UI, mostly because:

- it's easy to forget it as a consideration when fixing or adding UI
- it can be easy to cause regressions with existing exceptions (probably not as common in practice)
- tedious to figure out if an exception is needed i.e need to keep an eye out for how changes affect the test performance in Browserstack.
- diverging further and further away from the actual application. not overly concerned about this yet given the nature of the exceptions so far, but where do you draw the line?

Would be great to report the issues you've encountered related to test performance to Browserstack. Having the opportunity to have anything resembling a fix upstream would definitely be preferable.

Can you elaborate on how the implicit session timeouts this PR adjusts generally impact the outcome? My understanding comes from https://webdriver.io/docs/timeouts/#session-implicit-wait-timeout, which describes them as global-ish timeouts that wdio uses internally for functionality related to locating elements. Seems like they recommend against updating these (see https://webdriver.io/docs/autowait/#implicit-timeouts-not-recommended), so I'm curious about how doing so helps in the context of this PR.

cimigree:
In a subsequent PR I can document our test-specific flags and will work on reporting the test stability issues to BrowserStack somehow.

I took out the timeouts even though they seemed to speed things up. I have no idea why. They just seemed to make things work even though implicitTimeouts were supposedly 0; I think they maybe are reset somewhere somehow.

For now this PR drops test time from 30+ minutes to ~11, which seems worth merging as-is to unblock people and keep CI fast.

PR #1168, opened 2025-05-12
chore: always show leave project warning on invite

### LINKED ISSUE #1160: Leave project flow/ bring back
https://www.notion.so/digidem/Already-on-a-project-gets-project-invite-and-can-accept-no-warning-1f11b08162d580599d33efda56a48480?pvs=4

- Always show warning when being invite to a new project

ErikSin:
closes #1160 

Always show the leave project warning on an invite, since solo projects are now a valid project.

Not adding this to the develop branch since multi projects is being introduced right away meaning we won't need this code.

### DIFF
--- a/src/frontend/screens/Invites/InviteReceived.tsx
+++ b/src/frontend/screens/Invites/InviteReceived.tsx
@@ -7,12 +7,10 @@ import {
   useAcceptInvite,
   useRejectInvite,
   useSingleInvite,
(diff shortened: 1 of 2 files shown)


cimigree (approved):
This makes sense to me and I tested it out and it works fine. Interesting choice to use onSettled... Feels valid given your point and I am glad you left the comment.

PR #1169, opened 2025-05-12
fix: import categories while naming

### LINKED ISSUE #1159: Naming project/ Can’t import custom categories
**Describe the bug**
- User is mapping on their own, and is ready to name their project to invite others
- User goes to the Menu and taps on **+Invite**
- User taps on **Name My Project**
- User enters the name of the project
    - Tap on **Advanced Project Settings** + tap on **Import Categories**
    - User see a success screen confirming the categories were importe, tap on **OK**
- User taps on **Create Project**
    - button loads, then button returns back to **Create Project**
    - user can tap the button again and same result
- User cannot create a project
- 
**Expected behavior**
User should be able to create a project, import cat file and save project.

https://www.notion.so/digidem/Project-Creation-Can-t-import-custom-categories-1f11b08162d580da9cecc00cd2bf243f

cimigree:
Closes #1159 
This fixes an issue where importing a .comapeocat config file during project creation would cause an error and prevent navigation to the success screen — even though the categories were technically imported.

Previously, I was passing a configMetadata object to updateProjectSettings, assuming it was necessary to associate the config with the project. It turns out this isn't needed — the config is automatically applied after import (though backend confirmation on this behavior would be helpful).

The actual cause of the error was the logic attempting to extract and shape the config file for configMetadata. That code has been removed, since it's unnecessary and was blocking successful project creation UI.

Also, I tried to remove some of the repetition in the main create/update project method. Hope that makes sense.

On a different somewhat related note: the config file config-siekopai-sanpablo-v1.0.2.comapeocat used by Ximena in the video triggers a red screen error related to non-unique Mapbox category labels:
It is:
 ```
Mapbox [error] RNMBXStyleFactory | Failed to update: circleColor Set layer property "circle-color" failed:
[54]: Branch labels must be unique.
[match, [get, presetName], Animales muertos, #5c4445, Sin categoría, #0066FF, Cartucho, #db0d0d, Cazador ilegal, #f0230c, Sin Categoría, #0066FF, Sendero de cazadores, #d9023f, Tarima cazadores, #a66265, Tala ilegal, #c97f1e, Campamento ilegal, #0d6115, Asentamiento ilegal, #ff3030, Desechos de invasores, #a31303, Huellas invasores, #c7b10c, Pesca ilegal, #0662bf, Trampa pescador ilegal, #2e078a, Decontaminación / Limpieza, #3603a3, Canal de desechos, #d64a18, Mancha / Derrame de crudo, #361001, Excavadora, #6a6b1d, Piscina de desechos, #9506c4, Plataforma plataforma-petrolera, #140803, Pozo activo, #852c09, Pozo seco, #120202, Salida de desechos tóxicos, #D63E03, Trocha de sísmica, #4f3330, Tubería / Oleoducto, #70361d, Camera-trap, #653782, Sin Categoría, #0066FF, #F29D4B]
```
However, despite that, the importing still works and you can still see and use the custom categories.

### CODE REVIEW

File: src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx [hunk 1]
```diff
@@ -131,70 +130,51 @@ export const CreateProject: NativeNavigationComponent<'CreateProject'> = ({
...
+      navigation.navigate('ErrorBottomSheet');
+    };
 
     if (projectInfo.role === 'solo') {
-      if (fileUri) {
-        importProjectConfig.mutate(
-          {configPath: convertFileUriToPosixPath(fileUri)},
+      const updateSettings = () =>
+        updateSettingsMutation.mutate(
+          {name: projectName},
```
ErikSin, line 146: so, I think this is on the line of being a style preference vs an actual problem, so im happy to hear your side if you think we should keep it this way.

But i find it a little confusing that we are creating 3 functions (`onProjectCreated`, `onError`, `updateSettings`) inside `handleCreateOrUpdateProject`. It feel more straightforward to create those function outside of `handleCreateOrUpdateProject`. What do you think?

ErikSin (approved):
Good job, thanks

PR #1170, opened 2025-05-13
chore: E2E docs & config clarifications

cimigree:
Documentation for E2E exceptions. 
Some refinements in our E2E config to make builds in Browserstack have clearly viewable links to which github run they are for.

### DIFF
--- a/docs/EndToEndTests/E2ETestingNotes.md
+++ b/docs/EndToEndTests/E2ETestingNotes.md
@@ -8,7 +8,7 @@ The environment variable `EXPO_PUBLIC_E2E_TEST=true` is used in CI and for the b
 
 ### Auth Flow (Passcode Input)
 
-- `src/frontend/contexts/AuthContext.tsx`: Disables ``FlagSecureModule` during E2E tests to allow screen recording on BrowserStack.
+- `src/frontend/contexts/AuthContext.tsx`: Disables `FlagSecureModule` during E2E tests to allow screen recording on BrowserStack.
 - `src/frontend/screens/AuthScreen.tsx`: Hides the `CoMapeoLogoSvg`, which was delaying input responsiveness.
 
 ### Map Style Fallback
--- a/docs/EndToEndTests/E2ETestingNotes.md
+++ b/docs/EndToEndTests/E2ETestingNotes.md
@@ -1,0 +1,24 @@
+# Additional Notes for E2E Testing Performance on BrowserStack
+
+Here find environment-specific changes made to reduce execution time for E2E tests on BrowserStack.
+
+## E2E-Specific UI Behavior
+
+The environment variable `EXPO_PUBLIC_E2E_TEST=true` is used in CI and for the build type "test" to modify app behavior and improve test performance in the following ways:
+
+### Auth Flow (Passcode Input)
+
+- `src/frontend/contexts/AuthContext.tsx`: Disables ``FlagSecureModule` during E2E tests to allow screen recording on BrowserStack.
+- `src/frontend/screens/AuthScreen.tsx`: Hides the `CoMapeoLogoSvg`, which was delaying input responsiveness.
+
+### Map Style Fallback
+
+- `src/frontend/hooks/server/maps.ts`: Switches to a public Mapbox style instead of fetching from the internal map server, which can fail with 502 errors in CI (issue #1008).
+
+### Audio Recording
+
+- `src/frontend/screens/Audio/AudioRecording/index.tsx`: Replaces the `AnimatedBackground` with a blank `<View />` in E2E mode. This animation was causing unnecessary Appium rendering delays that seem to happen with animations.
+
+### GPS Pill
+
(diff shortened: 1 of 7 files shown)


docs/EndToEndTests/E2ETestingNotes.md
```diff
@@ -0,0 +1,24 @@
...
+
+Here find environment-specific changes made to reduce execution time for E2E tests on BrowserStack.
+
+## E2E-Specific UI Behavior
+
+The environment variable `EXPO_PUBLIC_E2E_TEST=true` is used in CI and for the build type "test" to modify app behavior and improve test performance in the following ways:
+
+### Auth Flow (Passcode Input)
+
+- `src/frontend/contexts/AuthContext.tsx`: Disables ``FlagSecureModule` during E2E tests to allow screen recording on BrowserStack.
```
achou11: super minor
```suggestion
- `src/frontend/contexts/AuthContext.tsx`: Disables `FlagSecureModule` during E2E tests to allow screen recording on BrowserStack.
```

wdio.config.js
achou11, line 1: This is the configuration that gets applied for every run that happens right? If so, the two things that I think are important to include in build/session information are: 

1. the commit associated with a run/build
2. the time in which it was triggered

I would suggest using something like the (short) commit hash as opposed to a PR number because:

- it's possible to do runs outside the context of a PR. The relevant workflow that runs these tests only does so on pull requests but that could change, which would make the currently used values a bit confusing.

- multiple runs can happen within a PR and it'd be helpful to be able to distinguish which specific commit a run is associated with.

For pull requests, I _think_ the sha of interest would be accessed in the github workflow using `github.event.pull_request.head.sha` (https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request)

wdio.config.js
```diff
@@ -15,12 +18,12 @@ const config = {
       'browserstack',
       {
         app: process.env.BROWSERSTACK_APP_URL,
-        buildIdentifier: `${process.env.BUILD_NUMBER || `build-${new Date().toISOString()}`}`,
+        buildIdentifier: prNumber,
```
achou11: this has to be unique (https://webdriver.io/docs/browserstack-service/#buildidentifier). would suggest using their built-in`BUILD_NUMBER` and using the `buildName` to provide more contextual information.

wdio.config.js
```diff
@@ -15,12 +18,12 @@ const config = {
...
         browserstackLocal: true,
         testObservability: true,
         testObservabilityOptions: {
           projectName: 'CoMapeo',
-          buildName: `CoMapeo E2E Tests - ${new Date().toISOString()}`,
+          buildName: `PR #${prNumber} – ${prTitle}`,
```
achou11: I think this should contain the more contextual info that I mentioned in a separate comment. For example:

```js
`${prTitle || 'Build'} (${gitSHA}) at ${timestamp}`
```

wdio.config.js
```diff
@@ -35,8 +38,8 @@ const config = {
       'appium:autoGrantPermissions': true,
       'bstack:options': {
         projectName: 'CoMapeo',
-        buildName: 'CoMapeo Android Build',
-        sessionName: 'Pixel 7 - Android 13',
+        buildName: `PR #${prNumber} – ${prTitle}`,
```
achou11: similar comment about updating this to use the git SHA instead of a PR number.

wdio.config.js
```diff
@@ -35,8 +38,8 @@ const config = {
...
+        sessionName: `Spec Run – ${timestamp}`,
```
achou11: would it be helpful for this to contain the git SHA?

achou11 (changes requested):
Mostly some suggestions around the changes to the webdriver config. Not overly adamant about it but I think they could help with identifying and debugging builds/runs

cimigree:
@achou11 This kind of thing is what will be visible in the left hand column now and the title for the build and tests and video:
chore: E2E docs & config clarifications 2b96519 2025-05-13T19:53:08.221Z
I took out the PR number
Screenshot 2025-05-13 at 4 22 28 PM

wdio.config.js
```diff
@@ -15,12 +18,12 @@ const config = {
       'browserstack',
       {
         app: process.env.BROWSERSTACK_APP_URL,
-        buildIdentifier: `${process.env.BUILD_NUMBER || `build-${new Date().toISOString()}`}`,
+        buildIdentifier: timestamp,
```
achou11: curious why use this instead of the what's provided by [browserstack](https://webdriver.io/docs/browserstack-service/#buildidentifier)? 

> DATE_TIME: Generates a Timestamp with every execution. Eg. 05-Nov-19:30

achou11 (approved):
nice! the documented exceptions related to perf are very helpful!

wdio.config.js
cimigree: I thought (from looking at the docs) that it did not have the time... But maybe I looked too quickly? Trying now.

PR #1174, opened 2025-05-13
fix: fix deletion of photo attachment when editing observation

### LINKED ISSUE #1172: Deleting added photo attachment while editing causes all photo attachments to disappear
**Describe the bug**

Deleting newly added photos while editing an existing observation causes all of the photo attachments to disappear.

**To Reproduce**

1. Create an observation with at least 1 photo attachment
2. Go to observations list and select the created observation
3. Edit the observation
4. Add another photo
5. Delete the photo from (4)

Note that the none of the attachments in the observation editor are present anymore.

If you leave the editing flow after (5), the original attachments are visible.
If you confirm the edits after (5), the original attachments are no longer viewable.

**Expected behavior**

Only the photo of interest is removed from the displayed list of attachments in the observation editor.

**Screenshots**

https://github.com/user-attachments/assets/3c63052c-6ee2-4d5d-8416-589a12fce159

 - Device: Pixel 6
 - OS: Android 15

**Additional context**

I could reproduce this issue on the currently published Google Play version (1.2.2), so it's been around for a while I guess.

achou11:
Fixes #1172


ErikSin:
I don't think i fully understand why the error was happening originally. Can you elaborate?

achou11:
> I don't think i fully understand why the error was happening originally. Can you elaborate?

yep! thought i had put something in the description but forgot.

Based on on the setup for the relevant zustand store, `attachments` is defined as `(Photo | Audio)[]`. If you expand `Photo` and `Audio` a bit, it's `(SavedPhoto | UnprocessedDraftPhoto | ProcessedDraftPhoto | AudioAttachment | UnsavedAudio)[]`.

This is the previous implementation's filtering logic that intends to remove a draft photo:

```ts
const newAttachments = get().attachments.filter(
  photo => 'originalUri' in photo && photo.originalUri !== uri,
);
```

Aside from the misleading variable naming of `photo` (it's really an attachment that could either be a photo or audio), only `ProcessedDraftPhoto` contains the `originalUri` field. The `'originalUri' in photo` check returns `false` for everything but that type which causes the filter check to return false (via short-circuiting). This means that any attachment that isn't of type `ProcessedDraftPhoto` is omitted from the `newAttachments` result, which is how the bug occurs.

When it comes to removing the relevant draft photo, the check should be aligned with the following statement: Given a set of attachments, if there are any that are of type `ProcessedDraftPhoto` AND it has the same `originalUri` as we're trying to find, omit them. Keep the rest.

Here's the updated check in this PR (as of https://github.com/digidem/comapeo-mobile/pull/1174/commits/b722a4635c3b6e03d538069604d02a05d9497178):

```ts
const newAttachments = get().attachments.filter(attachment => {
  // We currently only allow deletion of processed draft photos
  if (!isProcessedDraftPhoto(attachment)) return true;

  return attachment.originalUri !== uri;
});
```

We check to make sure that the attachment is of the relevant type first. If it's not, return `true` because we want to keep it. If it is, we keep it based on whether it's `originalUri` matches the uri of interest.

ErikSin (approved):
im suprised we did not catch this earlier...

PR #1175, opened 2025-05-13
test: scaffold Exchange screen test

EvanHahn:
_This PR description is based on existing understanding of things and might not be very good for "posterity"._

This is identical to the code in [the `910/setup-integration-testing` branch][0], with a few small changes:

- It is rebased against `develop`.

- A TODO was added (see dea713f35a96030b48f3dc35c96c86d75bbf1bd0).

- `@digidem/types` is included all the time, not just in integration tests (see 80a1911035dc90da795bf449c0460ca7fd729204). This is the most controversial, but I believe it's a good idea. `@digidem/types` includes types for various modules, which I think is marginally helpful to have across the app. For example, if we ever pull in `random-access-storage` into the main app, we won't have to install additional types.

Partially addresses #910.

Once merged, we should be able to delete the existing `910/setup-integration-testing` branch.

[0]: https://github.com/digidem/comapeo-mobile/tree/910/setup-integration-testing


src/frontend/App.tsx
```diff
@@ -158,22 +166,28 @@ const App = () => {
...
-      }
-      activeProjectIdStore={persistedActiveProjectIdStore}
-      metricsDiagnosticsStore={persistedMetricsDiagnosticsStore}
-      localeStore={persistedLocaleStore}>
-      <AppNavigator permissionAsked={permissionsAsked} />
-    </AppProviders>
+    <LocaleStoreProvider value={persistedLocaleStore}>
+      <IntlProvider>
+        {/* ServerLoading requires internationalization to be set up */}
+        <ServerLoading messagePort={messagePort}>
```
achou11: noting that i'm not happy with is the being moved (something I did). reason for doing so was because it requires access to the nodejs-mobile bridge, which made the usage of the other providers trickier in the context of the tests we're trying to support. if there's a straightforward way of mocking out the bridging details that it relies on, we can consider moving it back to where it was before. 

also should test that the relocation doesn't result in any significant regressions in the app (most likely startup). before this change, it was placed deeper in the tree and since it conditionally renders internally, moving it can have subtle effects on how things are set up.

achou11: separately, if there's a way we can get rid of this altogether, that'd be even better! this was something that was carried over from mapeo mobile but there may be better approaches for it that we just never considered exploring.

src/frontend/App.tsx
```diff
@@ -158,22 +166,28 @@ const App = () => {
...
-      manualEntryCoordinateFormatStore={
-        persistedManualEntryCoordinateFormatStore
-      }
-      activeProjectIdStore={persistedActiveProjectIdStore}
-      metricsDiagnosticsStore={persistedMetricsDiagnosticsStore}
-      localeStore={persistedLocaleStore}>
-      <AppNavigator permissionAsked={permissionsAsked} />
-    </AppProviders>
+    <LocaleStoreProvider value={persistedLocaleStore}>
+      <IntlProvider>
```
achou11, line 183: related to my comment about the ServerLoading, it's unfortunate that these are not in AppProviders.

tests/integration/helpers/react.tsx
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  const persistedManualEntryCoordinateFormatStore =
+    createManualEntryCoordinateFormatStore({
+      persist: true,
+    });
+
+  const persistedActiveProjectIdStore = createActiveProjectIdStore({
+    persist: true,
+  });
```
achou11: unless I'm misunderstanding how Jest does isolation, i think using `persist: true` will eventually be prone to issues because they rely on a singleton instance of MMKV. See #1073 .

`createSecurityStore` is the exception here since it uses expo-secure-store, which handles the instance setup internally.

achou11 (commented):
No objections to the inclusion of `@digidem/types` at the root level.

src/frontend/App.tsx
EvanHahn: Could `<ServerLoading>` mock out the `nodejs-mobile-react-native` calls it needs? Something like this?

```diff
diff --git a/src/frontend/ServerLoading.tsx b/src/frontend/ServerLoading.tsx
index 0a15ff8b..540c1f5f 100644
--- a/src/frontend/ServerLoading.tsx
+++ b/src/frontend/ServerLoading.tsx
@@ -8,14 +8,20 @@ import {FatalError} from './screens/FatalError';
 
 export const ServerLoading = ({
   messagePort,
+  requestServerStatus,
+  subscribeToServerStatus,
   children,
-}: React.PropsWithChildren<{messagePort: MessagePortLike}>) => {
+}: React.PropsWithChildren<{
+  messagePort: MessagePortLike,
+  requestServerStatus: () => unknown;
+  subscribeToServerStatus: (msg: StatusMessage) => (() => unknown);
+}>) => {
   const [serverStatus, setServerStatus] = React.useState<StatusMessage>({
     value: 'STARTING',
   });
 
   React.useEffect(() => {
-    const subscription = nodejs.channel.addListener('server:status', msg => {
+    const unsubscribe = subscribeToServerStatus(msg => {
       if (msg.value === 'STARTED') {
         messagePort.start();
       }
@@ -24,11 +30,10 @@ export const ServerLoading = ({
     });
 
     // In case the server starts before us (we miss the original
-    // `server-started` event), prompt the server to re-send.
-    nodejs.channel.post('get-server-status');
+    // "server started" event), prompt the server to re-send.
+    requestServerStatus();
 
-    // @ts-expect-error - incorrect types on nodejs.channel
-    return () => subscription.remove();
+    return unsubscribe;
   }, [messagePort, setServerStatus]);
 
   // Don't render any children while the backend is starting - this avoids
```

tests/integration/helpers/react.tsx
EvanHahn: Stopped persisting in a611fce087fe08e1610eeecbc477321bbfadc689.

EvanHahn:
I believe I've responded to comments, and I rebased against the latest `develop`.

src/frontend/App.tsx
achou11: yeah that's probably sufficient for now!

EvanHahn: Done in a49781c551b874d8383ddca736690b7c7c1a729a. (This change is _slightly_ different because it might fire more `requestServerStatus`es than before due to the `useEffect` callback. Also, I used `removeListener` instead of `subscription.remove()` because it sidesteps a `ts-expect-error`. I think that's fine but let me know if that's an issue.)

src/frontend/App.tsx
EvanHahn: Let me know if you'd like me to fix that as part of this PR.

EvanHahn:
Looks like I broke something. I'll take a look in the next couple of days.

EvanHahn:
Hoping to have time to do this in the next few days.

src/frontend/App.tsx
EvanHahn: Made a followup to fix this in 4929096129f72357780c96cfc7c40a3d1c6f0c49.

EvanHahn:
CI is back to green!

EvanHahn:
Rebased against `develop`.

src/frontend/App.tsx
achou11: think it's okay to do as a potential follow-up if it turns out that these are needed for more advanced tests to run. briefly looked into it and it becomes a slightly larger piece of work in order for `ServerLoading` to be more test compatible (due to need for message port)

src/frontend/App.tsx
```diff
@@ -158,22 +179,31 @@ const App = () => {
...
+        {/* ServerLoading requires internationalization to be set up */}
+        <ServerLoading
+          messagePort={messagePort}
+          requestServerStatus={requestServerStatus}
+          subscribeToServerStatus={subscribeToServerStatus}>
```
achou11, line 188: just highlighting that moving this here _may_ have user-facing consequences due to it conditionally rendering internally. from testing it, i haven't noticed anything too drastic but noting just in case a more notable regression is found.

src/frontend/screens/Exchange.test.tsx
achou11, line 1: running this test alone takes more than a minute on my machine 😬 (`npx dotenv -- jest src/frontend/screens/Exchange.test.tsx`). probably an optimization to look into as a follow-up

tests/integration/helpers/react.tsx
```diff
@@ -0,0 +1,142 @@
...
+        persistedDrafObservationStore={persistedDraftObservationStore}
+        metricsDiagnosticsStore={persistedMetricsDiagnosticsStore}
+        securityStore={persistedSecurityStore}
+        manualEntryCoordinateFormatStore={
+          persistedManualEntryCoordinateFormatStore
+        }
+        coordinateFormatStore={persistedCoordinateFormatStore}
+        trackStore={persistedTrackStore}>
+        {children}
+      </AppProviders>
```
achou11, line 132: not necessary to add in this PR, but we _may_ need to wrap this with the intl-related providers (similar to what's in `App.tsx`)

achou11 (approved):
noted a few non-blocking comments but overall think it's ready to be merged an iterated on. feel free to merge when you're ready!

EvanHahn:
Thanks for the review! I'll get this merged in the next few days.

tests/integration/helpers/react.tsx
EvanHahn: I'm going to omit that for now, but maybe it'll become necessary quickly.

EvanHahn: Turns out it was necessary! See #1205.

src/frontend/screens/Exchange.test.tsx
EvanHahn: Looked into this in #1246.

Issue #1043, opened 2025-03-17
Exposing Photo Metadata


ximenabb:
**Description**
To introduce Photo Metadata, we are updating the current View Photo screen which a person gets to anytime they tap on a Photo Thumbnail in an Observation.

**Feature Spec**
https://www.notion.so/digidem/Exposing-Observation-Data-GPS-Photos-6f9cd1cf9f0e40ed8eda96b6a70f3aa3?pvs=4

**User Requirements**
https://www.notion.so/digidem/1881b08162d580369a76fe6903715cf8?v=1c81b08162d58099b1bd000c34450044&pvs=4

**To Do**
<img width="256" alt="Image" src="https://github.com/user-attachments/assets/223d91d3-0e9d-4431-9d77-80796e6d94f4" />

- [ ] Open the photo detail view when tapping photo thumbnail
- [ ] When viewing the photo metadata, user can go back to the Observation by tapping on the Back icon
- [ ] In photo detail view, display the following
- Header display the copy *Photo Info*
- Display *Validated by CoMapeo*
- Date and timestamp
- GPS coordinates of the photograph
- Device meta data (device type, camera details including aperture, and photo size)
- Observation ID
- Device ID
- [ ] Cannot tap the items listed

See design notes in Notion : https://www.notion.so/digidem/Exposing-Observation-Data-GPS-Photos-6f9cd1cf9f0e40ed8eda96b6a70f3aa3

**Design**
https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4249-2489&t=3QL9RmVl7rzLHWgP-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=2081%3A1527

achou11:
not really sure what the technical criteria is for determining the validity of the image. is that something that's well-defined? i noticed the counterpart state which is manually uploaded, but not really sure how that fits into the current state of how core works.

achou11:
Aldo's provided an updated [Figma link](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4671-2810&t=T4b7uUblkEAOMYID-0) to account for some feedback I've provided about the availability of some of the exposed metadata

achou11:
[...]

PR #1179, opened 2025-05-14
feat: update observation photo preview screen

achou11:
Towards #1043

I'm expecting there to be a few more changes needed in this PR before it can be merged, but clarification is needed in a few places.

- Most of the metadata mentioned in the issue is not exposed because we need to upgrade core to account for those newly added fields. That will be done in a separate PR. For now, this PR is mostly focused around the UI changes for the relevant screens, and we render as much of the necessary info as possible for the time being.

- The designs didn't seem to indicate any changes to how deletion of a draft photo works. I've kept it the same as what currently exists (button in the top-right of the screen). The major difference is the implementation of the confirmation flow, which is now a separate navigation screen, which aligns with other similar flows.

- If you're testing the deletion of draft photos, you may run into #1172. That's being fixed in #1174.

- Still not sure what "validated by CoMapeo" means or the proper way to calculate that. Appreciate any insights.

---

### Observation create view

- View draft photo

    observation-create_processed-draft-photo

- Draft photo deletion interaction 

    observation-create_processed-draft-photo-delete-confirmation

    https://github.com/user-attachments/assets/05c62871-6c80-405a-ae95-33038ea49e5a
    
### Saved observation view

- Viewing saved observation photo 

    view-observation_edit-observation_saved-photo

### Edit observation view

- View draft photo

    observation-edit_processed-draft-photo

### Error placeholders

- Draft photo
    
    error-placeholder_processed-draft-photo

- Saved photo

    error-placeholder_saved-photo


src/frontend/sharedTypes/navigation.ts
```diff
@@ -43,7 +43,16 @@ export type RootStackParamsList = {
   CoordinateFormat: undefined;
   Experiments: undefined;
   PhotoPreviewModal: {
+    createdByDeviceId?: string;
+    observationDocId?: string;
     photo: SavedPhoto | ProcessedDraftPhoto;
+    validatedByCoMapeo?: boolean;
+  };
+  ConfirmDeletePhoto: {
+    onSuccess?: () => void;
```
achou11, line 50: doing this aligns with our agreed pattern of allowing callback nav params for certain kinds of screens.

In a separate branch, I experimented with a non-callback based approach that seems to work, and aligns with some of the suggestions that React Navigation provides to avoid callbacks as nav params.

https://github.com/digidem/comapeo-mobile/commit/a4eb915d6914716108caa43cb7aa620c40fae6ca

Not sure if it's really that much better, but it works and isn't overly complicated.

src/frontend/sharedTypes/navigation.ts
```diff
@@ -43,7 +43,16 @@ export type RootStackParamsList = {
   CoordinateFormat: undefined;
   Experiments: undefined;
   PhotoPreviewModal: {
+    createdByDeviceId?: string;
```
achou11: I opted to make this a nav param as opposed to something that this screen fetches internally, with the intention that this screen does as little work as possible when it comes to fetching data.

Open to changing but I think it's generally worked out okay so far 😄

src/frontend/sharedComponents/Images/TrulyContainedImage.tsx
achou11, line 1: this was painful to figure out but pretty happy with the result! React Native's behavior for `Image`s with `resizeMode="contain"` is pretty rigid, and doesn't account for dynamically sized containers/images very well. In most cases, it will add padding to fill-in the spaces that are unoccupied in the container due to the resizing of the image, which is undesirable in our case.

src/frontend/sharedComponents/Images/GracefulImage.tsx
achou11, line 1: Got annoyed with the amount of repetition that was needed to elegantly handle image errors. Decided to create this abstraction to make it easier to implement.

Open to name changes. This was unfortunately the best I could come up with...

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -131,7 +134,45 @@ export const Editor = ({
...
+                        photo={att}
+                        onPress={() => {
+                          if (!observationId) {
+                            navigate('PhotoPreviewModal', {
+                              photo: att,
+                            });
+                            return;
+                          }
+
+                          // TODO: Ideally use the core-react hooks but the error handling makes it tricky
```
achou11: didn't want to go down a rabbithole of making changes to something like the Editor component to make this more elegant. in the context of what this is for, I think this is generally okay for now.

alternatively, this could be removed if we decide to make the PhotoPreviewModal fetch the `createdByDeviceId` value internally as opposed to supplying it via a nav param. Wasn't too keen on doing that as an initial implementation, but open to considering it if necessary.

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -131,7 +134,45 @@ export const Editor = ({
...
+                                validatedByCoMapeo: true,
+                              });
+                            })
+                            .catch(() => {
+                              // TODO: Does it make sense to navigate to the screen if we fail to retrieve `createdByDeviceId`?
+                              navigate('PhotoPreviewModal', {
+                                photo: att,
+                                observationDocId: observationId,
+                                validatedByCoMapeo: true,
+                              });
```
achou11: Highlighting this. not sure how often this would happen and I figured navigating to the screen with some missing info is better than not doing so at all.

separately, should this log to sentry if this happens?

src/frontend/sharedTypes/navigation.ts
```diff
@@ -43,7 +43,16 @@ export type RootStackParamsList = {
...
+    observationDocId?: string;
     photo: SavedPhoto | ProcessedDraftPhoto;
+    validatedByCoMapeo?: boolean;
```
achou11: depending on how this is determined, may not need to be a nav param. but my lack of certainty has led me to use this as such for now.

achou11:
Just realized that one thing I haven't tested yet is viewing synced photos, especially from the device that receives them. Most notably should test that if the media sync settings are set to previews only, that rendering the photo in the preview screen works as expected.

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  projectId,
+  testID,
+  ...otherImageProps
+}: Props) => {
+  const [displayedVariant, setDisplayedVariant] =
+    useState<PhotoVariant>('original');
+
+  const {
+    data: attachmentUrl,
+    error,
```
achou11: does using this error actually work? i'm under the impression that in the case of the initial query failing, the use of a suspense hook here is actually going to propagate to the nearest error boundary (for which we have none!!!), meaning that this component won't be rendered.

achou11: ~just tested this by setting the "variants" to fetch to some nonsense and it works as intended! kind of surprised because i thought react query would've attempted to trigger an error boundary 🤔~ EDIT: nevermind, changing the values at the hook level isn't testing what I'm asking about. to properly test, I need to simulate core throwing an error when trying to build the attachment url

src/frontend/sharedTypes/navigation.ts
cimigree: I like the way it is in here better if that is what you are asking.

src/frontend/sharedTypes/navigation.ts
cimigree: The validated by Comapeo? In Observation Metadata it is a given if the location is known and determined during taking the observation, not when it is added by the user. Maybe that will be the same thing here?

src/frontend/sharedTypes/navigation.ts
cimigree: Will all 5+ (not sure how photo metadata will be achieved/ received/ broken down) properties be passed in here also for the same reason? That seems like a lot...

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
cimigree: Ok I don't think I can do that.

src/frontend/sharedComponents/Editor/index.tsx
cimigree: All of this is happening in onPress? Seems difficult to read and a little fragile. You are only catching errors at one level. Is this the best way?

src/frontend/screens/PhotoPreviewModal.tsx
```diff
@@ -1,133 +1,174 @@
...
+          <View style={{flex: 1, borderRadius: 10, overflow: 'hidden'}}>
+            <GracefulImage
+              src={photo.originalUri}
+              renderError={() => <ImageErrorPlaceholder />}
+            />
+          </View>
+        )}
+      </View>
+      <View style={{gap: 20}}>
+        {validatedByCoMapeo && (
```
cimigree: On my two devices, the photo looks good but the data properties don't quite align with the icons. Can you fix that?
<image src="https://github.com/user-attachments/assets/ffdd1d52-0d4a-4cb3-bdf1-7a5a336212fb" width="250" />

src/frontend/sharedComponents/Images/GracefulImage.tsx
cimigree: Frankly I find the wrappers around wrappers around wrappers a little confusing. I assume you had to do it this way? If there is a way to pull at least one layer back on the wrappers, I think that would be preferable, more readable, easier to follow... Not sure about testing though. Frankly I am not really sure what this component is doing so I can't really hazard a suggestion as to naming.

cimigree (approved):
I can't seem to find the sync settings (preview, etc that you asked about). I think they got accidentally removed from the new menu? Oops?

Are landscaped photos supposed to work? I never tried before so not sure. See the attached photo with icon misalignment.

I really don't know what to say about this. I find the logic confusing and the wrappers difficult to trace and understand, but perhaps that is the best architecture? I don't honestly know. Perhaps it is fine for now and when we get the metadata, it could be refactored? I feel like you know a lot more about how the app works and maybe it has to be this complicated.

achou11:
> I can't seem to find the sync settings (preview, etc that you asked about). I think they got accidentally removed from the new menu? Oops?

Yeah it was removed, I temporarily added it back locally to test what I was interested in 😄 

> Are landscaped photos supposed to work? I never tried before so not sure. See the attached photo with icon misalignment.

Based on the designs, yes! However, I think the fact that it doesn't show in landscape in this preview is more related to how the photo is taken because we don't allow configuring the rotation before saving the image. In other words, I think how it gets displayed in this screen is outside the scope of this PR because it's information that's encoded when you take the photo.

> I really don't know what to say about this. I find the logic confusing and the wrappers difficult to trace and understand, but perhaps that is the best architecture? I don't honestly know. Perhaps it is fine for now and when we get the metadata, it could be refactored? I feel like you know a lot more about how the app works and maybe it has to be this complicated.

I think this is a fair criticism! Can look into removing some of the abstractions (e.g. the graceful image component), if that's what you're referring to.

src/frontend/sharedComponents/Images/GracefulImage.tsx
achou11: Yeah I think we can afford to get rid of this abstraction if preferable. It was mainly introduced to introduce a consistent way of displaying images that may fail to load (e.g. anything that relies on a network).

Without this, every image that could potentially fail to load will need similar boilerplate of adding an `onError()` handler for the image, and then a stateful variable to indicate that an error occurred and something else should be rendered.

src/frontend/sharedTypes/navigation.ts
achou11: I suspect that the photo metadata will live within the `photo` nav prop, so no additional nav params should be needed to access those when they're introduced.

src/frontend/sharedComponents/Editor/index.tsx
achou11: the top level `catch` should also handle catches from the internal promise (note the returning of the internal promise i.e. `return projectApi.$originalVersionIdToDeviceId(...`).

whether this is the best way is debatable! like the comment mentions, i think it's partially a symptom of how this component is set up, but maybe moving the fetching to the photo preview screen is what we should do anyways.

achou11:
> Based on the designs, yes! However, I think the fact that it doesn't show in landscape in this preview is more related to how the photo is taken because we don't allow configuring the rotation before saving the image. In other words, I think how it gets displayed in this screen is outside the scope of this PR because it's information that's encoded when you take the photo.

To follow-up on this, I guess we could try to be clever and decide whether or not to rotate the image based on something like the dimensions, but I think that's kind of fragile (how do you know which way to rotate??). Would rather we fix the source of the issue, which I believe is related to the photo-taking implementation.

src/frontend/sharedTypes/navigation.ts
achou11: Ah just found the note in notion that provides some insight on this:

> Validated by CoMapeo (data is data that has been collected automatically and not edited)  accompanied by the Check Dot icon

honestly, still not quite sure what that means on a technical level. Will ask for clarification and post here

achou11: From Aldo:

> To my understanding any photo taken using CoMapeo is considered Validated.
> In the instance (I think in the future because this is not possible today) that a photo is uploaded manually than we can't consider it Validated. We will instead show "Uploaded Manually".

Which makes more sense, but as far as I know, there's currently no definitively way to derive this information from a technical level (e.g. a metadata field like `takenByCoMapeo`).

> In Observation Metadata it is a given if the location is known and determined during taking the observation, not when it is added by the user. Maybe that will be the same thing here?

I think it's different in this case because it's referring to the validity of the photo, not the observation. As mentioned above, there can be a future where some of the photos associated with an observation may not have been created via a CoMapeo application, which is the thing that this UI is trying to highlight I guess.

achou11: based on discussions, seems like we're going to go with checking that the photo has a timestamp and location as part of the metadata for now. this definitely has its limitations (that I've raised concerns about) but we don't really have the pieces in place to do this check more properly.

src/frontend/screens/PhotoPreviewModal.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/1179/commits/13823bd9eae2ab7533d85ce1483dd883822cd73c

src/frontend/sharedComponents/Editor/index.tsx
achou11: ended up moving the loading of this info into the screen. see https://github.com/digidem/comapeo-mobile/pull/1179/commits/f7961cf39c5b0eafb3b4c2d5457efa244241d0c4

src/frontend/sharedTypes/navigation.ts
achou11: implemented ^ via https://github.com/digidem/comapeo-mobile/pull/1179/commits/713efd12e2b687377ccca4b4c23e48f460a4d766

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
achou11: came to the conclusion that attempting to use the `error` here doesn't work as intended in the case of the initial query resulting in an error. Ended up moving the error handling to the consuming component and using error boundaries. see https://github.com/digidem/comapeo-mobile/pull/1179/commits/a660e64bd62f50a8318adf439093cd694d2e61ef

for future reference, a quick way to force test suspense errors is to implement a bad query in the component. for example, I added this to the component that uses the core-react hooks:

```ts
const badQuery = useSuspenseQuery({ queryKey: ['bad'], queryFn: () => throw new Error('Failed') })
```

This allowed me to test how the error boundaries I added are used.

src/frontend/screens/ConfirmDeletePhoto.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {type NativeRootNavigationProps} from '../sharedTypes/navigation';
+
+const m = defineMessages({
+  title: {
+    id: 'screens.ConfirmDeletePhoto.title',
+    defaultMessage: 'Delete this photo?',
+  },
+  deleteImage: {
+    id: 'screens.ConfirmDeletePhoto.deleteImage',
+    defaultMessage: 'Delete Image',
```
achou11: Should I use `Photo` instead of `Image` here? This is from the existing language but feels slightly inconsistent with the wording we're using elsewhere

achou11:
@cimigree think I addressed most of the feedback! Regarding some of the component abstractions, decided to remove the `GracefulImage` one but kept the rest

src/frontend/screens/PhotoPreviewModal.tsx
cimigree: FYI, They align with regular size text and display but not with enlarged display.
<image src="https://github.com/user-attachments/assets/625c6b30-4ca5-4cb8-ac1a-445a5538ac8e" width="250" />

src/frontend/screens/ConfirmDeletePhoto.tsx
cimigree: Yes I think so! Photo makes more sense to me to be consistent.

cimigree:
> @cimigree think I addressed most of the feedback! Regarding some of the component abstractions, decided to remove the `GracefulImage` one but kept the rest

Thanks that is helpful to drill down to only one to two additional layer(s) of abstraction.
Will be eager to see this with all of the metadata.

src/frontend/screens/ConfirmDeletePhoto.tsx
achou11: addressed via https://github.com/digidem/comapeo-mobile/pull/1179/commits/afeb6f13b9e9fe2b6fd74632e4d10f19fbf4ce4b

src/frontend/screens/PhotoPreviewModal.tsx
achou11: interesting - seems to vary between devices...i'm not noticing as much of a misalignment (if any) on my Pixel devices:

- Pixel 6, with max font size:

  image
  
- Pixel 2, with max font size:

    image

But with my Samsung, it's definitely looking similar to yours in terms of misalignment:

- Samsung, with max font size:

    image

---

Not entirely sure what a good fix is. Could do some math to adjust the padding of the icons based on the font multiplier I guess...

achou11: Think I figured it out! See https://github.com/digidem/comapeo-mobile/pull/1179/commits/edc77f16a57376279874bfd13f6f86cdf181d252

- Pixel 6 (max text size):

    image

- Pixel 2 (max text size):

    image

- Samsung (max text size):

    image

PR #1183, opened 2025-05-15
feat: block user from switch projects while tracks active

ErikSin:
image

### DIFF
--- a/src/frontend/screens/AllProjects.tsx
+++ b/src/frontend/screens/AllProjects.tsx
@@ -12,6 +12,7 @@ import {SecondaryButton} from '../sharedComponents/Buttons';
 import AddProjectIcon from '../images/AddProject.svg';
 import {FlatList} from 'react-native';
 import {useTracking} from '../hooks/useTracking';
+import {useActiveProject} from '../contexts/ActiveProjectContext';
 
 const m = defineMessages({
   navTitle: {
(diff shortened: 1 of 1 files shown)


src/frontend/screens/AllProjects.tsx
```diff
@@ -44,6 +48,15 @@ export const AllProjects: NativeNavigationComponent<'AllProjects'> = () => {
         data={data}
         renderItem={({item}) => {
           function handlePress() {
+            if (currentProjectId === item.projectId) {
+              popTo('Menu');
+              return;
+            }
```
achou11, line 54: out of curiosity: is this generally a desired behavior? I notice that something similar is done in other parts of the app but it's always caught me off guard a bit 😅 guess my expectation is that nothing should happen or that it shouldn't be pressable in the first place (with the UI reflecting the reason quite clearly)

achou11 (approved):

src/frontend/screens/AllProjects.tsx
ErikSin: I kind of made this decision unilaterally, but based on previous patterns.

My reasoning was that we explicitly never have disabled button. Buttons either dissappear (such as when loading) or cause an error state if the intended action cannot happen. But they never have a disabled state

There also is no indicator as to what project your on, on this page. So i thought it could be a little confusing to have a disabled button but no indicator that its disabled because you were on that project already.

Based on that, I thought this pattern was the simplest!

PR #1187, opened 2025-05-19
chore: tracking using location context

ErikSin:


cimigree:
I think this is a superb idea in general. Just have some questions/ comments about some specifics and usage
- Permissions:
  - the value is updated in the locationProvider, but should it be checked (like in UserTooltipMarker and MapScreen)?
- ThrottledLocation
  - I am unclear about its purpose as it does not seem to be used? Or is it going to be used in the GPS Pill? If it is not used directly in the components, do we need it? Or am I missing something (which is highly possible!)?
- SharedLocationContext
  - Are we deleting this and changing its usage in the Tabs and MapScreen too?

PR #1189, opened 2025-05-19
fix: update UI for GPS pill

### LINKED ISSUE #1107: Update to GPS pill
**Description**
Minor changes to the GPS pill.

**Todo**
No Gps

Image

Searching for GPS

Image

GPS with metric
Image
**Todo**

- [ ] Update the 3 states of the GPS pill as designed

achou11:
Fixes #1107 

Currently, this PR strictly focuses on the styling of the pill itself. It does not update its positioning relative to the other UI components to match the reference designs. Can consider doing that if desired.

---

- Loading:

    image

- Error:

    image

- Good:

    image

- Stacked view (just to get an understanding of size + spacing consistency):

    image


src/frontend/lib/styles.ts
achou11, line 1: the updated colors here are only used by the `GPSPill` component, so figured it could make sense to update these values for future usage elsewhere.

src/frontend/sharedComponents/GPSPill.tsx
```diff
@@ -19,70 +28,104 @@ type GPSPillProps = {
...
+      const isE2E = process.env.EXPO_PUBLIC_E2E_TEST === 'true';
+
+      backgroundColor = DARK_GREY;
+      text = '--';
+      icon = (
+        <View testID={props.iconTestID}>
+          <UIActivityIndicator
+            // Animations seem to cause issues with test performance on Browserstack
+            animating={!isE2E}
+            hidesWhenStopped={false}
```
achou11, line 68: Testing to see if this approach can be used to avoid the animation perf overhead that animations have in browserstack, without needing to more drastically conditionally render something like what was being done prior.

ErikSin (approved):
Looks good!

PR #1191, opened 2025-05-20
feat: save relevant EXIF data for photos that are taken 

achou11:
Towards #1043 

Extracts and persists the relevant EXIF fields from the captured photo. As of now, this PR is solely focused on storing that information. Was planning to do a stacked PR to display that information in the photo preview screen as a follow-up.

- Updates core + schema to versions that support the `photoExif` field on observation attachments

- Introduces a small refactoring of the hooks used for creating attachments. Before, we had a single `useCreateAttachment()` hook that was handling both audio files and photos. This PR replaces that with two separate hooks: `useCreatePhotoAttachment()` and `useCreateAudioAttachment()`, which provides more isolated implementations to account for the differing types when creating attachments. Also introduces more thorough error handling in the areas that are affected by these changes.


src/frontend/hooks/server/media.ts
```diff
@@ -1,67 +1,76 @@
...
+        driveDiscoveryId: blob.driveId,
+        hash: blob.hash,
+        name: blob.name,
+        type: blob.type,
+      };
+
+      // Should not happen but just in case...
+      if (blob.type !== 'photo') {
+        return baseAttachment;
+      }
```
achou11, line 43: wondering if I should handle this differently. In most cases, this won't happen but to be respectful of TypeScript types, I have this for now.

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -151,7 +163,7 @@ export const ObservationCreate = ({
     [addNewTrackLocations, addNewTrackObservation, value],
   );
 
-  const createObservation = React.useCallback(async () => {
```
achou11, line 154: removing the top-level async to account for some of the errors that are thrown early in this function, which don't need to be wrapped with async. similar to what #1016 addresses, although in a different context

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -86,13 +92,32 @@ export const CameraView = ({onAddPress}: Props) => {
...
         if (!pic) return;
+
+        let mediaMetadata: MediaMetadata = {
+          location,
+          timestamp: Date.now(),
+        };
+
+        if (pic.exif) {
+          try {
+            const extractedExif = parse(PhotoEXIFSchema, pic.exif);
```
achou11, line 113: since it's not immediately clear an initial glance, when valibot parses a schema defined with `v.object()`, it removes unknown fields from the parsed result, which is desired in this situation.

refer to the valibot docs to understand more about `object` (compared to `looseObject` and `strictObject`): https://valibot.dev/api/object/

src/frontend/lib/exif.ts
```diff
@@ -0,0 +1,28 @@
...
+  ISOSpeedRatings: v.optional(v.number()),
+  ImageLength: v.optional(v.number()),
+  ImageWidth: v.optional(v.number()),
+  Make: v.optional(v.string()),
+  Model: v.optional(v.string()),
+  Orientation: v.optional(v.number()),
+  ShutterSpeedValue: v.optional(v.number()),
+} satisfies Required<{
+  [field in keyof PhotoEXIF]: v.GenericSchema<PhotoEXIF[field]>;
+}>) satisfies v.GenericSchema<PhotoEXIF>;
```
achou11, line 29: the first `satisfies` is used to make sure that the object schema we're defining accounts for all of the fields that are defined by the relevant `@comapeo/schema` type. for example, if we forgot to add the schema definition of `ApertureValue`, we'd get a type error saying that the field definition is missing.

the second one is ensuring that the schema definition produces an output that matches the actual type. for example, if we defined `ApertureValue` as just `v.number()`, we'd get a type error saying that the output type doesn't match the type from `@comapeo/schema`.

pretty happy I figured this out!

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -82,17 +89,58 @@ export const CameraView = ({onAddPress}: Props) => {
...
+            GPSAltitude:
+              typeof location.coords.altitude === 'number'
+                ? location.coords.altitude
+                : undefined,
+            // TODO: Should we also write `GPSDateStamp` and `GPSTimeStamp
+          } satisfies {
+            [key in Extract<keyof PhotoEXIF, `GPS${string}`>]?: PhotoEXIF[key];
+          },
+        }
+      : BASE_CAMERA_CAPTURE_OPTIONS;
```
achou11: definitely want some visibility on this. basically, not all devices will add gps-related EXIF fields when taking a photo. See the expo-camera commit that's referenced for some additional context.

instead, we make sure that those fields are written to if we have a location to work with.

need some feedback about whether this is sufficient or if we need to write to more fields, e.g. the gps date stamp and time stamp

ErikSin: Do you know if this overwrites the exif data if it does exist? Or does it only insert it when it doesn't exist?

achou11: Good question: I don't know for sure but my guess is that it would overwrite it, which probably isn't ideal. I did think about this but there's no effective way of making sure that this data exists within the image otherwise, because we can't conditionally apply this before the picture is created.

open to thoughts about the potential overwriting is okay. i think in the case of phones taking photos, i'd be curious to know how many actually include the gps info by default because from what i can gather, it's almost never included.

achou11: hard to tell from the discussion around it, but it seems like `expo-camera` deliberately omits fields related to the GPS due to "system limitations", so it might be the case that for the fields of interest here, there wouldn't be existing values to overwrite 🤷

ErikSin: The more I think about this the more i don't think it actually matters. The point of this feature is to give some validity to our photos. Whether the GPS comes from the "pure" exif data or whether it comes from this `additionalExif` data doesnt really change the validity of the data. We can easily show that the `additionalExif` is still valid data that was collected at the time of taking the photo. And i imagine (though im not actually sure) that the "pure" exif data, if it exists, gets the information from the phones native gps anyway - aka the same source.

ErikSin: So yeah I think i would add `GPSDateStamp` and `GPSTimeStamp`

ErikSin (approved):
Besides adding those extra fields to the `additionalExif`, it looks good to go.

The edit observation logic (which i wrote) is very convoluted...its out of scope of this PR...but we should consider refactoring it...

src/frontend/sharedComponents/CameraView.tsx
ErikSin: Ok, yeah on digging the [gps info is just set via android location](https://developer.android.com/reference/androidx/exifinterface/media/ExifInterface?hl=en#setGpsInfo(android.location.Location)), which is what expo uses, so it doesn't matter if it rewrites it or not

achou11: yeah generally agree with your assessment! one thing to note is that there's technically room for potential loss of precision when it comes to the GPS info, just because we throttle the location information we receive for usage in the app. it'll always be more precise to retrieve the location on the spot rather than use something that's potentially throttled 😄 

don't think it'll be a huge issue but just noting that for future reference.

> So yeah I think i would add GPSDateStamp and GPSTimeStamp

I was afraid you were going to confirm this 😂 the formats used in the spec are subtly cursed (basically will need to do something like [this](https://github.com/digidem/mapeo-mock-data/blob/25462bdff06d3d71eb38e8403aa12d0ce0c7a3d9/lib/faker.js#L60-L85))

achou11: Addressed via https://github.com/digidem/comapeo-mobile/pull/1191/commits/5deead9612136664e6b4df6d6d555bf0310c67d4

PR #1195, opened 2025-05-23
fix: fix issue with GPS-related EXIF data not being written into photo

achou11:
Towards #1043

Should've caught this in #1191 but it turns out that `skipProcessing` needs to be `false` in order for the fields specified in `additionalExif` to actually be written into the photo that's created. Without this, the existing code will not actually write the relevant EXIF values into core when saving the observation.

This PR offers the bare minimum in terms of fixing the issue, although it may warrant some deeper thinking about whether it's the proper solution since it does comes with potentially user-facing consequences.

From Expo Camera's [docs](https://docs.expo.dev/versions/v52.0.0/sdk/camera/#camerapictureoptions): 

> If set to true, camera skips orientation adjustment and returns an image straight from the device's camera. If enabled, quality option is discarded (processing pipeline is skipped as a whole). Although enabling this option reduces image delivery time significantly, it may cause the image to appear in a wrong orientation in the Image component (at the time of writing, it does not respect EXIF orientation of the images).
>
> Note: Enabling skipProcessing would cause orientation uncertainty. Image component does not respect EXIF stored orientation information, that means obtained image would be displayed wrongly (rotated by 90°, 180° or 270°). Different devices provide different orientations. For example some Sony Xperia or Samsung devices don't provide correctly oriented images by default. To always obtain correctly oriented image disable skipProcessing option

### DIFF
--- a/src/frontend/sharedComponents/CameraView.tsx
+++ b/src/frontend/sharedComponents/CameraView.tsx
@@ -2,7 +2,6 @@ import {captureException} from '@sentry/react-native';
 import React from 'react';
 import {View, StyleSheet, Text} from 'react-native';
 import {
-  CameraCapturedPicture,
   CameraPictureOptions,
   CameraView as ExpoCameraView,
   useCameraPermissions,
@@ -125,7 +124,13 @@ export const CameraView = ({onAddPress}: Props) => {
         }
 
         onAddPress({
-          capturePromise: rotatePhoto(pic, accelerometerMeasurement.current),
+          capturePromise: resizePhoto({
+            uri: pic.uri,
+            width: pic.width,
+            height: pic.height,
+            rotation: getPhotoRotation(accelerometerMeasurement.current),
+            keepMeta: !!pic.exif,
+          }),
           mediaMetadata,
         });
       })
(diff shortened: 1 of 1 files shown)


ErikSin:
Do we care about the orientation issues since we are manually determining that anyways?

achou11:
> Do we care about the orientation issues since we are manually determining that anyways?

My understanding is that the orientation issue is only relevant when `skipProcessing: true`, which is what we currently do. If anything, we could potentially remove the code that does the manual orientation because it wouldn't be needed, but haven't looked too deeply at what else would be affected.

achou11:
Noticed that there's a potential bug where the saved image actually doesn't contain any of the original EXIF data due to the default behavior of the resizing library we use:

> keepMeta: If true, will attempt to preserve all file metadata/exif info, except the orientation value since the resizing also does rotation correction to the original image. Defaults to false, which means all metadata is lost. Note: This can only be true for JPEG images which are loaded from the file system (not Web).

See https://github.com/digidem/comapeo-mobile/pull/1195/commits/4ec96ce7672c1aee2093029a42a29a503ef0d68e for potential fix.

achou11:
Put up an alternative PR that more thoroughly handles issues related to photo-taking that I've come across while working in this space in recent weeks: https://github.com/digidem/comapeo-mobile/pull/1201. Can close this one if we're interested in considering that one instead.

ErikSin:
Based on looking at both these branches I am in favor of #1201!

achou11:
Closing in favor of #1201

PR #1199, opened 2025-05-27
fix: coordinator can edit all observations

### LINKED ISSUE #1190: Coordinator cannot edit observations of others
**Describe the bug**
As a coordinator on a project, I cannot edit the observations created by others on the project and they should be able to.

Coordinators can:

- Take and share observations
- Edit observations created by others
- Invite devices to the project (to be Coordinators or Participants)
- Manage project details
- Import config (categories)
- Leave project

Participants can:

- Take and share observations
- Edit observations they created
- View project data
- Leave project

Participants **cannot**:
- Edit observations created by others
- Invite other devices to the project
- Import config (categories)
- Edit project details (e.g. project name)

cimigree:
closes #1190 

Uses useOwnRoleInProject hook call to check for the coordinator id to determine if an observation can be edited...
I thought about using useProjectRoleAndDetails, but that makes an extra hook call that is not needed, so I opted against it.

And now a coordinator can edit the observations of others on projects they are in even if they did not create the observations


ErikSin (changes requested):
I think we should use this [hook now](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/hooks/server/useIsMyDocument.ts).

Also I think we forgot to include delete. I will confirm, but I think a coordinator should also be able to delete an observation

cimigree:
@ErikSin 
I made a hook since it got a little complex (I forgot to include creator id) and I used it to edit and delete observations but also edit and delete tracks (Tracks previously could be deleted by anyone -- that didn't seem right? But maybe that was as intended?).

Once the new PR related to tests is merged into develop (https://github.com/digidem/comapeo-mobile/pull/1204), I will merge develop in and then the E2E tests will pass

src/frontend/hooks/server/useCanEditorDelete.ts
ErikSin, line 1: Can you change this file name to `useCanEditOrDelete.ts` (just capitalize the "or")

src/frontend/screens/Track/index.tsx
```diff
@@ -71,7 +73,7 @@ export const TrackScreen = ({
         dockContent={
           <ActionButtons
             handleDelete={deleteTrack}
-            isMine={true}
+            isMine={canDelete}
```
ErikSin: can we change "isMine" to "canDelete"?

src/frontend/screens/Observation/index.tsx
```diff
@@ -68,15 +65,8 @@ export const ObservationScreen: NativeNavigationComponent<'Observation'> = ({
...
-  const {data: createdByDeviceId} = useDocumentCreatedBy({
-    projectId: projectId,
-    originalVersionId,
-  });
-  const isMine =
-    deviceInfo?.deviceId !== undefined &&
-    createdByDeviceId !== undefined &&
-    deviceInfo.deviceId === createdByDeviceId;
+
+  const canDelete = useCanEditOrDelete(originalVersionId);
```
ErikSin: I know this is me who originally did this. But i think we should just use the hook directly inside the button component since that component has access to the `originalVersionId`. Can delete is not needed for this component

ErikSin (approved):
some minor changes, but overall looks good

PR #1200, opened 2025-05-27
chore: updated location provider

### LINKED ISSUE #1194: Update Location Provider to unify the location being used by map
Currently the Map component has several things that utilize a live location: 
1. GPS Pill
2. CurrentMapTrackLayer
3. UserToolTip (For Tracks)
4. UserLocation.

All of these are getting their location for various different sources. This is causing problems such as #938 .

To fix this we can put location into a zustand store, and propogate the store via a context (similar to how we deal with persisted state). By putting it into a context we can start and stop to location subscription using the react life cycle via `useEffect`, which will make it easier to control and avoid the subscription running when we do not want it to. And by using a zustand store, we wont cause any unnecessary re-renders of all the children as the zustand manages rerenders via `useStore`

### Background
In the past we avoided putting location into a context for 2 reasons.
1. Location updates frequently and with react any update in state of a context causes all its children to re-render. Because of react navigation, the context lives at the top level, meaning it would cause ALL screen to re-render even if it wasn't using location
2. We didn't want the location subscription to always be on. Having the context as a parent of the navigator meant that it was always mounted, so even if there were screens not using it, the subscription would still be running in the background

We can avoid number one with zustand. And with 2, our solutions was to have each component just run their own subscription. That way when a component unmounted, the subscription would stop. While this makes sense, what we didn't consider is the the Map screen is ALWAYS mounted. Since it is the first screen in the stack navigation, it is always mounted in the background. And the map screen uses 3 `useLocation()` hooks (for the GPS pill, the map itself, and CurrentTrackLayer.) That means we are starting 3 separate subscriptions that are ALWAYS running. By putting it into a context, we are guaranteeing only one subscription is running.

towards #938 

## To Do
- [ ] Create Location Store (See Types Below)
- [ ] Create Location Context to propogate Store
- [ ] Create useLocationState hooks (this store does not have any actions)
- [ ] Start (and stop) location subscription in a useEffect in Context
- [ ] Move locationPermission and ProviderStatus subscription into this context, and manage via a useEffect as well.
- [ ] remove useLocation()
- [ ] remove useLocationProviderStatus()
- [ ] update gpsPill, CurrentTracksLayer, and userToolTip to use location from this new context
```ts
type LocationState = {
  location: LocationObject | undefined;
  throttledLocation: LocationObject | undefined;
  locationPermission: PermissionStatus | undefined;
  providerStatus: LocationProviderStatus | undefined;
};
```

ErikSin:
closes #1194 

Replaces `useLocation` and `useLocationProviderStatus` hooks.

We should also refactor how location is being used in some places(specifically with the map), but I think it is out of scope for this PR.


src/frontend/contexts/LocationContext.tsx
```diff
@@ -0,0 +1,284 @@
...
+        store.setState(prev => ({
+          ...prev,
+          throttledMapLocation: location,
+        }));
+        return;
+      }
+
+      const timeElapsed = location.timestamp - lastThrottledLocation.timestamp;
+
+      if (timeElapsed < 1000) return;
```
cimigree: Maybe make this magic number a constant?

src/frontend/contexts/LocationContext.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+      const coords = getCoords(location);
+      const lastCoords = getCoords(lastThrottledLocation);
+      const ruler = new CheapRuler(lastCoords[1], 'meters');
+      const distance = ruler.distance(coords, lastCoords);
+
+      if (distance > 5) {
```
cimigree: Another magic number that maybe should be a constant?

src/frontend/contexts/LocationContext.tsx
```diff
@@ -0,0 +1,284 @@
...
+          ...prev,
+          throttledMapLocation: location,
+        }));
+        return;
+      }
+    },
+  );
+}
+
+function useCheckPermissionOnAppStateChange(store: StoreApi<LocationState>) {
```
cimigree, line 216: Could we ask permission again here?

cimigree (approved):
Overall I think this is a great PR and will be an improvement for the app.
Some (non blocking) suggestions.
Right now, we only check permission status on app resume (getForegroundPermissionsAsync) — we do not prompt the user again if permission is denied.
So the GPS pill show a spinner indefinitely if permission is off and there's no way to recover.
Should we re-request permission? I feel like this used to happen?

The throttledMapLocation is not use in a lot of places and instead they use the raw location. Is it definitely expected that Map is the only one that throttles? Should UserTooltipMarker, UserLocation etc use the throttledLocation? Any chance you would consider other names for throttledLocation and location? I find them a little confusing but I can adapt.

Maybe add a comment in useMostAccurateLocationForObservation that it is running its own subscription since now we have just one that everywhere else uses.

Maybe put magic numbers (1000ms and 5m) in a constant?

src/frontend/contexts/LocationContext.tsx
ErikSin: We could. My hesitation is that it would be annoying to ask after the user has intentionally went into the setting to turn it off. This check is designed to check when the user has changed the persmission outside of the app (aka in the settings). They would only do that intentionally. So I don't think it then makes sense to ask them to turn it on again.

ErikSin:
> The throttledMapLocation is not use in a lot of places and instead they use the raw location. Is it definitely expected that Map is the only one that throttles? Should UserTooltipMarker, UserLocation etc use the throttledLocation? Any chance you would consider other names for throttledLocation and location? I find them a little confusing but I can adapt.

The map is the only place that will use the throttled location. All the other components should be using the most accurate and updated locations

ErikSin:
> Maybe add a comment in useMostAccurateLocationForObservation that it is running its own subscription since now we have just one that everywhere else uses.

This will be updated in a subsequent PR - it won't have its own subscription soon. I just didn't want this PR to get too big

PR #1201, opened 2025-05-27
fix: do not skip processing of camera photo

achou11:
Towards #1043 

Quick experiment based on some discussion that came from #1195. Basically wanted to test if we could set `skipProcessing: false` for when taking any photos, in order to avoid any chance of losing metadata but also to fix some tangential issues.

I think the original reasoning to skip processing was to avoid potential issues with lower-powered devices. That may still be a valid concern but I haven't tested it. I'm a bit skeptical because we end up doing some amount of post-processing of the initially taken image anyways (via the resizer library), which probably has some level of computational rigor as well.

---

To experiment: I took photos with my phone turned 90 degrees such that my phone was oriented as though I was taking a  "landscape" photo. The expectation is that the picture is portrayed in landscape format.

- Here's side-by-side of the thumbnails. First one was with the code from the develop branch, and the second one with code from this branch.

    image

- Here's a full-sized preview of the photo taken using the develop branch

    image

- Here's a full-sized preview of the photo taken using the changes in this PR

    image

_NOTE: the above photos are two distinct photos, not the same photo. Tried my best to take similar looking photos but maybe it's more useful to take very different ones?_

---

The result from this branch matches the expected orientation of the photo and how it's displayed. The less desirable behavior of the former was actually highlighted in https://github.com/digidem/comapeo-mobile/pull/1179 and I noted that it had more to do with how we take the photo than the changes in that PR (see this [comment](https://github.com/digidem/comapeo-mobile/pull/1179#issuecomment-2884747937))


ErikSin:
I have a very slow phone (Samsung a03s). It is slow enough that you can time how long each photo takes to process. I got some interesting results (but all in favor of doing this).

First the amount of time it took for the CameraView to navigate away after the photo was taken was about 6 seconds for both this branch AND develop (which is pretty abysmal IMO 💩). Which means that the time it take to manually rotate it with react native resizer and with expo is about the same. 

But the interesting thing is that the processing of the photo afterwards was significantly different. On this branch it took about 2 seconds for the photo loader to turn into a photo thumbnail, while it took develop 5/6 second. I thought it was maybe a fluke but I tried it multiple times...and same results.

Your changes shouldn't be effecting the after processing...

Regardless, I am in favour of doing this and getting rid of the unnecessary rotation logic.

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -88,18 +74,20 @@ export const CameraView = ({onAddPress}: Props) => {
...
-        }
-      : BASE_CAMERA_CAPTURE_OPTIONS;
-
     ref.current
-      .takePictureAsync(captureOptions)
+      .takePictureAsync({
+        base64: false,
+        exif: true,
+        skipProcessing: false,
+        quality: 0.75,
```
achou11, line 82: this is preserving the quality setting that was used in the previous implementation. is it still preferable to do this though?

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -122,7 +110,7 @@ export const CameraView = ({onAddPress}: Props) => {
         }
 
         onAddPress({
-          capturePromise: rotatePhoto(pic, accelerometerMeasurement.current),
+          capturePromise: Promise.resolve({uri: pic.uri}),
```
achou11, line 113: somewhat lazy but updating the thing that uses `capturePromise` can become a potentially larger change in how the draft observation stuff works. preferred to leave that as is for the scope of this PR.

ErikSin (approved):
LGTM

Issue #1181, opened 2025-05-15
Tracks UX improvements


ximenabb:
**Description**
Improvements to the Tracks feature

View clickable prototype : https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=5471-3571&t=JZAINjSzPyFCMMvC-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=5471%3A3571&show-proto-sidebar=1

Feature spec : https://www.notion.so/digidem/Tracks-UX-improvements-14c1b08162d5803a922eef06556a16de

**Todo**

<img width="1004" alt="Image" src="https://github.com/user-attachments/assets/df39dd15-86e3-4d6e-b8f9-b4296facd4f8" />

### Tracks tab - GPS on

**Tracks Tab - Ready**
<img width="277" alt="Image" src="https://github.com/user-attachments/assets/10278661-e1a3-46ed-93aa-86d6487acccf" />
- [x] Add copy "Ready to record new track?"

**Tracks Tab - Recording**
<img width="270" alt="Image" src="https://github.com/user-attachments/assets/54e15fda-6910-44cc-918b-daf5d18d4bcd" />
- [x] Add copy "Move one meter to see track."

**Tracks Tab - Error: Tracks Stopped without Moving**

<img width="266" alt="Image" src="https://github.com/user-attachments/assets/9fd821fb-e2d0-43b3-b558-54acd1e7eb45" />
If user hasnt moved or created an observation

- [x] Display error bottom sheet
- [x] Display copy as designed
- [x] Tapping the "Exit Tracks" button stops the tracks feature, doesnt save a track
- [x] Tapping the "Continue Recording" keeps the recording going

### Recording UI 

<img width="572" alt="Image" src="https://github.com/user-attachments/assets/053da00d-9237-4fef-a4d0-8e03cb713fe3" />

- [x] Update the Tracks line
- [x] Add the Current user location as the last point of the line always, so that the line matches with where user actually is. When saving the track, make sure to add this point to the track.
- [x] Display track in map view or while recording - as designed

### ~~Attaching a  Category~~

<img width="882" alt="Image" src="https://github.com/user-attachments/assets/0ccb02d9-c3b8-434e-a1a9-d615ef44d5b1" />

~~**Category selector**~~

- [x] ~~Tapping on "stop tracks', display Category selector ~~
[...]

PR #1202, opened 2025-05-28
chore: simplify tracking hooks

ErikSin:
toward #1181 

I simplified the tracks hook and made tracks more declarative.

First I removed state that did not need to be persisted from the Tracks Context. I removed `distance`, `isTracking` and `trackingSince` because all these states can be derived so there was no point in setting them in persisted state.

Then I updated the useTracks hook, and turned it into 2 hooks, `useCurrentTrackState` and `useStartStopTracks`. in the `useCurrentTrackState` I created the state `hasActiveTrack`, `trackingSince`, and `totalDistance`. These states are determined in a declarative way, by simply being based off the location history. `hasActiveTrack` is true if there is more than 1 location history, `trackingSince` is based on the timestamp of the first location, and `totalDistance` is calculated based on `locationHistory`. Previously we were manually setting these values when the user started the track. One notable change is renaming `isTracking` to `hasActiveTrack`. Previously, we set `isTracking` to false when the user stopped tracking. Now, to make the state more declarative, `hasActiveTrack` returns whether there are any tracks in the persisted state. This reflects that even after the user stops tracking, we can't simply clear the track data — as they still have to view it and save it. So the new name better expresses the actual intent: checking if there is an existing track, not just whether tracking is currently active. In other words there can be an existing track but the user is no longer actively tracking. Knowing whether the user is actively tracking is not necessary based on our UI. When a user stops tracking we navigate to the `SaveTracks` page, and that page does not need to be aware of whether it is tracking or not. Now we just need to clear tracks whenever we navigate away from this page to reset the `hasActiveTrack` state.

I also separated the relevant actions for starting a track into its own hook. Ending the track also returns the total distance of the track. This will be used later as in #1181, where we will be warning the user when they stop a track and when the track is less than 1m. 

Lastly i updated the `calculateDistance` function. Previously it was using cheap-ruler. Cheap ruler required us to calculate the distance between each point and add them together. I opted to use [turf length](https://turfjs.org/docs/api/length) instead as it is a more established library, and it can just calculate the distance using the whole array instead of just pairs of points.


src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -8,12 +8,10 @@ import * as v from 'valibot';
 
 import {MMKVZustandStorage} from '../hooks/persistedState/createPersistedState';
 import {LocationHistoryPoint} from '../sharedTypes/location';
-import {calculateTotalDistance} from '../utils/distance';
 
 const TrackStateSchema = v.intersect([
```
achou11: no longer need the `v.intersect(...)` here.

```ts
const TrackStateSchema = v.object({
  description: v.string(),
  locationHistory: v.array(
    // TODO: Create schema for this
    v.object({
      latitude: v.number(),
      longitude: v.number(),
      timestamp: v.number(),
    }),
  ),
  observationRefs: v.array(
    // TODO: Create schema for this
    v.object({
      docId: v.string(),
      versionId: v.string(),
    }),
  ),
});
```

src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -30,16 +28,6 @@ const TrackStateSchema = v.intersect([
...
-  v.union([
-    v.object({
-      isTracking: v.literal(true),
-      trackingSince: v.date(),
-    }),
-    v.object({
-      isTracking: v.literal(false),
-      trackingSince: v.null(),
-    }),
-  ]),
```
achou11: i think the removal of this warrants a bump in the `version` for the zustand persisted store, since we want to prompt a migration of existing values. 

from what I understand, without migrating the values you can encounter a scenario where the type of the value doesn't actually match the value you have at runtime. In this case, the app code will probably still work because the only difference is the removal of fields and the code won't try to use those fields due to the updated type, but it can still result in gnarly bugs.

src/frontend/hooks/useTracking.ts
```diff
@@ -7,50 +7,58 @@ import {
...
+    hasActiveTrack: locationHistory.length > 0,
+    totalDistance: calculateTotalDistance({
+      points: locationHistory,
+      units: 'kilometers',
+    }),
+    trackingSince: locationHistory[0]
+      ? new Date(locationHistory[0].timestamp)
+      : null,
+  };
+};
```
achou11: are `totalDistance` and `trackingSince` relevant when there's no location history? thinking this hook could be updated to account for that. Couple ways of going about it but my initial thought is something like this:

```ts
export const useCurrentTrackState = () => {
  const locationHistory = useTrackState(state => state.locationHistory);

  if (locationHistory.length === 0) {
    return null;
  }

  return {
    locationHistory,
    totalDistance: calculateTotalDistance({
      points: locationHistory,
      units: 'kilometers',
    }),
    trackingSince: new Date(locationHistory[0]!.timestamp),
  };
};
```

You lose an explicit `hasActiveTrack` field but I think using `null` return value to represent that can be sufficient. if you really want that field, can update the example implementation to something like:

```ts
export const useCurrentTrackState = () => {
  const locationHistory = useTrackState(state => state.locationHistory);

  if (locationHistory.length === 0) {
    return {hasActiveTrack: false}
  }

  return {
    hasActiveTrack: true,
    locationHistory,
    totalDistance: calculateTotalDistance({
      points: locationHistory,
      units: 'kilometers',
    }),
    trackingSince: new Date(locationHistory[0]!.timestamp),
  };
};
```

but i don't find the additional field all that useful in this case.

src/frontend/hooks/useTracking.ts
achou11, line 1: I have some doubts about the utility of `useStartStopTracks()`. From what I can tell, you can achieve something that does a better job of error handling by wrapping `Location.startLocationUpdatesAsync()`/`Location.stopLocationUpdatesAsync()` as mutations and then making use of the `onSuccess`/`onError` callbacks.

For example, with the relevant mutations implemented:

```ts
function useStartLocationUpdates() {
  return useMutation({
    mutationFn: async () => {
      return Location.startLocationUpdatesAsync(LOCATION_TASK_NAME, {
      });
    },
  });
}

function useStopLocationUpdates() {
  return useMutation({
    mutationFn: async () => {
      return Location.stopLocationUpdatesAsync(LOCATION_TASK_NAME);
    },
  });
}
```

Consuming components like `StartStopTrack` could look like this:

```ts
function StartStopTrack() {
  const {clearCurrentTrack} = useTrackActions();

  // In theory this could work, but maybe it's better
  // to do the calculation within the mutation definition (see note below)
  const totalTrackDistance = useTrackState(state =>
    calculateTotalDistance({
      points: state.locationHistory,
      units: 'meters',
    }),
  );

  const startLocationUpdates = useStartLocationUpdates();
  const stopLocationUpdates = useStopLocationUpdates();

  function startTracks() {
    startLocationUpdates.mutate(undefined, {
      onError: err => {
        Sentry.captureException(err);
      },
    });
  }

  function endTracks() {
    stopLocationUpdates.mutate(undefined, {
      onSuccess: () => {
        // Might need to do the calculation here instead i.e.
        // if (calculateTotalDistance(...)) { ...
        if (totalTrackDistance > 1) {
          navigation.navigate('SaveTrack');
          return;
        }

        clearCurrentTrack();
      },
      onError: err => {
        Sentry.captureException(err);
        // Maybe desirable to clear the track here but not sure off the top of my head...
        // clearCurrentTrack();
      },
    });
  }
}
```

Might look a little more verbose but this feels more compositional and prevents issues that could be caused by the hiding of some of the internals (e.g. the current implementation does not sufficiently account for errors that may arise from the Location method calls).

Could create wrapper hooks for something functionally equivalent to `endTracks()` and `startTracks()` in the above example if you want to avoid the repetition, but that seems premature.

achou11 (changes requested):
Haven't tested the changes on device just yet so an initial code review for now!

src/frontend/hooks/useTracking.ts
ErikSin: I think this causes a mismatch in the 2 `locationHistory` values. So this hook is a utility hook to have the other derived states of location history (based off location history). `locationHistory` for `useTrackState(state => state.locationHistory);` always returns a array (even if it is empty). 

If we did it how you are suggesting, we now have 2 different `locationHistory` values, which i think will cause confusion. We could just get rid of `LocationHistory` completely from this hook, and have the consuming component use useTrackState, but I like the convenience of having it available.

achou11: > I think this causes a mismatch in the 2 locationHistory values

sorry what are the 2 values in this case? are you referring to if you used this hook and the useTrackState hook at the same time? i.e.

```ts
function Example() {
  const locationHistory = useTrackState(state => state.locationHistory)
  const currentTrack = useCurrentTrackState()
  

  // Mismatch between `locationHistory` and `currentTrack?.locationHistory`?
}
```

if so, that's a fair point - can be a bit confusing, although I think there's some level of confusion you can run into in either case because there's now kind of a soft rule about using one hook or the other but not both at the same time.

If you prefer to keep all the fields, you can still update the return type to represent the explicit states a bit more strictly e.g.

```ts

type CurrentTrackState =
  | {
      hasActiveTrack: false;
      locationHistory: TrackState['locationHistory'];
      totalDistance: 0;
      trackingSince: null;
    }
  | {
      hasActiveTrack: true;
      locationHistory: TrackState['locationHistory'];
      totalDistance: number;
      trackingSince: Date;
    };

export const useCurrentTrackState = (): CurrentTrackState => {
  const locationHistory = useTrackState(state => state.locationHistory);

  if (locationHistory.length === 0) {
    return {
      hasActiveTrack: false,
      locationHistory,
      trackingSince: null,
      totalDistance: 0,
    };
  }

  return {
    hasActiveTrack: true,
    locationHistory,
    totalDistance: calculateTotalDistance({
      points: locationHistory,
      units: 'kilometers',
    }),
    trackingSince: new Date(locationHistory[0]!.timestamp),
  };
};
```

re: the confusion about two location history values: you may have already pointed it out in the previous comment, but you can update the hook to accept a parameter instead of using the track state hook internally, such that the usage would look like this from a consuming component:

```ts
function Example() {
  const locationHistory = useTrackState(state => state.locationHistory)
  
  // `currentTrackState` would not have a `locationHistory` field
  const currentTrackState = useCurrenTrackState(locationHistory)
}
```

So the "true" location history always comes from the consuming component in this case.

achou11: > re: the confusion about two location history values: you may have already pointed it out in the previous comment, but you can update the hook to accept a parameter instead of using the track state hook internally, such that the usage would look like this from a consuming component

I guess at that point it's really not a hook but rather just some kind of utility function to get derived state 😅 maybe that's something to consider rather than having a hook?

ErikSin: Sorry, when i was saying 2 values, i was pointing out that `const {locationHistory} = useCurrentTrackState()` could be undefined. Technically its not a seperate value, it just doesnt exist in the context of the hook. But what I was trying to point out was that `locationHistory` from `useTrackState` always returned something. There was 2 uses of locationHistory that were unsynced. Thats what i meant by "2 values", which was the incorrect way of saying that

> I guess at that point it's really not a hook but rather just some kind of utility function to get derived state 😅 maybe that's something to consider rather than having a hook?

I think more inclined to do this. I put it into a hook because in the zustand world everyone suggests putting derived values into custom hooks...

ErikSin: > I think more inclined to do this. I put it into a hook because in the zustand world everyone suggests putting derived values into custom hooks...

Actually its has just become quite verbose...so im just going to stick to the hook!

src/frontend/contexts/TrackStoreContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1202/commits/4d6e39307492675de8182fb351cc891e56cd1f00

src/frontend/contexts/TrackStoreContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1202/commits/4d6e39307492675de8182fb351cc891e56cd1f00

src/frontend/hooks/useTracking.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1202/commits/105439352cb78aadb644d66e36f543fbc337f1bf

src/frontend/hooks/useTracking.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1202/commits/05afdb9e4552eb18fa414489f2c2d468cbce301d

src/frontend/hooks/useTracking.ts
achou11, line 1: thoughts on renaming this file to match the exported hook name?

alternatively, rename this to something like `tracks` and move the hooks from `useStopLocationUpdates` and `useStartLocationUpdates` into this file. not sure how helpful it is to have a file per function in this case 😄

src/frontend/hooks/useStopLocationUpdate.ts
achou11, line 1: very nitpick but file name doesn't match the hook name. `useStopLocationUpdates.ts`

src/frontend/hooks/useTracking.ts
```diff
@@ -1,56 +1,39 @@
...
+      hasActiveTrack: true;
+      locationHistory: TrackState['locationHistory'];
+      totalDistance: number;
+      trackingSince: Date;
+    };
 
-    if (!hasTrackWithMoreThan1Point) {
-      clearCurrentTrack();
-    }
+export const useCurrentTrackState = (): CurrentTrackState => {
```
achou11: some hesitation about the `hasActiveTrack` and `trackingSince` being strictly derived from the location history. guess it depends on what we're trying to communicate, but couldn't there be a notable difference between the time when someone enables the tracking and when the first location is recorded? From what I understand, the location history is written to in the background by the background location updater so we don't really have control of when it fires and writes to the location history state.

Maybe the disparity isn't notable in practice, in which case this change is a nice simplification!

achou11: hm yeah I think this could be a potential issue, at least in terms of UI. On my older phone, I pressed `Start Tracks`  and it took about 5 seconds for the first location to be recorded. Up until that time, the `Start Tracks` button remained on the screen so I could interact with it even though I had already pressed it before.

Once the track-in-progress UI appeared, it seemed like the timestamp of the first location more-or-less matched when i actually pressed the button, so I think the potential `trackingSince` issue I alluded to probably isn't too much of an issue.

Guessing this is maybe why the `isTracking` was an explicit state before 😅

achou11: I guess a potential solution is to immediately write to the track store when pressing button, with a location that comes from our shared location hook. That way the derivation of the relevant states can remain the same as you currently have it, but the UI will update more immediately.

achou11 (dismissed):
One concern about the UI responsiveness, which might require some additional changes. Otherwise the changes look great!

src/frontend/hooks/useTracking.ts
gmaclennan: Just stumbled across this and wanted to add my two-cents based on field experience:

1. The location from our shared location hook at the time that the start track button is pressed might not be recent or accurate, so I don't think it is a good idea to use that location, as suggested by Andrew. I think we should rely on the background location updates and capture the locations from that.
2. The issue of a delay between the tracking starting and the first location update is something that does need addressed. In many field situations it could be several seconds or even several minutes between the user pressing "start track" and the first location update, due to GPS signal availability (you generally do not notice this when you are in an area with cell phone service, because location updates use cellphone towers to hugely increase GPS signal capture speeds). Also in some phones the motion API is not going to start actual tracking until the user starts moving.

TL;DR this needs to use something other than the first location update for the start track time, and using the location from the shared location hook is probably not the best solution because of the risk of stale/inaccurate locations for the start of the track.

achou11: Thanks for that input! Based on that, it seems like restoring an explicit field for the "tracking since" in the store state seems like the most reasonable option.

ErikSin: @gmaclennan I think that true if we use use the [SharedLocationContext](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/contexts/SharedLocationContext.tsx) which only updates every 3m.  But if we use expo's [getCurrentPositionAsync](https://docs.expo.dev/versions/latest/sdk/location/#locationgetcurrentpositionasyncoptions), that shouldn't give us the latest and most accurate location, would it not?

I would like to keep it as declarative as possible, so trying to avoid inserting a manual time if possible

gmaclennan: No because in the field the device might not have a location when the user starts tracking - it can take several minutes to acquire a location. That function could take a while to resolve.

gmaclennan: This is not to do with being declarative. There is a state: the user is either tracking or not tracking. If this state can be derived from another state, like the location history, then it's good practice to reduce duplicate state. However I don't think it can be reliably derived from the length of location history, and trying to manipulate the location history array to make it represent the state of whether the user is currently tracking seems counterproductive, and for the reasons above would introduce other errors that would be very hard to track down later.

ErikSin: updated here: https://github.com/digidem/comapeo-mobile/pull/1202/commits/5bc1d2659f811e408d4c3647433a50d8d7627f43

gmaclennan (changes requested):
Since I accidentally ended up here, I ended up looking at some of the rest of the code. I'm concerned about the performance implications of deriving the distance, rather than storing it in state. `turf.length()` is an expensive function, and as I understand it we are calling it frequently whenever the track updates. Running the [cheap ruler benchmarks](https://github.com/mapbox/cheap-ruler/blob/main/bench/bench-distance.js) shows that turf.distance is about 35x slower than cheap ruler, in addition, the previous code was cumulatively calculating the distance, so only a very small calculation on each state update (adding the distance from the new points). A track can include tens of thousands of points, and I am concerned that repeatedly calculating the distance of the entire track is going to significantly lock up the JS thread, particularly on slow devices.

PR #1203, opened 2025-06-02
feat: expose more photo metadata

### LINKED ISSUE #1043: Exposing Photo Metadata

achou11:
Closes #1043 

- Introduces the usage of `expo-image` in order to facilitate the retrieval of file information, most notably the storage size of the image. This is only being used by anything that uses `TrulyContainedImage`, which at the moment is limited to screen displaying the photo and its metadata.

---

- Processed draft photo (scrolled down a bit due to vertical overflow)
    
    - For new observation
    
        image
        
    - For editing observation
    
         image

- Saved photo (scrolled down a bit due to vertical overflow)

    - Portrait
    
        image

    - Landscape
    
        image


src/frontend/lib/exif.ts
```diff
@@ -8,6 +8,8 @@ export const PhotoEXIFSchema = v.object({
   DateTime: v.optional(v.string()),
   ExposureTime: v.optional(v.number()),
   Flash: v.optional(v.number()),
+  // @ts-expect-error Need changes to schema to fix
+  FNumber: v.optional(v.number()),
```
achou11: this is available in latest schema (1.7.0) but core doesn't support the schema just yet.

src/frontend/lib/file-system.ts
```diff
@@ -53,3 +54,37 @@ export async function selectFile({
...
+/**
+ * Returns the storage size of an image rendered using `expo-image`.
+ * This is specific to Expo because their `Image` component uses the Expo file system abstraction for caching.
+ *
+ * @param imageURL The URL of the image.
+ *
+ * @returns The storage size of the image, in bytes.
+ */
+export async function getExpoImageStorageSize(
+  imageURL: string,
```
achou11, line 67: we get storage size in the following situations:

1. From a processed draft image

2. From a saved image.

The URL in (1) is a file URI e.g. starts with `file://`, which means we can instantly get more granular info using expo-file-system.

The URL in (2) is in the form of a remote URL e.g. `http://...`. In this case, we provide it as a lookup key for expo-image to use in order to figure out if it's cached or not (using their Image component automatically makes use of their caching). Once we get a cache path for it, then we can get more granular information about corresponding image file in the cache using expo-file-system.

src/frontend/sharedComponents/Images/TrulyContainedImage.tsx
```diff
@@ -12,32 +15,28 @@ export function TrulyContainedImage({
...
         // There seems to be a bug related to images that have non-network sources (e.g. `file://...`)
         // where the loading the image won't even be attempted unless the image has a known height prior.
         // We trigger the load attempt by initially setting the height to a non-zero value and then removing it once it loads.
-        {aspectRatio, height: aspectRatio ? undefined : 1},
+        // Note: with React Native's image, the initial height can be set to 1 but with Expo's Image, it needs to be a larger value or else
+        // there's some layout glitchiness each time it loads...
+        {
+          aspectRatio,
+          height: aspectRatio ? undefined : 50,
+        },
```
achou11, line 35: a rather gnarly difference. haven't had the energy to look into the issues to see if there's anything about workarounds or updates.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+
+  const coordinateFormat = useCoordinateFormat();
+
+  // TODO: This check needs to be updated in the case of the saved photo using attachment fields that will
+  // be available in an unreleased version of core.
+  const isValidatedByCoMapeo =
+    photo.type === 'photo'
+      ? true
+      : typeof photo.mediaMetadata.timestamp === 'number' &&
+        photo.mediaMetadata;
```
achou11: just noting this will need to be updated to make use of the `createdAt` and `position` fields that may exist on an observation attachment. Not available in core just yet though.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+  const exif =
+    photo.type === 'photo' ? photo.photoExif : photo.mediaMetadata.photoExif;
+
+  const [imageStorageSize, setImageStorageSize] = useState<number | undefined>(
+    undefined,
+  );
+
+  // TODO: For saved photo, use attachment's `createdAt` field (available in unreleased version of core)
+  const timestamp =
+    photo.type === 'processed' ? photo.mediaMetadata.timestamp : undefined;
```
achou11: similar comment about needing to update this based on changes to core/schema.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+
+  const coordinates =
+    photo.type === 'processed' && photo.mediaMetadata.location
+      ? {
+          lon: photo.mediaMetadata.location.coords.longitude,
+          lat: photo.mediaMetadata.location.coords.latitude,
+        }
+      : photo.type === 'photo'
+        ? // TODO: Use attachment's `position` field (not yet implemented in schema)
+          undefined
```
achou11: another case of needing to update this based on changes to core/schema

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+
+  const deviceDetails = exif
+    ? getDeviceDetailsText({make: exif.Make, model: exif.Model})
+    : null;
+
+  const cameraDetails = exif
+    ? getCameraDetailsText(
+        {
+          // @ts-expect-error Need to update schema
+          fNumber: exif.FNumber,
```
achou11: this is already available in schema, but core needs to be updated to use it.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+              source={photo.originalUri}
+              onLoad={event => {
+                getExpoImageStorageSize(event.source.url)
+                  .then(size => {
+                    setImageStorageSize(size);
+                  })
+                  .catch(err => {
+                    captureException(err);
+                  });
+              }}
```
achou11: there's a good reason for why I'm not wrapping this in a `useQuery()`: getting the storage size of network-based image (e.g. `http://...`) won't work if the image hasn't been loaded (and subsequently cached) by expo-image. We must wait for guarantees about the image being loaded successfully before attempting to fetch this information via expo.

Tangentially, I could wrap the usage of `getExpoImageStorageSize()` here with a `useMutation()`, but opted not to because we don't really need any of the state management benefits of the loading logic right now.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+                  projectId={projectId}
+                  onLoad={event => {
+                    getExpoImageStorageSize(event.source.url)
+                      .then(size => {
+                        setImageStorageSize(size);
+                      })
+                      .catch(err => {
+                        captureException(err);
+                      });
+                  }}
```
achou11: similar comment about why we're doing this via `onLoad` instead of alternative approaches.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+                size={20}
+                color={NEW_DARK_GREY}
+                allowFontScaling
+              />
+            }>
+            <BodyText selectable style={styles.primaryInfoText}>
+              {
+                // TODO: Showing this ID is a temporary measure
+                // Ideally we use a format that is more human readable/friendly
+                observationDocId.slice(0, 15)
```
achou11, line 268: decision to truncate was based on Gregor's recommendation as a temporary measure.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+          size={20}
+          color={NEW_DARK_GREY}
+          allowFontScaling
+        />
+      }>
+      <BodyText selectable style={styles.primaryInfoText} numberOfLines={1}>
+        {
+          // TODO: Showing this ID is a temporary measure
+          // Ideally we use a format that is more human readable/friendly
+          createdByDeviceId.slice(0, 15)
```
achou11, line 390: same thing about this being a recommended short-term approach.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+      displayedParts.push(
+        layout === 'horizontal'
+          ? formatMessage(m.landscape)
+          : formatMessage(m.portrait),
+      );
+    } catch (err) {
+      captureException(err);
+
+      // TODO: What happens if layout cannot be determined or is missing?
+      // Should we still try to display something?
```
achou11: highlighting this. unlikely we'll run into this but basically if the orientation value from the EXIF is unrecognized, we won't attempt to show an orientation to the user (e.g. no `Portrat` or `Landscape` text will show).

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+    width: number;
+    height: number;
+    storageSize?: number;
+  },
+  formatMessage: IntlShape['formatMessage'],
+): string | null {
+  const displayedParts: Array<string> = [];
+
+  // TODO: Should this be translated?
+  displayedParts.push(`${width} x ${height}`);
```
achou11, line 461: part of me thinks this could be something that should be translated, but I really have no idea.

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+
+  if (typeof storageSize === 'number') {
+    displayedParts.push(
+      formatMessage(m.imageStorageSize, {
+        value: Math.max(bytesToMegabytes(storageSize), 0.01).toFixed(2),
```
achou11, line 466: rounding to the hundredth made sense to me, with a minimum value of `0.01` being shown. can't imagine many images will actually be smaller than that but just in case.

src/frontend/screens/PhotoPreviewModal/index.tsx
achou11: wondering in the case of a saved photo, if the timestamp field is missing, should we fall back to the associated observation's timestamp? this was something that was suggested when talking to Gregor, but I kind of have mixed feelings about doing that.

achou11:
noting that that i've been introduced to the new [criteria](https://www.notion.so/digidem/Exposing-Observation-Metadata-GPS-Photos-6f9cd1cf9f0e40ed8eda96b6a70f3aa3#2091b08162d5804d9896fd081eee0a40) around what "validated by CoMapeo" means in the context of photos, so will need to do some work to update this PR to align with that.

achou11:
Coverting to a draft because:

1. it still needs some adjustments based on evolving UX requirements
2. relies on some updates to schema/core

src/frontend/screens/PhotoPreviewModal/index.tsx
ErikSin: No, i think for simplicity, lets not fall back to the observation timestamp. there are less than 200 users rn (and im pretty sure 14 of them are jen's phones 🤣). So i dont think we should over complicate the code for such a small user base

src/frontend/screens/PhotoPreviewModal/index.tsx
```diff
@@ -0,0 +1,497 @@
...
+          },
+          t,
+        )
+      : null;
+
+  return (
+    <ScrollView contentContainerStyle={{padding: 20, gap: 20}}>
+      <View>
+        {photo.type === 'photo' ? (
+          <Suspense fallback={null}>
```
ErikSin, line 142: Is there a reason we don't just want to use the suspense boundary that every other screen uses (aka the one that wraps the navigator)

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
```diff
@@ -9,9 +9,15 @@ type Props = {
...
+export function CoreBlobImage({
+  driveId,
+  name,
+  projectId,
+  testID,
+  onLoad,
+}: Props) {
   const [cannotLoadImage, setCannotLoadImage] = useState(false);
   const [displayedVariant, setDisplayedVariant] =
     useState<PhotoVariant>('original');
```
ErikSin, line 23: we need to account for when they have their media manager turned on and they only have photo variant `preview` available

src/frontend/screens/PhotoPreviewModal/index.tsx
ErikSin: Also to further simplify, i think we should get rid of the ErrorBoundary, and just let the component internally handle the error (see comment below)

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
ErikSin, line 1: I think it would simplify the component to manually check for an error in the query, and update  

```ts

if (cannotLoadImage || queryErrror) {
    return <ImageErrorPlaceholder testID={testID} />;
  }

```

src/frontend/sharedComponents/Images/TrulyContainedImage.tsx
```diff
@@ -12,32 +15,28 @@ export function TrulyContainedImage({
...
       ]}
       onLoad={event => {
```
ErikSin, line 37: What if instead of exposing the onLoadEvent callback we expose an `onImageStorageSizeLoaded` event that returns the image size, where internally inside this component we do `getExpoImageStorageSize`. It would just simplify this component a bit, and really the consuming component does not need any other on load info

src/frontend/screens/PhotoPreviewModal/index.tsx
ErikSin, line 1: I think we should encapsulate all the logic for determine the photo-metadata into a hook. It will simplifiy this component, and be testable

ErikSin (commented):
I think overall the UI looks good.

2 architectural changes I would suggest is encapsulating the logic to exract the data into a function/hook

And i think we should not expose the entire onLoad event, and have the image component internally calculate the size

src/frontend/screens/PhotoPreviewModal/index.tsx
achou11: > Is there a reason we don't just want to use the suspense boundary that every other screen uses (aka the one that wraps the navigator)

Yeah probably worth just doing that for simplicity and consistency. i had started with granular suspense boundaries since that's what initially came to mind and then realized that we don't really do that elsewhere for similar screens 😄 definitely would remove the need for creating extra suspense-boundary components

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
achou11: does this not address that concern?

https://github.com/digidem/comapeo-mobile/blob/caae9071aeeffb20b2a2ab73d11cca175ed7305a/src/frontend/sharedComponents/Images/CoreBlobImage.tsx#L43-L48

ErikSin: lol, i dont know how i missed that....BUT i think we should just show the preview regardless (aka dont do this check). Now the screen for viewing the photo is even smaller, and the user cannot manipulate or zoom on the photo, there will be no noticeable difference to show the preview vs showing the original

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
achou11: what I learned through testing this is that without specific configuration of react query, a suspense query will throw if it fails and there hasn't been a previously successful query:

https://github.com/TanStack/query/blob/2b9b122b98e9eae16c8c29fd6fb720814d503526/packages/react-query/src/useSuspenseQuery.ts#L28

https://github.com/TanStack/query/blob/2b9b122b98e9eae16c8c29fd6fb720814d503526/packages/react-query/src/suspense.ts#L11-L19

This means that if this query initially failed for example, it would propagate to the nearest error boundary that's defined in the app, which I think is currently very high up.

Handling the `error` variable in the context of suspense hooks is dealing with _subsequent_ failed queries by default. If you want granular error handling for queries that fail initially, you need to either:

1. Update the query (or query client?) options to do a different check compared to the referenced internal implementation above

2. Create an error boundary.

(2) is much easier in my opinion and generally good practice when it comes to how Suspense works.

This is kind of a subtle gotcha of react query that I wasn't familiar with until this PR.

src/frontend/sharedComponents/Images/TrulyContainedImage.tsx
achou11: I think it's helpful to expose the onLoad as-is because the need for size information is usage-specific. As a shared component, this ideally could be used anywhere and I don't know if having a specific prop about getting the size of the image is strictly necessary in the other potential cases.

Open to updating this component if it becomes needed more often elsewhere but otherwise think it's an unnecessary change for now.

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
achou11: > Now the screen for viewing the photo is even smaller, and the user cannot manipulate or zoom on the photo, there will be no noticeable difference to show the preview vs showing the original

In terms of the image that's displayed, yeah probably not a huge difference. However, some of the metadata that's exposed like the image dimensions and resolution are specific to the image that's loaded and displayed.

I think there should be a preference for displaying information about the original image if possible, and displaying the preview if we don't have it available, as this implementation currently does.

achou11: in other words, is it okay to always be showing metadata about a preview variant of an image, even though we may have the original? i would guess that it's not ideal.

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
achou11: to summarize, to comprehensively handle errors from suspense queries, both of these need to be done:

- create an error boundary _somewhere_. ideally as close as you can, depending on the desired UX and granularity of handling that's needed. this accounts for errors from suspense queries that initially fail.
 
- do something with `error` variable returned by a suspenseful query. this refers to errors that occur _after_ successful queries have already happened.

src/frontend/screens/PhotoPreviewModal/index.tsx
achou11: > Also to further simplify, i think we should get rid of the ErrorBoundary, and just let the component internally handle the error (see comment below)

The error boundary is needed for more co-located error handling of suspense queries. See https://github.com/digidem/comapeo-mobile/pull/1203#discussion_r2130078249 for explanation.

achou11: I think I'm going to push back on removing the granular suspense boundary here. most of the screen's information is immediately available so it should be preferable to show the screen immediately if possible (one of the main benefits of suspense). blocking the whole screen based on a specific portion of the screen (in this case, the image loaded via HTTP) doesn't seem ideal to me. in most cases, the loading of the image will also be almost immediate, but given that it's a separate transport from the RPC layer we have, I think it's safer to treat it as though it's slightly more fragile.

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -124,7 +124,6 @@ export const Editor = ({
                           // TODO: Does it make sense to provide the `observationDocId` in this case?
                           // Reasoning for not doing so is because the photo isn't actually saved yet, so it's technically not
                           // officially associated with the observation being created/edited.
```
achou11, line 140: would like some thoughts on this. for draft photos that are created while editing an observation, we do not show the associated observation ID for the reasoning mentioned in this comment. wondering if that should be reconsidered though.

src/frontend/screens/PhotoPreviewModal/index.tsx
achou11: Creating another hook wasn't necessary (no other hooks are needed internally) and instead I extracted helper functions related to extracting photo info. see `lib/photos.ts` (with tests in `lib/photos.test.ts`)

src/frontend/sharedComponents/Editor/index.tsx
ErikSin: I think for simplicity lets leave it as is and if we get feedback from the field we can change it!

src/frontend/sharedComponents/Images/CoreBlobImage.tsx
ErikSin: No, your right, lets leave it as is

ErikSin (approved):
As I mentioned in my comments, exposing the `onLoad` and having the extra suspense feel a little over complicated. But upon conferring with Jen and Ximena, Im just going to approve as it is working as expected.

That being said, can you wait to merge it? I think this touches a lot of files, and v4 is still being QA'd so I would like to not complicate cherry-picking

achou11:
> That being said, can you wait to merge it?

I'd be okay with you merging it on my behalf whenever you're ready for it, if that's helpful.

ErikSin:
> > That being said, can you wait to merge it?
> 
> I'd be okay with you merging it on my behalf whenever you're ready for it, if that's helpful.

ok, yes i will do that. Feel free to manually close original issue on your sprint board.

ErikSin:
@achou11 can you update this from develop? I am going to merge it

achou11:
> @achou11 can you update this from develop? I am going to merge it

Done. Didn't spend the time to actually re-test on a device though as I'm trying to maintain focus on desktop. Would you mind doing that if possible before merging?

ErikSin:
Yup, it looks good to me!

PR #1205, opened 2025-06-03
test: add first sync screen assertion

EvanHahn:
*I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/comapeo-mobile/pull/1205/files?w=1).*

This adds a test assertion for the sync screen. The assertion itself is simple--all it does is check for text on screen--but some additional work was required.

It turned out that the component wasn't actually being rendered! The [Safe Area Context library][0] wasn't including its children in tests, which meant that nothing was rendering. Fixing that uncovered several other problems, which I fixed:

- I had to mock:
  - The Safe Area Context library, as mentioned, using [their built-in Jest mock][1]
  - `expo-location`
  - Native modules
  - Local discovery. I mocked our module rather than mapping lower-level things; for example, I could have mocked `NetInfo` and `AppState` and `Zeroconf`. I can probably find a way to change this but it feels tedious.
  - SVG imports, which otherwise import as `{ testUri: './path.svg' }`, which doesn't render and gives an error. I considered using [a library][2] to do this for me, but the mock was so simple that I just did it myself.
- I had to fix two resource leaks:
  - Device diagnostic metrics needed to be disabled, matching app diagnostic metrics
  - Tanstack Query needed to have garbage collection disabled. See the comment in `helpers/react.tsx`
- I had to include the i18n providers (as @achou11 [predicted](https://github.com/digidem/comapeo-mobile/pull/1175#discussion_r2109516470))

This change should have no user impact.

[0]: https://appandflow.github.io/react-native-safe-area-context/
[1]: https://appandflow.github.io/react-native-safe-area-context/testing/
[2]: https://www.npmjs.com/package/react-native-svg-mock


EvanHahn:
I'm surprised that CI is failing...it doesn't look related.

achou11:
> I'm surprised that CI is failing...it doesn't look related.

yeah separate issue being addressed in another PR

achou11:
any thoughts on what's documented here and whether it would make sense to introduce a more proper setup in this PR?

https://reactnavigation.org/docs/testing/

Mostly just thinking about how we're currently relying on something that's considered an antipattern, but honestly not sure about the technical details of getting something more idiomatic in place.

jest.setup.js
```diff
@@ -1,12 +1,38 @@
...
+      status: 'granted',
+      expires: 'never',
+      granted: true,
+      canAskAgain: true,
+    }),
+  ),
+  getLastKnownPositionAsync: jest.fn(() => Promise.resolve(null)),
+  watchPositionAsync: jest.fn(() => Promise.resolve({remove: jest.fn()})),
+};
+jest.mock('expo-location', () => mockExpoLocation);
```
achou11: soft recommendation to move this to `__mocks__/expo-location.ts`

tests/integration/mocks/svg-mock.tsx
achou11, line 1: can this be moved to the `__mocks__` directory?

achou11 (approved):
couple of non-blocking suggestions but seems like good progress overall!

jest.setup.js
EvanHahn: Ah, didn't know we already had one of those. Moved in d164b79225803ee98c03ecc170215eabc01ba85d.

tests/integration/mocks/svg-mock.tsx
EvanHahn: Moved in cd0782c2b552dc182bf8627135547af9496c33cc.

EvanHahn:
Things broke after merging with `develop`. Investigating...

EvanHahn:
Should be fixed. (edaffc9025f9334260de5cd96442535a832308d0 added a new app provider, and I needed to mock something additional.)

EvanHahn:
> any thoughts on what's documented here and whether it would make sense to introduce a more proper setup in this PR?
> 
> https://reactnavigation.org/docs/testing/
> 
> Mostly just thinking about how we're currently relying on something that's considered an antipattern, but honestly not sure about the technical details of getting something more idiomatic in place.

If anything, I think this is worth doing in a follow-up. I generally agree with the philosophy the docs you linked, but I don't see fixing it as very related to this task. Personally, I'd file an issue to investigate this and move on.

As a small improvement, we could remove the `@ts-expect-error` like this:

<details>
<summary>Code</summary>

```diff
diff --git a/src/frontend/screens/Exchange.test.tsx b/src/frontend/screens/Exchange.test.tsx
index 2c91ebbb..f88c9c31 100644
--- a/src/frontend/screens/Exchange.test.tsx
+++ b/src/frontend/screens/Exchange.test.tsx
@@ -32,13 +32,9 @@ describe('Exchange screen', () => {
       goBack: jest.fn(),
     };
 
-    render(
-      <SyncScreen
-        // @ts-expect-error Not ideal but bare minimum navigation needed by the screen
-        navigation={navigation}
-      />,
-      {wrapper: appProviders.wrapper},
-    );
+    render(<SyncScreen navigation={navigation} />, {
+      wrapper: appProviders.wrapper,
+    });
 
     // TODO: Add assertions
   });
diff --git a/src/frontend/screens/Exchange/index.tsx b/src/frontend/screens/Exchange/index.tsx
index 46ae10d8..b18de7b6 100644
--- a/src/frontend/screens/Exchange/index.tsx
+++ b/src/frontend/screens/Exchange/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {useSyncState} from '@comapeo/core-react';
+import {PickDeep} from 'type-fest';
 
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
 import {
@@ -23,7 +24,9 @@ const m = defineMessages({
   },
 });
 
-export const SyncScreen = ({navigation}: NativeRootNavigationProps<'Sync'>) => {
+export const SyncScreen = ({
+  navigation,
+}: PickDeep<NativeRootNavigationProps<'Sync'>, 'navigation.goBack'>) => {
   const wifiStatus = useLocalDiscoveryState(state => state.wifiStatus);
 
   const hasInternetAccess = useNetInfo().isConnected;
```
</details>

PR #1212, opened 2025-06-04
chore: audio improvements

### LINKED ISSUE #1182: Audio UX Improvements
**Description**
Improvements to the Audio Recording feature

See clickable prototype : https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=5698-3212&t=xP9pskMLuvQFL2yC-1&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1&starting-point-node-id=5698%3A3212&show-proto-sidebar=1

Feature spec : https://www.notion.so/digidem/Audio-UX-Improvements-1791b08162d58073b66bf39764a25281

**Todo**

## Updated UX Flow
1. From New Observation, tapping Audio instantly starts a new audio recording.
2. The audio recording continues until it reaches 5 minutes or a person taps on the Stop button
3. When recording is stopped, the Saved Audio screen appears. At this point, the Audio Recording is attached to the Observation but the Observation itself is not yet saved.
    1. A person can Play to the Audio Recording
    2. A person can Delete the Audio Recording
    3. A person can Close the Audio Recording and return to the New Observation editor.
4. When a person closes out of Saved Audio, they see the Audio Thumbnail version of the recording now attached to the Observation.

## Updated Recording Audio UI

Image

- [x] When tapping Audio icon, display bright blue from the beginning (it ends with the same color as well. no more color progression)

## Recording Saved to Observation / Preview Audio UI

Image

- [x] When tapping on Stop, recording automatically saved
- [x] Display "Recording Saved!"
- [x] Display preview, user can playback, see timestamp and when recording was saved
- [x] Tapping on "Back to Editing" , user goes back to Observation editor, sees the recording saved
- [x] Tapping on "Delete" , user goes back to Observation editor, sees the recording is not saved

## New Observation Editor UI, featuring Audio Recording Thumbnail UI

Image

- [ ] Update Audio Recording thumbnail as designed

## View Audio Recording UI (via Observation Detail screen)

Image

- [ ] Tapping on the saved recording, open Playback screen
- [ ] Display Playback screen as designed

**Designs**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=5698-3285&t=K9U6PUl5Q1R9Y7mr-0

**Video walkthrough**
https://drive.google.com/file/d/1ic_Of39byrCvP3oTdTrNIMrsLzAdtfZS/view

### LINKED ISSUE #1222: Edit Obsv/ cant save when Audio recorded + deleted
Not fixed - https://github.com/digidem/comapeo-mobile/issues/1127

- Create an observation with a photo and Save it
- Go back to the observation and tap on Edit
- Create an audio recording
- Delete that recording
- Go back to the editor and tap on save
- Screen will open saying “something went wrong”

cimigree:
Closes #1182 
Closes #1222 

This PR updates the audio recording and playback flow for observations according to the specifications outlined here:
https://www.notion.so/digidem/Audio-UX-Improvements-1791b08162d58073b66bf39764a25281
Using architecture discussed here:
https://www.notion.so/digidem/Audio-Refactor-1fb1b08162d5803a940be625c06c4c1a

### Feature Updates

* Audio recording starts immediately after tapping the microphone button.
* The animated recording background is simplified to a single color.
* Audio is now saved immediately after stopping, removing the extra confirmation screen.
* A delete button is shown for draft recordings — deletion is immediate with no confirmation.

###  Playback Screens (Draft & Saved -- new names)

* Play/stop toggle
* Progress bar
* Duration and timestamp
* I made the decision to show duration as 0:00 / 2:23 (eg) and the first part increases as it plays.

###  Restrictions

*  Audio cannot be deleted once it is part of a saved observation.
*  Audio can only be shared from the saved observation screen thumbnail.

### Thumbnails

* **Draft thumbnails** display:

  * Duration
  * Relative time (e.g., “a few seconds ago”)
* **Saved thumbnails** display:

  * Relative time only which is now a property that comes from the audio attachment

### Technical Fixes

* Added retry logic for `startRecording()` to address a race condition immediately after permissions are granted.
* Logic for turning an attachment into Audio happens in the Attachment playback screen now.
* Updated E2E tests and documentation to reflect new flow and validations. Unfortunately, we can't test playing a saved audio from an observation because of known URL difficulties within Browserstack. See this issue: https://github.com/digidem/comapeo-mobile/issues/1008


src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin, line 1: I think it would be a bit simpler if this hook did not expose `isStopping`, and instead allow the consuming component to just deal with that.

In otherword, in `AudioRecording/index` just create this state since `stopRecording` is async

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -27,50 +27,60 @@ const m = defineMessages({
...
+    };
+    start();
+  }, [startRecording]);
+
+  const finishRecording = React.useCallback(async () => {
+    const result = await stopRecording();
+    if (result?.uri && result.createdAt) {
+      navigation.replace('AudioDraftPlaybackScreen', {
+        uri: result.uri,
+        createdAt: result.createdAt,
```
ErikSin: i think we should account for errors here, and if no `uri` or `createdAt` then navigate to the error bottom sheet and do an early return

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -27,50 +27,60 @@ const m = defineMessages({
...
+    const sub = AppState.addEventListener('change', state => {
+      if (state !== 'active' && isRecording) {
+        finishRecording();
+      }
+    });
+    return () => sub.remove();
+  }, [isRecording, finishRecording]);
 
   React.useEffect(() => {
     const subscription = AppState.addEventListener('change', nextAppState => {
```
ErikSin: I think this is a typo, there are now 2 of the same subscription

src/frontend/screens/Audio/AudioRecording/index.tsx
```diff
@@ -27,50 +27,60 @@ const m = defineMessages({
...
   navigation,
 }: NativeRootNavigationProps<'AudioRecording'>) {
   const isE2E = process.env.EXPO_PUBLIC_E2E_TEST === 'true';
-  const {startRecording, stopRecording, status} = useAudioRecording();
-
+  const {startRecording, stopRecording, status, isStopping} =
+    useAudioRecording();
   const timeElapsed = status?.durationMillis || 0;
   const isRecording = !!status?.isRecording;
+  const [recordingReady, setRecordingReady] = React.useState(false);
```
ErikSin: i think we should unify the loading state. Just have one loading state for both starting and stopping the recording And that loading state just  changes the button into a loader. 

Currently there are 2 different loading state hooks, and 2 different UIs for loading which i think over complicates this component

src/frontend/sharedTypes/navigation.ts
```diff
@@ -108,14 +108,17 @@ export type RootStackParamsList = {
...
   };
   AudioRecording: undefined;
-  AudioPlaybackUnsavedReview: {uri: string; duration: number};
-  AudioPlaybackUnsavedPreview: {uri: string};
-  DeleteAudioBottomSheet: {
-    onPressDelete: () => void;
+  AudioDraftPlaybackScreen: {
     uri: string;
+    createdAt: number;
+    fromEditorPreview?: boolean;
```
ErikSin: can we rename this to make it more explicit as to what exactly it is doing. Perhaps `showRecordingSavedText`. Also I dont think we should make it an optional param

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -162,15 +162,11 @@ export const Editor = ({
...
+                    <AudioDraftThumbnail
                       key={att.uri}
                       size={size}
-                      onPress={() =>
-                        navigate('AudioPlaybackUnsavedPreview', {uri: att.uri})
-                      }>
-                      <PlayArrow width={48} height={48} />
-                    </ThumbnailContainer>
+                      audio={att}
+                    />
```
ErikSin: I think we should just get rid of `DraftAudioThumbnail` as you are just prop forwarding. Instead lets just copy and paste your code into the parent component. 
```ts
<ThumbnailContainer
      size={size}
      onPress={() =>
           // nav action
      }
      containerStyle={styles.container}
      accessibilityLabel="Play audio recording.">
      <PlayArrow width={48} height={48} />
      <DateDistance
        date={new Date(observation.createdAt)}
        style={{fontSize: 12, fontWeight: 400, color: DARK_GREY}}
      />
    </ThumbnailContainer>
```
I understand creating a component for the Saved Audio as that uses a hook

ErikSin (changes requested):
overall looks good. Some minor, but blocking comments

src/frontend/sharedComponents/Editor/index.tsx
cimigree: Ok

ErikSin:
Can you update from develop @cimigree ?

src/frontend/screens/Audio/AudioRecording/index.tsx
ErikSin: Sorry, what I was trying to communicate was a little unclear. This is what i think would be simpler:

```tsx
const [isLoading, setIsLoading] = React.useState(true);

React.useEffect(() => {
    const start = async () => {
      await startRecording();
      setIsLoading(false);
    };
    start();
  }, [startRecording]);
  
  const finishRecording = React.useCallback(async () => {
    setIsLoading(true);
    try {
      const result = await stopRecording();

      if (!result?.uri || !result.createdAt) {
        navigation.replace('ErrorBottomSheet');
        return;
      }

      navigation.replace('AudioDraftPlaybackScreen', {
        uri: result.uri,
        createdAt: result.createdAt,
        showRecordingSavedText: true,
      });
    } finally {
      setIsLoading(false);
    }
  }, [stopRecording, navigation]);
  
  //And then in dockContent:
  dockContent={
          isLoading ? (
            <View style={styles.center}>
              <UIActivityIndicator
                color={WHITE}
                size={PRIMARY_CONTROL_DIAMETER}
              />
            </View>
          ) : (
            <TouchableOpacity
              onPress={finishRecording}
              style={audioStyles.basePressable}
              accessibilityLabel="Stop recording audio.">
              <View style={audioStyles.stop} />
            </TouchableOpacity>
          )
        }>

```

Then you can get rid of `isStopping` and `isRecording` as they are encapsulated in 1 loading state. And you can get rid of:

```ts
if (!recordingReady) {
    return (
      <View
        style={{
          flex: 1,
          backgroundColor: DARK_GREY,
          justifyContent: 'center',
          alignItems: 'center',
        }}>
        <UIActivityIndicator color={WHITE} />
      </View>
    );
  }


```

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -1,50 +1,65 @@
...
+      setRecordingInstance(recording);
     } catch {
-      navigate('ErrorBottomSheet');
+      try {
+        // Retry once after short delay — helps avoid timing issues when mic permission was just granted
+        await new Promise(res => setTimeout(res, 200));
+        const {recording} = await Audio.Recording.createAsync(
+          Audio.RecordingOptionsPresets.HIGH_QUALITY,
+          stat => setStatus(stat),
+        );
```
ErikSin: I found another way to do this without relying on a `setTimeout`

In the consuming component `AudioRecording/index.tsx` use `Audio.usePermissions()`:

```ts
//AudioRecording/index.tsx

const [audioPermission] = Audio.usePermissions();


//then check permission before starting
React.useEffect(() => {
    const start = async () => {
      if (!audioPermission) return;
      if(!audioPermission.granted){
      //this shouldnt happen so we could log into sentry if it does
          navigation.replace('AudioAskPermissionBottomSheet')
          return     
       }
      await startRecording();
      setIsLoading(false);
    };
    start();
  }, [startRecording, audioPermission]);
  
  
  
  //And then in the dock content:
  
dockContent={
          isLoading || !audioPermission ? (
            <View style={styles.center}>
              <UIActivityIndicator
                color={WHITE}
                size={PRIMARY_CONTROL_DIAMETER}
              />
            </View>

```

ErikSin (changes requested):
2 small things, but i think they are still blocking.

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -1,50 +1,63 @@
...
 export function useAudioRecording() {
-  const [recordingPromise, setRecordingPromise] =
-    useState<Promise<Audio.Recording> | null>(null);
+  const [recordingInstance, setRecordingInstance] =
+    useState<Audio.Recording | null>(null);
   const [status, setStatus] = useState<Audio.RecordingStatus | null>(null);
   const {navigate} = useNavigationFromRoot();
+  const {addAudio} = useDraftObservation();
 
   const reset = useCallback(async () => {
```
ErikSin: We don't actually ever use reset, so lets just get rid of it!

ErikSin (approved):
One small - non blocking comment and then you can merge!

PR #1215, opened 2025-06-06
chore: update usage of location in map component

ErikSin:
Update the map component with the new location hook. The GPS Pill and Current Tracks now consumes their own location from context, allowing it to update consistently, without effecting the throttled updating of the map.

I also created the `SavedLocationProvider`. We want to center the user in the last known location if there location is turned off and they are opening the app. Previously we were trying to do that via `useLastKnownLocation`. This only works if the app is already open, so when the user was opening the app, the location was undefined. This was centering the user in the middle of the ocean which was confusing ui.

Now the last saved location is persisted to state, and it updates every time the user closes the app, or when the user updates their permission or turn off their location service. I also updated how the map is being centered on first open. Previously we were using `MapBox.Camera DefaultSettings`. This had buggy implementation as it often being overridden `MapBox.Camera's` `centerCoordinate` Prop. Now we are setting the center coordinate programatically to ensure that it is only be set once (on first open).


src/frontend/contexts/LocationContext.tsx
```diff
@@ -214,19 +215,38 @@ function startWatchPosition(store: StoreApi<LocationState>) {
...
   React.useEffect(() => {
     let isCancelled = false;
+
     const handleAppStateChange = async (nextAppState: AppStateStatus) => {
+      if (isCancelled) return;
+
+      const location = store.getState().location;
+      // update saved location when app goes in the background
+      if (nextAppState === 'background' || nextAppState === 'inactive') {
+        if (location) setSavedLocation(location);
```
cimigree, line 229: Minor architectural note: setSavedLocation() is called in multiple places (useCheckPermissionOnAppStateChange, usePollProviderStatus, etc.), sometimes without checking whether the value has actually changed.

Would it make sense to debounce or skip redundant calls to avoid repeated writes to persistent storage?

src/frontend/contexts/LocationContext.tsx
```diff
@@ -244,43 +264,52 @@ function useCheckPermissionOnAppStateChange(store: StoreApi<LocationState>) {
       isCancelled = true;
       subscription.remove();
     };
-  }, [store]);
+  }, [store, setSavedLocation]);
 }
 
 function usePollProviderStatus(store: StoreApi<LocationState>) {
   const queryClient = useQueryClient();
+  const {setSavedLocation} = useSavedLocationActions();
```
cimigree, line 272: Minor suggestion: the logic in usePollProviderStatus and useCheckPermissionOnAppStateChange overlaps some — both update location state and may call setSavedLocation(). Wondering if this could be consolidated for fewer duplicated conditionals.

cimigree (commented):
Overall this PR introduces a solid improvement to how location is persisted and how the  map camera is initialized (I think -- see below).

The saved location fallback is promising, but behavior is inconsistent when starting the app or creating a project — on develop, the initial placement on the map felt smoother. Might be worth revisiting the fallback logic or ensuring setCamera() doesn't run before GPS is ready?

Obviously need to fix the merge conflicts.

None of these are blocking, but worth tracking as follow-up.
Some things I noticed with testing and comparing.
The initial location on map sometimes starts in Africa instead of saved or fallback coordinates — this behavior was actually better on develop. This happens when I force close and reopen or when I name the default project or create a new project.

The user location dot did not follow my movement on this branch and seemed more responsive on develop.

The GPS pill and location modal update constantly  — also present on develop although maybe worse on this branch, so not introduced here but I guess I was hoping fixing the location stuff would have fixed it? But maybe it is expected? I don't know. I attached a video.
https://photos.app.goo.gl/SXTgS3Uk285vZPLB9

There was a Console warning seen frequently on this branch that I did not see on develop:
`MaxListenersExceededWarning: Possible EventEmitter memory leak detected.`

Also, this file: src/frontend/screens/LocationInfoScreen.tsx
Looks like useLastKnownLocation() is still used here. Since this PR introduces SavedLocationContext as the new source of truth, it might be worth replacing these references to fully migrate off the old location system.

Same applies to CoordinateFormat.tsx, which also uses it for previewing coordinates. Might be worth removing useLastKnownLocation entirely once all usages are cleaned up.

In this file: src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts
Not blocking, but this hook creates its own watchPositionAsync subscription rather than relying on LocationContext. It may be intentional due to its one-off filtering logic, but worth revisiting whether all location updates should be owned centrally for clarity and consistency.

Same pattern exists in DraftObservationContext.tsx. Maybe document what is not using the new location stuff and why?

ErikSin:
> The saved location fallback is promising, but behavior is inconsistent when starting the app or creating a project — on develop, the initial placement on the map felt smoother. Might be worth revisiting the fallback logic or ensuring setCamera() doesn't run before GPS is ready?

and

> The initial location on map sometimes starts in Africa instead of saved or fallback coordinates — this behavior was actually better on develop. This happens when I force close and reopen or when I name the default project or create a new project.

fixed here: https://github.com/digidem/comapeo-mobile/pull/1215/commits/3b14dd86e92e5017dbce49942c10fbf48afad6d5

ErikSin:
> The GPS pill and location modal update constantly — also present on develop although maybe worse on this branch, so not introduced here but I guess I was hoping fixing the location stuff would have fixed it? But maybe it is expected? I don't know. I attached a video.
> https://photos.app.goo.gl/SXTgS3Uk285vZPLB9

This is actually the desired behaviour. The GPS pill should always indicate the CURRENT accuracy, so the user can make an informed decision about when they want to take an observation. Previously it was throttling the accuracy, and only updating every 5m, which was needed for the map but should not have been applied to the GPS pill

ErikSin:
> Also, this file: src/frontend/screens/LocationInfoScreen.tsx Looks like useLastKnownLocation() is still used here. Since this PR introduces SavedLocationContext as the new source of truth, it might be worth replacing these references to fully migrate off the old location system.
> 
> Same applies to CoordinateFormat.tsx, which also uses it for previewing coordinates. Might be worth removing useLastKnownLocation entirely once all usages are cleaned up.

`useLastKnownLocation` and `useNonReactiveSavedLocation` are actually slightly different use cases. `useLastKnownLocation` just finds the last known location in GPS. This allows us to use a location object without having it constantly updating and causing react re-renders like `useLocationState` does. For the coordinateFormat and LocationInfoScreen, we just need one stable location, and we want it to be as up to date as possible. On the otherhand `useNonReactiveSavedLocation` is actually used for placing the user on the map initially. `useNonReactiveSavedLocation` only updates when the app goes in the background. So if the user hasnt put the app in the background and moved a very far distance away from the initial location, it will still only return that initial location. While useLastKnownLocation will update with the most up to date location without being reactive and causing unnecessary re-renders

src/frontend/contexts/LocationContext.tsx
ErikSin: im not sure how we could do that as both of these checks are separate subscriptions. Did you have something in mind?

src/frontend/contexts/LocationContext.tsx
ErikSin: We could, but i think it would complicate the code. There could be the case that the user turns off their location, updates their permission, and goes into the background. This would cause 3 separate writes to persisted state all of the same location. But i think having persisted state being written 3 times in a row is not a problem. It is not an expensive process, and won't slow down the app. And the saved location does not cause re-renders (aka it would not cause 3 state updates).

The alternative would be write debounce code, and then also check that previous locaiton does not equal the new location. This feels like it would be a lot of code for a small use case.

ErikSin:
> There was a Console warning seen frequently on this branch that I did not see on develop:
> `MaxListenersExceededWarning: Possible EventEmitter memory leak detected.`

This feels like a blocker to me. I unfortunately cannot re-create, so let me know if you can figure out the conditions in which this happened

src/frontend/contexts/LocationContext.tsx
cimigree: That makes sense — I agree it's not a performance issue. Okay I guess then my concern is more about understanding when and why setSavedLocation() is being triggered.

Totally fine to keep the way it is for now, but maybe worth a small comment somewhere noting that redundant writes are intentional and not harmful — just to avoid future me or someone else trying to “fix” it unnecessarily or wondering if something’s wrong.

src/frontend/contexts/LocationContext.tsx
cimigree: Yeah totally fair — they need to be separate subscriptions, I get that. I wasn’t thinking of consolidating the effects themselves, just whether the overlapping logic inside them (like checking locationServicesEnabled, updating location, calling setSavedLocation, etc.) could maybe live in a shared helper or callback.

Might not be worth it. Just a thought

cimigree:
Good news! I could not get the memory leak again. Perhaps it was because the branch had not been updated with develop and had the merge conflict previously and that is what I had been testing with. I think I did everything I did before so I would not worry about it at this time.

cimigree (approved):

PR #1216, opened 2025-06-06
Chore: exchange settings

### LINKED ISSUE #1132: Move Exchange Settings (Media Manager) to Exchange screen
**Description**
Right now, Exchange Settings (aka Sync Settings) lives under Project Settings. With the introduction of Multi-projects, we need to move this feature to make it device-wide. Since it's part of the Exchange process, we will move it to the exchange screen.
We are also taking the opportunity to update the Exchange screen

Notable updates :
- Star icon for Exchange everything, Grey leaf icon for Preview Only
- Text and link to update the exchange settings

Initial Issue reported :
https://www.notion.so/digidem/Move-Exchange-Settings-to-Exchange-screen-1e31b08162d580a8940cf7c619c4c0e8

Figma files : https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6086-3499&t=EUZhGecd77bNbCqo-0

**Todo**

## Exchange screen UI / Exchange everything

Image

Image

- [ ] Update screen UI and copy as seen on Figma
- [ ] When no devices found, Exchange icon in grey
- [ ] When devices are found, Exchange icon turns orange and display Star icon on top
- [ ] Display text regarding "Exchange everything" text
- [ ] When tapping on "Change Settings" go to Exchange Settings
- [ ] When exchanging, do not display the "Change Settings" link
- [ ] When exchaning and "up to date" bring back the  "Change Settings" link

## Exchange screen UI / Previews Only

Image

- [ ] Update screen UI and copy as seen on Figma
- [ ] Display Grey Leaf icon on top of Orange Exchange icon
- [ ] When devices are found, Exchange icon turns orange and display Leaf icon on top
- [ ] Display text regarding "Exchange Preview Only" text
- [ ] When tapping on "Change Settings" go to Exchange Settings
- [ ] When exchanging, do not display the "Change Settings" link
- [ ] When exchaning and "up to date" bring back the  "Change Settings" link

## Exchange Settings

Image

- [ ] "Change Settings" is  Exchange Settings bottom sheet open
- [ ] Display the bottom sheet  UI and copy as seen on Figma
- [ ] Display orange icon next to Exchange Everything
- [ ] Display grey leaf icon next to Previews only

## No wifi

Image

- [ ] Display UI and copy as seen on Figma

cimigree:
Closes #1132  
Design: [Figma – Exchange Redesign](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=0-1&p=f&t=mpDByL06EFdl8waf-0)  
Initial Issue: [Notion – Move Exchange Settings to Exchange Screen](https://www.notion.so/digidem/Move-Exchange-Settings-to-Exchange-screen-1e31b08162d580a8940cf7c619c4c0e8)

---

## Summary

This PR moves **Exchange Settings** out of Project Settings and into the **Exchange screen**, making the setting device-wide. It also introduces a redesigned user experience for Exchange.

## Exchange Screen Updates

- Extracted some of the render logic into pure components because the main screen was too too long and big.
- Added an Exchange Settings Bottom Sheet
- Uses comapeo/core-react directly
- Updates the No wifi screen per design
- Deletes old stuff no longer used

##  Adds Tests

- Added end-to-end tests under `tests/e2e/specs/exchange/`
- Adds documentation for the tests


src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+}: {
+  setting: MediaSyncSetting;
+  icon: React.ReactElement;
+  title: string;
+  description: string;
+  selected: MediaSyncSetting;
+  onSelect: (s: MediaSyncSetting) => void;
+  testID: string;
+}) => {
+  const isSelected = selected === setting;
```
achou11: can specify `isSelected` as a prop instead of doing the calculation internally:

```suggestion
  isSelected,
  onSelect,
  testID,
}: {
  setting: MediaSyncSetting;
  icon: React.ReactElement;
  title: string;
  description: string;
  isSelected: boolean;
  onSelect: (s: MediaSyncSetting) => void;
  testID: string;
}) => {
```

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+          styles.radioOuter,
+          {borderColor: isSelected ? DARK_GREY : NEW_DARK_GREY},
+        ]}>
+        {isSelected && (
+          <View
+            style={[styles.radioInner, {backgroundColor: DARK_GREY}]}
+            testID={`EXCHANGE.radio-selected-${setting}`}
+          />
+        )}
+      </View>
```
achou11: seems easier to use the (un)checked radio icons from Material in this case (which we use elsewhere)

image

src/frontend/screens/Exchange/ExchangeSettingsCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <BodyText variant="smallMeta" style={{color: NEW_DARK_GREY}}>
+        {mediaDescription}
+      </BodyText>
+      <BodyText variant="smallMeta" style={{color: NEW_DARK_GREY}}>
+        {storageDescription}
+      </BodyText>
+      {showChangeSettingsLink && (
+        <Button variant="text" onPress={onPressChangeSettings}>
+          <HeaderText variant="header6" style={styles.changeLink}>
+            Change Settings
```
achou11: should be translatable

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+    id: 'screens.ExchangeSettingsBottomSheet.previewsDesc',
+    defaultMessage: 'Reduced smaller size photos. No audio included.',
+  },
+});
+
+export const ExchangeSettingsBottomSheet = () => {
+  const {mutate: setMediaSyncSetting} = useSetMediaSyncSetting();
+  const currentSetting = useGetMediaSyncSetting();
+  const {formatMessage: t} = useIntl();
+  const [selected, setSelected] = useState<MediaSyncSetting>('everything');
```
achou11: initial value should be instantiated based on the actual value
```suggestion
  const [selected, setSelected] = useState<MediaSyncSetting>(() => {
    return currentSetting 
  });
```

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const {goBack} = useNavigationFromRoot();
+
+  const handleSave = () => {
+    setMediaSyncSetting(selected);
+  };
+
+  React.useEffect(() => {
+    if (currentSetting) setSelected(currentSetting);
+  }, [currentSetting]);
```
achou11: is this needed? seems less necessary if you update the state initialization of `selected` as I suggested elsewhere.

otherwise, can only imagine this being potentially needed if the value from `useGetMediaSyncSetting` is updated from somewhere outside of this, which I don't think happens.

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+            selected={selected}
+            onSelect={setSelected}
+            testID="EXCHANGE.option-everything"
+          />
+          <OptionCard
+            setting="previews"
+            icon={<GreyLeaf width={30} height={30} />}
+            title={t(m.previewsTitle)}
+            description={t(m.previewsDesc)}
+            selected={selected}
```
achou11: example of changes if using a `isSelected` prop as suggested elsewhere
```suggestion
            isSelected={selected === "everything"}
            onSelect={setSelected}
            testID="EXCHANGE.option-everything"
          />
          <OptionCard
            setting="previews"
            icon={<GreyLeaf width={30} height={30} />}
            title={t(m.previewsTitle)}
            description={t(m.previewsDesc)}
            isSelected={selected === "previews"}
```

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+export const ExchangeSettingsBottomSheet = () => {
+  const {mutate: setMediaSyncSetting} = useSetMediaSyncSetting();
+  const currentSetting = useGetMediaSyncSetting();
+  const {formatMessage: t} = useIntl();
+  const [selected, setSelected] = useState<MediaSyncSetting>('everything');
+  const {goBack} = useNavigationFromRoot();
+
+  const handleSave = () => {
+    setMediaSyncSetting(selected);
```
achou11: seemingly not specified in the designs but should also:

1. close the screen upon success

2. handle any errors

src/frontend/screens/Exchange/NoWifiDisplay.tsx
```diff
@@ -1,66 +1,165 @@
...
+              style={{
+                position: 'absolute',
+                right: 0,
+                bottom: 0,
+                backgroundColor: WARNING_RED,
+              }}>
+              <HeaderText variant="header4" style={{color: WHITE}}>
+                !
+              </HeaderText>
+            </Circle>
```
achou11: should this be using a vector icon instead?

src/frontend/screens/Exchange/NoWifiDisplay.tsx
```diff
@@ -1,66 +1,165 @@
...
+    borderRadius: 6,
   },
-  descriptionText: {
-    fontSize: 24,
-    textAlign: 'center',
+  wifiCardTextContainer: {
+    flexShrink: 1,
+    flexGrow: 0,
+    flexBasis: 'auto',
+    maxWidth: '80%',
```
achou11: why a percentage-based max width instead of defining a padding on the containing view? measurements based on percentages will lead to variation between devices due to differences in viewport sizes, which is sometimes useful but not sure if that's desired in general for most of the designs.

src/frontend/screens/Exchange/WifiCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Circle color="#CCE0FF" radius={14} style={styles.signalIndicator}>
+          <Icon size={16} color={BLACK} />
+        </Circle>
+        <View style={styles.textContainer}>
+          {ssid ? (
+            <BodyText style={styles.wifiName}>{ssid}</BodyText>
+          ) : (
+            <>
+              <BodyText>--</BodyText>
+              <BodyText style={styles.wifiName}>No Wi-Fi</BodyText>
```
achou11: should be translatable

src/frontend/sharedComponents/icons/index.tsx
```diff
@@ -130,20 +130,21 @@ export const WifiOffIcon = ({
...
   color = 'white',
   style,
 }: FontIconProps) => (
-  <MaterialIcon color={color} style={style} name="wifi" size={size} />
+  <MaterialIcon
+    color={color}
+    style={style}
+    name="wifi"
+    size={size}
+    testID="wifi-icon"
```
achou11: usually should be a prop to avoid ID clashes (e.g. if you render multiple at once)

src/frontend/screens/Exchange/ExchangeScreenContent.tsx
achou11, line 1: more of a design concern but maybe it's worth confirming that the vertical overflow behavior in this case is okay:

image

This is on a Pixel 2 with the default font size. Personally wouldn't care too much but mentioning it since it hides an actionable button in this example (`Change Settings`).

src/frontend/screens/Exchange/ExchangeSettingsCard.tsx
achou11, line 1: could just be me, but the text sizing here seems a bit small. basing this on the relative sizing compared to the big header text in the screen, for example:

designs:

image


what's implemented:

image

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,210 @@
...
+    defaultMessage: 'Exchange Previews Only',
+  },
+  previewsDesc: {
+    id: 'screens.ExchangeSettingsBottomSheet.previewsDesc',
+    defaultMessage: 'Reduced smaller size photos. No audio included.',
+  },
+});
+
+export const ExchangeSettingsBottomSheet = () => {
+  const {mutate: setMediaSyncSetting} = useSetMediaSyncSetting();
```
achou11: maybe out of scope, but i don't think there's a strong need to have this custom mutation anymore. it used to be needed because the `mutationKey` option was being used, but the new UX that this PR introduces makes that no longer the case. I would suggest deleting `useSetMediaSyncSetting` entirely and using `useSetIsArchiveDevice` directly instead:

```ts
const {mutate: setIsArchiveDevice} = useSetIsArchiveDevice();

const handleSave = () => {
  // 🟠 Can import the `isArchiveDevice()` helper if preferred,
  // but honestly think that could also be deleted entirely in favor of a direct check like this
  setIsArchiveDevice({isArchiveDevice: selected === 'everything'});
};

**Review by achou11 — CHANGES_REQUESTED** [review]
Overall UI changes seem solid! Provided some relatively minor suggestions and pointed out a couple places that require updates.

File: src/frontend/screens/Exchange/ExchangeSettingsCard.tsx (hunk 3 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsCard.tsx]: Oof. That was an oversight!

File: src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx (hunk 5 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: Yes you are right. Maybe I should not be working all day on a Friday! Sorry for all the need for extra comments!

File: src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx (hunk 6 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: Yes. I did this...

File: src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx (hunk 7 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: Ok I wasn't sure if it should close or be a manual close. Will do.

File: src/frontend/screens/Exchange/ExchangeSettingsCard.tsx
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsCard.tsx:1]: My bad. I was looking at the height instead of the fontsize.

File: src/frontend/screens/Exchange/ExchangeScreenContent.tsx
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeScreenContent.tsx:1]: Okay I think I figured it out. Can you check again?

File: src/frontend/screens/Exchange/NoWifiDisplay.tsx (hunk 8 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/NoWifiDisplay.tsx]: Yeah good idea.

File: src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx (hunk 4 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: Of course. Sorry.

File: src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx (hunk 12 above)
Comment (cimigree) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: I got rid of the file and just used comapeo/core-react directly in the files.

File: src/frontend/screens/Exchange/ExchangeScreenContent.tsx
Comment (achou11) [reply src/frontend/screens/Exchange/ExchangeScreenContent.tsx:1]: much better! now it doesn't overflow by default 👍  (unless I set some rather extreme maximums for font/display size)

File: src/frontend/screens/Exchange/ExchangeScreenContent.tsx [hunk 13]
```diff
@@ -138,10 +143,12 @@ export const ExchangeScreenContent = ({syncState}: {syncState: SyncState}) => {
   const {projectApi, projectId} = useActiveProject();
   const progress = useDataSyncProgress({projectId});
   const startSync = useStartSync({projectId});
+  const {data: isArchive} = useIsArchiveDevice();
 
-  const ssid = useLocalDiscoveryState(state => state.ssid);
+  const currentMediaSetting = isArchive ? 'everything' : 'previews';
+  const [syncWasStopped, setSyncWasStopped] = React.useState(false);
```
achou11: why a separate state for this as opposed to using `syncState.data.isSyncEnabled`? From what I can tell, you should be able to use `!syncState.data.isSyncEnabled` wherever `syncWasStopped` is used

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,198 @@
...
+          numberOfLines={undefined}>
+          {description}
+        </BodyText>
+      </View>
+      {icon}
+    </Pressable>
+  );
+};
+
+ExchangeSettingsBottomSheet.navTitle = m.save;
```
achou11: is a nav title needed for this screen?

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
```diff
@@ -0,0 +1,198 @@
...
+  isSelected,
+  onSelect,
+  testID,
+}: {
+  setting: MediaSyncSetting;
+  icon: React.ReactElement;
+  title: string;
+  description: string;
+  isSelected: boolean;
+  onSelect: (s: MediaSyncSetting) => void;
```
achou11: maybe stylistic but can make this prop have less knowledge and let the consumer handle more:

```
  onSelect: () => void;
```

From the consuming component
```tsx
<OptionCard
  setting="everything"
  onSelect={() => {
    setSelected("everything")
  }}
/>

Comment (achou11) [reply src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx]: alternatively, can use a generic to make sure that the argument type that's passed to `onSelect` matches the value used in the `setting` prop:

```tsx
const OptionCard = <Setting extends MediaSyncSetting>({
  setting,
  onSelect,
}: {
  setting: Setting;
  onSelect: (s: Setting) => void;
}) => {
  // ...
}
```

```tsx
<OptionCard
  setting="everything"
  onSelect={(value) => {
    // Value here is always `"everything"`
    setSelected(value)
  }}
/>

<OptionCard
  setting="previews"
  onSelect={(value) => {
    // Value here is always `"previews"`
    setSelected(value)
  }}
/>
```

Generally think my other suggestion is preferable though.

achou11 (changes requested):
really nice PR overall! have one last concern about a potentially unnecessary state variable but otherwise the changes look good 👍

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
cimigree: Ok.

src/frontend/screens/Exchange/ExchangeSettingsBottomSheet.tsx
cimigree: Nope. Fixed.

src/frontend/screens/Exchange/ExchangeScreenContent.tsx
cimigree: Ok fixed.

achou11 (approved):
might be worth updating with develop branch and checking that CI passes, but otherwise lgtm

PR #1217, opened 2025-06-09
test: Exchange screen simulated sync

EvanHahn:
Following up on d7563f5ebf93dc6c561d7f66ef6bf7cf855db487 and e199928271b45cd6b874607e217d768617aeaf79, this fleshes out tests for the Exchange screen. It adds two major tests:

1. What does the screen look like in "solo mode", before anyone else has joined? This is a simple test.

2. How does syncing work between a simulated peer? This test invites a new (fake) device to the project and tests syncing observations. This required a lot of scaffolding, much of which was adapted from [similar test code in `@comapeo/core`][0].

This task required mocking several more parts of the system:

- `@gorhom/bottom-sheet` and `react-native-device-info` needed mocking because of errors. These were easily fixed by using their built-in mocking code.

- `react-native-indicators` and `react-native-progress` gave warnings because the component animating, therefore updating outside of an `act` call. Mocking these components should have no effect on tests other than silencing those warnings.

- Relatedly, our `useFormattedTimeSince` hook kept causing warnings because it kept re-rendering with the current time. I moved the "update the current time" functionality into a new hook, `useCurrentTime`, which I then mocked to avoid the warnings.

- Our very-fake navigation mock needed to be made a little more realistic because of the `useNavigation` hooks used in the Exchange content component.

- Using this caused `expo-file-system`'s `documentDirectory` property to be accessed. I needed to define this.

[0]: https://github.com/digidem/comapeo-core/blob/a996995bf5c40ed530e29156c06fee9d497789a5/test-e2e/utils.js


tests/integration/helpers/core.ts
```diff
@@ -33,6 +38,7 @@ export function createManager() {
     clientMigrationsFolder,
     fastify,
   });
+  await manager.setDeviceInfo(deviceInfo);
```
EvanHahn, line 46: These managers aren't very useful (for sync) without device info, so I added this here.

tests/integration/helpers/react.tsx
```diff
@@ -151,7 +153,10 @@ export function createAppProvidersWrapper(
           }
           coordinateFormatStore={persistedCoordinateFormatStore}
           trackStore={persistedTrackStore}>
-          {children}
+          {/* We don't use navigation in our tests, but we set up the context. */}
```
EvanHahn: @achou11: Your comment about using more realistic navigation was prescient. This doesn't get us all the way there but it's a step.

src/frontend/screens/Exchange/index.tsx
```diff
@@ -23,7 +24,9 @@ const m = defineMessages({
   },
 });
 
-export const SyncScreen = ({navigation}: NativeRootNavigationProps<'Sync'>) => {
+export const SyncScreen = ({
+  navigation,
+}: PickDeep<NativeRootNavigationProps<'Sync'>, 'navigation.goBack'>) => {
```
EvanHahn: This fixes an error when passing the `navigation` prop in tests. I think this type is more exact, too.

src/frontend/screens/Exchange.test.tsx
```diff
@@ -1,45 +1,281 @@
...
+
+async function connectPeers(
+  managers: ReadonlyArray<MapeoManager>,
+): Promise<() => Promise<void>> {
+  await tellPeersAboutEachOther(managers);
+  await waitForPeersToBeConnected(managers);
+  return () => stopPeerDiscovery(managers);
+}
+
+async function tellPeersAboutEachOther(
```
EvanHahn: This function is adapted from [this test code in `@comapeo-core`](https://github.com/digidem/comapeo-core/blob/a996995bf5c40ed530e29156c06fee9d497789a5/test-e2e/utils.js#L56-L64).

src/frontend/screens/Exchange.test.tsx
```diff
@@ -1,45 +1,281 @@
...
+      const {name, port} = await manager.startLocalPeerDiscoveryServer();
+      for (const otherManager of managers) {
+        if (otherManager === manager) continue;
+        otherManager.connectLocalPeer({address: '127.0.0.1', name, port});
+      }
+    }),
+  );
+}
+
+async function waitForPeersToBeConnected(
```
EvanHahn: This function is adapted from [this test code in `@comapeo-core`][0].

[0]: https://github.com/digidem/comapeo-core/blob/a996995bf5c40ed530e29156c06fee9d497789a5/test-e2e/utils.js#L172-L209

src/frontend/screens/Exchange.test.tsx
```diff
@@ -1,45 +1,281 @@
...
+  managers: ReadonlyArray<MapeoManager>,
+): Promise<void> {
+  await Promise.all(
+    managers.map(manager =>
+      manager.stopLocalPeerDiscoveryServer({force: true}),
+    ),
+  );
+}
+
+async function inviteToProject(
```
EvanHahn: This function is adapted from [this test code in `@comapeo-core`][0].

[0]: https://github.com/digidem/comapeo-core/blob/a996995bf5c40ed530e29156c06fee9d497789a5/test-e2e/utils.js#L100-L146

src/frontend/screens/Exchange.test.tsx
```diff
@@ -1,45 +1,281 @@
...
+
+jest.mock('../hooks/useCurrentTime');
 
 describe('Exchange screen', () => {
   let appProviders: ReturnType<typeof createAppProvidersWrapper>;
-  const {manager, fastifyController} = createManager();
-  const {client, server, stop} = setUpIPC({manager});
+  let manager: MapeoManager;
+  let client: MapeoClientApi;
+  let onTeardown: Array<() => unknown> = [];
```
EvanHahn, line 24: I moved to a "here's a list of cleanup functions to run" approach.

EvanHahn:
The CI failure seems unrelated, though I can verify that next week.

src/frontend/__mocks__/react-native-indicators.ts
achou11, line 1: out of curiosity, is it usually preferable to mock these as null as opposed to a basic `<View />`?

src/frontend/screens/Exchange/index.tsx
achou11: this probably isn't preferable since it's another thing that has to be thought about when creating these screen components and arbitrarily limits how `navigation` is used within the implementations.

seeing this prompted me to try to get a bit further with the testing setup. this seems to work and get closer to what React Navigation recommends:

```tsx
import {createNativeStackNavigator} from '@react-navigation/native-stack';

// In the test...

// 🟠 Don't need to use the `Pick` here but maybe it's helpful 🤷 
const Stack = createNativeStackNavigator<Pick<AppStackParamsList, 'Sync'>>();

const renderSyncScreen = () => {
  const {unmount} = render(
    <Stack.Navigator>
      <Stack.Screen name="Sync" component={SyncScreen} />
    </Stack.Navigator>,
    {
      wrapper: appProviders.wrapper,
    },
  );
  
  // 🟠 Rest of the stuff omitted for brevity...
});

File: tests/integration/helpers/react.tsx (hunk 2 above)
Comment (achou11) [reply tests/integration/helpers/react.tsx]: wondering if this should be co-located with tests that rely on navigation as opposed to having it created in the wrapper by default. extending the exploration i mentioned earlier, a test that rendered a navigation screen would look like this:

```tsx
const {unmount} = render(
  <NavigationContainer>
    <Stack.Navigator>
      <Stack.Screen name="Sync" component={SyncScreen} />
    </Stack.Navigator>
  </NavigationContainer>,
  {
    wrapper: appProviders.wrapper,
  },
);
```

honestly have no idea if that's considered better or not. might be fine to leave it as you have it!

tests/integration/helpers/react.tsx
```diff
@@ -151,7 +153,10 @@ export function createAppProvidersWrapper(
...
+          <NavigationContainer ref={rootNavigationRef}>
```
achou11: wondering if this should be creating its own `rootNavigationRef` as opposed to the singleton that the actual app uses. seems like using this is prone to issues around test isolation

achou11 (commented):
Haven't looked through everything but overall seems encouraging! kind of a worried about having to extend / maintain this but that's more of a existential concern than something specific to this PR 😄

tests/integration/helpers/react.tsx
achou11: upon further inspection, wondering if we even need to have that defined at all! seems like we used to need it but it's no longer being used in the app anywhere...

src/frontend/__mocks__/react-native-indicators.ts
EvanHahn: No, I think making them `<View>`s is better. Changed in 7363f422bc84f56b1d8ed2cf10a90ed4a297019e.

src/frontend/screens/Exchange/index.tsx
EvanHahn: I like this much better. Changed in 7ddab6e79c981bd44030a634b602b4e043fa8aa9.

tests/integration/helpers/react.tsx
EvanHahn: I like this much better. Changed in 7ddab6e79c981bd44030a634b602b4e043fa8aa9.

tests/integration/helpers/react.tsx
EvanHahn: Removed in 1b7f17428b820a17267edb0f02b3aa26df835d2f.

EvanHahn:
I believe I've addressed all comments.

src/frontend/screens/Exchange.test.tsx
achou11, line 1: can you move the core helpers into `test/integration/helpers/core.ts`?

achou11 (approved):
great work on this! assuming last request for changes are made and CI passes, feel free to merge when ready!

achou11:
might be worth waiting for https://github.com/digidem/comapeo-mobile/pull/1216 to get merged fist and update this PR with those changes. guessing it might affect some parts of the test you've introduced

src/frontend/screens/Exchange.test.tsx
EvanHahn: Done in cb057ca405afd2ca17a1de45a18b18240e06c405.

PR #1220, opened 2025-06-10
fix: observation list scrolling

### LINKED ISSUE #1155: Scroll issues in observation list when many observations exist
**Describe the bug**

Scrolling through the observations list when you have several observations feels very broken. It's almost impossible to quickly scroll through it without the list items disappearing and being overtaken by a loading UI. It also seems that the scroll position eventually gets reset to the top of the list in this case, which is difficult to work with.

**To Reproduce**
1. Create lots of observations. I used the test data generator to generate 100 observations
2. Go to the observation list
3. Scroll through the list with some haste and conviction.

**Expected behavior**

The list items should always remained displayed. They should never be fully removed in place of a loading indicator.

The scroll position should be aligned with the actions taken by the user. It should never reset to unexpected positions.

**Screenshots**

https://github.com/user-attachments/assets/65f8e76a-8aaa-48f0-8845-1d4e372fef8e

cimigree:
closes #1155 

### Fixes scroll jump in Observations list

**What was happening:**
When scrolling quickly through the Observations list, the screen would reset to the top and items would disappear while loading. This made it feel broken and hard to scroll.

**What I changed:**
Wrapped the inside of `ObservationListItem` in a `<React.Suspense>` boundary so that only the individual item shows a fallback while loading — not the entire list. This stops the list from unmounting and resetting.

**Why it works:**
The Suspense boundary catches the loading state from the queries inside the list item (`useManyDocs`, `useIsMyDocument`) and prevents it from bubbling up and resetting the `FlatList`.

**No other screens or logic were changed.**

https://photos.app.goo.gl/VSiuSahBfMgcpkau5

### DIFF
--- a/src/frontend/screens/ObservationsList/ObservationListItem.tsx
+++ b/src/frontend/screens/ObservationsList/ObservationListItem.tsx
@@ -17,6 +17,7 @@ import {sharedStyles} from './SharedStyle.ts';
 import {HeaderText} from '../../sharedComponents/Text/HeaderText.tsx';
 import {BodyText} from '../../sharedComponents/Text/BodyText.tsx';
 import {useIsMyDocument} from '../../hooks/server/useIsMyDocument.ts';
+import {UIActivityIndicator} from 'react-native-indicators';
 
 interface ObservationListItemProps {
   style?: ViewStyleProp;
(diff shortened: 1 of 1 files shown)


achou11 (approved):
Definitely an improvement! Think there are probably a variety of other improvements that could be made to make the scrolling experience better, but can be addressed later. Fixing the scroll reset is helpful.

Maybe I'm too aggressive of a scroller but posting this just for posterity, in case additional improvements are considered later.

https://github.com/user-attachments/assets/a1aad5a2-e7a4-43ee-ae79-347e132b8629

PR #1221, opened 2025-06-10
feat: Add sentry logging to backend

### LINKED ISSUE #1033: Add Sentry logging
## Description

We need to be able to save logs for debugging.

## Tasks

- [ ] Options to supply sentry init during MapeoManager init
- [ ] Logging impl that pushes to sentry in addition to STDOUT
- [ ] Tests?

### LINKED ISSUE #1209: Backend Sentry - Minimal setup

We want to start collecting console logs and errors from the nodejs-mobile backend to help with debugging.

- Integrate the sentry [node.js SDK](https://docs.sentry.io/platforms/javascript/guides/node/)
- Set up a new project in sentry and use it's key to configure the nodejs side
- Run in app and check for collection
- Place logging behind a flag

We talked about pushing logs to the front end sentry instance but then we'd miss the instrumentation and the RPC bridge would have trouble pushing that much volume.

**Additional context**

We don't need to integrate Sentry into comapeo-core for the minimal setup - we just need to initialize it in mobile on the backend and link it to a separate project under our Sentry Awana org. That will ensure we get informed of uncaught errors and stack traces. Because the Sentry Node.js SDK does not support offline usage out-of-the-box, this minimal setup will only work when the user is online.

RangerMauve:
Closes #1209 

Also closes https://github.com/digidem/comapeo-core/issues/1033

Tested in a separate nodejs-mobile project

- Adds automatic tracing to HTTP
- Saves all logs to sentry
- Doesn't send PII

Is there an existing consent flag for this tracking on the frontend? Should we have a way to opt in/out?


RangerMauve:
I think I'm missing something regarding how to add dependencies to the backend. I saw that the root package.json seemed to depend on comapeo/core so I tried adding sentry at that level, but that seems to be missing in the backend.

However trying to install the dependency inside src/backend/package.json leads to this error:

```
> npm i --save @sentry/node
npm error code 1
npm error path /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild
npm error command failed
npm error command sh -c node install.js
npm error /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:132
npm error     throw new Error(`Expected ${JSON.stringify(versionFromPackageJSON)} but got ${JSON.stringify(stdout)}`);
npm error           ^
npm error
npm error Error: Expected "0.19.5" but got "0.25.2"
npm error     at validateBinaryVersion (/home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:132:11)
npm error     at /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:285:5
npm error
npm error Node.js v20.18.3
```

Is there a different way to install dependencies or is node 20 not the correct version to use?

achou11:
> I think I'm missing something regarding how to add dependencies to the backend. I saw that the root package.json seemed to depend on comapeo/core so I tried adding sentry at that level, but that seems to be missing in the backend.
> 
> However trying to install the dependency inside src/backend/package.json leads to this error:
> 
> ```
> > npm i --save @sentry/node
> npm error code 1
> npm error path /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild
> npm error command failed
> npm error command sh -c node install.js
> npm error /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:132
> npm error     throw new Error(`Expected ${JSON.stringify(versionFromPackageJSON)} but got ${JSON.stringify(stdout)}`);
> npm error           ^
> npm error
> npm error Error: Expected "0.19.5" but got "0.25.2"
> npm error     at validateBinaryVersion (/home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:132:11)
> npm error     at /home/mauve/src/comapeo-mobile/src/backend/node_modules/esbuild/install.js:285:5
> npm error
> npm error Node.js v20.18.3
> ```
> 
> Is there a different way to install dependencies or is node 20 not the correct version to use?

We have a [test](https://github.com/digidem/comapeo-mobile/blob/a0806aa6609de7798521f309657dcc37cafb19c6/src/frontend/packageJson.test.ts) that does some checks on the frontend and backend package.json files (due to some strict requirements we want to enforce).

The error you're seeing is because you're using a non-exact version of `@sentry/node` at the _top-level_ package.json. From what I can tell, you're only intending to install it in the backend, in which case you only need to install the module via the `src/backend/package.json` and leave the top-level one alone. We don't enforce a strict version for deps used in the backend from what I remember. The test only makes sure that the relevant core-related deps between top-level and backend match

achou11:
sorry, i was commenting on the error that CI was raising. Not super sure about the esbuild-related error when trying to install in the backend though 🤔 I was able to install it locally without any issue (I'm using node 20.19.2).

achou11:
> I saw that the root package.json seemed to depend on comapeo/core

Core is installed as a _dev_ dependency at the root level with the intent of guaranteeing the same version that's used as prod dep in the backend in order for the types to align when using other core-related deps in the frontend code e.g. schema, ipc, and core-react.

in general, my loose guide for installing new deps:

- If it's purely used in the frontend e.g. react native application or tooling related to that, update the root package.json

- If it's purely used in the backend, update the backend package.json. if the dep is core-related, there's a good chance that the root level package.json also needs to be updated in order for correct resolution of types

- If it's used in both, update both package.json files.

gmaclennan (changes requested):
Looks good so far, and I think there are a few other integrations to make with the frontend. Initially I think we can do all of these as startup flags for the node process by passing them to [`initializeNodejs()`](https://github.com/digidem/comapeo-mobile/blob/e199928271b45cd6b874607e217d768617aeaf79/src/frontend/initializeNodejs.ts#L8-L9) and setting them as startup flags, which are then [parsing in the backend](https://github.com/digidem/comapeo-mobile/blob/e199928271b45cd6b874607e217d768617aeaf79/src/backend/index.js#L19-L20):

- [`sentryEnvironment`](https://github.com/digidem/comapeo-mobile/blob/e199928271b45cd6b874607e217d768617aeaf79/src/frontend/App.tsx#L41-L47), which helps us separate noise from dev and QA from production events.
- The same user ID for [`initialScope`](https://docs.sentry.io/platforms/javascript/guides/node/configuration/options/#initialScope) as defined by [`getSentryUserID()`](https://github.com/digidem/comapeo-mobile/blob/e199928271b45cd6b874607e217d768617aeaf79/src/frontend/App.tsx#L55-L56) - that will help us match frontend and backend events. It's a random rotating ID - by default Sentry creates a persistent random ID (in the frontend at least)
- [`metricsIsEnabled`](https://github.com/digidem/comapeo-mobile/blob/e199928271b45cd6b874607e217d768617aeaf79/src/frontend/App.tsx#L123) - note that this will mean changes come into effect only when the app restarts, but I think that's an ok compromise for now. We should set the init option [`enabled`](https://docs.sentry.io/platforms/javascript/guides/node/configuration/options/#enabled) based on this.

RangerMauve:
I'm still getting a weird error where it says it can't find `@sentry/node` depsite it being in the `src/backend/package.json` and the `src/backend/node_modules`. I ran the install with `--ignore-scripts`. Do I need to add the module in the build somewhere or maybe use a different method to import it?

achou11:
> I'm still getting a weird error where it says it can't find `@sentry/node` depsite it being in the `src/backend/package.json` and the `src/backend/node_modules`. I ran the install with `--ignore-scripts`. Do I need to add the module in the build somewhere or maybe use a different method to import it?

TL;DR - you should either:

1. Move the Sentry code you've added to `src/backend/index.js` (not `loader.js`)

2. Update the [script](https://github.com/digidem/comapeo-mobile/blob/feat/initial-backend-logs/scripts/build-backend.mjs) that builds the backend to make sure that `@sentry/node` isn't purged from the included `node_modules` directory that's loaded into the android assets.

My guess is that (1) will be preferable but I don't have full context.

---

When building the backend for the app we [bundle the code using the `src/backend/index.js` file as the entry point](https://github.com/digidem/comapeo-mobile/blob/8bc4a1e80646d8886ac19b07413c08e07437658b/src/backend/package.json#L9). This is for performance + space reasons, as we want to avoid having to include a full `node_modules` directory loaded into the android app. The `loader.js` is not transformed and ran as-is by nodejs-mobile when starting the backend.

We still have a `node_modules` directory that's included as part of the assets loaded into the app, but we purge it and only keep things that are needed for static resolution e.g. maps or categories files ([link](https://github.com/digidem/comapeo-mobile/blob/8bc4a1e80646d8886ac19b07413c08e07437658b/scripts/build-backend.mjs#L96-L116))

RangerMauve:
I'll try (1) but I worry it won't work as well because sentry needs us to run the init before everything else so it can attach itself onto dependencies

achou11:
> I'll try (1) but I worry it won't work as well because sentry needs us to run the init before everything else so it can attach itself onto dependencies

hm yeah I see the note in their setup guide about that. part of me wonders if we can get rid of the separate `loader.js` and just move that initialization code into the `index.js`.

EDIT: we tell nodejs-mobile which script to run [here](https://github.com/digidem/comapeo-mobile/blob/8bc4a1e80646d8886ac19b07413c08e07437658b/src/frontend/initializeNodejs.ts#L24) - could just be a matter of pointing it to the bundled index.js file

gmaclennan:
> hm yeah I see the note in their setup guide about that. part of me wonders if we can get rid of the separate `loader.js` and just move that initialization code into the `index.js`.

The separate loader code was copied over from the original manyverse code in Mapeo Legacy. I believe it was needed so that `os.homedir()` and `process.cwd()` are defined before other modules load. I'm not sure if it's needed for our code as it is right now.

RangerMauve:
TODOs:

- only log errors on prod
- double check redactions for ids.

RangerMauve:
We don't have a way to distinguish log level in the core logger ATM so I'll make logging a flag.

https://github.com/digidem/comapeo-core/blob/main/src/logger.js

src/backend/loader.js
```diff
@@ -14,4 +16,54 @@ os.homedir = () => nodejsProjectDir
...
 process.env = process.env || {}
 
-import './index.js'
+const { values } = parseArgs({
+  options: {
+    sentryEnvironment: { type: 'string' },
+    sentryUserId: { type: 'string' },
+    metricsIsEnabled: { type: 'string' },
+  },
+  strict: false,
```
gmaclennan: This is ok as-is, but just to note that we could change remove `strict: false` and avoid the typeof checks below - we would need to add rootKey parsing here (must parse all arguments) and we would need to change the arg for metricsIsEnabled to an actual boolean arg (e.g. true if present, rather than a string arg with value `true|false`).

src/frontend/App.tsx
```diff
@@ -133,6 +133,9 @@ persistedMetricsDiagnosticsStore.instance.subscribe((current, previous) => {
   }
 });
 
+// Need to know if metrics are enabled before starting node
+initializeNodejs(metricsIsEnabled, sentryEnvironment, sentryUserId);
```
gmaclennan: nitpick, but in general our pattern has been to pass an object, rather than rely on positional parameters. It's slightly mute with Typescript, but in general passing `options` as an object removes the chance of bugs from getting the parameter positions mixed up (this has happened to me plenty)

src/backend/loader.js
```diff
@@ -14,4 +16,54 @@ os.homedir = () => nodejsProjectDir
...
+  throw new Error('backend did not receive metricsIsEnabled')
+if (typeof sentryUserId !== 'string')
+  throw new Error('backend did not receive sentryUserId')
+if (typeof sentryEnvironment !== 'string')
+  throw new Error('backend did not receive sentryEnvironment')
+
+const sentryDebug = sentryEnvironment === 'development'
+const initialScope = sentryUserId ? { user: { id: sentryUserId } } : undefined
+
+const logLevels = ['error']
```
gmaclennan, line 25: This removes the need for @ts-expect-error below - it ensures that these levels are typed as string literals, rather than just a generic string.

```suggestion
/** @type {Array<"error" | "log" | "warn">} */
const logLevels = ['error']
```

src/backend/loader.js
```diff
@@ -14,4 +16,54 @@ os.homedir = () => nodejsProjectDir
...
+  _experiments: {
+    enableLogs: true,
+    beforeSendLog: (log) => {
+      if (!log.attributes) log.attributes = {}
+      log.attributes.user = { id: sentryUserId }
+      return log
+    },
+  },
+  tracesSampleRate: 1.0,
+  // @ts-expect-error Unclear where to import type from their docs
```
gmaclennan:
```suggestion
```

src/backend/src/app.js
```diff
@@ -24,6 +25,13 @@ const MAPBOX_ACCESS_TOKEN =
   'pk.eyJ1IjoiZGlnaWRlbSIsImEiOiJjbHRyaGh3cm0wN3l4Mmpsam95NDI3c2xiIn0.daq2iZFZXQ08BD0VZWAGUw'
 const DEFAULT_ONLINE_MAP_STYLE_URL = `https://api.mapbox.com/styles/v1/mapbox/outdoors-v11?access_token=${MAPBOX_ACCESS_TOKEN}`
 
+// needed to have logs captured by sentry
+debug.log = function log(...args) {
+  // @ts-expect-error InspectOpts shouldn't be undefined but might be
+  return console.log(formatWithOptions(debug.inspectOpts, ...args))
```
gmaclennan: This change fixes the undefined issue. We still need the @ts-expect-error, because the way [debug parses env variables](https://github.com/debug-js/debug/blob/33330fa8616b9b33f29f7674747be77266878ba6/src/node.js#L124-L145) means that the opts could have incorrect values, but I think it's ok to skip validating that since we are not setting them from env at this time.

```suggestion
  return console.log(formatWithOptions(debug.inspectOpts || {}, ...args))
```

src/backend/loader.js
```diff
@@ -14,4 +16,54 @@ os.homedir = () => nodejsProjectDir
...
+Sentry.init({
+  dsn: 'https://5326989762cd5899283975f5459524c1@o4507148235702272.ingest.us.sentry.io/4509442300641281',
+
+  enabled: metricsIsEnabled === 'true',
+  sendDefaultPii: false,
+  debug: sentryDebug,
+  environment: sentryEnvironment,
+  initialScope,
+  _experiments: {
+    enableLogs: true,
```
gmaclennan: I think we should only enable logs if `sentryEnvironment !== 'production'`, e.g. only for dev and QA for now - until we have added the consent opt-in for "help make comapeo better".

gmaclennan (changes requested):
Some very minor suggestions, but the one thing to change before we merge is to only capture logs in QA and dev for now.

src/backend/loader.js
RangerMauve: I think I'd prefer to keep it as is so we don't need to keep the two sets of arguments in sync. Another option could be to extract the argument parsing into a separate module to import into both places. Is it safe to import libraries in the loader like that or would it mess with the bundler?

gmaclennan: Yes the bundler should extract it as a separate file that is imported by both

src/backend/src/args.js
```diff
@@ -0,0 +1,35 @@
...
+
+  if (typeof metricsIsEnabled !== 'boolean')
+    throw new Error('backend did not receive metricsIsEnabled')
+  if (typeof sentryUserId !== 'string')
+    throw new Error('backend did not receive sentryUserId')
+  if (typeof sentryEnvironment !== 'string')
+    throw new Error('backend did not receive sentryEnvironment')
+  if (typeof values.rootKey !== 'string')
+    throw new Error('backend did not receive root key from front end')
+  if (typeof values.version === 'string')
```
gmaclennan:
```suggestion
  if (typeof values.version !== 'string')
```

gmaclennan: This will fix the typecheck that is failing in CI.
```suggestion
  if (typeof version === 'string')
```

src/backend/src/args.js
```diff
@@ -0,0 +1,35 @@
...
+    rootKey,
+  } = values
+
+  if (typeof metricsIsEnabled !== 'boolean')
+    throw new Error('backend did not receive metricsIsEnabled')
+  if (typeof sentryUserId !== 'string')
+    throw new Error('backend did not receive sentryUserId')
+  if (typeof sentryEnvironment !== 'string')
+    throw new Error('backend did not receive sentryEnvironment')
+  if (typeof values.rootKey !== 'string')
```
gmaclennan:
```suggestion
  if (typeof rootKey !== 'string')
```
This will fix the typecheck failing CI - it ensures that the returned value has a type-guard to prevent it being `undefined`.

gmaclennan (changes requested):
Some small changes needed, go ahead and merge without further review after fixing.

RangerMauve:
All changes addressed, CI is passing. Merging

PR #1223, opened 2025-06-10
chore: do type checking on two exhaustiveness errors

EvanHahn:
*This change should have no impact on functionality. It is not urgent.*

This removes a `@ts-expect-error` comment that can be replaced with type checking.

Notably, this comment suppressed a misspelled variable name that would've been caught by type checking.

I noticed this while working on #1217.


achou11 (approved):
nice catch!

PR #1225, opened 2025-06-10
chore: update production dependencies

### LINKED ISSUE #1218: Update Production Dependencies
Update production dependencies to @latest. Make sure to pin the production dependencies to an exact version.

Can use npm run upgrade-deps:production

Ignore dependencies related to expo/react native.

cimigree:
close #1218 
##  Monthly Production Dependency Upgrade

This PR updates production dependencies as part of our regular maintenance process using `npm run upgrade-deps:production`, `depcheck`, and `expo-doctor`.

---

### Updated Packages

| Package                                | From    | To      |
| -------------------------------------- | ------- | ------- |
| `@expo-google-fonts/rubik`     | 0.2.3 |  0.4.1 |
| `@gorhom/bottom-sheet`                 | 5.1.2   | 5.1.6   |
| `@react-navigation/bottom-tabs`        | 7.3.10  | 7.3.14  |
| `@react-navigation/native`             | 7.1.6   | 7.1.10  |
| `@react-navigation/stack`              | 7.2.10  | 7.3.3   |
| `@tanstack/react-query`                | 5.74.4  | 5.80.6  |
| `expo-build-properties` | 0.13.2 | 0.13.3 |
| `react-hook-form`                      | 7.56.0  | 7.57.0  |
| `react-native-confirmation-code-field` | 7.4.0   | 8.0.1   |
| `react-native-safe-area-context`       | 5.4.0   | 5.4.1   |
| `react-native-screens`                 | 4.10.0  | 4.11.1  |
| `react-native-share`                   | 12.0.9  | 12.0.11 |
| `react-native-svg`                     | 15.11.2 | 15.12.0 |
| `valibot`                              | 1.0.0   | 1.1.0   |
| `zustand`                              | 5.0.3   | 5.0.5   |

---

### Skipped Packages

| Package                    | Reason for Skipping                                                                                                    |
| -------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| `@rnmapbox/maps`           | Known crash issue on Android when New Architecture is disabled that Andrew reported (https://github.com/rnmapbox/maps/issues/3796) |
| `react-native-mmkv`        | v3 is a breaking change; staying on v2.12.2 for now                                                                    |
| `expo-*` & `react-native`  | Managed by Expo SDK; intentionally pinned to SDK 52 compatibility                                         |
| `@sentry/react-native`     | Skipped upgrading beyond 6.10.0 to avoid introducing potential breaking changes until fully validated  by expo                |

---

###  Removed Packages

These were identified as unused by `depcheck` and confirmed via manual review:

| Package                              | Reason                            |
| ------------------------------------ | --------------------------------- |
| `expo-status-bar`                    | Not used anywhere in the codebase |
| `react-native-android-open-settings` | Not imported or referenced        |
| `react-native-linear-gradient`       | Not used in components or styles  |
| `assert` | Not used anywhere |
| `debug`                                | Took it out of direct usage but it is a subdependency |

---

###  Compatibility Checks (`expo-doctor`)

* Ran `npx expo-doctor` and reviewed 2 warnings:

#### Validate Packages Metadata

| Package                                                | Warning                      | Action Taken                                      |
| ------------------------------------------------------ | ---------------------------- | ------------------------------------------------- |
| `react-native-restart`                                 | Untested on New Architecture | App does not use New Architecture; safe to ignore |
| `react-native-indicators`                              | Marked as unmaintained       | Retained; still works reliably                    |
| Multiple internal deps (`@comapeo/*`, `valibot`, etc.) | No metadata in RN Directory  | Safe to ignore — internal or utility packages     |

#### Version Mismatches with SDK, all downgrades

| Package                          | Installed | Expected | Action                                       |
| -------------------------------- | --------- | -------- | -------------------------------------------- |
| `@react-native-picker/picker`    | 2.11.0    | 2.9.0    | Kept newer version — no regressions observed |
| `react-native-gesture-handler`   | 2.25.0    | \~2.20.2 | No issues observed — skipped                 |
| `react-native-safe-area-context` | 5.4.1     | 4.12.0   | Works fine — skipped                         |
| `react-native-screens`           | 4.11.1    | \~4.4.0  | Retained newer version                       |
| `react-native-svg`               | 15.12.0   | 15.8.0   | Newer version working fine                   |

### Some fixes needed based on package changes

* Updated `InputPasscode` to fix type errors.
* Adjusted usage of `isLastFilledCell` in `PasscodeInput` to match the updated function signature.
* Had to add a patch to `react-native-confirmation-code-field` so that the masking of the passcode would work.
* Updated `@expo-google-fonts/rubik` to `0.4.1` and fixed font paths in `app.json` to reflect new file structure requirements.

### Manual QA
- Creating and editing Tracks and Observations (used 100 test data plus manually done)
- Recording audio and updated permissions
- Using the passcode
- Updating config file
- Exchanging data between devices via sharing and sync
- Switching languages
- Connecting and syncing with a remote server
- General navigation and edge case testing across the app


package.json
```diff
@@ -33,33 +33,31 @@
...
     "@sentry/react-native": "6.10.0",
-    "@tanstack/react-query": "5.74.4",
+    "@tanstack/react-query": "5.80.6",
     "@turf/helpers": "7.2.0",
     "@turf/random": "7.2.0",
-    "assert": "2.1.0",
-    "buffer": "6.0.3",
     "cheap-ruler": "4.0.0",
-    "debug": "4.4.0",
+    "debug": "4.4.1",
```
achou11: I think this should be removed. it's only used in one place (`AddPhoto.tsx`) and I don't think it's really needed there.

src/frontend/sharedComponents/PasscodeInput.tsx
```diff
@@ -54,8 +54,8 @@ export const PasscodeInput = React.forwardRef<TextInput, PasscodeInputProps>(
         textChild = (
           <MaskSymbol
             maskSymbol={maskValues ? '*' : symbol}
-            isLastFilledCell={isLastFilledCell({index, value: inputValue})}>
-            {symbol}
+            isLastFilledCell={isLastFilledCell(inputValue, index)}>
+            {maskValues ? '*' : symbol}
```
achou11: went down a small rabbithole because this looked odd compared to the [example](https://github.com/retyui/react-native-confirmation-code-field/blob/05d73d0345dd1ea5f10624323466f9a7b7d6fedc/examples/DemoCodeField/src/MaskExample/index.js) from the library. tried updating to make it like the example:

```tsx
textChild = maskValues ? (
  <MaskSymbol
    maskSymbol={'*'}
    isLastFilledCell={isLastFilledCell(inputValue, index)}>
    {symbol}
  </MaskSymbol>
) : (
  symbol
);
```

but noticed an issue where the mask was being unapplied undesirably (especially when typing quickly), which is probably the issue you encountered and hence why this code is here.

with some digging, think i narrowed it down to an issue with the library. if you're curious, here's a patch that I think fixes things as desired. It adds a small patch to the library such that the change I presented above works as desired.

<details>
<summary>Patch diff</summary>

```diff
diff --git a/patches/react-native-confirmation-code-field+8.0.1+001+fix-mask-symbol-logic-issue.patch b/patches/react-native-confirmation-code-field+8.0.1+001+fix-mask-symbol-logic-issue.patch
new file mode 100644
index 00000000..7a783acd
--- /dev/null
+++ b/patches/react-native-confirmation-code-field+8.0.1+001+fix-mask-symbol-logic-issue.patch
@@ -0,0 +1,14 @@
+diff --git a/node_modules/react-native-confirmation-code-field/esm/MaskSymbol.js b/node_modules/react-native-confirmation-code-field/esm/MaskSymbol.js
+index 68347ca..4fe674a 100644
+--- a/node_modules/react-native-confirmation-code-field/esm/MaskSymbol.js
++++ b/node_modules/react-native-confirmation-code-field/esm/MaskSymbol.js
+@@ -5,7 +5,8 @@ export function MaskSymbol({ isLastFilledCell, children: symbol, maskSymbol, del
+     'use memo';
+     const [visibleFlag, setFlag] = useState(true);
+     const toggleVisibility = useCallback(function toggleMaskVisibility() {
+-        setFlag((prev) => !prev);
++        // setFlag((prev) => !prev);
++        setFlag(false)
+     }, []);
+     useTimeout(toggleVisibility, delay);
+     useEffect(() => {
diff --git a/src/frontend/sharedComponents/PasscodeInput.tsx b/src/frontend/sharedComponents/PasscodeInput.tsx
index c6e5f6e8..4bd33d1d 100644
--- a/src/frontend/sharedComponents/PasscodeInput.tsx
+++ b/src/frontend/sharedComponents/PasscodeInput.tsx
@@ -51,12 +51,14 @@ export const PasscodeInput = React.forwardRef<TextInput, PasscodeInputProps>(
       let textChild;
 
       if (symbol) {
-        textChild = (
+        textChild = maskValues ? (
           <MaskSymbol
-            maskSymbol={maskValues ? '*' : symbol}
+            maskSymbol={'*'}
             isLastFilledCell={isLastFilledCell(inputValue, index)}>
-            {maskValues ? '*' : symbol}
+            {symbol}
           </MaskSymbol>
+        ) : (
+          symbol
         );
       } else if (isFocused) {
         textChild = <Cursor />;
```
</details>

achou11: not adamant about fixing this as I described, but it does read a little easier to me (albeit with the need for a patch)

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -54,7 +54,7 @@ export const InputPasscode = ({
   const inputRef = useBlurOnFulfill({
     value: inputValue,
     cellCount: CELL_COUNT,
-  });
+  }) as React.RefObject<TextInput>;
```
achou11: would it make sense to update the `PasscodeInput` component type declaration to avoid needing the type coercion here? Maybe not worth doing...

```diff
-export const PasscodeInput = React.forwardRef<TextInput, PasscodeInputProps>(
+export const PasscodeInput = React.forwardRef<TextInput | null, PasscodeInputProps>(
```

achou11 (dismissed):
Super helpful summary! Really appreciate having it as a reference. Good to see the removal of a couple unused deps.

Provided a couple of non-blocking suggestions, but otherwise liking how this looks 👍

package.json
```diff
@@ -33,33 +33,31 @@
...
+    "@react-navigation/native": "7.1.10",
+    "@react-navigation/stack": "7.3.3",
     "@rnmapbox/maps": "10.1.36",
     "@sentry/react-native": "6.10.0",
-    "@tanstack/react-query": "5.74.4",
+    "@tanstack/react-query": "5.80.6",
     "@turf/helpers": "7.2.0",
     "@turf/random": "7.2.0",
-    "assert": "2.1.0",
-    "buffer": "6.0.3",
```
achou11, line 55: not ideal but I believe this needs to be restored. it's used [here](https://github.com/digidem/comapeo-mobile/blob/14592b9d22c4f6762e4197107887e3f9104a6563/src/frontend/utils/base64.ts), which is used [here](https://github.com/digidem/comapeo-mobile/blob/14592b9d22c4f6762e4197107887e3f9104a6563/src/frontend/screens/Observation/Buttons.tsx)

achou11 (changes requested):
sorry, I think the `buffer` module needs to be restored unfortunately. other than that, my previous review still stands 😄

package.json
achou11: the reason it wasn't caught by depcheck is because `buffer` can also refer to the [built-in node module](https://nodejs.org/api/buffer.html). however, react native's runtime doesn't support this module by default (to my knowledge) and thus needs this package to be installed in order for this to work.

package.json
cimigree: Sounds good. I just got rid of the usage. Seemed like someone added it during development and that it isn't still needed, but please let me know if I am wrong.

package.json
cimigree: I am very glad you caught that! I am not sure how I missed it!

cimigree:
@achou11, So it looks like assert was also required as a dependency, here in particular: babel.config.js
Also, adding that back fixed the issue with the lockfile test failing.
Also, FYI, debug is a subdependency of many, many, many other dependencies, so it is not truly removed, just removed as one of our prod dependencies.

achou11:
> @achou11, So it looks like assert was also required as a dependency, here in particular: babel.config.js Also, adding that back fixed the issue with the lockfile test failing.

I'm surprised about the `assert` being needed. the one used in `babel.config.js` can and ideally should reference the [built-in Node assert module](https://nodejs.org/docs/latest-v20.x/api/assert.html). maybe update that file to import from `node:assert` explicitly to avoid that confusion, and see if removing the `assert` module is okay?

> Also, FYI, debug is a subdependency of many, many, many other dependencies, so it is not truly removed, just removed as one of our prod dependencies.

Yes I'm aware - this is fine + intended

patches/README.md
```diff
@@ -86,3 +86,11 @@ As this feature is not used in CoMapeo, this can be safely hardcoded to `false`.
...
 
 Refer to https://github.com/n4kz/react-native-indicators/issues/43 for details.
+
+## `react-native-confirmation-code-field`
+
+### [Fix mask symbol logic issue](./react-native-confirmation-code-field+8.0.1+001+fix-mask-symbol-logic-issue.patch)
+
+Fixes a bug in the `MaskSymbol` component where the mask (`*`) briefly un-hides when typing quickly. This patch sets the `visibleFlag` to `false` immediately, preventing the undesired flicker.
+
+See: [Reviewer context](https://github.com/recruiters/github/blob/master/README.md).
```
achou11: wrong URL?

achou11 (approved):
My note on the `assert` question is not blocking - just me being particular about shaving deps if possible 😄 doesn't have any runtime consequences from what I can tell.

patches/README.md
cimigree: Ugh. I meant to put in this PR as the link. Fixing.

cimigree:
@achou11 Ok I removed assert and changes the Readme and import in the babel file. Any idea why the lockfile test failed that time?

achou11:
> @achou11 Ok I removed assert and changes the Readme and import in the babel file. Any idea why the lockfile test failed that time?

have no idea. the `Not Found` error that's there is a bit vague 🤷

PR #1230, opened 2025-06-14
test: add more tests for exchange screen

EvanHahn:
This tests syncing with a remote archive and the "no wifi" screen.


EvanHahn:
Rebased against the latest `develop`.

src/frontend/screens/Exchange.test.tsx
ErikSin, line 1: Can we move this file into the `screens/exchange` folder?

src/frontend/screens/Exchange.test.tsx
```diff
@@ -51,7 +49,15 @@ describe('Exchange screen', () => {
 
   const Stack = createNativeStackNavigator<Pick<AppStackParamsList, 'Sync'>>();
 
-  const renderSyncScreen = () => {
+  const renderSyncScreen = ({
```
ErikSin, line 52: Can we update this to mount the entire navigator and navigate to the sync screen, as we discussed in our meeting?

ErikSin (approved):
2 small comment. 1 kind of blocking, but if the tests pass I don't think it is worth a re-review, so can merge after addressed.

I will admit that the cloud server stuff is a little out of my depth...but the tests are working so that feels good enough.

src/frontend/screens/Exchange.test.tsx
EvanHahn: We discussed this and decided not to do this as part of this PR, but might do it later.

src/frontend/screens/Exchange.test.tsx
EvanHahn: Done in 1839f55efbbaf2c4633cf8c00d8206e8de6c2110.

PR #1231, opened 2025-06-16
feat: update @comapeo/core to 4.0.0

achou11:
Introduces `@comapeo/core@4.0.0`, which introduces the following changes:

- Fixes data export file extension names (https://github.com/digidem/comapeo-core/pull/1039)

- Adds the optional `presetRef` field to the `Track` schema type (https://github.com/digidem/comapeo-schema/pull/302)

- Adds the required `external` field to the `Observation` `Attachment` schema type (https://github.com/digidem/comapeo-schema/pull/299)

- Adds the optional `FNumber` field to the `photoExif` field in the `Observation` `Attachment` schema type (https://github.com/digidem/comapeo-schema/pull/293)

- Adds the optional `createdAt` and `position` fields to the `Observation` `Attachment` schema type (https://github.com/digidem/comapeo-schema/pull/295)

---

App-specific changes as a result of the above:

- We no longer store GPS-related EXIF fields when saving an observation photo attachment (see 3728aa2f1eacedf91c2568761d4a886196bb7ef5)

- We store the `FNumber` EXIF field when saving an observation photo attachment (see 3728aa2f1eacedf91c2568761d4a886196bb7ef5)

- We set `external: false` to all attachments created in the app (see 4f9a77261616fc6d2294a5d11cd322acbe9269e5)

- We no longer use the `createdAt` and `location` metadata fields when creating a blob. Instead we set the `createdAt` and `position` fields to attachments (see 4f9a77261616fc6d2294a5d11cd322acbe9269e5 and  4c91aef2454153983a9eab48b5b3e8feca3c31fc)

    - ~Was debating whether I should introduce this change as part of #1203 . Still open to doing that if preferred!~ EDIT: Thinking that it should stay here because we're removing the usage of the blob api to store similar info.
    
Note that the `presetRef` field for tracks is still left unused in this PR. This is intentional and will be addressed in a separate PR when the related features are being implemented.


src/frontend/hooks/server/media.ts
```diff
@@ -25,25 +25,36 @@ export function useCreatePhotoAttachment({projectId}: {projectId: string}) {
...
         },
       });
 
-      const baseAttachment = {
-        driveDiscoveryId: blob.driveId,
-        hash: blob.hash,
-        name: blob.name,
-        type: blob.type,
-      };
+      const position = photo.mediaMetadata.location
```
cimigree, line 31: I wonder if anyone will ever want position on Audio.

src/frontend/hooks/server/media.ts
```diff
@@ -68,6 +78,8 @@ export function useCreateAudioAttachment({projectId}: {projectId: string}) {
         hash: blob.hash,
         name: blob.name,
         type: blob.type,
+        external: false,
+        createdAt: new Date(file.createdAt).toISOString(),
```
cimigree, line 82: It will be great to have created at here as that is part of the new audio design.

src/frontend/hooks/server/media.ts
```diff
@@ -25,25 +25,36 @@ export function useCreatePhotoAttachment({projectId}: {projectId: string}) {
...
-        driveDiscoveryId: blob.driveId,
-        hash: blob.hash,
-        name: blob.name,
-        type: blob.type,
-      };
+      const position = photo.mediaMetadata.location
+        ? expoLocationToAttachmentPosition(photo.mediaMetadata.location)
+        : undefined;
+
+      const createdAt = new Date(photo.mediaMetadata.timestamp).toISOString();
```
cimigree, line 35: Is there any chance that this will every be zero or undefined? Should there be a validation or fallback?

cimigree (approved):
This seems like it makes things simpler. One nonblocking comment. I imagine further handling of these fields will happen in the PhotoPreviewModal and subsequent PRs. I did basic QA for tracks, audio, and photo but imagine you will have more to do with the Photo metadata PR.

src/frontend/hooks/server/media.ts
achou11: i think we may eventually want to! just didn't include it as part of this PR since it requires updating a couple of things elsewhere. think it can be done as a follow-up when needed

src/frontend/hooks/server/media.ts
achou11: think chances of it being zero are near...zero 😄 think it's okay to leave as-is.

shouldn't be undefined based on the typescript typing for it.

PR #1238, opened 2025-06-18
test: update Jest dependencies

EvanHahn:
This updates various Jest dependencies to their latest versions.

Notably, this does not update `jest-expo` as it's not compatible with our Expo version.

I did this because I hoped it would improve test performance. I don't think it did, but I still think it's a useful change.


EvanHahn:
Rebased against the latest `develop`.

ErikSin (approved):

PR #1239, opened 2025-06-18
fix: remove leave project

### LINKED ISSUE #1234: Remove "leave project" button
https://www.notion.so/digidem/Leave-project-Update-flow-20a1b08162d58017860cd418743200d3

To do
 - leave Project button for now
 - remove the screen

cimigree:
closes #1234 
For now, since we don't want folks to uninstall comapeo and lose everything just to leave one project, we are removing the leave project button. 
In the future the backend and codesign will add the ability to leave a project, but this is what was decided upon for now.
Since leaving a project will be very different, I also removed the HowToLeaveProject screen as it will be irrelevant in the future.
While I was in there, I updated the text to BodyText and the TouchableOpacity of the Device Card as TouchableOpacity from react native gesture handler is deprecated.
While I am removing things, I also took out the test that was related to not joining a project if you were already on a project.


achou11 (approved):
Might be worth re-running the failed e2e test workflow just to make sure, but seems like flakiness.

PR #1242, opened 2025-06-18
fix: passcode and audio permission

### LINKED ISSUE #1226: Audio permission and passcode confusing/ incompatible
**Describe the bug**
When someone has a passcode set and they go to enable the audio permissions in an Observation, they get stuck in a cycle of entering the passcode over and over and can't get out of it. 

**To Reproduce**
1. Enable a passcode
2. Click on an Observation
3. Hit the microphone to record an audio piece for your observation
4. You will see the permission dialogue. Hit allow. Hit While Using the app.
5. You will see Enter your passcode. Enter it. You will see the Audio permission again.
6. If you click Allow, you will have to enter the passcode again. And then you will see the Audio permission again.
7. The only way to get out of it is to click "Not Now" which seems counter intuitive.

**Expected behavior**
Hit the microphone to record audio, have to allow permission, maybe(?) once enter passcode, or maybe not, and then get back to the app with the permission enabled.

**Additional context**
In Tracks, you get pulled out to settings, then you enter your passcode, then you are back in the app, which seems much more desirable.
Another fix was created for the share dialog... That fix is in AuthContext.

cimigree:
closes #1226

This PR fixes an edge case in the audio permission flow: When using permissions with audio, the passcode would pop up and when authenticated would navigate back, which always brought the user back to the permissions modal and then they would be stuck in a loop.

To fix this, I created a small hook (inspired by the one we use for sharing) that tracks when a modal is open and prevents triggering the auth screen while doing in-app permission checks. That means if the user taps "Allow" in-app — which doesn’t technically leave the app — they’re no longer forced to re-enter their passcode. And if they do go to settings to enable mic access, they’ll now only be prompted for their passcode after returning, instead of before leaving, which used to happen.

In addition, I added a useFocusEffect in the AudioAskPermissionBottomSheet that only runs when passcode is enabled. This is because in the passcode flow, the app technically becomes "active" before the user completes authentication. Without this, we’d miss the permission change that happened in settings and never navigate forward.

I'm also using a hasNavigatedRef to make sure we don’t double-navigate, since the screen can become active in multiple ways (settings return, passcode unlock, etc). This makes the logic more predictable and hopefully avoids flickering or race conditions.

There are four situations that should be tested...
1. With no passcode enabled:
 - Hit microphone
 - Hit Allow
 -  Hit While Using the App
You should be taken right to the recording screen
2. With no passcode enabled:
- Hit microphone
 - Hit Allow
 - Don't Allow
 - Hit Go to Settings
 - Enable microphone permissions
 - Hit the back button to return to the app
 You should be taken right to the recording screen
 (If you don't enable permissions the modal will still be open)
 
 3. With passcode enabled
- Hit microphone
 - Hit Allow
 -  Hit While Using the App
You should be taken right to the recording screen and never have to enter your passcode
4.  With passcode enabled:
 - Hit microphone
 - Hit Allow
 - Don't Allow
 - Hit Go to Settings
 - Enable microphone permissions
 - Hit the back button to return to the app
 - You should then have to enter your passcode
 Now you should be taken right to the recording screen

### DIFF
--- a/src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
+++ b/src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
@@ -66,14 +66,20 @@ export const AudioAskPermissionBottomSheet = ({
     });
   }, [replace]);
 
+  // We need *both* the AppState listener and the useFocusEffect because passcode being active shifts the timing.
+  //
+  // In the no-passcode flow, the app becomes "active" *after* returning from settings —
+  // so the AppState change is enough to detect permission changes.
+  //
+  // But in the passcode flow, the app becomes "active" *before* the user finishes authenticating.
+  // So we use useFocusEffect to re-check permissions after the user actually lands back on this screen.
+  //
+  // Without both, we’d miss the permission change in one of the two flows.
+
   // Covers re-entering from Auth screen (passcode is active flow)
   useFocusEffect(
     React.useCallback(() => {
-      if (
-        passcode &&
-        !hasNavigatedRef.current &&
-        permission.status !== 'granted'
-      ) {
+      if (passcode && permission.status !== 'granted') {
         checkAndNavigateIfGranted();
       }
     }, [checkAndNavigateIfGranted, permission.status, passcode]),
--- a/src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
+++ b/src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
@@ -1,12 +1,6 @@
 import * as React from 'react';
 import {BottomSheetWrapper} from '../../sharedComponents/BottomSheetWrapper';
-import {
-  AppState,
-  AppStateStatus,
-  Linking,
-  StyleSheet,
-  View,
-} from 'react-native';
+import {Linking, StyleSheet, View} from 'react-native';
 import {defineMessages, useIntl} from 'react-intl';
 import AudioPermission from '../../images/observationEdit/AudioPermission.svg';
 import {HeaderText} from '../../sharedComponents/Text/HeaderText';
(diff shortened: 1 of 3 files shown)


src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -55,44 +51,54 @@ export const AudioAskPermissionBottomSheet = ({
...
-      );
-
-      return () => {
-        isCancelled = true;
-        subscription.remove();
-      };
-    }, [replace]),
+      if (
+        passcode &&
+        !hasNavigatedRef.current &&
```
ErikSin: i dont think we need to check `hasNavigatedRef.current` here as you are already doing it in `checkAndNavigateIfGranted`

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
```diff
@@ -55,44 +51,54 @@ export const AudioAskPermissionBottomSheet = ({
...
+    Audio.getPermissionsAsync().then(newPermission => {
+      setPermission(newPermission);
+      if (newPermission.status === 'granted' && !hasNavigatedRef.current) {
+        hasNavigatedRef.current = true;
+        replace('AudioRecording');
+      }
+    });
+  }, [replace]);
+
+  // Covers re-entering from Auth screen (passcode is active flow)
```
ErikSin, line 79: can you add a comment explaining why we need both the app state subscriber and the useFocusEffect. It took some experimenting for me to understand why they were different things.

ErikSin (approved):
2 small non blocking changes!

src/frontend/screens/Audio/AudioAskPermissionBottomSheet.tsx
cimigree: Ok

PR #1247, opened 2025-06-23
fix: new project screens

### LINKED ISSUE #1232: Update UX - Invite Collaborator and Create New Project
**Problem:** 

Path #1

- User is mapping on their own and want to create a project
- They go to the main menu, and on the project card taps +Invite
- The project is created
- User goes to All Projects….and sees the project they created
- The “my solo project” card is not there

Path #2 

- User is mapping on their own and want to create a project
- They go to the main menu, and to All Projects
- Tap on +Create new project
- The project is created
- User goes back to all project
- The project created AND “my solo project” appear

Issue is the Create new project screen is the same, and it causes confusion

To do:

See Prototype : https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6323-3746&t=rXSaYc4L17jgrdav-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6323%3A3746

See details : https://www.notion.so/digidem/The-same-new-project-screen-2-different-results-20e1b08162d580d283a8c996a17bd47a?source=copy_link#20e1b08162d5818b9d82c0a1a8c2ccd4

## Update of "My Solo Path" 

- [ ] Tapping on "+Invite" , display the "Invite Collaborators" screen as designed 
Image

- [ ] Tapping on "Name My Project" , display the "Name My Project" screen as designed 
- [ ] Name screen "Name My Project"
Image

- [ ] Update Success screen as designed
- [ ] Display the "Update Category" button
- [ ] When tapping  "Update Category"  - go to the Categories screen

Image

- [ ] Update "Device Accepted" screen as designed
- [ ]  tap on “View Team” and go to the Your Team screen
- [ ]  tap on “Invite another device ” and go back to the invite flow

Image

## Update of "Start New Project" 
- [ ] Tapping on "All Projects"  and then "Start New Project", display the start new project screen as designed 

Image

- [ ] Tapping on "Start" , display the "Start New Project" screen as designed 
- [ ] Name screen "Start New Project"

Image

- [ ] Update Success screen as designed
- [ ] When tapping  “Start Mapping” and go to the map screen
- [ ]  tap on “View Project” and go to the Project Settings

Image

cimigree:
closes #1232 

This PR differentiates the flow a bit between starting a new project and naming the default solo project.
A lot of the files are emoji/ svgs, test files, or minor changes, so try not to be overwhelmed!
It adds different emojis and explanatory text to each flow.  It adds a start new project explanatory screen similar to the Invite Collaborators screen
See the Figma prototype, Notion document, (links can be found in the ticket) and [Figma Screens ](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6323-3606&t=8hilPgNwHq9fWdlE-0)
I had to add or change various SVG to use. One was a duplicate -- the Success -- so I deleted the extra one and updated its usage. Others came from openmoji as requested by design, and the intro page looks a little different as a result. and an attribution was added on the About CoMapeo page. Here [they are used in Figma](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6601-3526&t=j885jZ7PeAel6Dny-0): 
Also, there is a more simple invite accepted screen in the designs, so that was also updated and the route params simplified.


src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
```diff
@@ -1,107 +1,94 @@
...
   navigation,
   route,
-}) => {
+}: NativeRootNavigationProps<'CreateProject'>) => {
   const {formatMessage: t} = useIntl();
-  const [advancedSettingOpen, setAdvancedSettingOpen] = React.useState(false);
-  const [configFileResult, setConfigFileResult] =
-    React.useState<ConfigFileImportResult | null>(null);
 
   const action = route.params.action;
```
ErikSin: I think a more declarative approach would actually be register this component to 2 different names

Eg:

```tsx
export const CreateOrNameSoloProject = ({
  navigation,
  route,
}) => {
}: NativeRootNavigationProps<'CreateProject' | 'NameSoloProject'>) => {
     const action = route.name
     
     //...rest of the code stays that same
}

```

And then the navigation actions would be declarative without having to give props. So instead of `navigation.navigate('CreateProject', {action: 'CreateNewProject' | 'UpdateSolo'})` it would be `navigate('CreateProject')` or `navigate('NameSoloProject')` without having to feed it props

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
```diff
@@ -43,57 +61,101 @@ export const ProjectCreated = ({
...
+    newlyCreated: {
+      message: m.nowAdded,
+      secondaryLabel: m.goToMap,
+      secondaryAction: () => handleGoToMap(),
+      primaryLabel: m.viewProject,
+      primaryAction: () => handleViewProject(),
+    },
+  };
+
+  const setup = screenOptionsObject[route.params.type];
```
ErikSin: can we call this something a little more descriptive? Maybe `screenActionsAndLabels`

src/frontend/screens/Settings/ProjectSettings/YourTeam/InviteCollaborators.tsx
```diff
@@ -132,22 +127,24 @@ const styles = StyleSheet.create({
...
   },
   bulletItem: {
     flexDirection: 'row',
-    columnGap: 8,
+    alignItems: 'center',
+    gap: 12,
   },
   bulletIcon: {
-    marginTop: 6,
+    marginTop: 0,
```
ErikSin: is `0` necessary? Or can we just take it out completely?

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
```diff
@@ -43,57 +61,101 @@ export const ProjectCreated = ({
...
+    ]);
+  }
 
-      return CommonActions.reset({
-        routes,
-        index: routes.length - 1,
-      });
-    });
+  function handleViewProject() {
+    resetTo([{name: 'Home'}, {name: 'Menu'}, {name: 'ProjectSettings'}]);
```
ErikSin: Resetting the nav stack like this is becoming increasingly fragile (and I know that i started this trend). I am concerned that if we do this with most of the actions it will be hard to maintain. It is prone to error as there is no typesafety, and it requires developer knowledge of the previous screens, and if the screen order changes we will not catch it easily.

An alternative would be to use replace instead of navigate.

For the create project flow:

AllProject>navigate("StartNewProject)>replace("CreateProject")>replace(ProjectCreated). Then the "Start Mapping" button you can [popToTop](https://reactnavigation.org/docs/stack-actions#poptotop), which will take the user back to the map and reset the stack. And for the "View Project" button you can `replace("ProjectSettings)`. Now from project settings, if they press back, they will be brought back to AllProjects.

For the Solo Project Flow:

Menu>navigate("InviteCollaborators")>replace(NameMyProject)>Replace(ProjectCreated). For the "Update Categories Set" button, you can replace("Config"). And for the Invite Device button, you can replace(InviteDevice). And then when they press back from the Config or InviteDevice screen they will go back to the Menu (which is maybe not 100% desired, but i think it is better than the alternative).

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
ErikSin, line 1: Same comment about making this more declarative by registering this component to 2 nav screen. So instead of passing a prop so the screen know whether it is "updatedSolo" or "newlyCreated", instead just naming the screen that. IT will just make the navigation action a little easier to read

ErikSin (changes requested):
Overall it looks good, but there are 2 things that I think we should reconsider with navigation that will make it a little less fragile

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/index.tsx
cimigree: ok no more props.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
cimigree: Unfortunately, using the replace does not work. That replaces the current screen only. When you go back using the back button (which is the way to exit) you end up back on the naming screen and then the interstitial screen.
I tried a number of things and what I landed on is the only way it works. Of course, that just makes everything go back to the Map, but I assume that is ok and preferable to going through the stacks.

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
cimigree: ok done

src/frontend/screens/Settings/CreateOrJoinProject/CreateProject/ProjectCreated.tsx
ErikSin: I played around with it a bit, and you have to use replace for several actions. Here is what it would look like:
https://github.com/digidem/comapeo-mobile/pull/1259

cimigree: Ok thanks.

src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+      'hardwareBackPress',
+      () => {
+        if (mutationIsPending) {
+          return true;
+        }
+      },
+    );
+
+    return () => subscription.remove();
+  }, [mutationIsPending]);
```
ErikSin: i realize that we also need to disable the back header

ErikSin (approved):
One small comment (that is kind of blocking), but does not need a re-review so I am going to approve

Issue #1245, opened 2025-06-19
Data Export - File Naming - Observations + Media


ximenabb:
Related to : https://github.com/digidem/comapeo-core/issues/1054

Issue #1122, opened 2025-04-28
Exporting Data + Output


ximenabb:
**Description**
Many users need to use data collected with CoMapeo outside the app, for further analysis or for generating visualizations or reports from the data. Exporting data in CoMapeo is also important if a user wants to stop using CoMapeo and transfer their data to another service.
User, Coordinator or Participant, can export 3 options of files
- Observations only
- Observations and attachments
- Tracks only

Exporting is Project-specific

**Feature spec**
https://www.notion.so/digidem/Data-Export-mobile-fe0a747ef76e42c5a3ca32736bd4e82b?pvs=4

**Tech proposal by Gregor**
https://www.notion.so/digidem/Data-Export-mobile-fe0a747ef76e42c5a3ca32736bd4e82b?pvs=4#1ba1b08162d580829c33f74e8a3c1678

**User requirements**
https://www.notion.so/digidem/Data-Export-mobile-fe0a747ef76e42c5a3ca32736bd4e82b?pvs=4#1891b08162d5808b9b83c6c4b0f61eab

**Todo**

<img width="730" alt="Image" src="https://github.com/user-attachments/assets/7ba0b7cf-3e82-45c6-82e7-12916bb3d0e9" />

- UI Changes : [ Download Observations ] button is replacing the role card that previously sat atop the observations list. Instead we are adding the role icon to the project title in the header. In the future we may add more utility to the project title, a user may be able to tap it to see more project details. But that is not part of this work and is for future consideration. 

### Exporting
- [ ] Add roles icon next to the name of the project, at the top. example "(Icon)2025 Community Science Colla…"
- [ ] Display **Download Observations** button under project name

### Export selector

- [ ] When on the Observations list, when **Download Observations** tapped, open bottom sheet
- [ ] Display bottom sheet as designed
- [ ] Add 3 possible options : All Observations, All Observations with Media, All Tracks
- [ ] For Observation witg media, remove the mention of Video
- [ ] Display **Download** button
- [ ]  when tapped open Android share sheet
- [ ] Display **Close** button
- [ ] When tapped, go back to the observations list

### Outputs

When exporting, follow the naming convention:

- [ ] All Observations : **CoMapeo_[Project Name]_Obsvns_YYYY-MM-DD.geojson**
- [ ] Observations + Media : **CoMapeo_[Project Name]_Obsvns_YYYY-MM-DD.zip**
- [ ] Folder sorting : 
[...]

PR #1248, opened 2025-06-23
fix: data export file name

cimigree:
towards #1245 
When exporting observations with media, the export options include tracks and that is hard coded to true.
Thus, when the file is returned, the type includes the word tracks, as it is being sent to the useExportZipFile no matter what.
I am not sure why that was done.
Instead, I have set the tracks boolean to false, as the type is for Observations and not tracks with Media.
It looks like the folder for media is supposed to just be about Observations based on the [original ticket](https://github.com/digidem/comapeo-mobile/issues/1122)

>  Folder name with media : CoMapeo_[Project Name]_Obsvns_YYYY-MM-DD_Media

I can correct that on the front end. However, I cannot correct where the word Media goes (currently it is like this: **CoMapeo_New project_Media_Mon Jun 23...** and it is supposed to look like the above with Media at the end.
Also, I cannot control how the date is formatted. It looks like it is not being formatted according to the Specs. That just comes in from the back end. It looks like this, 
CoMapeo_New project_Obsvns_Mon Jun 23 2025.zip instead of as YYYY-MM-DD

I am not sure what exactly was expected and so not sure if that Boolean was true for a reason, but it appears to be making the file naming come through incorrectly.

### DIFF
--- a/src/frontend/hooks/server/projects.ts
+++ b/src/frontend/hooks/server/projects.ts
@@ -118,7 +118,7 @@ export function useExportObservationsAndShare({
         exportOptions: {
           lang,
           observations: true,
-          tracks: true,
+          tracks: false,
           attachments: true,
         },
       });


ErikSin:
Im just going to confirm this at the meeting tomorrow, but I assumed (probably incorrectly) that when we selected that option we wanted to export everything

ErikSin (approved):

PR #1250, opened 2025-06-24
fix: app crash during tracks

### LINKED ISSUE #1233: App relaunch while track- app crash
**Problem:** 
- Start Track
- Create observations
- Close CoMapeo while Tracks is active
- Relaunch CoMapeo
- App closes again
- Try to open app again, and stays open

Todo
- Is the crash because Tracks is ongoing?

cimigree:
closes #1233 
Prevents app crash caused by new functions in App.tsx
Here is the error:
 ERROR  Warning: TypeError: channel.removeListener is not a function (it is undefined)

This error is located at:
    in ServerLoading (created by App)
    in IntlProvider (created by IntlProvider)
    in IntlProvider (created by App)
    in App (created by RootApp)
    in FeedbackWidgetProvider (created by RootApp)
    in ReactNativeProfiler (created by RootApp)
    in RCTView (created by View)
    in View (created by __Sentry.TouchEventBoundary)
    in __Sentry.TouchEventBoundary (created by RootApp)
    in RootApp (created by withDevTools(RootApp))
    in withDevTools(RootApp)
    in RCTView (created by View)
    in View (created by AppContainer)
    in RCTView (created by View)
    in View (created by AppContainer)
    in AppContainer
    in main(RootComponent), js engine: hermes

I also checked for it on Sentry in the rc build just to be sure. Here it is there:
https://awana-digital.sentry.io/issues/6673926544/?environment=qa&query=&referrer=issue-stream&stream_index=0

Prevents error on undefined channel on app relaunch.

This is actually something Andrew was worried about as he was reviewing the PR that added this code:
https://github.com/digidem/comapeo-mobile/pull/1175

My change only runs the remove listener if it is defined. This is a temporary solution imo.
I don't think this is a long term, good solution to the problem, but finding one is a little out of my depth.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -35,6 +35,7 @@ import {getAppLanguageTag} from './lib/intl';
 import {IntlProvider} from './contexts/IntlContext';
 import {ServerLoading} from './ServerLoading';
 import type {StatusMessage} from '../backend/src/status';
+import type {EventEmitter} from 'events';
 
 type SentryEnvironment = 'development' | 'qa' | 'production';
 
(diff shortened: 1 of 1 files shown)


EvanHahn:
This was my fault—I made a change (d7563f5ebf93dc6c561d7f66ef6bf7cf855db487) that I thought was equivalent and it wasn't. Just spoke to Erik who has a fix.

Here's the logical difference as I see it:

Before:

```js
const subscription = nodejs.channel.addListener('server:status', msg => {
  // ...
});
// @ts-expect-error - incorrect types on nodejs.channel
return () => subscription.remove();
```

After:

```js
nodejs.channel.addListener('server:status', listener);
return () => {
  nodejs.channel.removeListener('server:status', listener);
};
```

The latter needs a reference to the channel on teardown, which seems to get deleted in some cases. The former does not.

PR #1255, opened 2025-06-25
test: add test ID to SVG mocks

EvanHahn:
*This change should have no impact today.*

We mock SVGs in our Jest tests, but don't pass them the `testID` prop.  This isn't causing problems now, but could in the future if we ever target an SVG by test ID. This fixes that.


ErikSin (approved):
Thanks Evan!

PR #1256, opened 2025-06-25
fix: properly removes node js channel subscription

### LINKED ISSUE #1233: App relaunch while track- app crash

ErikSin:
closes #1233 
and closes #1250

Nodejs mobile has improper types. Previously we were removing the subscription be overwriting the ts via `ts-expect-error`. Evan had removed it as he thought we could just close the channel directly (which has a typing) - and this was causing the app to crash. This Pr re-introduces the old architecture.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -163,10 +163,9 @@ const requestServerStatus = () => {
 };
 
 const subscribeToServerStatus = (listener: (msg: StatusMessage) => unknown) => {
-  nodejs.channel.addListener('server:status', listener);
-  return () => {
-    nodejs.channel.removeListener('server:status', listener);
-  };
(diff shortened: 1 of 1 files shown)


cimigree (approved):
I verified that the app no longer crashes when backgrounding and reopening a track, the splash screen transitions normally (no blank screen anymore. Yay), and server APIs work as expected after relaunch. I also confirmed that the FatalError screen is reachable when simulating an error. This seems like an appropriate fix here given the broken typings in nodejs-mobile-react-native. Looks good to me.

PR #1258, opened 2025-06-26
chore: tracks improvements

### LINKED ISSUE #1181: Tracks UX improvements

cimigree:
closes #1181 
Uses this [Figma Prototype](https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=5471-3571)
From this [Notion Doc](https://www.notion.so/digidem/Tracks-UX-improvements-14c1b08162d5803a922eef06556a16de)

## Tracks Feature UX Improvements

Implements key UX and architecture updates to the **Tracks** feature, based on designs and spec provided:

### New Behavior and Logic

* **Track always ends with current location**

  * When stopping tracking, user's current location is added as the final point.
  * This ensures the track line reaches their actual location.

* **Auto-dismiss or bottom sheet if no movement**

  * If user hasn't moved >1m or added observations, a `DidNotMoveBottomSheet` appears on "Stop Tracks".
  * Offers options to **Exit Tracks** (deletes track) or **Continue Recording** (resumes tracking).

* **Category (Preset) is now required**

  * Track creation and editing both support and display the selected preset.
  * Preset includes icon, and affects styling in the editor and saved detail views.

---

####  Editor Updates

* `Editor` component now accepts and displays:

  * `trackDurationMs` and `trackDistance` (formatted)
  * Updated styling for Tracks, matching designs with the distance and the time included
  * Presets and their related props are passed in

####  Track screens

* `SaveTrackScreen`

  * Pulls duration/distance from context
  * Passes track preset icon and name to `Editor`
* `TrackEditScreen`

  * Computes duration/distance from saved track
  * Uses fallback icon if none preset
* `TrackScreen`

  * Uses `TrackStats` with distance & duration below map
  * Renders preset icon if available

#### Metrics

* New utility `trackMetrics.ts`

  * `getLocationHistoryFromTrack()` – normalizes track location format
  * `getTrackDurationAndDistance()` – returns durationMs + total distance
* Used in SaveTrack, EditTrack, and Track detail views

####  TrackStats component

* Extracted and made reusable
* Accepts `distance`, `durationMs`, optional `backgroundColor` and `center` prop
* Supports internationalized "km" suffix

####  Category Chooser refactor

* Split `PresetChooser` into:

  * `ObservationCategoryChooser.tsx`
  * `TrackCategoryChooser.tsx`
* Shared grid component: `CategoryGrid.tsx`
* Improves clarity, avoids having to pass in params, clear navigation

---

### Test Updates

* **Unit Tests**

  * `TrackStoreContext.test.tsx` updated to include `preset` in all state assertions

* **End-to-End (Appium / WebDriverIO)**

  * Updated flow for no-movement track behavior:

    * Verifies `DidNotMoveBottomSheet` appears
    * Tests both “Exit Tracks” and “Continue Recording” flows
  * Skips distance/movement validation due to BrowserStack GPS limitations

---

### Styling & UX

* Map line styling updated in three places so used a shared styling module


src/frontend/lib/trackMapStyles.ts
```diff
@@ -0,0 +1,21 @@
+import {LineJoin, LineLayerStyle} from '@rnmapbox/maps';
+import {COMAPEO_BLUE, WHITE} from './styles';
+
+export const getTrackLineStyles = () => {
```
ErikSin: I dont think you need to export this as a function. You can just export BASE_TRACK_LINE_STYLE and OVERLAY_TRACK_LINE_STYLE

src/frontend/contexts/TrackStoreContext.tsx
```diff
@@ -9,11 +9,21 @@ import * as v from 'valibot';
 import {MMKVZustandStorage} from '../hooks/persistedState/createPersistedState';
 import {LocationHistoryPoint} from '../sharedTypes/location';
 import {calculateTotalDistance} from '../utils/distance';
+import {Preset} from '@comapeo/schema';
+
+const PresetSchema = v.object({
+  docId: v.string(),
+  name: v.string(),
+  geometry: v.array(v.string()),
```
ErikSin: geometry must include 'line', im not sure if that possible in valibot....

src/frontend/hooks/useTracking.ts
```diff
@@ -23,34 +26,61 @@ export function useTracking() {
     Location.startLocationUpdatesAsync(LOCATION_TASK_NAME, {
       accuracy: Location.Accuracy.Highest,
       activityType: Location.LocationActivityType.Fitness,
+      distanceInterval: 1,
+      timeInterval: 5000,
```
ErikSin: Was there a reason for adding this timeInterval

src/frontend/hooks/useTracking.ts
```diff
@@ -23,34 +26,61 @@ export function useTracking() {
...
-   * @returns {boolean} True if location history has more than one entry (track should be kept), false otherwise.
+   * @returns [distance and number of points] to determine next behavior.
    */
-  const cancelTracking = useCallback(() => {
-    setTracking(false);
+  const evaluateTrackStatus = useCallback(() => {
+    const hasMovedEnough = distance > 0.001;
+    const hasMultiplePoints = locationHistory.length > 1;
+    return {hasMovedEnough, hasMultiplePoints};
+  }, [distance, locationHistory]);
```
ErikSin: this is a redundant check. There has to be at least 2 points for there to be any distance. I would opt to get rid of this function completely. And in the places you used `evaluateTrackStatus` use `const distance = useTrackState(store=>store.distance)` instead

src/frontend/screens/MapScreen/MapLayers/TracksMapLayer.tsx
```diff
@@ -25,20 +25,12 @@ export const TracksMapLayer = () => {
       onPress={handlePress}
       id="tracks"
       shape={convertTracksToFeatures(tracks)}>
-      <LineLayer id="trackLines" style={styles.lineLayer} existing />
+      <LineLayer id="trackLineBase" style={base} existing />
+      <LineLayer id="trackLineOverlay" style={overlay} existing />
```
ErikSin: this line should actually be a black line. the current track is the dotted line, and any saved tracks are a black solid line

src/frontend/screens/SaveTrack/SaveTrackButton.tsx
```diff
@@ -34,12 +35,22 @@ export const SaveTrackButton: FC = () => {
...
+                },
+              }
+            : {}),
         },
       },
       {
         onSuccess: () => {
           clearCurrentTrack();
-          navigation.goBack();
+          navigation.navigate('Home', {
```
ErikSin: we need to use `popTo` here

src/frontend/screens/SaveTrack/SaveTrackScreen.tsx
```diff
@@ -28,17 +40,31 @@ export const SaveTrackScreen = () => {
...
+        preset ? (
+          <PresetCircleIcon iconId={preset.iconRef?.docId} size="medium" />
+        ) : (
+          <TrackIcon style={styles.icon} />
+        )
+      }
+      onPressPreset={
+        preset ? () => navigation.navigate('TrackCategoryChooser') : undefined
+      }
+      presetDisabled={preset ? false : true}
```
ErikSin: why are we disabling the preset here?

src/frontend/screens/MapScreen/TrackBottomSheet/DidNotMoveBottomSheet.tsx
```diff
@@ -0,0 +1,101 @@
...
+            renderIcon={({color, size}) => (
+              <MaterialIcons name="delete" size={size} color={color} />
+            )}
+          />
+
+          <SecondaryButton
+            fullSize
+            text={formatMessage(m.keepRecording)}
+            onPress={() => {
+              navigation.navigate('Home', {
```
ErikSin: you should use `popTo` here or the navigation stack will have 2 maps screen on it

src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleSelect = (preset: Preset) => {
+    setTrackPreset(preset);
+    navigation.navigate('SaveTrack');
+  };
+
+  React.useLayoutEffect(() => {
+    navigation.setOptions({
+      headerLeft: props =>
+        existingPreset ? (
```
ErikSin: this check won't work as trackPreset is optional (it is not optional for observations). So the user could be editting a track that has no preset, and then the go to change the preset and then they decide to press back - which will cause the confirm delete bottoms sheet to pop up and they will get stuck on the screen unless they choose a preset

src/frontend/utils/trackMetrics.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const sorted = [...locations].sort((a, b) => a.timestamp - b.timestamp);
+  const first = sorted[0];
+  const last = sorted[sorted.length - 1];
+  const durationMs = first && last ? last.timestamp - first.timestamp : 0;
+  const distance = calculateTotalDistance(sorted);
+
+  return {durationMs, distance};
+}
```
ErikSin, line 30: There is never a time you use these 2 functions seperately. For simplicity I think you should export just the 1 function, `getTrackDurationAndDistance` and just encapsulate the other function inside of it

src/frontend/hooks/server/track.ts
```diff
@@ -39,3 +40,12 @@ export function useDeleteTrackMutation() {
     docType: 'track',
   });
 }
+
+export function useGetPresetFromTrack(track: Track) {
```
ErikSin: one of the principles that we are trying to follow is that functions should be agnostic and take the minimal amount of props. In this case, the prop should be `trackPresetRefId:string|undefined`. In otherwords it does not need the entire track object if it is only using the `presetRef` Id

ErikSin (changes requested):
overall looks good, there are a few things that need to be addressed.

1. The user should be able to edit a catergory of a track after its been saved (same as an observation)
2. Watch out for the use of navigate vs popTo. To be safe you can just always use popTo and it will act like navigate if the screen is not already in the stack

src/frontend/hooks/useTracking.ts
cimigree: I just thought it would make sense to limit the updating of the location to every 5 seconds. I can remove.

src/frontend/screens/SaveTrack/SaveTrackScreen.tsx
cimigree: Because I thought it could not be edited.

cimigree: I am not super clear on the requirements here. Will EVERYONE have presets for tracks from the get go? I had the preset disabled if it was just the track icon, which would happen if someone had no track preset. Which is what it says below. If there is no preset, then they would not need to press the preset to change the preset.

src/frontend/utils/trackMetrics.ts
cimigree: But there is a time when I use just one of the functions, the second one. I think changing it makes things more complicated, since I do use getTrackDurationAndDistance with non-schema data (from state in SaveTrackScreen), where we already have the locationHistory array. If I collapse the two functions, I’d need to add unnecessary conversions in those places just to extract data that’s already in the correct format or the passing of various parameters and a lot of if checks, so I either I would like to leave it or I need you to show me how to do what you ask in a way that makes it less complicated instead of more so.

src/frontend/screens/SaveTrack/SaveTrackScreen.tsx
cimigree: I will change.

src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
cimigree: Ok I think I found a way to deal with that.

cimigree:
### Summary of Changes


* Changed some navigations to popTo as appropriate
* Updated the Saved track to just be a black line
* Took out the points check
* Enables **editing the category (preset)** of a saved Track, similar to Observation editing.
* Updates `TrackEdit` screen to:
  * Load `presetRef` from the saved Track.
  * Store `preset` and `docId` in `TrackStoreContext`.
  * Navigate to `TrackCategoryChooser` to update preset.
* Updates `TrackCategoryChooser` to:
  * Use `docId` from store to decide navigation behavior
* Used `TrackStoreContext` to track the `docId` of the track being edited.
  * This avoids passing navigation params or trying to check route values in `TrackCategoryChooser`.
  * Follows the same pattern as observation draft handling.
* Preset is only set via `setTrackPreset`; it's read from store when saving the track.
  * Mirrors how observations manage the draft preset, I think.
* Cleaned up context + test to support `docId` addition and preset validation.

src/frontend/screens/SaveTrack/SaveTrackScreen.tsx
ErikSin: It based on whether there are presets available (via the categories). So we need to check if there are trackPresets. and if there are no track presets then disable it

cimigree: Ok I think I handled all of the cases for that.

src/frontend/hooks/useTracking.ts
```diff
@@ -23,34 +24,51 @@ export function useTracking() {
...
-    const hasTrackWithMoreThan1Point = locationHistory.length > 1;
 
-    if (!hasTrackWithMoreThan1Point) {
-      clearCurrentTrack();
+    if (location?.coords) {
+      addNewLocations([
+        {
+          latitude: location.coords.latitude,
+          longitude: location.coords.longitude,
+          timestamp: location.timestamp ?? Date.now(),
```
ErikSin: is this not a redundant check? location.timestamp is a required field

src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      navigation.navigate('SaveTrack');
+    }
+  };
+
+  React.useEffect(() => {
+    if (trackPresets.length === 0) {
+      navigation.goBack();
+    }
+  }, [trackPresets.length, navigation]);
```
ErikSin: I think we can take this out. We do the check before navigating anyways. I understand the reasoning, as there is a chance that a dev navigates to this page without there being track presets...but i think because there are only a few use cases, it doesn't matter too much...

If you are worried, we could pass the array of trackPresetIds, as a nav param. That way the developer has to explicitly check and find the ids before even navigating. I dont think it is necessary, but if you feel worried then feel free to add a nav check like that

src/frontend/utils/trackMetrics.ts
ErikSin: Sorry, i only searched for the one function. But `getTrackDurationAndDistance` always requires 
`getLocationHistoryFromTrack`. Would it not make sense to just encapsulate `getLocationHistoryFromTrack` inside `getTrackDurationAndDistance`?

src/frontend/hooks/useTracking.ts
```diff
@@ -23,34 +24,51 @@ export function useTracking() {
...
-    setTracking(false);
+  const endTracking = useCallback(() => {
     Location.stopLocationUpdatesAsync(LOCATION_TASK_NAME).catch(err => {
       Sentry.captureException(err);
     });
-    const hasTrackWithMoreThan1Point = locationHistory.length > 1;
 
-    if (!hasTrackWithMoreThan1Point) {
-      clearCurrentTrack();
+    if (location?.coords) {
```
ErikSin, line 46: we should probably check the accuracy, and make sure the accuracy is within 3m...otherwise it wont align with the rest of the tracks which have a high accuracy. I think there is actually a better way to do this, but prehaps out of scope for this PR, so ill create an issue

ErikSin (changes requested):
Some small comment, but the actual blocking change are the icons. see picture:

Image

cimigree:
About this:

> Sorry, i only searched for the one function. But getTrackDurationAndDistance always requires
getLocationHistoryFromTrack. Would it not make sense to just encapsulate getLocationHistoryFromTrack inside getTrackDurationAndDistance?

It doesn't actually. That is the whole reason it is set out as a different function. It requires the location history, which comes from state in one case. In that case, I don't have a track, because a track doesn't exist yet.
In the case when it is used, there is no location history in state and I need to get it from the track and use that function. It is only used once, so I can take it out of there and put it right in the file where it is used if you would prefer.

src/frontend/utils/trackMetrics.ts
cimigree: It doesn't actually. That is the whole reason it is set out as a different function. It requires the location history, which comes from state in one case. In that case, I don't have a track, because a track doesn't exist yet.
In the case when it is used, there is no location history in state and I need to get it from the track and use that function. It is used in two different files which is why I made it a function so I wouldn't have to repeat the logic twice. I can just repeat it in the two files though instead of defining it in here if you would prefer.

ErikSin: ohh i see i see. Ok, no what you have makes sense then!

src/frontend/hooks/useTracking.ts
cimigree: [bb2965d](https://github.com/digidem/comapeo-mobile/pull/1258/commits/bb2965d825aefadf44bab49cce706c42d3e519ff)

src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
cimigree: [bb2965d](https://github.com/digidem/comapeo-mobile/pull/1258/commits/bb2965d825aefadf44bab49cce706c42d3e519ff)

src/frontend/hooks/useTracking.ts
cimigree: [bb2965d](https://github.com/digidem/comapeo-mobile/pull/1258/commits/bb2965d825aefadf44bab49cce706c42d3e519ff)

cimigree:
Oops about the icon but it is a very simple fix.
This look ok to you?
<image src="https://github.com/user-attachments/assets/6b42af19-f72e-42d3-aaeb-b0241daa4655" width="350" />

ErikSin (approved):
Thanks! Exciting to have the new Track UI!

src/frontend/hooks/useTracking.ts
gmaclennan: @ErikSin Why 3m? It's rare / impossible for GPS to be as accurate as 3m

PR #1259, opened 2025-06-26
chore: show replace usage

ErikSin:
demonstrating how to use replace instead of resetting the navigation stack

### DIFF
--- a/src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/index.tsx
+++ b/src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/index.tsx
@@ -7,6 +7,7 @@ import {
   StyleSheet,
   View,
   TouchableWithoutFeedback,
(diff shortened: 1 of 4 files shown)


src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/index.tsx
```diff
@@ -90,19 +91,17 @@ export const CreateOrNameSoloProject = ({
...
 
-      event.preventDefault();
-    });
-
-    return () => {
-      unsubscribe();
-    };
-  }, [navigation, mutationIsPending]);
+    return () => subscription.remove();
+  }, [mutationIsPending]);
```
ErikSin, line 104: `beforeRemove` was causing the screen not to navigate properly. replace "removes" the screen. The on-success of the mutation was called before the mutationIsPending state was changed, so this was causing the screen not to be removed.

PR #1260, opened 2025-06-30
fix: obs blank space

### LINKED ISSUE #999: Obvs w track - no pic / blank space
**Describe the bug**
 Create an Observation while a Track is being recorded
Dont add a picture to that Observation
Save Observation

When looking at the Observation, there's a blank space

https://www.notion.so/digidem/Obvs-w-tracks-no-pics-random-space-18e1b08162d5804284e4d5bf361a7829

Image

- Possible to not have the blank space?

cimigree:
closes #999 

This PR fixes an issue where there was a random extra blank space on the observation screen.
The fix requires not only removing the unneeded flex: 1 from the section, but also only renders the notes if there are notes. The typeof conditional was always returning true.
I also took out the bottom padding for the section and added it only where needed so that there would not be an awkward space between divider lines.

<image src="https://github.com/user-attachments/assets/3904becc-2590-4b3a-8c0c-3e8160396e3d" width="250" />


ErikSin (changes requested):
Sorry, i think there is a little confusion about what is wanted. So we don't want to get rid of the spacing entirely. Essentially, if there is NO scroll, then we want the observation details and the 2 buttons to hug the bottom of the screen. But if there is a scroll then we do not want there to be an unnecessary space which forces the scroll view to be longer. The solution would be a flex box with justifyContent of spaceBetween. This will add space if there is room, but will get rid of the space when there is alot of content

cimigree:
@ErikSin This is what you meant? It looks ok to me, but would be good if you could also check various scenarios on your devices to see if it looks how you want it to. Still ok to only render notes when there are notes? Or do you want them rendered no matter what?

ErikSin:
Yeah it seems to be working:

here is it BEFORE the fix:
image


and AFTER:
image


But im not sure if the justify content fixed it or it was just the notes check .... im testing it without the justifyContent now

ErikSin (approved):
It seems like the `justifyContent:space-between` is actually not needed and it was the notes that was causing the problem. You can take that out before merging

PR #1261, opened 2025-06-30
fix: tracks location permission fixes

### LINKED ISSUE #1164: Tracks/ if location is disabled/ odd flow
**Describe the bug**
Set location permissions to Dont Allow
Tap on Tracks
App asks to grant location
User grants “allow only while using the app”
Go back to the app
App says to enable location permission
Click in ENABLE
User selects “allow all the time” 
Go back to app
User still sees the Enable button
Tap it
Back to the Locations permission screen
And that loops continues

**Workaround**
If user restarts the app, the app is now set to “allow all the time”  and Tracks works

**Expected behavior**
-Click in ENABLE / User selects “allow all the time” 
-when user goes back to then app, Tracks should be ready to start

See video
https://www.notion.so/digidem/Tracks-if-location-is-disabled-odd-flow-1e81b08162d5806b8ce0ccc4b4dc8760

cimigree:
closes #1164 

The problem is the app was only checking permissions once when the screen first mounted. It did not detect changes made in system settings after navigating away and returning.

Now it checks for permissions when app state becomes active. Adds a shared method for both permission checks to use. 

The other one, focus, is still for sure needed to cause the modals to open and if the user uses the in app permission granting.

Similar to the fix required in #1242


src/frontend/screens/MapScreen/TrackBottomSheet/index.tsx
```diff
@@ -19,18 +19,44 @@ export const TrackBottomSheet = React.memo(() => {
...
     React.useCallback(() => {
-      Location.getForegroundPermissionsAsync().then(permission =>
-        setForegroundPermission(permission),
-      );
-
-      Location.getBackgroundPermissionsAsync().then(permission =>
-        setBackgroundPermission(permission),
-      );
-    }, []),
+      checkPermissions();
```
ErikSin: I don't think we need this useCallback anymore as checkPermissions is already memoized

ErikSin (approved):
I small non blocking comment

PR #1263, opened 2025-07-01
fix: shows preset color ring

### LINKED ISSUE #1257: Observation list should show the category color as a border for a category
**Describe the bug**
When choosing a category from the category list (whether creating a new observation or editing the category of an existing observation) the border color for a category should correspond to the actual category color set in the categories file

See details : https://www.notion.so/digidem/Observation-list-should-show-the-category-color-as-a-border-for-a-category-1ed1b08162d58042a8fefbf26e7f7f9b

cimigree:
closes #1257 
As far as I can tell, we haven't shown the color ring around the presets/ categories since May of 2024 [when the issue included the to do item of removing the color prop](https://github.com/digidem/comapeo-mobile/issues/340). I am not sure why that was done.

So, this PR reinstates the color prop. Passes color to the PresetCircleIcon component wherever it is used, which then passes it to the Circle component.
I increased the border width to 3 px because that is [what it appears to be in Figma](https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=3379-985&t=ttzOfc0HnvSjPCG6-0).

---

<image src="https://github.com/user-attachments/assets/53aa4a5a-59f9-486d-838a-942f2e128f0c" width="250" />
<image src="https://github.com/user-attachments/assets/16d70a1c-9449-4dec-929f-009b806b7840" width="250" />
<image src="https://github.com/user-attachments/assets/82eb783b-6e8c-4817-9ecc-d0da7e45a4c5" width="250" />
<image src="https://github.com/user-attachments/assets/e1f3595b-7c2b-4e64-95ea-948717d6bd26" width="250" />

### DIFF
--- a/src/frontend/sharedComponents/icons/Circle.tsx
+++ b/src/frontend/sharedComponents/icons/Circle.tsx
@@ -6,7 +6,7 @@ import {ViewStyleProp} from '../../sharedTypes';
 
 const BORDER_DEFAULTS = {
   color: '#EAEAEA',
-  width: 1,
+  width: 3,
 } as const;
 
 export const Circle = ({
--- a/src/frontend/sharedComponents/icons/Circle.tsx
+++ b/src/frontend/sharedComponents/icons/Circle.tsx
@@ -6,7 +6,7 @@ import {ViewStyleProp} from '../../sharedTypes';
 
 const BORDER_DEFAULTS = {
   color: '#EAEAEA',
-  width: 3,
+  width: 1,
 } as const;
 
 export const Circle = ({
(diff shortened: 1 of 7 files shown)


src/frontend/sharedComponents/icons/Circle.tsx
```diff
@@ -6,7 +6,7 @@ import {ViewStyleProp} from '../../sharedTypes';
 
 const BORDER_DEFAULTS = {
   color: '#EAEAEA',
-  width: 1,
+  width: 3,
```
achou11: Relatively minor, but I think this is going to unintentionally affect any other components that are using the `Circle` icon (counted 6 based on a quick search)

achou11 (changes requested):
one comment that probably requires changes

src/frontend/sharedComponents/icons/Circle.tsx
cimigree: Oops! Good catch. Sorry about that. Fixed in [bf7a7df](https://github.com/digidem/comapeo-mobile/pull/1263/commits/bf7a7df058d2771199db04abf7bc36bf7f302868)

achou11 (approved):
assuming CI passes, lgtm!

PR #1270, opened 2025-07-03
fix: screen unresponsive when track running

cimigree:
- There was unresponsive UI (or extremely delayed touch handling) when the TrackBottomSheet was visible and E2E test failures and manual interactions being blocked.
- Because of a non memoized checking of permissions, there were unnecessary or repeated re-renders, making the app appear frozen or unresponsive.
- So, this PR updates useFocusEffect to use a memoized callback
- That is the way I originally had it my PR, so that is what I tested with the most. https://github.com/digidem/comapeo-mobile/pull/1261
- This ensures the checkPermissions() effect is stable and only re-runs when intended, preventing unnecessary re-renders or re-evaluation loops


ErikSin (approved):
This does not makes any sense to me...react navigation is such a black box

PR #1271, opened 2025-07-03
fix: map wrong size and placement

### LINKED ISSUE #1184: Map only shows in the bottom left corner of the screen
**Describe the bug**

About half the time when I open CoMapeo Mobile 1.3.0, the map only shows in the bottom left corner of the screen. The rest of the screen is blank. (See screenshot where you can see the tiny map and my location within it underneath the scale bar).

When this happens, if I close CoMapeo and open it again, the map renders fine. 

**To Reproduce**

I haven't pinned down exactly when this happens, but the following just worked for me.

1. Clear data and cache from CoMapeo in Android Settings -> Applications.
2. Open CoMapeo, give permissions, and move past the intro screens.
3. See error.

**Screenshots**

Image

 - Device: Samsung Galaxy S25 Ultra
 - OS: Android version 15
 - Version: 1.3.0 (Updated on May 14, 2025)

cimigree:
closes #1184 

This PR reliably initializes full-screen Mapbox on Android cold-start
On Android, 10.x of @rnmapbox/maps, MapView defaults to SurfaceView and sets the size on whatever size it sees at mount time—and never re-measures. On a totally “cold” launch (cache + data cleared) that initial mount often reports 0×0 (or 1×1), so you get a tiny map stuck in the bottom-left. Android 15’s layout/timing changes make this happen almost 50% of the time.
This PR works around the race by:

**1. Measuring the container first**
Wrap the map in a onLayout that captures its real width/height into local state (dims).

**2. Delaying the MapView mount**
We only render <Mapbox.MapView> once dimensions are not null, so that the very first setFixedSize(w,h) matches the true screen size.

**3. Forcing a fresh surface via key**
Whenever dimensions change (which only happens once, on first layout), we bump a mapKey so React tears down any previous SurfaceView and builds a new one at the correct size.

Together these changes seem to make the map fill the screen 100% of the time on a cold Android 15 start. I verified 10/10 clean launches on a Pixel 8 device.

**How to test**

1. Install on an Android 15 device/emulator.
2. Clear app data & cache.
3. Open the app; observe the map fill the screen immediately.
4. Repeat steps 2–3 at least 5× to confirm no tiny‐map edge cases.

If we ever become able, we should upgrade to @rnmapbox/maps@^11.1.0 and switch to textureView={true} instead of surfaceView. That will resize with flex layout and let us remove all of this measurement+key stuff.

**References**

* **Mapbox v10 docs** [(surfaceView defaults to `true` on Android, no `textureView` override)](https://rnmapbox.github.io/docs/components/MapView#surfaceview)

* **Android SurfaceView API reference** [(describes how a `SurfaceView` creates its own surface at mount and doesn’t automatically re-measure on later layout passes)](https://developer.android.com/reference/android/view/SurfaceView)


cimigree:
@ErikSin Moved the resize logic to a hook and tested.

ErikSin (approved):
Hard to recreate, but doesnt break anything so i am going to approve

PR #1275, opened 2025-07-10
fix: passcode limit attempts

### LINKED ISSUE #1227: Passcode/no limit on the number of passcode attempts
Via Atredis report - https://www.notion.so/digidem/CoMapeo-Security-Audit-1b31b08162d58078a52cfa2308dc24bb

**No Limit on Passcode Attempts**
**SEVERITY: MEDIUM**
Finding Overview
CoMapeo does not limit the number of incorrect passcodes that may be attempted, allowing an
attacker with physical access to the device to brute-force the CoMapeo passcode manually or by
automated means.

Finding Detail
CoMapeo has an opt-in feature to set a five-digit passcode that is required when launching the
application and when switching to it from another application but there is no limit set on the number
of attempts that can be attempted. By enumerating common values (00000, 12345, etc.) and those
that hold significance to a target user (e.g., important dates), an attacker with physical access to the
user's device may be able to manually perform a dictionary attack against their passcode.
Alternatively, the attacker can automate a brute-force attack by simulating key events with Android
Debug Bridge (ADB). This attack does not require root access to the device but does require that
USB debugging is enabled, and that the attacker's computer is trusted by the device.
The proof-of-concept Python script below iterates over all possible passcodes, sending each one as
simulated input to a USB-connected device. Using this method, an attacker can exhaust the key
space (100,000 possible passcodes) in several hours.
import os
for i in range(0, 99999):
 guess = str(i).zfill(5)
 os.system("adb shell input text '" + guess + "'")
Passcode brute-force script

**Recommendation(s)**
Modify the passcode verification functionality to temporarily disallow further attempts after a certain
number of failed attempts (e.g., five). Have the application increase the lockout period as the number
of failed attempts rises.

**Todo**

https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6770-33760&t=uPFFZX5XOuOR2f3i-0

**Notification 1**

Image

- [ ] Display screen as designed
- [ ] Display "Try again in 1 minute"
- [ ] show this message after 5 failed login attempts.

**Notification 2**

Image

- [ ] show this message after 5 failed login attempts.
- [ ] Display "Try again in 3 minutes"
- [ ] After Notification 1 , the user can make 2 more attempts. If the 7th attempt fails, show this message.

**Notification 3- Final**

Image

- [ ] Display "Try again in 5 minutes"
- [ ] After Notification 2, the user can try 1 more time. If the 8th attempt fails, show this message. 
- [ ] Show this message after every failed attempt thereafter.

cimigree:
closes #1227 

### Behavior

* Lockout triggers after a defined number of failed attempts:

  * 5 attempts → lock for 1 minute
  * 7 attempts → lock for 3 minutes
  * 8+ attempts → lock for 5 minutes
* Lockout message appears:
  `"Try again in X minute(s)"`
* During lockout:

  * Input is disabled
  * Attempts are blocked with a `"LOCKED_OUT"` error
* After lockout period ends, input is automatically re-enabled.

### `AuthContext.tsx`

* Centralizes all passcode-related decision logic (validation, obscure mode, lockout, etc.) in `authenticate()`.
* `LOCKED_OUT` errors are thrown when appropriate so UI can respond.

---

### `SecurityStoreContext.tsx`

* Now stores:

  * `failedAttempts`
  * `lockUntil`
* Adds:

  * `incrementAndGetAttempts()`
  * `resetFailedAttempts()`
  * `setLockout()`
* Automatically clears attempts and lockout when passcode is removed.

---

### `EnterPassToTurnOff.tsx`

* Refactored to no longer use the generic `InputPasscode` component. This was shared with the set/ confirm passcode. It didn't make sense to me to use that for a passcode authorization screen. It was too different after adding the new lockout state and error recovery logic and involved having to pass too many props and becoming overly complex.
* Uses `PasscodeInput` directly and handles validation inline.
* Implements lockout behavior using `usePasscodeLockout()` hook.

---

###  Unit Tests

**`SecurityStoreContext.test.tsx`**

* Validates:

  * Initial state
  * Passcode setting
  * Obscure code rules
  * Resetting state

**`security.test.ts`**

* Tests validation logic (`PasscodeSchema`)
* Tests lockout logic (`getLockoutThreshold`, `getRemainingLockoutMinutes`)

### E2E Tests (Appium/BrowserStack)

**Updated:**
`check-passcode-requirements.test.ts`

* Simulates 5 failed attempts
* Confirms lockout message appears
* Waits 1 minute, confirms login is possible again
* Verifies input is reset properly on retry


cimigree:
@ErikSin, In your version since we only calculate minutes once, the message doesn't change in the way that I told you it did. It's just a static message now. Just wanted to make sure you weren't expecting the behavior I had promised was going to happen because that was based on the code that I wrote and should be now ready for re-review.

ErikSin:
> @ErikSin, In your version since we only calculate minutes once, the message doesn't change in the way that I told you it did. It's just a static message now. Just wanted to make sure you weren't expecting the behavior I had promised was going to happen because that was based on the code that I wrote and should be now ready for re-review.

Hmm, im not sure i see how the code has changed (in terms of the message being static) I believe the way it was [previously done](https://github.com/digidem/comapeo-mobile/blob/c42ad8736a42e573e737bb0bfd2f1407db4ae7a7/src/frontend/hooks/usePasscodeLockout.ts#L22) was also a static message. Im open to changing it if you don't think it should be static, but I don't think the message being static has changed from the way you had it

src/frontend/contexts/AuthContext.tsx
```diff
@@ -73,19 +76,39 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
     (passcodeValue, validateOnly = false) => {
       if (validateOnly) return passcodeValue === passcode;
 
+      const isLockedOut = lockUntil !== null && Date.now() < lockUntil;
```
ErikSin: You can take out this null check

src/frontend/lib/security.ts
```diff
@@ -8,3 +9,16 @@ export const PasscodeSchema = pipe(
...
+  const msRemaining = lockUntil - Date.now();
+  return msRemaining > 0 ? Math.ceil(msRemaining / 60000) : 0;
+}
+
+export function getLockoutThreshold(attempts: number): number | null {
+  const match = PASSCODE_LOCKOUT_THRESHOLDS.find(
+    t => t.attempts === attempts || (attempts > 8 && t.attempts === 8),
+  );
+  return match ? match.minutes : null;
+}
```
ErikSin: This function is a little hard to read. I think it can be simplified and does not need `PASSCODE_LOCKOUT_THRESHOLDS`

```ts
export function getLockoutThreshold(attempts: number): number | null {
  if (attempts === 5) return 1;
  if (attempts === 7) return 3;
  if (attempts >= 8) return 5;

  return 0 //I think using 0 makes it easier to use instead of null, but i think that is just a preference thing at this point
}

```

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -14,10 +14,14 @@ const SecurityStateSchema = v.variant('passcode', [
...
     passcode: PasscodeSchema,
     obscureCodeEnabled: v.boolean(),
+    failedAttempts: v.number(),
+    lockUntil: v.number(),
   }),
   v.object({
     passcode: v.null(),
     obscureCodeEnabled: v.literal(false),
+    failedAttempts: v.number(),
+    lockUntil: v.number(),
```
ErikSin: We can be a little more strict with the typing and make both of these `v.literal(0) `

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -82,6 +93,19 @@ export function createSecurityStore({persist} = {persist: false}) {
...
+      const current = store.getState().failedAttempts + 1;
+      store.setState({failedAttempts: current});
+      return current;
+    },
+
+    resetFailedAttempts: () => {
+      store.setState({failedAttempts: 0, lockUntil: 0});
+    },
+
+    setLockout: (lockUntil: number) => {
```
ErikSin: to be a little clearer can we call this `setLockUntil`?

src/frontend/lib/security.test.ts
```diff
@@ -22,3 +27,33 @@ describe('PasscodeSchema', () => {
...
+    expect(getRemainingLockoutMinutes(past)).toBe(0);
+  });
+});
+
+describe('getLockoutThreshold', () => {
+  test.each([
+    [1, null],
+    [4, null],
+    [5, 1],
+    [7, 3],
```
ErikSin: we should test "6 attempts" here because that is the odd one out

ErikSin (commented):
Im not sure if your worried about the static message. IMO i don't think its a big deal, but if you would like to change that then it would require a re-review. Otherwise I think it looks good. Just waiting to hear on what you want to do before i approve it.

There are a few non-blocking comments and also the spacing of the error message is cutting of the SVG:

Image

src/frontend/lib/security.ts
cimigree: I just liked the idea of having a constant instead of magic numbers. But I changed it as you prefer.

cimigree:
> Im not sure if your worried about the static message. IMO i don't think its a big deal, but if you would like to change that then it would require a re-review. Otherwise I think it looks good. Just waiting to hear on what you want to do before i approve it.
> 
> There are a few non-blocking comments and also the spacing of the error message is cutting of the SVG:
> 
> Image

I guess I imagined the static message changing. It only changes when you force close and reopen and it does that in this version also. Sorry about the confusion.

I triple checked this though...
The logo is cut off on develop which is why I was not worried about it. My bad.


I did fix it this pass through in a way I hope that makes sense, but I just wanted you to know that it was not the spacing of the error message (that just replaces an existing message anyway).

ErikSin:
> I guess I imagined the static message changing. It only changes when you force close and reopen and it does that in this version also.

Im happy to keep it as is, but if you would like to change it to a dynamic timer, let me know (as I don't want to approve it if we are changing it). Otherwise I will approve it as is.

cimigree:
> > I guess I imagined the static message changing. It only changes when you force close and reopen and it does that in this version also.
> 
> Im happy to keep it as is, but if you would like to change it to a dynamic timer, let me know (as I don't want to approve it if we are changing it). Otherwise I will approve it as is.

No it's fine. I think everyone will understand and the timing will change if they force close.

ErikSin (approved):

PR #1279, opened 2025-07-16
fix: wait for backend start before sending RPC

### LINKED ISSUE #1063: Local Discovery server sometimes not starting
The local discovery server can sometimes fail to start. This is because the RPC message that starts the server can be sent before the backend nodejs process is started. This is more likely as additional code is added to the backend and nodejs startup is slower, particularly on older phones.

This is the likely cause of issues with devices not being able to discover each other.

gmaclennan:
This fixes https://github.com/digidem/comapeo-core/issues/1063

The `ServerLoading` component had been largely protecting us from this issue, because it does not render any of the app until the server has started, which ensures that not RPC messages are sent before the server is started. However, the LocalDiscoveryController is initiated outside the React render tree, and the RPC message sent by `mapeoApi.startLocalDiscoveryServer()` can happen before the backend nodejs process has started, which results in the server never starting.

This is not an issue if it only happens on one phone out of a pair of phones trying to connect, since the connection can happen in the other direction, however if it happens on both phones, then they will never be able to connect to each other, so they will not show as available devices for invite or sync.

This issue is more likely to occur on slower phones or when additional code is added to the backend, which slows in initial nodejs startup.

This PR adds an internal queue to the frontend MessagePort implementation that waits for the backend to start before sending any RPC messages. It moves the logic to listen for server start out of the react tree and into a new `ServerStateStore`. It uses `React.useSyncExternalStore()` to use this state in the `ServerLoading` component which is now much simpler and mainly responsible for showing an error screen if there is a fatal server error. `ServerLoading` also currently does not render the app while the backend is starting, but it does not need to do this now, and could just render the app and the user would see the observation list loading animation as the server starts. However I have left this behavior as-is for now.

A similar queue is not needed in the backend, because we only post RPC messages to the frontend after receiving a message first from the frontend, so it is guaranteed that the frontend process is started and listening before the backend sends anything.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -34,7 +34,7 @@ import {getAppLanguageTag} from './lib/intl';
 import {IntlProvider} from './contexts/IntlContext';
 import {ServerLoading} from './ServerLoading';
 import {createSavedLocationStore} from './contexts/SavedLocationContext';
-import {createServerStateStore} from './lib/serverStateStore.ts';
+import {createServerStateStore} from './lib/ServerStateStore.ts';
 
 type SentryEnvironment = 'development' | 'qa' | 'production';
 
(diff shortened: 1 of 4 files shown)


RangerMauve (approved):
LGTM code-wise

PR #1280, opened 2025-07-16
fix: store hashed passcode

### LINKED ISSUE #1228: App Passcode Stored in Plaintext
Via Atredis report - https://www.notion.so/digidem/CoMapeo-Security-Audit-1b31b08162d58078a52cfa2308dc24bb

**Finding Overview**
CoMapeo's in-app passcode is stored on the device filesystem in unencrypted form. This allows an
attacker with root access to the device to extract the passcode.
**Finding Detail**
An opt-in passcode feature is available to lock/unlock the CoMapeo interface. If enabled, the
passcode must be provided when the app is started or switched back to. However, the passcode is
stored in plaintext on the device using the Persist middleware of the Zustand library. In the following
code snippet, the set function is implemented by CoMapeo's Zustand Persist store and writes the
passcode, in plaintext, to the application's mmkv.default file. This snippet is from comapeomobile/src/frontend/hooks/persistedState/usePersistedPasscode.ts:
setPasscode: passcodeValue => {
 const obscureCode = get().obscureCode;
 if (passcodeValue === OBSCURE_PASSCODE) {
 throw new Error('passcode is reserved');
 }
 if (passcodeValue !== null && passcodeValue === obscureCode) {
 throw new Error('passcode is already being used obscure code');
 }
 if (!validPasscode(passcodeValue)) {
 throw new Error('passcode not valid');
 }
 if (passcodeValue === null) {
 set({passcode: null, obscureCode: null});
 return;
 }
 set({passcode: passcodeValue});
},
passcode persisted to filesystem in plaintext
An attacker with physical access to a rooted device can therefore retrieve the plaintext passcode
using the Android Debug Bridge (ADB), as shown below:
Atredis Partners – Open Technology Fund CoMapeo Platform Security Assessment
Atredis Partners - Confidential Page 26
$ adb shell su -c "cat /data/user/0/com.comapeo/files/mmkv/mmkv.default" | grep -
Eoa 'passcode":"\d+"'
passcode":"12345"
Extraction of passcode from device

**Recommendation(s)**
If the device is not shared by other users, delegate app locking to the user's device credentials
(passcode and/or biometrics). When app locking is enabled, require users to have at least one device
credential set. Require the user to authenticate using a valid device credential every time the app is
started or switched to from another app. Consider combining this functionality with the
recommendation given in Expo SecureStore Not Protected by User Credentials to tie use of the root
key to the device credentials when app locking is enabled.
If the device is shared with other users, CoMapeo should only store a hash of the user’s passcode
and compare that value. During the unlock process, compare the stored hash against a hash of the
entered passcode. A strong password hashing algorithm should be used for this process.

cimigree:
closes #1228 

### What this PR Does

* Replaces legacy passcode handling with a secure and modern implementation using `expo-crypto` + `expo-secure-store`.
* Introduces salted SHA-256 hashing for passcodes.
* Ensures **backward compatibility** by detecting and upgrading legacy passcodes (5-digit plain text) on next successful login.
* Updates `setPasscode` and `authenticate` logic to be fully async.
* Adds validation using `valibot` schemas for both hashed and input formats.
* Adds new unit tests.

---

### Why Not bcryptjs / react-native-bcrypt?

* `bcryptjs` is **pure JS**, and relies on native code that’s incompatible with our React Native / Expo project setup.
* `react-native-bcrypt` is outdated (last touched 9 years ago), unmaintained, and failed in our environment with:

  ```
  Error: Illegal callback: undefined
  ```

* `expo-crypto` and `expo-secure-store` are maintained, secure, and compatible with **Expo and Android** even if they are slightly more complicated to use. We even had both installed and setup with mocks for testing.

---

### How to Test Legacy Passcode Migration

1. **Start with a legacy passcode** (5-digit plain text string) by starting the app on the develop branch or the [fix/: passcode limit attempts](https://github.com/digidem/comapeo-mobile/pull/1275) and creating a passcode. Then change branches and rerun `npm run android`

2. **Launch the app** and enter the passcode.

   * On successful authentication, the app will:

     *  Detect the legacy format
     * Hash it using `SHA-256` + salt
     * Save the hashed version to `SecureStore`

3. Subsequent logins will use the new hash format as the password will have been updated in the SecureStore.


src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -61,18 +72,36 @@ export function createSecurityStore({persist} = {persist: false}) {
...
     },
 
+    updateToHashedPasscode: async (passcode: string) => {
+      const salt = generateSalt();
+      const hashed = await hashPasscode(passcode, salt);
+      store.setState({passcode: hashed});
+      await setItem(
+        STORAGE_KEY,
+        JSON.stringify({...store.getState(), passcode: hashed}),
+      );
```
ErikSin: This bypasses Zustand. Im not sure how it is done under the hood, but there is a potential that you are not using the zustand store anymore by doing this. To be safe, we should use the zustand [migrate function instead](https://zustand.docs.pmnd.rs/integrations/persisting-store-data#migrate), which accounts for changes in the store

src/frontend/constants.ts
```diff
@@ -74,3 +74,12 @@ export const projectColors: projectColor[] = [
...
 ];
+
+export const PASSCODE_LOCKOUT_THRESHOLDS: {
+  attempts: number;
+  minutes: number;
+}[] = [
+  {attempts: 5, minutes: 1},
+  {attempts: 7, minutes: 3},
+  {attempts: 8, minutes: 5}, // 8 and above get 5 minutes
+];
```
ErikSin: can remove

src/frontend/contexts/AuthContext.tsx
```diff
@@ -73,21 +78,32 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
   }, [passcode, shareDialogIsOpen, isAudioPermissionModalOpen]);
 
   const authenticate: AuthContextType['authenticate'] = React.useCallback(
-    (passcodeValue, validateOnly = false) => {
-      if (validateOnly) return passcodeValue === passcode;
-
-      const isLockedOut = Date.now() < lockUntil;
+    async (passcodeValue, validateOnly = false) => {
+      const isLockedOut = lockUntil !== null && Date.now() < lockUntil;
```
ErikSin: merge conflict: can remove null check

src/frontend/contexts/AuthContext.tsx
```diff
@@ -73,21 +78,32 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
+      if (passcodeValue && passcode) {
+        if (safeParse(PasscodeInputSchema, passcode).success) {
+          isCorrect = passcodeValue === passcode;
+          await updateToHashedPasscode(passcodeValue);
+        } else {
+          isCorrect = await verifyPasscode(passcodeValue, passcode);
+        }
+      }
+
+      if (validateOnly) return isCorrect;
```
ErikSin: See comment about migration - this screen should not be dealing with this check

src/frontend/lib/security.ts
```diff
@@ -22,3 +30,32 @@ export function getLockoutThreshold(attempts: number): number | null {
...
+    Crypto.CryptoDigestAlgorithm.SHA256,
+    toHash,
+  );
+  return `${salt}:${hash}`;
+}
+
+export async function verifyPasscode(
+  input: string,
+  stored: string,
+): Promise<boolean> {
```
ErikSin: When there are 2 or more parameters, we should use an object so it is less ambigous. Eg:

```ts
hashPasscode({passcode: string,
salt: string})

verifyPasscode({input:string, stored:string})


```

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
ErikSin, line 1: We should refactor this to use the new Bottom Sheet Architecture.

We should also add a loading state and error state since adding a passcode is now async

src/frontend/screens/AuthScreen.tsx
```diff
@@ -78,9 +78,9 @@ export const AuthScreen = ({
     }
   }
 
-  function validatePass(passValue: string) {
+  async function validatePass(passValue: string) {
```
ErikSin, line 78: We need to account for this loading state now

ErikSin (changes requested):
The architecture looks really good! There are just 2 blocking things which are side effects of using this new architecture:
1. We should use the migrate function when changing the store...The way you are doing it may work, but I think we should follow the official way of doing it as it will be less prone to errors
2. authenticate and setPassword are now 2 async functions. The UI needs to reflect that. Otherwise there will be times when things are loading, but it looks like nothing is happening. On faster phones you may not notice it, but on slower phones it will just look buggy. We should add loaders. I will ask design what that should look like for the authentication screen.

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -39,35 +44,66 @@ function createInitialState(): SecurityState {
...
+): Promise<SecurityState> => {
+  if (version !== 0) {
+    return persistedState as SecurityState;
+  }
+  try {
+    const raw = persistedState as LegacyPasscodeState;
+    // not sure if I need this if check here.
+    if (
+      typeof raw.passcode === 'string' &&
+      v.safeParse(PasscodeInputSchema, raw.passcode).success
```
ErikSin: v.safeParse is not necessary.

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -39,35 +44,66 @@ function createInitialState(): SecurityState {
   };
 }
 
+type LegacyPasscodeState = {
```
ErikSin, line 54: we can also add a:
```ts
 /**
 * @deprecated Removed in v1. Use SecurityState instead.
 */
```

Just so that is explicit that it should not be used

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -39,35 +44,66 @@ function createInitialState(): SecurityState {
...
+  passcode: string;
+  obscureCodeEnabled?: boolean;
+  failedAttempts?: number;
+  lockUntil?: number;
+};
+
+const migrate = async (
```
ErikSin, line 61: Migration being async is going to cause another race condition.

So if the user has v0, the migration is going to run. Since Zustand is a synchronous store, it will just return the default state (where passcode===null). And then run the migration, and then [hydrate the store with the updated has passcode](https://zustand.docs.pmnd.rs/integrations/persisting-store-data#hydration-and-asynchronous-storages). This means that when the migration happens the app will read as having no passcode which means the security will be completly bypassed

We will have to [keep track of the hydration status and block the UI](https://zustand.docs.pmnd.rs/integrations/persisting-store-data#how-can-i-check-if-my-store-has-been-hydrated) until hydration has been complete

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -39,35 +44,66 @@ function createInitialState(): SecurityState {
...
+  persistedState: unknown,
+  version: number,
+): Promise<SecurityState> => {
+  if (version !== 0) {
```
ErikSin: can we be a little more explicit here and do `if (version === 1)` instead

ErikSin (changes requested):
Unfortunately the async migration is going cause a security risk (where the first migration will bypass security). We have to add blocking ui to wait for the hydration to complete

I think the `AuthContext` could check for hydration status, and return a loader until the store has been hydrated

src/frontend/screens/AuthScreen.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/e3333d86fa5366b6f24df88f9d2b035fb09bafe0

src/frontend/contexts/SecurityStoreContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fc6cf5a166309d411168a49357b60f84d4876eab

src/frontend/contexts/SecurityStoreContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fc6cf5a166309d411168a49357b60f84d4876eab

src/frontend/contexts/SecurityStoreContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fc6cf5a166309d411168a49357b60f84d4876eab

cimigree:
> Unfortunately the async migration is going cause a security risk (where the first migration will bypass security). We have to add blocking ui to wait for the hydration to complete
> 
> I think the `AuthContext` could check for hydration status, and return a loader until the store has been hydrated

https://github.com/digidem/comapeo-mobile/pull/1280/commits/6822e65a87efd3cfa2be07916b74625dd5413ee4 & https://github.com/digidem/comapeo-mobile/pull/1280/commits/4497ce692eb9e17a24cac858074bae86b01cab76

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -25,7 +30,10 @@ const SecurityStateSchema = v.variant('passcode', [
   }),
 ]);
 
-export type SecurityState = v.InferOutput<typeof SecurityStateSchema>;
+export type SecurityState = v.InferOutput<typeof SecurityStateSchema> & {
+  _hasHydrated: boolean;
+  setHasHydrated: (hydrated: boolean) => void;
+};
```
ErikSin: What is the reasoning for not adding `_hasHydrated` to the schema? And what the reasoing behind adding `setHasHydrated` to the persisted state instead of just making it an action like all the other functions?

src/frontend/contexts/AuthContext.tsx
```diff
@@ -31,7 +32,12 @@ export const useAuthContext = () => {
 };
 
 export const AuthProvider = ({children}: {children: React.ReactNode}) => {
-  const {passcode, obscureCodeEnabled, lockUntil} = useSecurityState();
+  const {
+    passcode,
+    obscureCodeEnabled,
+    lockUntil,
+    _hasHydrated: hasHydrated,
+  } = useSecurityState();
```
ErikSin: it would be a good idea to use [useShallow](https://zustand.docs.pmnd.rs/guides/prevent-rerenders-with-use-shallow) here to prevent unnecessary re-renders

src/frontend/contexts/AuthContext.tsx
```diff
@@ -3,18 +3,19 @@ import {AppState, AppStateStatus, NativeModules} from 'react-native';
...
 import {useSecurityState, useSecurityActions} from './SecurityStoreContext';
 import {useIsAudioPermissionModalOpen} from '../hooks/useAudioPermissionTracker';
 import {getLockoutThreshold} from '../lib/security';
+import {Loading} from '../sharedComponents/Loading';
 
 export type AuthState = 'unauthenticated' | 'authenticated' | 'obscured';
 
 type AuthContextType = {
   authenticate: (
     passcodeValue: string | null,
```
ErikSin: i don't think `null` is actually ever being used (and doesn't really make sense), I think we should make `passcodeValue:string`

src/frontend/contexts/AuthContext.tsx
```diff
@@ -117,6 +132,8 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
   );
 
   return (
-    <AuthContext.Provider value={contextValue}>{children}</AuthContext.Provider>
+    <AuthContext.Provider value={contextValue}>
+      {!hasHydrated ? <Loading /> : children}
```
ErikSin: Sorry, I realize that this `hasHydrated`/Loading state logic should actually live in the security store context. We dont want ANY of the security store to be accessible until it is hydrated.

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
```diff
@@ -37,47 +38,96 @@ const m = defineMessages({
...
+          {<ErrorIcon width={60} height={60} color={RED} />}
+        </View>
+
+        <HeaderText variant="header2" style={styles.title}>
+          {t(m.title)}
+        </HeaderText>
+        {loading ? (
+          <View style={styles.loading}>
+            <UIActivityIndicator size={32} />
+          </View>
```
ErikSin, line 83: this loader should replace the buttons not the text. By replacing the buttons we are blocking the user from pressing the button and causing the functions to run more than once

src/frontend/contexts/AuthContext.tsx
```diff
@@ -73,21 +78,30 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
   }, [passcode, shareDialogIsOpen, isAudioPermissionModalOpen]);
 
   const authenticate: AuthContextType['authenticate'] = React.useCallback(
```
ErikSin, line 78: I think because you created the `verifyPasscode` function we should get rid of the `validateOnly` prop in this function. And anywhere we need to `validateOnly` function we should just use `verifyPasscode`

src/frontend/screens/AppPasscode/EnterPassToTurnOff.tsx
```diff
@@ -41,14 +41,15 @@ export const EnterPassToTurnOff: NativeNavigationComponent<
...
-      return;
-    }
-    navigate('DisablePasscode');
+  function validate(passcode: string): void {
+    authenticate(passcode, true).then(success => {
+      if (!success) {
+        setError(true);
+        return;
+      }
+      navigation.navigate('DisablePasscode');
```
ErikSin: we should use `verifyPasscode` here instead

ErikSin (changes requested):
Upon further reflection we should be putting the hasHydrated loader instead the Security Store Provider, not the Auth Provider

src/frontend/screens/AppPasscode/EnterPassToTurnOff.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/screens/AppPasscode/ConfirmPasscodeSheet.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/contexts/AuthContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

src/frontend/contexts/SecurityStoreContext.tsx
cimigree: I didn’t add _hasHydrated to the schema because I think it is and should be ephemeral and we don’t want it persisted – it gets reset on each app load and it should, I think. Same reason I initially put setHasHydrated next to it in the store state (kind of like other ephemeral state values). You had suggested moving setHasHydrated to be a proper action, which makes sense, so I’ve done that.

https://github.com/digidem/comapeo-mobile/pull/1280/commits/fe300c33d4abe977505437451db935bb3926fb78

cimigree:
@ErikSin To get the tests to pass, I had to pass hydrated as true when persist is false so that the screens of the rest of the app would show up. And I had to make persist false for the helper that the exchange test uses. Otherwise they won't pass and the screens don't show up.

src/frontend/contexts/SecurityStoreContext.tsx
ErikSin: RIght now it is still being set to persisted state. [This line here](https://github.com/digidem/comapeo-mobile/blob/60f799538ea41bd488cb4879ecb2e8fbd02877cf/src/frontend/contexts/SecurityStoreContext.tsx#L100) you are setting the base state inside `createPersistedState`. If you don't want it in persisted state i think you would have to create a completly different store.

I think having it persisted is actually fine. When migration starts it ALWAYS returns the initial default state, then when the migration completes it hydrates the store and returns the migrated state. The initial default state of `_hasHydrated` is false. So it will ALWAYS return false until the migration completes

src/frontend/contexts/AuthContext.tsx
```diff
@@ -73,21 +76,24 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
-      if (validateOnly) return passcodeValue === passcode;
-
+    async passcodeValue => {
       const isLockedOut = Date.now() < lockUntil;
       if (isLockedOut) {
         throw new Error('LOCKED_OUT');
       }
 
+      const isCorrect =
+        passcodeValue && passcode
```
ErikSin: we do not need to check for `passcodeValue` anymore

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -25,7 +31,9 @@ const SecurityStateSchema = v.variant('passcode', [
   }),
 ]);
 
-export type SecurityState = v.InferOutput<typeof SecurityStateSchema>;
+export type SecurityState = v.InferOutput<typeof SecurityStateSchema> & {
+  _hasHydrated: boolean;
+};
```
ErikSin: We should include _hasHydrated in the schema. The Schema is simply use for typing, and does not effect whether it is persisted or not

src/frontend/lib/security.ts
```diff
@@ -22,3 +30,39 @@ export function getLockoutThreshold(attempts: number): number | null {
...
+
+export async function hashPasscode({
+  passcode,
+  salt,
+}: {
+  passcode: string;
+  salt: string;
+}): Promise<string> {
+  const toHash = `${salt}:${passcode}`;
+  const hash = await Crypto.digestStringAsync(
```
RangerMauve, line 47: This is sadly insecure, you need to do several iterations for safety at this point since hash generation can do billions of iterations per second. Let's do [13 for now](https://dev.to/mochafreddo/finding-the-optimal-bcrypt-rounds-for-your-machine-2kpf). The salt is also too small, lets do 10 bytes. The salt should be applied at each round of hashing.

Try https://github.com/emiraydin/react-native-bcrypt or maybe adapt their algo: https://github.com/emiraydin/react-native-bcrypt/blob/master/src/bcrypt/impl.js#L387

RangerMauve (changes requested):
More info on password hashing: https://stackoverflow.com/a/20659101

Would you like help with this bit?

src/frontend/contexts/SecurityStoreContext.tsx
```diff
@@ -36,38 +44,84 @@ function createInitialState(): SecurityState {
...
+
 export function createSecurityStore({persist} = {persist: false}) {
   let store: StoreApi<SecurityState>;
 
+  const baseState: SecurityState = {
+    ...createInitialState(),
+    _hasHydrated: !persist,
+  };
+
```
ErikSin: instead of doing this can we just explictly set _hasHydrated to true in the non persisted store and leave a comment as to why:

```ts
if (persist) {
    store = createStore(
      createPersistedState(createInitialState(), {...config option}
    );
  } else {
    // We need to set hasHydrated to true here as there is no async/hydration needed for non-persisted state (you can probably word this better than me)
    store = createStore(() => {...createInitialState(), _hasHydrated:true});
  }
```

src/frontend/lib/security.ts
RangerMauve: Actually, sorry before we get into this how big are the passwords?

cimigree: 5 digits

ErikSin: @RangerMauve I see your concern, and I think it makes alot of sense. And I think we need to consider the likelihood of this kind of threat when balancing how much dev work should go into this. The likehood that some government agent or extractive company agent is going to actually take the phone, get root access, extract the encrpyrted passcode, and then unencrypt is pretty low. And at that point I think we have bigger problems to worry about. 

So I think this is not necessary at this point, and the low level encryption, plus our timeout when using a passcode is good enough for now. And if we do want to change the encrpytion at a later point, we can do it as a follow up

ErikSin (approved):
Oops sorry, i meant to approve it earlier

RangerMauve (approved):
Given pin size the current hashing is good enough to prevent timing attacks.

PR #1281, opened 2025-07-16
fix: e2e tests long reboot and timezone

cimigree:
If the adding of a remote server fails, we reboot the app. This PR
Adds browser wait after rebooting app after adding remote server fails.
Increases browser wait time after rebooting app after removing a remote server fails.
Gets device timezone to check for date when adding a remote archive.


ErikSin (approved):

tests/e2e/specs/remote-archive/add-remote-success-and-removal.test.ts
```diff
@@ -1,13 +1,17 @@
...
 import {output} from '../../utils/naming';
 
 let archiveAdded = false;
 let cleanupNeeded = false;
+let deviceTimezone: string;
 
 describe('Remote Archive - Add Success Flow', () => {
+  before(async () => {
+    deviceTimezone = await getDeviceTimezone();
+  });
```
gmaclennan, line 14: Since all these tests run on the same device, you only need to get the timezone once at the start of this describe block. A minor optimization, but I know appium can be slow!

cimigree: I am very confused. I thought that is what I was doing. The before hook in Mocha actually runs once before all tests in the describe block,  I thought, and as you are suggesting. https://mochajs.org/#hooks

PR #1283, opened 2025-07-21
chore: user zustand store for server state

gmaclennan:
This is a non-urgent follow-up to #1279 to switch the implementation to use a zustand store. The functionality should be the same, but this reduces some boilerplate and matches our existing use of Zustand for stores.


cimigree (approved):
I notice this store doesn't use a context wrapper...
Is that because:
Testing this would require mocking the Node.js channel anyway?
Or should this eventually be wrapped in context for consistency?

But in the meantime, nice work reducing boilerplate!

gmaclennan:
It doesn't use context because it's only used by one component (`<ServerLoading>`) so there's no need to pass it via context, because it can be passed directly to that component. This PR doesn't change any of that, it just changes how the store is implemented (switching from a custom useSyncExternalStore() to a zustand store)

PR #1295, opened 2025-07-24
chore: turn on Sentry tracing for non-production environments

gmaclennan:
This PR ensures that Sentry tracing is not turned on in production. Once app metrics opt-in is added, we can also turn it on when the user opts-in.

This PR also adds:

- User Interaction Tracing, which adds traces for when the user touches the screen. In most places in the app, a user interaction results in a navigation change, and Sentry by default will only show the navigation transaction, rather than the user interaction, but the touch event will be recorded as a breadcrumb.
- React Navigation integration, which adds transactions for navigation events.

The app start trace is not working at the moment. I suspect is has something to do with how we have implemented our react navigation stack, where we [create a screen group](https://github.com/digidem/comapeo-mobile/blob/a0bb5a99c6a1ee68143874a14315bdcaaa88bf63/src/frontend/Navigation/Stack/index.tsx#L53-L56), since this will return different elements on each render, and I think that is confusing Sentry. It could also be related to how our providers will conditionally render children, which could also result in navigation re-mounts, confusing the Sentry app start metrics. Anyhow, we can look at fixing this particular issue in the future.

This integration is a pre-requisite for setting up distributed tracing of mapeo-core interaction, which I will add in a follow-up / stacked PR.


src/frontend/AppNavigator.tsx
```diff
@@ -1,18 +1,38 @@
...
+  permissionAsked,
+  navigationIntegration,
+}: {
+  permissionAsked: boolean;
+  navigationIntegration:
+    | ReturnType<(typeof Sentry)['reactNavigationIntegration']>
+    | undefined;
+}) => {
+  const containerRef =
+    React.useRef<NavigationContainerRef<AppStackParamsList>>(null);
```
ErikSin, line 23: should we just use the [useNavigationContainerRef](reactnavigation.org/docs/navigation-container#ref)?

src/frontend/App.tsx
```diff
@@ -46,16 +46,30 @@ if (applicationId?.endsWith('.dev') || applicationId?.endsWith('.pre')) {
...
   initialScope: {user: {id: sentryUserId}},
 });
 
+if (appMetricsOptIn) {
+  Sentry.setTag('appMetricsOptIn', 'true');
+  navigationIntegration = Sentry.reactNavigationIntegration({
+    enableTimeToInitialDisplay: true,
+    ignoreEmptyBackNavigationTransactions: false,
+  });
+  Sentry.getClient()?.addIntegration(navigationIntegration);
```
ErikSin, line 70: What is this line for? I don't see it in the [documentation for react navigation setup](https://docs.sentry.io/platforms/react-native/tracing/instrumentation/react-navigation/)

ErikSin (approved):
1 non blocking comment and 1 question!

src/frontend/App.tsx
gmaclennan: Under the [Initialization](https://docs.sentry.io/platforms/react-native/tracing/instrumentation/react-navigation/#initialization) section, it's shown under a `Sentry.init()` option `integrations`. I used the `addIntegration()` function because the code was easier to read than conditionally defining an `integrations` array (particularly getting the typescript types correct). This is [documented here](https://docs.sentry.io/platforms/react-native/integrations/plugin/#adding-an-integration), but the documented method doesn't currently work, but [will be fixed in the next Sentry SDK release](https://github.com/getsentry/sentry-react-native/issues/5017).

src/frontend/AppNavigator.tsx
gmaclennan: Good catch, yes. I was following some docs from somewhere but not sure where, and this function is better. I'll update.

gmaclennan: Ah yes, I remember: I was confused about the difference between `createNavigationContainerRef()` and `useNavigationContainerRef()`. The [docs are very vague about this](https://reactnavigation.org/docs/typescript/#annotating-ref-on-navigationcontainer), and the [source code](https://github.com/react-navigation/react-navigation/blob/8127603ec7e60a6c1a907cc51a497868a6925de0/packages/core/src/useNavigationContainerRef.tsx#L6) is not clear to me. That's why the safest option felt to use a regular ref - then I know how it works.

PR #1296, opened 2025-07-24
chore: Add Sentry distributed tracing for RPC calls

gmaclennan:
Adds [Sentry distributed tracing](https://docs.sentry.io/platforms/react-native/tracing/trace-propagation/#what-is-distributed-tracing) for RPC calls to the backend. This gives visibility to performance of backend API calls, and provides valuable information to help debug errors.

~This currently uses an unreleased build of rpc-reflector from https://github.com/digidem/rpc-reflector/pull/25, the goal being to evaluate the new hook API. This currently also uses an unreleased build of @comapeo/ipc based on the [`update-rpc-reflector`](https://github.com/digidem/comapeo-ipc/tree/chore/update-rpc-reflector) branch. The changes to exports, so that client and server are separated, are important for avoiding bundle issues in react native when Metro tries to bundle node-specific deps from the server code.~

This does seem to work and produces [traces in Sentry](https://awana-digital.sentry.io/explore/traces/trace/ff970f37cc0a428d8102f0b4b81ee441/?node=span-beed74cb48be4d12&project=4507170965618688&source=traces&statsPeriod=30d&targetId=8cec5ce3c5e359cf&timestamp=1753307097) which include both frontend and backend spans.

~One implementation question I have: I think it makes send to move `createMapeoClient()` to a separate file now that it is a lot of code. Where should I put this file?~

This PR updates @comapeo/core, @comapeo/core-react and @comapeo/ipc in order to correctly integrate the changes in rpc-reflector, namely the change to an ESM module instead of CJS.

Several patches to rpc-reflector and comapeo/ipc are no longer needed because these have been fixed upstream.

Also, this adds the metro resolver [config option `unstable_enablePackageExports`](https://metrobundler.dev/docs/configuration/#unstable_enablepackageexports-experimental), which allows Metro to follow the `package.json` `"exports"` field for `@comapeo/ipc` and `rpc-reflector`, which ensures that Metro does not try to bundle backend code. This option defaults to `true` anyway for `metro@0.82.0` (we are on `v0.81.0`).


ErikSin:
I dont have any strong reasoning for this, but I think we can put `createMapeoClient` in it's own file in the `lib` folder

src/frontend/App.tsx
```diff
@@ -92,11 +91,11 @@ const appDiagnosticMetrics = new AppDiagnosticMetrics({
   },
 });
 
+export function noop() {}
```
RangerMauve: Is this the best place to define it? Feels more appropriate in some sort of utils folder/file

RangerMauve (approved):
Minor nit pick but LGTM

src/frontend/App.tsx
gmaclennan: Good catch, this was left-over from quick prototyping.

ErikSin (approved):
Looks good!

ErikSin:
Im going to merge this so i can build the next RC

PR #1304, opened 2025-07-24
chore: remove useOnBackgroundedForegrounded() hook

gmaclennan:
This hook is used to communicate with the backend when the app goes into the foreground or background. However, the backend already has an API to detect app foreground and background, which we already use. This change removes frontend code and reduces traffic over the bridge. Behaviour should remain the same.


achou11 (approved):
Changes make sense but probably could benefit from a manual test. From what I understand, not sure if e2e passing is enough of a confirmation, since this affects how sync is started and shut down.

achou11:
Just did a manual test setting up two devices, creating a project, and exchanging data. occasionally closed/suspended the app to test whether any notable behavior changed based on this PR's changes. Didn't catch anything though.

PR #764, opened 2024-09-30
fix: use more modern Zeroconf backend

EvanHahn:

PR #1306, opened 2025-07-28
fix: fixes and improvements to network service discovery

gmaclennan:
This PR implements some fixes to our network service discovery implementation (e.g. mDNS discovery):

- The function to stop publishing a service was listening to the incorrect event (`remove`, which is when a discovered service is removed). This is the cause of errors seen in Sentry for discovery timeouts. Additionally, the timeouts never worked correctly, because the app stops publishing as it goes into the background, and react native pauses all timers when the app is in the background. I don't think any of this is actually the cause of any bugs, but I am unsure.
- The underlying react-native-zeroconf library does not have any handlers for errors when registering or unregistering a service (e.g. publishing/unpublishing a service), which would mean that any error would just result in the functions to publish and unpublish would never resolve (or timeout), and we would not capture any error. This PR adds rudimentary events for these errors. It's not the neatest fix, but it works without major changes to the library code (implemented with a patch).
- react-native-zeroconf uses an alternative MDNS implementation, which [we tried](https://github.com/digidem/comapeo-mobile/pull/764) but it does not work, and it is also now unmaintained. This PR removes that implementation to reduce APK size and also remove possible errors that might have been introduced by keeping this package around.

None of this directly fixes any known bugs, but addresses several places where bugs might happen, and introduces a way to capture errors with service registration and unregistration.


src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -403,13 +401,20 @@ function unpublishZeroconf(
...
   name: string;
 } {
   const address = addresses[0];
+  if (!address) return null;
+  // The service name can include a suffix ' (1)' when there is a conflict,
+  // which could happen when the device is still unpublishing the current
+  // service when it tries to start publishing again, e.g. after the user leaves
+  // the app and returns. Because the service name is the peer name, we need to
+  // remove the suffix when connecting to the peer.
+  const name = serviceName.replace(/[^a-fA-F0-9].*/g, '');
```
RangerMauve, line 417: This is a really good catch 👍

RangerMauve (approved):
LGTM. Did you try testing the app locally to make sure discovery still worked?

gmaclennan:
> LGTM. Did you try testing the app locally to make sure discovery still worked?

Not yet, I was only testing in the emulator. I'll charge my two androids tonight and test tomorrow - that's why I left it draft for now.

app.json
```diff
@@ -110,7 +110,8 @@
         "android.permission.FOREGROUND_SERVICE",
         "android.permission.FOREGROUND_SERVICE_LOCATION",
         "android.permission.CAMERA",
-        "android.permission.RECORD_AUDIO"
+        "android.permission.RECORD_AUDIO",
+        "android.permission.CHANGE_WIFI_MULTICAST_STATE"
```
RangerMauve, line 114: I was getting this when trying to run the app: 

```
 Possible Unhandled Promise Rejection (id: 0):
Error: Exception During Scan: WifiService: Neither user 10422 nor current process has android.permission.CHANGE_WIFI_MULTICAST_STATE.
```

RangerMauve:
Tested locally and discovery is not working. Backend is starting the server but no candidates are going through. Going to dig into it more to see where we have the hiccup

RangerMauve:
Phones are advertising on local net (can see them from avahi on my main machine) but they aren't getting candidates from the react-native side to the backend. Drilling to see where

RangerMauve:
Can you describe what you changed in `node_modules/react-native-zeroconf/dist/index.js`? 

I'm not seeing peers seeing each other at the Android level so it might also be an issue with discovery scanning not running.

gmaclennan:
> Can you describe what you changed in `node_modules/react-native-zeroconf/dist/index.js`?

Here: https://github.com/gmaclennan/react-native-zeroconf/pull/1/files

I added event listeners for the new `RNZeroconfServiceRegisterError` and `RNZeroconfServiceUnregisterError` events that I created in the java code, for handling failures when publishing or unpublishing (these were previously just ignored).

gmaclennan:
@RangerMauve when I test this on my two devices (Samsung A13 and Moto4), it works without issue. Are you sure you did not do what I did initially, which is start with devices I had used before, and were already on the same project, and tested this from the invite screen, where I did not see the other device because it was already a member of the project?

RangerMauve:
Doing more tests with this to confirm. Will document exact flow.

ErikSin:
I tested this on my devices and discovery and invites are both working as expected. I tried several invites. I have 2 phones and tried where each phone is both the invitee and invitor

RangerMauve:
YEah can confirm I wasn't able to reproduce the issues I had yesterday accross my devices.

PR #1310, opened 2025-07-30
fix: don't debug log in production

### LINKED ISSUE #1309: Slowdown from excessive logging in the backend
In testing locally, I am seeing significant slowdown of the app in v5 internal builds compared with v3. I believe this is due to debug logging being turned on for the backend. This seems to be particularly the case for indexing: currently indexing happens in the main thread and is synchronous, and the debug logging logs a line for each record index. For the default presets, which is generally always present, there are 100s of translation records to index, which seems to lock the device up.

Debug logging is [always turned on in the backend](https://github.com/digidem/comapeo-mobile/blob/abc3a1329da7fead47ddc559f04256e863e222b9/src/backend/src/app.js#L33), which logs huge amounts of data to the device adb log, whether or not Sentry logging is enabled.

I think the short-term solution is to not turn on logging with `debug.enable()` in the backend in production, and not turn on Sentry logging in production. We will still have the issue in QA, and I think some of the excessive timeouts we are seeing is because the event loop is blocked during indexing and initial project sync, partly because of the logging overhead.

The medium-term solution is to temporarily turn off debug logs for indexing and sync state updates, because these are excessive. We can comment them out and turn them on when needed for local debugging.

The medium-long-term solution is to replace the `debug` library we currently use for debugging with a logger that supports log levels that can be turned on and off, and switch to a more performant logger such as pino, which logs in a separate worker thread to avoid blocking the main thread.

gmaclennan:
Fixes #1309

### DIFF
--- a/src/backend/loader.js
+++ b/src/backend/loader.js
@@ -8,6 +8,7 @@ import path from 'path'
 import parseArgs from './src/args.js'
 import { createRequire } from 'module'
 import debug from 'debug'
+import { formatWithOptions } from 'node:util'
 const require = createRequire(import.meta.url)
 /** @type {import('./types/rn-bridge.js')} */
 const rnBridge = require('rn-bridge')
--- a/src/backend/loader.js
+++ b/src/backend/loader.js
@@ -7,6 +7,7 @@ import os from 'os'
 import path from 'path'
 import parseArgs from './src/args.js'
 import { createRequire } from 'module'
+import debug from 'debug'
 const require = createRequire(import.meta.url)
 /** @type {import('./types/rn-bridge.js')} */
 const rnBridge = require('rn-bridge')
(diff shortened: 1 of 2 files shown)


src/backend/loader.js
```diff
@@ -23,14 +24,23 @@ const initialScope = sentryUserId ? { user: { id: sentryUserId } } : undefined
...
 const logLevels = ['error']
-
 let enableLogs = false
 
 if (sentryEnvironment !== 'production') {
   logLevels.push('log', 'warn')
   enableLogs = true
 }
 
+if (enableLogs) {
```
RangerMauve, line 35: Is `loader.js` running in the same bundle as the MapeoManager? If not we should do this update inside `src/backend/index.js`. Else it looks good to me,

RangerMauve (approved):

src/backend/loader.js
gmaclennan: They are separate bundles, however rollup is also splitting the bundles for shared code like `debug`, so it's the same debug import being used by both the loader.js bundle and the app.js bundle. I verified this by checking that debug logging is still enabled in my dev build, and by also checking `debug.enabled()`.

RangerMauve (approved):

PR #1311, opened 2025-07-30
fix: Do not stop the discovery server

gmaclennan:
Stopping the discovery server attempts to close the `node:net` server in the backend. Closing a node server awaits any active connections, and currently we never close peer connections, so if a device has any active incoming peer connections, the server will never actually close. Also, server close is non-cancellable, so there is no way to start it again. The short-term solution to this is to never try to close the discovery server.

This is an initial fix for https://github.com/digidem/comapeo-core/issues/1075


RangerMauve (approved):
👍👍

gmaclennan:
I merged without the e2e tests passing because I confirmed that they did pass for a [previous commit](https://github.com/digidem/comapeo-mobile/actions/runs/16650805596/job/47123950126), and since that commit the only change was removing an unused var so that eslint passes. The e2e test failure looks like a flakey test related to the remote archive and unrelated to this PR.

PR #1312, opened 2025-07-30
feat: app usage stats

### LINKED ISSUE #1284: App Usage Stats /  Interstitial
**Description**
App Usage Statistics: how an individual user is using the app, eg how long a user had the app open, what screens and features they use (eg number of shares), but not information about data collected (because this is always in the context of a project which other participants must know about). This data can be very detailed eg which screens a user is visiting.
Goal :  integrate a passive opt-in for a fixed period of time to track app usage behavior

**Feature spec**
https://www.notion.so/digidem/Data-Metrics-App-Usage-Statistics-1f11b08162d580bb9cd8cb757f864ac3

**Interactive prototype**
https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6296-3475&t=N69UmW1JW4FtjHC9-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6296%3A3475&show-proto-sidebar=1

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6861-3560&t=qDenZB70eMbSHKWm-0

**To Do**

Image

- [ ] Add screen  7 days after first app open and completion of onboarding screens
- [ ] Integrate screen as designed, see figma for latest version
- [ ] Tapping "No, not now" closes the screen and lands the user where they were before the screen appeared
- [ ] If tapped "No, not now" , display screen after 3 months and up to 3 times.
- [ ] Tapping "Yes, count me in" go to Success screen below
- [ ] if tapped  "Yes, count me in" gather data for up to 12 months at which point the feature gets turned off

Image

- [ ] Integrate screen as designed, see figma for latest version
- [ ] Tapping "Close" closes the screen and lands the user where they were before the screen appeared

cimigree:
closes #1284 

### Summary

This PR introduces a new  screen that prompts users to share how they use CoMapeo. The screen is conditionally shown based on onboarding and prompt timing logic, and is meant to support opt-in usage data collection.

### Key Features

* **New Screen**: `AppUsagePromptInterstitial`

  * Displays a privacy-first usage opt-in prompt
  * Automatically appears 7 days after onboarding and then again after 3 months if not opted in previously
  * Respects user dismissal and tracks `promptCount` and `lastPromptAt` state

* **State Management**:

  * Backed by `useAppUsageStatsPromptStore` (Zustand)
  * Logic encapsulated in `shouldShowAppUsagePrompt(...)`

* **Navigation Integration**:

  * Logic lives in the RootStack for when to show (or not show) the screen
  * Automatically skips the screen if the prompt should not be shown

* **Testing**:

  * Unit tests for logic
  * Tests for Context
  * My first integration test! That was a bear!


src/frontend/contexts/AppUsageStatsPromptContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+  persist as createPersistedState,
+} from 'zustand/middleware';
+import {MMKVZustandStorage} from '../hooks/persistedState/createPersistedState';
+
+export interface AppUsageStatsPromptState {
+  optedIn: boolean | null; // null = never asked, true = opted in, false = opted out
+  completedOnboardingAt: number | null;
+  lastPromptAt: number | null;
+  promptCount: number;
+}
```
ErikSin: We should use valibot as it makes migrations much simpler

src/frontend/screens/AppUsagePromptInterstitial.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const {setOptedIn} = useAppUsageStatsPromptActions();
+  const showPrompt = shouldShowAppUsagePrompt(
+    useAppUsageStatsPromptState(s => s),
+  );
+
+  React.useEffect(() => {
+    if (!showPrompt) {
+      navigation.replace('Home', {screen: 'Map'});
+    }
+  }, [showPrompt, navigation]);
```
ErikSin: when screens are being conditionally rendered, [we do not want to manually navigate](https://reactnavigation.org/docs/auth-flow/#dont-manually-navigate-when-conditionally-rendering-screens)

src/frontend/screens/AppUsageSharingSuccess.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const {data} = useOwnDeviceInfo();
+  const deviceName = data?.name;
+  const {setOptedIn} = useAppUsageStatsPromptActions();
+  const {optedIn} = useAppUsageStatsPromptState(s => s);
+
+  React.useEffect(() => {
+    if (optedIn) {
+      navigation.replace('Home', {screen: 'Map'});
+    }
+  }, [optedIn, navigation]);
```
ErikSin: Same comment as above, dont navigate when conditionally rendering

src/frontend/screens/AppUsagePrompt.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+              options={{headerShown: false}}
+            />
+            <RootStack.Screen
+              name="Home"
+              component={HomeTabs}
+              options={{headerShown: false}}
+            />
+          </RootStack.Navigator>
+        </AppUsageStatsPromptProvider>
+      </NavigationContainer>,
```
ErikSin: Your tests are not properly testing the actual use case. In the actual nav stack, we are conditionally rendering these screens based on the `AppUsageStat` state . Here you are ALWAYS rendering the screens. I think the best thing would be to mount the entire AppStack and set the `AppUsageStatsPromptProvider` state (similar to how your doing it, but [directly using the hooks](https://github.com/testing-library/react-hooks-testing-library) in your tests.

ErikSin (changes requested):
We are missing the use case where the use upgrades the app and already has done onboarding. If that's the case, `completedOnboardingAt` will always be null.

The suggest fix is to run a useEffect. In that useEffect check for deviceName. And if there is a deviceName AND completedOnboardingAt is null, set it completedOnboardingAt to Date.now() - So it will run 7 days after the user has upgraded

src/frontend/screens/AppUsagePrompt.test.tsx
ErikSin: I think skip these test for now, and lets create a ticket to address it later. I think i can update the testing environment to make this easier in the future. I ran some experiments and it just not working the way they should be

src/frontend/contexts/AppUsageStatsPromptContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1312/commits/afcecf5fc76487ff038257a91009544d94cf1b5a

src/frontend/screens/AppUsagePromptInterstitial.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1312/commits/afcecf5fc76487ff038257a91009544d94cf1b5a

src/frontend/screens/AppUsageSharingSuccess.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1312/commits/afcecf5fc76487ff038257a91009544d94cf1b5a

src/frontend/screens/AppUsagePrompt.test.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1312/commits/afcecf5fc76487ff038257a91009544d94cf1b5a

cimigree:
I added the useEffect in the RootStack. That file will need some adjustment when the other PR gets integrated...
I realized that I missed a requirement. This one: 

> if tapped "Yes, count me in" gather data for up to 12 months at which point the feature gets turned off

So I added an opted in date (and expiry date) to the storage. Not sure if I should have included the expiry date. I did for now but can see calculating that on the fly. I wasn't sure what was the best way.

ErikSin:
Can you update from develop and take out these screens from navigation (aka make them inaccessible to the user)? I am going to use these screen eventually, but instead of maintaing this pr, lets just merge it and not make it accessible to the user but allow us to add these screens really easily when needed

ErikSin:
Can we also take out all uses of the AppUsagePromtContext. Again Im going to use it eventually, but for now lets just hide until i can properly connect it to posthog

cimigree:
@ErikSin , I think I removed all usages as requested. What do you think?

ErikSin (approved):
Thank you

PR #1315, opened 2025-08-04
tests: Add timeout for remote archive invalid url

RangerMauve:
The CI errors related to remote archive seem to be from the connection attempt to `example.com` taking too long to error out.

(IIUC) this should help avoid the false negative. Testing locally it just needed like 2-5 seconds to show the error.

### DIFF
--- a/tests/e2e/specs/remote-archive/add-remote-error.test.ts
+++ b/tests/e2e/specs/remote-archive/add-remote-error.test.ts
@@ -21,7 +21,10 @@ describe('Remote Archive - Add Error Flow', () => {
     const urlInput = await $(byResourceId('RA.url-inp'));
     await urlInput.setValue('example.com');
     await $(byResourceId('OBS.edit-save-btn')).click();
-    await expect($(byTextMatches('example.com'))).toBeDisplayed();
+    await $(byTextMatches('example.com')).waitForExist({
+      timeout: 10000,
+      reverse: false,
+    });
 
     await expect($(byTextMatches('Invalid URL'))).toBeDisplayed();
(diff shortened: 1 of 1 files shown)


gmaclennan:
Some follow-up issues here I think, @RangerMauve can you open the core-related issues.

1. I don't think is actually an invalid URL error (e.g. `URL.canParse(url) === false`) - I think we are making a request to the given URL/server and validating if it is a correctly configured remote archive server. An error here could come from either URL being correct and the server being broken, or the URL being incorrect (but valid) and not pointing to a remote archive server.
2. We should add some kind of initial check for a remote archive server, like checking a header or a public endpoint that returns known data (do we already have one?). Currently we just try to send a `PUT` that includes all the project encryption keys - if a user mis-types a URL then this information is sent to the wrong server.
3. For the tests we should use a domain we control, since we are actually making requests to the server for tests. We can set up a simple cloudflare worker to return a predetermined status code and body.

RangerMauve:
@gmaclennan Can confirm it's not an invalid URL error but an error trying to connect to a valid domain that isn't comapeo-cloud. In my local testing I found that using an actual invalid domain `examplecom` errore out immediately. Should we maybe switch this test to that instead? I don't think the current test is providing much over the existing comapeo-core tests

gmaclennan:
Yes I suggest we change this test to what is says, a test for an invalid URL. It could be worth adding an e2e test for adding a URL to a server that is not a comapeo remote archive server, but as Mauve says, that's covered in backend tests which should be enough for now. It will be useful to have an e2e test here for an invalid server once we have an error screen for that state.

RangerMauve:
I'll make a new PR

PR #1319, opened 2025-08-05
chore: Github workflow to update core

gmaclennan:
A quick workflow to automate updating core. We can trigger this from the Optic release workflow in core whenever a new release is published. This will avoid missing releases of core in mobile, and avoid human error like forgetting `--save-exact` or forgetting to update both the front-end and back-end deps.


.github/workflows/update-core.yml
```diff
@@ -0,0 +1,110 @@
...
+          else
+            echo "No update needed. Current version: $current_version, Latest version: $latest_version"
+            exit 1
+          fi
+        env:
+          current_version: ${{ steps.current-version.outputs.current_version }}
+          latest_version: ${{ steps.latest-version.outputs.latest_version }}
+
+      - name: Update @comapeo/core in frontend
+        run: npm install --save-exact @comapeo/core@latest
```
ErikSin: we should install this as a dev dependency here

.github/workflows/update-core.yml
```diff
@@ -0,0 +1,110 @@
...
+
+env:
+  BRANCH_NAME_PREFIX: 'chore/update-core-v'
+  COMMIT_MESSAGE_PREFIX: 'chore: update @comapeo/core to v'
+  PR_COMMENT_IDENTIFIER: '<!-- update-core -->'
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types: [update-core]
```
ErikSin, line 15: perhap we should comment where this post request is being made (aka the workflow in core)

ErikSin (approved):
2 non blocking comments

.github/workflows/update-core.yml
gmaclennan: good catch - this should be blocking!

PR #1321, opened 2025-08-05
chore: only run e2e tests when source-code files change

gmaclennan:
Small change to reduce unnecessary e2e tests. Not sure if it will work as intended, so pushing to test. The action used [dorny/paths-filter](https://github.com/dorny/paths-filter) is [used by Sentry](https://github.com/getsentry/sentry/blob/38e0b080b5039ab1896c5d6bb889811c65a1dcd7/.github/workflows/backend.yml#L36) so I think it's fairly reliable and maintained.

### DIFF
--- a/.github/workflows/e2e-appium-browserstack.yml
+++ b/.github/workflows/e2e-appium-browserstack.yml
@@ -13,7 +13,44 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  files-changed:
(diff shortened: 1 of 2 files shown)


gmaclennan:
Not sure it makes sense to merge this - the amount of times e2e CI runs could be minimal, and it introduces the risk of adding new files that affect builds but aren't included in the filter here.

PR #1322, opened 2025-08-06
[DO NOT MERGE] feat: save/create document

gmaclennan:
This is a quick experiment to try the [ACTION_CREATE_DOCUMENT](https://developer.android.com/training/data-storage/shared/documents-files#create-file) intent with Android to enable saving a file to a location chosen with the system file picker.

I was looking into this because, on my phones at least, "sharing" a file (e.g. via an ACTION_SEND intent) does not show the Files app as a destination. This means ACTION_SEND cannot be used to save something to phone storage.

This is a quick-and-dirty local expo module to use the documented Android way of creating a new file on disk. I could not find any existing React Native modules to do this.

This is done as a proof-of-concept to see if this is possible if we get feedback that this functionality is essential.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,8 +84,8 @@ ios/*.env*.local
 *.mobileprovision
 
 # Expo
-android/
-ios/
+/android
+/ios
 
 # WebdriverIO local configuration
 wdio.local.config.js
--- a/modules/action-create-document/android/build.gradle
+++ b/modules/action-create-document/android/build.gradle
@@ -1,0 +1,43 @@
+apply plugin: 'com.android.library'
+
+group = 'expo.modules.actioncreatedocument'
+version = '0.6.3'
+
+def expoModulesCorePlugin = new File(project(":expo-modules-core").projectDir.absolutePath, "ExpoModulesCorePlugin.gradle")
+apply from: expoModulesCorePlugin
+applyKotlinExpoModulesCorePlugin()
+useCoreDependencies()
+useExpoPublishing()
+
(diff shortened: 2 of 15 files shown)


achou11:
Curious if using https://react-native-documents.github.io/docs/sponsor-only/picker/save-as-dialog?current-os=android would fulfill the needs. It uses the same intent for Android

EDIT: ah, for some reason I only now noticed that it's available for sponsors only 😅

EDIT 2: maybe no longer sponsors-only? Seems like the URL and the actual docs are potentially misaligned 🤔

gmaclennan:
Yeah looks like it's MIT now see https://github.com/react-native-documents/document-picker/pull/741

The module looks good, the code is pretty much identical to what is in this PR with a few extra catches, and it uses turbo modules (new architecture) and the advantage is it has the iOS implementation. I think we should use it if we need this functionality, which I am pretty sure we will need. There is not much you can do with the exported data otherwise (on my testing phone Google Drive was the only app that would accept the zip file, and it required me going throough the whole sign-in process.)

achou11:
> and it uses turbo modules (new architecture) and the advantage is it has the iOS implementation.

to clarify, does it require the new architecture? seems like it supports both based on a brief skimming of docs.

gmaclennan:
Unsure. The docs are from before open sourcing and the open sourcing PR mentions a re-write, and looking at the react docs for how to support the old arch with turbo modules, I don't think it has the necessary compat code.

ErikSin:
being addressed here: https://github.com/digidem/comapeo-mobile/pull/1357

PR #1324, opened 2025-08-06
Feat/sentry backend offline transport

RangerMauve:
Adds the offline transport so we can collect envelopes offline / flush them when back online.

Once it's looking good overall I'll release 1.0 of the transport and update the deps


src/backend/loader.js
```diff
@@ -31,6 +40,8 @@ if (sentryEnvironment !== 'production') {
   enableLogs = true
 }
 
+const sentryDB = new Database(dbDir)
```
gmaclennan, line 46: just to note that the parameter for `new Database()` is the path to a file, not a dir. The database will be created at the path `${dataDir}/sentry-logs`. I suggest naming it with `.db` extension to match what we've done with other databases.

gmaclennan (approved):
If this works, then great, I think good to merge, although note my comment about the DB path. Do you have a testing plan? Maybe build and test locally with wifi off then connecting and checking sentry, or reviewing some debug messages? Would be good to check this generally works before deploying.

ErikSin:
@RangerMauve what is the status of this?

RangerMauve:
@ErikSin Need to find time to do some testing to make sure it's actually working offline. Else it should be good to go.

ErikSin:
@RangerMauve ok great. im just clearing stale PRs, so just wanted to make sure i didn't close it by mistake. I will leave it open!

RangerMauve:
Tested locally and can confirm logs are getting queued when offline and flushed when online

RangerMauve:
Gonna release the transport to NPM then update deps and merge this

PR #1325, opened 2025-08-06
feat: validated by comapeo description

### LINKED ISSUE #1286: Exposing Photo Metadata - Updated
**Description**
Photo Info now includes a Validated by CoMapeo module. 
In addition to a photo’s metadata, the Validated by CoMapeo module describes important details about a photo’s authenticity.
Whenever tapped, the Validated by CoMapeo module expands to displays key metrics about the photo’s integrity.
This includes:
- amount of time the photo was attached after the observation was created;
- the distance from the original location of observation creation; and
- the device from where the photo originated.

When tapped again, the Validated by CoMapeo module collapses to its original state.

**Feature spec**
https://www.notion.so/digidem/Exposing-Photo-Metadata-Updated-20f1b08162d5806b8cd4dacc9190a2f1

**Interactive prototype**

https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6387-2477&t=AsEEotgx18kceXse-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6387%3A2477&show-proto-sidebar=1

**To Do**

Image

- [ ] Integrate the collapsible module as designed
- [ ] Add  "Y seconds/minutes/hours after observation"
- [ ] Add " X m/km from observation"
- [ ] Add "Taken by DeviceName"
- [ ] When tapped again, the Validated by CoMapeo module collapses to its original state.

ErikSin:
closes #1286

to make this photo modal component simpler I broke it up into 2 components: `DraftPhotoPreviewModal` and `AttachedPhotoPreviewModal`.

# Screen shots:
image

image

## Draft Photo Preview Modal
This screen only deals with unsaved photos. It does not need to worry about having the `observationId` passed as a nav param. And it does not show the observation Id, the author id, nor the "validated by comapeo" drop down

## Attached Photo Preview Modal
This screen only deals with saved photos (aka saved to the db). We can now pass the `observationId` as a required param. This allows us to easily determine the metadata of the photo without have to deal with too many ternaries, which would have been the case if these 2 screen were the same as it was in the original architecture. Another notable thing was that I took out the granular  `Suspense` which wrapped the photo component (to simplify the component)

I also created the `useGetCreatedBy` hook, which allows us to get info on who created a core document

## e2e testing
I spent a long time trying to get e2e tests to work. I was able to take a photo. But selecting a photo via appium was just not possible. I am not sure why but I dont think it is a wise use of my time to continue trying


tests/e2e/specs/onboarding/setup.test.ts
ErikSin, line 1: I left this in. This can be used in the future to set up comapeo if we want to test flows in isolation and need the device to be named

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -128,23 +128,27 @@ export const Editor = ({
                   return <ThumbnailLoader size={size} key={att.draftPhotoId} />;
                 }
                 if (isDraftPhoto(att)) {
+                  console.log('is draft photo');
```
cimigree: Please remove

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+import Octicons from 'react-native-vector-icons/Octicons';
+
+const m = defineMessages({
+  validatedByCoMapeo: {
+    id: 'screens.PhotoPreviewModal.validatedByCoMapeo',
+    defaultMessage: 'Validated by CoMapeo',
+  },
+  verifiedOriginal: {
+    id: 'screens.PhotoPreviewModal.verifiedOriginal',
+    defaultMessage: 'This is a verified unaltered original',
```
cimigree: I think there is supposed to be a period at the end.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+                }}>
+                <BodyText selectable style={sharedStyles.primaryInfoText}>
+                  {formatMessage(m.verifiedOriginal)}
+                </BodyText>
+
+                {photoTimeRelativeToObs && (
+                  <InfoItem
+                    icon={
+                      <MaterialIcons
+                        name="timer"
```
cimigree: This is not exactly what was in the designs though very similar.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+                  gap: 20,
+                  padding: 20,
+                  borderTopColor: DARK_GREY,
+                  borderTopWidth: 1,
+                }}>
+                <BodyText selectable style={sharedStyles.primaryInfoText}>
+                  {formatMessage(m.verifiedOriginal)}
+                </BodyText>
+
+                {photoTimeRelativeToObs && (
```
cimigree: If the value is 0, this section won't render, right? Maybe should change this to check for null or undefined.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  const photoTimeRelativeToObs = photoCreatedAt
+    ? calcPhotoTimeRelativeToObs({
+        photoCreatedAt: photoCreatedAt,
+        observationCreatedAt: new Date(observationCreatedAt).getTime(),
+      })
+    : undefined;
+
+  const distanceFromObservation =
+    photoCoordinates && lon && lat
```
cimigree: Coordinates could be 0, which would cause the block to skip, right? Instead I recommend using typeof, like typeof lon === 'number' && typeof lat === 'number' instead

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+            title={
+              <View style={{flexDirection: 'row', gap: 12}}>
+                <Octicons
+                  name="check-circle"
+                  size={20}
+                  color={NEW_DARK_GREY}
+                  allowFontScaling
+                />
+                <BodyText selectable style={sharedStyles.primaryInfoText}>
+                  {formatMessage(m.validatedByCoMapeo)}
```
cimigree: Is this supposed to be always present and conditional? There is no photoInfo.external (as there used to be in the old photo preview modal)?

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+              size={20}
+              color={NEW_DARK_GREY}
+              allowFontScaling
+            />
+          }>
+          <BodyText
+            selectable
+            style={sharedStyles.primaryInfoText}
+            numberOfLines={1}>
+            {memberInfo.deviceId.slice(0, 15)}
```
cimigree, line 313: Could the member info be undefined and throw an error here? Maybe there should be a fallback?

src/frontend/screens/PhotoPreviewModal/helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+    description: 'Describes image layout when taking photo horizontally',
+  },
+  portrait: {
+    id: 'screens.PhotoPreviewModal.portrait',
+    defaultMessage: 'Portrait',
+    description: 'Describes image layout when taking photo vertically',
+  },
+});
+
+export function calculateDistanceFromObservation({
```
cimigree: Spec says "X m/km from observation" but this only returns meters. I suggest you separate calculation from formatting and return raw meters from here but format the result in a tiny presenter that switches meters to kilometers. And also, maybe sets tiny distances to 0 within a certain range. When I took my photos I was literally sitting in one spot, but the distance shows as .96 and 1.47 from the observation. Maybe set to 0 if the distance is below the accuracy or below say 3-5 meters (or even 2?).

src/frontend/screens/PhotoPreviewModal/helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+}: {
+  photoLocation: [number, number];
+  observationLocation: [number, number];
+}): number {
+  const photoPoint = point(photoLocation);
+  const observationPoint = point(observationLocation);
+  return distance(photoPoint, observationPoint, {units: 'meters'});
+}
+
+export function calcPhotoTimeRelativeToObs({
```
cimigree, line 41: The spec says "Y seconds/minutes/hours after observation" but this only returns minutes with two decimals. Same comment about presenter but use it to calculate longer lags. Might make more sense to return the raw data (milliseconds) and let the presenter do all the converting in the other file.

src/frontend/screens/PhotoPreviewModal/helpers.ts
```diff
@@ -0,0 +1,134 @@
...
+  photoCreatedAt,
+  observationCreatedAt,
+}: {
+  photoCreatedAt: number;
+  observationCreatedAt: number;
+}): number {
+  const diffInMilliseconds = photoCreatedAt - observationCreatedAt;
+  const min = diffInMilliseconds / 1000 / 60;
+  return parseFloat(min.toFixed(2));
```
cimigree: Should we be using intl.formatNumber here? For locale-aware decimals and formatting?

src/frontend/screens/PhotoPreviewModal/helpers.ts
```diff
@@ -0,0 +1,134 @@
+import {point} from '@turf/helpers';
```
cimigree, line 1: Seems like this file would be a good place for unit testing.

src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -89,6 +89,7 @@ function ObservationListItemInner({
           <View style={styles.smallIconContainer}>
             <PresetCircleIcon
               iconId={preset?.iconRef?.docId}
+              testID={`OBS.${preset?.name}-list-icon`}
```
cimigree, line 92: Since we are not using this are you sure you want to leave in extra props?

src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -128,23 +128,27 @@ export const Editor = ({
...
                   return (
                     <ThumbnailContainer
                       key={att.draftPhotoId}
+                      accessibilityLabel="View draft photo."
```
cimigree, line 134: Should we leave this in if not using?

src/frontend/sharedComponents/Thumbnails/PhotoThumbnail.tsx
```diff
@@ -37,7 +37,10 @@ export const SavedPhotoThumbnailImage = ({
   const image = useAttachmentUrlQuery(photo, 'thumbnail');
 
   return (
-    <ThumbnailContainer size={size} onPress={image.error ? undefined : onPress}>
+    <ThumbnailContainer
+      accessibilityLabel="View saved photo."
```
cimigree, line 41: Should we leave in the accessibility label if not using?

src/frontend/sharedComponents/Thumbnails/ThumbnailContainer.tsx
```diff
@@ -15,6 +15,7 @@ type ThumbnailContainerProps = {
   size: number;
   accessibilityLabel?: string;
   containerStyle?: ViewStyle;
+  testId?: string;
```
cimigree, line 18: Not sure if best to leave these in or remove just to limit the changes of this PR and reduce unused/ unneeded props.

src/frontend/sharedTypes/navigation.ts
```diff
@@ -42,9 +42,12 @@ export type RootStackParamsList = {
   LanguageSettings: undefined;
   CoordinateFormat: undefined;
   Experiments: undefined;
-  PhotoPreviewModal: {
-    observationDocId?: string;
-    photo: SavedPhoto | ProcessedDraftPhoto;
+  DraftPhotoPreviewModal: {
```
cimigree, line 45: I like the clarification of splitting into two screens, but...
there’s a lot of duplication (image load logic, EXIF->strings, InfoItem composition).

Suggestion... 
move onImageLoad + imageLoadInfo into a hook such as useImageLoadInfo and reuse it in both screens.

Make a presentational component like PhotoPreviewBase or something and keep the Draft and Attached Modals as adapters that compute data using hooks.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+        width: event.source.width,
+      });
+    }
+  }
+
+  return (
+    <ScrollView contentContainerStyle={{padding: 20, gap: 20}}>
+      <View>
+        <View style={{flex: 1, borderRadius: 10, overflow: 'hidden'}}>
+          <ErrorBoundary fallback={() => <ImageErrorPlaceholder />}>
```
cimigree, line 146: You said you took out the suspense to make the component simpler, but are you sure it is not needed?

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+import {CoreBlobImage} from '../../sharedComponents/Images/CoreBlobImage.tsx';
+import {ImageErrorPlaceholder} from '../../sharedComponents/Images/ImageErrorPlaceholder.tsx';
+import {useActiveProject} from '../../contexts/ActiveProjectContext.tsx';
+import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
+import {BodyText} from '../../sharedComponents/Text/BodyText.tsx';
+import {sharedStyles} from './sharedStyles.ts';
+import {useGetCreatedBy} from '../../hooks/server/useGetCreatedBy.ts';
+import {useSingleDocByDocId} from '@comapeo/core-react';
+import {useAppLanguageTag} from '../../hooks/useAppLanguageTag.ts';
+import {Accordian} from '../../sharedComponents/Accordian.tsx';
```
cimigree, line 33: Not really part of this PR, but the spelling of Accordian is bothering me. It is Accordion... Just wanted to mention it.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+          }}>
+          <Accordian
+            style={{padding: 20}}
+            title={
+              <View style={{flexDirection: 'row', gap: 12}}>
+                <Octicons
+                  name="check-circle"
+                  size={20}
+                  color={NEW_DARK_GREY}
+                  allowFontScaling
```
cimigree: The default for allowFontScaling is true, so why are you including it here?

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+      setImageLoadInfo({
+        height: event.source.height,
+        width: event.source.width,
+      });
+    }
+  }
+
+  return (
+    <ScrollView contentContainerStyle={{padding: 20, gap: 20}}>
```
cimigree: There are a lot of inline styles and many of them (on first glance) are shared with the DraftPhotoPreviewModal. If you don't make a presentational component, you could at least extract some of these shared styles to named styles and put them in the sharedStyles file, which you made anyway

cimigree (changes requested):
I think the biggest fix(es) needed have to do with the formatting of the time and distance to match the requirements and to make sure they show if they are 0. Also, not sure but having the photo showing it was around a meter from the observation when it wasn't in my personal experience was disconcerting.
After that, next priority for me is reducing some of the code duplication with a presentational component or at least a hook with the shared logic to aid maintainability.
And third, it would be great to see at least a unit test in here if not an integration test since E2E was a no go.
Lots of other very small things aside from that.
This will be a cool feature to have.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: i spoke to Aldo and they told me to use the icons that i have here in the code

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: The suspense will be caught by the suspense at the navigation level. We could make the suspense more granular which is how it was originally. But i think for simplicity it was not needed

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: no, the `memberInfo.name` can be undefined, but the `deviceId` will always exist

src/frontend/screens/PhotoPreviewModal/helpers.ts
ErikSin: >When I took my photos I was literally sitting in one spot, but the distance shows as .96 and 1.47 from the observation. Maybe set to 0 if the distance is below the accuracy or below say 3-5 meters (or even 2?).

yes AND, this is not that simple. Part of the problem is accuracy of the GPS is constantly changing. So i agree with the sentiment that we should just set the distance to 0 if it is within a certain threshold to account for that. But we need to consider that the main use case for this is to be used in legal cases. So if we are falsely set the distance to 0, I think it would look bad to have a value that is not the quantifiable and potentially challenge the credibility of our other data. My sentiment is that it is easier to argue in court that being withing 50m of the observation would still be valid, but showing that we are changing the numbers could lead to a questioning of the credibility of the tool as a whole.

src/frontend/sharedTypes/navigation.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/722332c8882fc3da39f9f0bef1551c5f8b25be44

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/e33bbac162e8efc55e3fabb4bd39a67aa92bfcf3

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/464dffe8f3eca8dfc78e674f17c3c1c9856dc11f

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/2ecd6ed8e89ee8a3e1720e6787129c5acaaf3f46

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/2ecd6ed8e89ee8a3e1720e6787129c5acaaf3f46

src/frontend/screens/ObservationsList/ObservationListItem.tsx
ErikSin: I think it makes sense to keep. The test shouldnt break if there is a photo involved because it is showing the exact same icon, this reflects that

src/frontend/sharedComponents/Editor/index.tsx
ErikSin: Im using it for my e2e tests now!

src/frontend/sharedComponents/Thumbnails/PhotoThumbnail.tsx
ErikSin: Im using it for my e2e tests now!

src/frontend/sharedComponents/Thumbnails/ThumbnailContainer.tsx
ErikSin: Im using it for my e2e tests now!

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/3980126c58e0f32da6fe17b2b5c61dfdbe96ae64

src/frontend/sharedComponents/Editor/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/66496d9e0486976519fc873d7435e841eee95b0b

src/frontend/screens/PhotoPreviewModal/helpers.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/ba746e2c905751f626cd0eecb9534e6e0796f925

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/d9c17c98c7556b991755bbbe3815207d175f0b99

src/frontend/screens/PhotoPreviewModal/helpers.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/ba746e2c905751f626cd0eecb9534e6e0796f925

src/frontend/screens/PhotoPreviewModal/helpers.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/ba746e2c905751f626cd0eecb9534e6e0796f925

src/frontend/sharedTypes/navigation.ts
ErikSin: I opted not to use a presentational component because there are quite a few differences between the 2 screens. The `attachedPhotoView` has`Validated By Comapeo`, `id number`, and `userId`, and the things that are shared, the created at, location, and exif info were already abstracted into a reusable component. To avoid a bunch of conditional rendering, and abstractions of abstractions, i thought it made more sense to not use a presentational component

src/frontend/screens/PhotoPreviewModal/helpers.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1325/commits/1daa028a5506fae3f64047f35f0f3955a3795fc4

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+                      }>
+                      <BodyText selectable style={sharedStyles.primaryInfoText}>
+                        {metersFromObservation < 1000
+                          ? formatMessage(m.distanceFromObs, {
+                              distance: formatNumber(metersFromObservation, {
+                                style: 'unit',
+                                unit: 'meter',
+                              }),
+                            })
+                          : formatMessage(m.distanceFromObs, {
```
cimigree, line 236: You compare metersFromObservation < 1000, but in the else  you pass meters to formatNumber(..., { unit: 'kilometer' }) without dividing by 1000. That will return "1234 km" for 1234 m.

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+                          color={NEW_DARK_GREY}
+                          allowFontScaling
+                        />
+                      }>
+                      <BodyText selectable style={sharedStyles.primaryInfoText}>
+                        {metersFromObservation < 1000
+                          ? formatMessage(m.distanceFromObs, {
+                              distance: formatNumber(metersFromObservation, {
+                                style: 'unit',
+                                unit: 'meter',
```
cimigree, line 232: You could add maximumFractionDigits: 1 (or 2?) if you want here. I think the default is 3 and do we want that? Same for the kms.

cimigree (commented):
I hope you don't mind I tried to have a go at fixing the failing test(s). They do pass for me locally with the scrolling. And in the videos on browserstack you can see what is being looked for and is right there! The only thing I can think is that the error in the photo is overshadowing Browserstack's ability to see/ find things?
It's probably good you were able to extract some of the duplicated logic and styles. Thanks for adding the unit test(s) also. And fixing all the possible 0s not rendering stuff. Anyway, I do think the m/ km thing needs to fixed and we can't have a failing test from the get go in the E2E or our CI is just frustrating and useless. If we advanced to Android 14 or 15 they all might work? But I am not sure if that is true or if we want to do that. You could just take out that one last test that is failing (it('should open second photo and show photo create After observation').
Those two changes need to be made (m/1000) and the failing E2E test needs to be fixed somehow or removed but otherwise this is approved. 
And if you want me to undo the test changes I can, but I don't think they hurt anything and more tests seemed to pass after I added the changes. I was just trying to see if it helped. Now only the last one is failing.

ErikSin:
I took out the tests from ci and updated the km m formatting

cimigree (approved):

PR #1326, opened 2025-08-11
chore: onboarding improvements

### LINKED ISSUE #1290: Onboarding Improvements
**Description**
This update aims to address the issue  raised after recent field trips. Participants are needing more guidance with Projects upfront.*

Recent field trips over the last few months have given us insights that users get lost once they’ve onboarded to CoMapeo. To address this we are making updates to our onboarding flow which might add a few more screens but we hope are helpful to users in understanding their next steps.

Keep in mind Project Stats flow is also introduced - this will be in this ticket
—

**Feature spec**
https://www.notion.so/digidem/Onboarding-Improvements-21d1b08162d58005a2dcf20485696931?source=copy_link

**Interactive prototype**

https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6560-3595&t=FdhNx4sG9EgxuVIQ-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6560%3A3595&show-proto-sidebar=1

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6560-3524&t=ggEU3Eniqm0RlLbg-0

**ToDo**

## Welcome
Image

- [x] Update screen as designed

## Data & Privacy

Image

- [x] Update screen as designed

## Name Your Device

Image

- [x] Update screen as designed

## Success: Device is Ready

Image

- [x] Update screen as designed
- [x] Display device name as written in the Name Your Device step
- [x] Replace copy of "Join a project" button to "Next"

## Projects Intro

Image

- [x] Integrate screen as designed
- [ ] Tapping "Join an Existing Project" user goes to https://github.com/digidem/comapeo-mobile/issues/1291 Join Project screen
- [ ] Tapping "Start a New Project"  user goes to https://github.com/digidem/comapeo-mobile/issues/1291 Start Project screen
- [ ] Tapping "SGo to Map"  user goes to https://github.com/digidem/comapeo-mobile/issues/1291 Go to Map screen

cimigree:
closes #1290 

This PR updates the onboarding flow with new images, text, and navigation behavior, and introduces a revised approach for setting the active project ID. It also includes new integration and end-to-end (E2E) tests for these changes.

#### Onboarding Flow

* **Updated visuals and copy**: New images and updated text on the onboarding screen to make the first-use experience clearer.
* **New navigation step**: After the onboarding success screen, users are now taken directly to the **Projects Intro** screen.
* **Integration + E2E coverage**: Added tests to confirm the flow from onboarding → success → projects intro works as expected.

#### Active Project ID handling

* **Refactored `ActiveProjectContext`** to change how the active project ID is initialized:
  * Previously: A project (and its ID) was created automatically on first launch if none existed — even though it wasn’t needed yet.
  * Now: Since we will have explicit project onboarding, projects will only be created deliberately at the appropriate time.
  
#### Exchange Screen Test Updates
* The Exchange screen doesn’t run inside the RootStackNavigator, so I explicitly wrapped `SyncScreen` in `ActiveProjectProvider` for tests.
* I preset `activeProjectId` in the wrapper so it's clear and predictable.
* No functional changes were made to the `Exchange` UI logic — only the test helper was updated to work with the new context requirement.

---

## Why This Change Was Made

* This allows us to check for a project before deciding whether or not to render project onboarding screens.
* If someone joins an existing project, they won’t get an unnecessary default solo project — they’ll only get one if they choose to create it.
* Creating a project on app load was hidden, abstracted, and confusing. This change removes that “magic” and aligns project creation with explicit user action.

---

## Next Steps

* The next PR will complete the Project Onboarding flow and implement explicit project creation.


src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -49,58 +42,51 @@ export const Success = ({
...
+          <PrimaryButton
+            testID="ONBOARDING.go-to-project-btn"
+            fullSize
+            onPress={() => {
+              mutate({
+                name: deviceName,
+                deviceType: expoToCoreDeviceType(deviceType),
+              });
+            }}
+            text={status === 'success' ? t(m.next) : t(m.joinProject)}
```
ErikSin: why is this ternary here?

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -49,11 +51,11 @@ export const RootStackNavigator = () => {
...
       }}
       screenOptions={NavigatorScreenOptions}>
-      {deviceInfo.name
-        ? createDefaultScreenGroup({
-            intl: formatMessage,
-          })
-        : createOnboardingScreens({intl: formatMessage})}
+      {!deviceInfo?.name
+        ? createOnboardingScreens({intl: formatMessage})
+        : projects && projects.length === 0
```
ErikSin: i think this can be simplified to `project.length === 0` (aka you can take our `projects &&`)

tests/e2e/specs/flow.test.ts
ErikSin, line 1: i assume this is unintentional?

src/frontend/contexts/ActiveProjectContext.tsx
```diff
@@ -46,26 +42,20 @@ export const ActiveProjectProvider = ({
...
-  }, [activeProjectId, setActiveProjectId, createProject, mapeoApi]);
+  }, [activeProjectId, setActiveProjectId, mapeoApi]);
 
   if (!activeProjectId) {
-    return <Loading />;
+    return (
+      <ActiveProjectContext.Provider value={undefined}>
+        {children}
+      </ActiveProjectContext.Provider>
+    );
```
ErikSin: Im a little unsure about this

Previously, if you did not have a project, we created one right away. But in order for the user to never have to interact with the app while there was no project, we rendered a loader while it created the new project. And only rendered the children when there was an active project.

Now you are rendering the children even if there is no project. Which makes sense because there is a state now where the user does not have a project.

But then you have [this line](https://github.com/digidem/comapeo-mobile/blob/6cc08bf36bed6ee26135617572d5d980b33fee95/src/frontend/contexts/ActiveProjectContext.tsx#L92) which throws an error if it is undefined. So you have this react logic that looks like the children are rendered, but if the user tries to use the context an error will be thrown.

And the purpose of this context is to guarantee a activeProjectId. If activeProjectId could be undefined, then we would always have to check projectId exists everytime it is used, which is quite often, and it would cause a lot of laborious ts checking...

I think because of that we need to separate the concerns. We should only render this context if there is an activeProjectId. Therefore activeProjectId should be a prop of this context.

Unfortunately that means we cant wrap the onboarding screen because those screen will not have a projectId. 

But due to the limitations of react navigation, we can't just wrap some screen with a context, we have to wrap the entire navigator.

There are 2 ways that I think of dealing with this:
1. We can pass the projectId as a nav prop to every screen that needs. That way it is guaranteed that the projectId exists, as we would not be able to navigate to it otherwise.
2. We can [nest default screens in its own navigator](https://reactnavigation.org/docs/nesting-navigators/). Since we can wrap a navigator with a context, and make that entire navigator a stack screen, we can the control which screen are wrapped with this context.

Both these ways feel a little laborious so perhaps this current pattern is ok. 

Genuinely I do not know, so im curious what you think

src/frontend/Navigation/Stack/RootStack.test.tsx
ErikSin, line 1: nice integration tests!

src/frontend/screens/Exchange/index.test.tsx
```diff
@@ -95,7 +97,8 @@ describe('Exchange screen', () => {
   };
 
   test('when project is in "solo mode", renders a screen with info', async () => {
-    renderSyncScreen();
+    const projectId = await client.createProject({name: undefined});
```
ErikSin: i think the creation of the project should be in the 'beforeEach' logic, since that is what deals with the setup

ErikSin (commented):
Nice job with the integration tests.

I am unsure about how we initialize the active project Id so I would like your opinion

src/frontend/screens/Onboarding/Success.tsx
cimigree: Screenshot 2025-08-13 at 9 42 49 AM
I am interpreting this request to mean the button is supposed to say next after the success? But I was guessing. Could definitely be wrong!

tests/e2e/specs/flow.test.ts
cimigree: No it is intentional. I think I did not explain this well in this PR write up, but this PR goes with this one: https://github.com/digidem/comapeo-mobile/pull/1334. The issue for this sort of ends in the middle, so this PR does also. If you ran the code, you would see that if you don't have a project, the app just sort of stops on the ProjectsInfo screen so the E2E would all fail. 
The next PR picks up where this one leaves off and restores the E2E tests.

src/frontend/Navigation/Stack/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1326/commits/303c82240bdef32902c0f47c6bdef4525c604d8f

src/frontend/contexts/ActiveProjectContext.tsx
cimigree: I think I understand what the problem is. Rendering children with no ProjectId seems fallible... so I changed that and 
am pushing a proposed solution that I had thought of earlier but decided not to use because it seemed repetitive/ cumbersome. I did not understand the #2 suggestion but maybe what I am doing is similar in intent? https://github.com/digidem/comapeo-mobile/pull/1326/commits/303c82240bdef32902c0f47c6bdef4525c604d8f

src/frontend/screens/Exchange/index.test.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1326/commits/303c82240bdef32902c0f47c6bdef4525c604d8f

src/frontend/contexts/ActiveProjectContext.tsx
```diff
@@ -46,23 +42,13 @@ export const ActiveProjectProvider = ({
...
       })
       .catch(err => {
         // TODO: Surface error in UI
         console.error(err);
       });
-  }, [activeProjectId, setActiveProjectId, createProject, mapeoApi]);
+  }, [activeProjectId, setActiveProjectId, mapeoApi]);
 
   if (!activeProjectId) {
     return <Loading />;
```
ErikSin: this context shouldn't do this check anymore. We should do this check the level about (and that check is how we should determine whether to show the project onboarding screens).

I think `activeProjectId` should be a prop of this component

src/frontend/screens/Onboarding/Success.tsx
ErikSin: Hmm im not sure, i just asked ximena to clarify. Ill get back to you as soon as I know

ErikSin (commented):
The way you did it was actually now what I was thinking. But I think your way is better than what I had in mind.

One thing though is that we should do is refactor the activeprojectcontext. All that logic should live in the parent component since its the same logic to determine if the user should be in the join project screen.

Also, once we do that, we should update your other PR. I don't want to approve this until your other onboarding PR has been merged into this. I just don't want unfinished work in the dev branch

src/frontend/screens/Onboarding/Success.tsx
ErikSin: I just spoke to Ximena, it should just say "Next", never "Join a Project".

src/frontend/contexts/ActiveProjectContext.tsx
cimigree: I am not totally sure I understand what you wanted, but took a stab at it: https://github.com/digidem/comapeo-mobile/pull/1326/commits/63f8f8ca78c8fbbcdeb74394300823bedcc71524

src/frontend/screens/Onboarding/Success.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1326/commits/63f8f8ca78c8fbbcdeb74394300823bedcc71524

cimigree:
@ErikSin I took a stab at interpreting what you said/ guessing what you meant. I don't want to refactor the other PR until I know if it is indeed what you meant/ what you want so I would appreciate your feedback when you get a chance.

I had to add the suspense loader in the context for the exchange index test btw. If you don't like it there, I can move it into the test itself.

src/frontend/contexts/ActiveProjectContext.tsx
```diff
@@ -1,81 +1,26 @@
...
-  }
+};
 
+export function ActiveProjectProvider({activeProjectId, children}: Props) {
   return (
     <React.Suspense fallback={<Loading />}>
       <ProjectApiLoader projectId={activeProjectId}>
         {children}
       </ProjectApiLoader>
     </React.Suspense>
```
ErikSin: Sorry one more thing, I think we can simplify this even more and take out the ProjectApiLoader and the Suspense. Before it was at the top level of the app so there was no suspense wrapping it, but now it is pretty nested so we don't need this suspense

ErikSin (commented):
Yeah, I think how you did it made sense. Now the context doesn't have that unnecessary logic.

1 small suggestion, but otherwise "approved". I am not going to approve it on github because I do not want unfinshed code in the develop branch, but you can update your other PR now

src/frontend/contexts/ActiveProjectContext.tsx
cimigree: As I said, I put the suspense in for the test. Want me to move it into the test?

ErikSin: Sorry, im confused. This suspense was already there, i don't think you added it?

cimigree: It was there. I removed it before I made this PR update. The exchange tests failed. I added it back. The exchange tests pass.

ErikSin: Then yes, i dont think the suspense should just be here for a test

ErikSin (approved):

PR #1327, opened 2025-08-11
fix: Fix track distance calculation

gmaclennan:
There was a bug in the track distance calculation: Adding more than one location when there was already a location history would exclude from the distance calculation the distance between the last point in the location history and the newly added locations.

This PR adds a failing test to identify this issue, and then a fix, which also simplifies the implementation.


achou11 (approved):
Might want to add Erik as a reviewer but the change makes sense to me and makes the implementation much easier to understand 👍

PR #1328, opened 2025-08-11
fix: Remove track timer context and use hook

gmaclennan:
The track context was updating every 1000ms when a track was active, which causes a whole-app re-render every 1000ms. The track timer doesn't need to be in context, so switching to a hook fixes this.

This PR also makes a small change to the `useFormattedTimeSince()` hook to guarantee that `useFormattedTimeSince(null)` always returns `00:00:00`. The previous implementation had this behaviour in general, but there was an edge-case possibility of a 1 millisecond difference (due to comparing two separate `new Date()` calls) which would not have had any real-life consequences, but seemed worth cleaning up now, and hopefully clarifies the behaviour of passing `start=null` which may have been unclear before.


ErikSin (approved):
Thanks! I should have caught this earlier. Previously the context was scoped to just the tabs (which was still not ideal - but didn't cause the entire app to re-render). But because we changed the nature of the tabs, i just moved it to the top app level without much thought.

Issue #1291, opened 2025-07-23
Onboarding Improvements - Join a Project/Start Project/Go to Map


ximenabb:
**Description**
This update aims to address the issue  raised after recent field trips. Participants are needing more guidance with Projects upfront.

Recent field trips over the last few months have given us insights that users get lost once they’ve onboarded to CoMapeo. To address this we are making updates to our onboarding flow which might add a few more screens but we hope are helpful to users in understanding their next steps.

Keep in mind Project Stats flow is also introduced - this will be in this ticket
—

**Feature spec**
https://www.notion.so/digidem/Onboarding-Improvements-21d1b08162d58005a2dcf20485696931?source=copy_link

**Interactive prototype**

https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6560-3595&t=FdhNx4sG9EgxuVIQ-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6560%3A3595&show-proto-sidebar=1

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6560-3524&t=ggEU3Eniqm0RlLbg-0

**ToDo**

## Join a Project Interstitial

<img width="291" height="528" alt="Image" src="https://github.com/user-attachments/assets/06ccf7d7-7b79-461a-8437-609c6d45f1cb" />

- [ ] Integrate screen as designed
- [ ] Tapping on "Go Back" go to the Projects Intro screen https://github.com/digidem/comapeo-mobile/issues/1290

## Start New Project

<img width="296" height="525" alt="Image" src="https://github.com/user-attachments/assets/e8966a9d-165a-4c36-9b6f-597495d251e3" />

- [ ] Integrate screen as designed
- [ ] Tapping on "Go Back" go to the Projects Intro screen https://github.com/digidem/comapeo-mobile/issues/1290
- [ ] Tapping on "Start" go to Name My Project

## Start New Project - Name My Project

<img width="235" height="431" alt="Image" src="https://github.com/user-attachments/assets/f715b013-87cd-45ce-acf6-54dc0739438c" />

- [ ] Update screen as designed. 

## Map on Your Own

<img width="298" height="526" alt="Image" src="https://github.com/user-attachments/assets/d9ac3289-9618-4c79-aee1-90f0bf88afda" />

[...]

PR #1334, opened 2025-08-12
feat: project onboarding

cimigree:
closes: #1291 
# Project Onboarding: screens, navigation, and tests

This PR introduces the **Project Onboarding** feature and a small set of **presentation components** so I can reuse existing screens.

### New presentation components

* **`StartNewProjectBaseScreen`**

  * Pure, stateless UI. Accepts `TopIcon`, `onGoBack`, and `onStart`.
  * Lets us reuse the same screen layout with different icons (composition over abstraction).
  * Keeps props limited and agnostic (no leaking `navigation` into the presentational layer).

* **`CreateProjectForm`**

  * Handles only the “project name” field + passes an action for the submit.
  * Displays “Keep in mind” info rows.
  * Exposes `(projectName: string) => void` submit handler and `isPending` to block double taps.
  * Keeps form concerns decoupled from navigation / mutation logic.

### New onboarding screens

* **`Join a Project`** 

  * Just has text and a Go Back button

* **`StartNewProject`** (container)

  * Renders `StartNewProjectBaseScreen`.

* **`CreateProject`** (container)

  * Renders `CreateProjectForm`.
  * Calls `useCreateProject`, sets active project on success, and then because there is now a project, the MapScreen and regular app stack populates automatically.
  * I have no information in the ticket about which screens I should add next, though there is UX and info from design for the future (I think?). So I was not sure what to do here.
  * Blocks back nav while the mutation is pending and surfaces errors to `ErrorBottomSheet`.
  
* **`MapOnOwn`**

  * Creates a default project (localized name) and sets it active.
  * Blocks repeat taps and back nav while pending.
  * Lets the **RootStack** naturally flip to the default app flow once a project exists.

### Navigation

* **Root stack branching**

  * If device not named → onboarding (existing).
  * If device named **and** no projects → **project onboarding** (new).
  * Else → default app.

* Moved the ErrorBottomSheet to the RootStack so it could be used in the ProjectOnboarding and in the regular App Screens

### Integration test (Jest + RTL)

* adds multiple tests in the RootStack test for checking the functioning of creating a project and checking for various error states and edge cases

### E2E test (Appium/Webdriver/BrowserStack)

* **`tests/e2e/specs/onboarding/project-onboarding.test.ts`**
* This tests the different project onboarding screens back and forth but does not create a project
* Instead creates a default solo project so we can keep the rest of the flow
* Updated the documentation


cimigree:
@ErikSin I had to wrap the Provider in the main app with suspense in order to get the one test where I create a project to pass. It was not an issue of adding Suspense to the test itself which makes me think it is actually needed in the code and would be useful for real world use cases. It also passed fine when I had suspense in the Context, but I know you did not want it there.

src/frontend/sharedComponents/Projects/CreateProjectForm.tsx
ErikSin, line 1: I think we need to reconsider how we are reusing shared UI.

We are doing two different things, both which independently solve the same problem, but we are combining them which makes it a little confusing

Originally The `CreateOrJoinProject` rendered the same screen, and we conditionally changed the text and action based on the navigation name.

Now you have created this component, which does the same thing as above, which is to allow for the same UI with slight variations on the text and action.

But now the CreateOrJoinProject uses this component, so were using 2 different tactics to solve the same problem, when only one is needed.

I also think the 2 create project screens should actually be the same. I just sent a [message to design](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1755197488069659) to ask. 

If the 2 create project screen are the same, I am still in favor of using the original tactic of one screen conditionally rendered based on the nav name. While the code is slightly more verbose, it avoids nesting and  i think nesting makes things harder to debug and refactor in the future.

If the 2 create project screens are not the same, then I think we can use this, but I think we should split up the createOrJoinProject Screen to have a total of 3 different react components that use this component

src/frontend/sharedComponents/Projects/StartNewProjectBase.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <ScreenContentWithDock
+      dockContent={
+        <View style={styles.buttonsContainer}>
+          <SecondaryButton fullSize text={t(m.goBack)} onPress={onGoBack} />
+          <PrimaryButton fullSize text={t(m.start)} onPress={onStart} />
+        </View>
+      }>
+      <View style={styles.headerArea}>
+        <TopIcon width={90} height={90} color={DARK_ORANGE} {...iconProps} />
```
ErikSin: if were passing the TopIcon, why doesn't it just deal with props itself. So instead of passing icons and props we just pass an icon that has props set already.

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -85,7 +98,8 @@ export const RootStackNavigator = () => {
   if (projects.length === 0) {
     return (
       <RootStack.Navigator {...commonNavigatorProps}>
-        {createProjectOnboardingScreens()}
+        {createProjectOnboardingScreens({intl: formatMessage})}
+        {ErrorBottomSheetGroup}
```
ErikSin: i believe this wont work. I think you need to do `ErrorBottomSheetGroup()` -with the brackets at the end.

But i think to simplify it you can just do:
```ts
const ErrorBottomScreenOption: NativeStackNavigationOptions = {
  presentation: 'transparentModal',
  headerShown: false,
  animation: 'none',
  contentStyle: {backgroundColor: 'transparent'},
};


//in the 2 navigators that use it:
// It also doesn't need to be nested in a `group`. The group just allows for the screens to have shared options, but since there is only one screen we can just apply it directly to the screen
<RootStack.Navigator {...commonNavigatorProps}>
        {...createWhateverScreen()}
        <RootStack.Screen
          name="ErrorBottomSheet"
          options={ErrorBottomScreenOption}
          component={ErrorBottomSheet}
        />
  </RootStack.Navigator>

```

ErikSin (changes requested):
This is looking good. Sorry, there are some changes, but they are not your fault, they are due to a mistake in the way these were handed off. 

After speaking to aldo, The Start New Project and the Create Project Screen should be exactly the same as when the user is starting a project not in onboarding.

That means we can get rid of your sharedComponents/Project Screens, and just reuse the screens that were already created.

src/frontend/sharedComponents/Projects/StartNewProjectBase.tsx
ErikSin: This comment is not relevant anymore as we are getting rid of this componenent

src/frontend/Navigation/Stack/index.tsx
cimigree: You're right the Group isn’t necessary for a single screen. Sorry about that added complication. FWIW the previous version did work because it was a JSX element, not a function and I know that for extra sure because the test I wrote passed. https://github.com/digidem/comapeo-mobile/pull/1334/commits/787691e3c31bd979bcf73d2896698a6f58473ac7

src/frontend/sharedComponents/Projects/StartNewProjectBase.tsx
cimigree: Ok no longer relevant.

src/frontend/sharedComponents/Projects/CreateProjectForm.tsx
cimigree: @ErikSin, I was not able to use one screen for all three iterations of create or name solo project because there is no Project context and no project for when creating a project during on boarding for the first time and it is not possible to conditionally call hooks, as you know.
I know you don't want to nest components at all no matter what, so I had to create two different screens. I could not figure out how to do it without some kind of nesting, so that is why there is now createproject and namesoloproject files now.

ErikSin: > I know you don't want to nest components at all no matter what, so I had to create two different screens. 

I don't actually think this. If you look at my comment above, "If the 2 create project screens are not the same, then I think we can use this, but I think we should split up the createOrJoinProject Screen to have a total of 3 different react components that use this component". 

What i was saying is that **IF** we could avoid nesting, then we should. But since we have to make 3 different screens, we **should** use this component...

src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/NameSoloProject.tsx
```diff
@@ -0,0 +1,192 @@
...
+    <View style={styles.infoRow}>
+      <Icon width={20} height={26} style={styles.infoIcon} />
+      <BodyText variant="smallMeta" style={styles.infoText}>
+        {text}
+      </BodyText>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
```
ErikSin: If we are not going to use the shared component, then I think we should reuse these styles and share those between the components

src/frontend/Navigation/Stack/RootStack.test.tsx
ErikSin, line 1: Nice, these test are actually testing the correct behaviour. Good job with these

src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/NameSoloProject.tsx
```diff
@@ -0,0 +1,192 @@
...
+function InfoRow({Icon, text}: InfoRowProps) {
+  return (
+    <View style={styles.infoRow}>
+      <Icon width={20} height={26} style={styles.infoIcon} />
+      <BodyText variant="smallMeta" style={styles.infoText}>
+        {text}
+      </BodyText>
+    </View>
+  );
+}
```
ErikSin: If you don't want to use the shared screen, then i think this should be reuseble since they are identical

ErikSin (changes requested):
Just to reiterate, Im not against nested components/reusable components, I think we just need to not do it unnecessarily as would have been the case if we could have just used one screen and registered with 3 different nav components.

My suggestion was to use 1 reusable component for the 3 screens... But if we are going to keep it as 2 different screens, I think we should reuse the styles, and reuse some of the smaller pieces as per my comments

Also, can we add screenshots to the PR?

The main blocker though is the create project in the onboarding flow does not have a success screen. I know we are conditionally rendering, and that is why, but I think there is a simple solution.

in RootStackNavigator:

```ts
const [hasInitialized, setHasInitialize] = React.useState(false);

React.useEffect(() => {
    if (!hasInitialized && !storedActiveProjectId && fallbackProjectId) {
      setActiveProjectId(fallbackProjectId)
      setHasInitialized(true)
    } else {
         if(!hasInitialized){
            setHasInitialized(true)
         }
    }

  
  }, [storedActiveProjectId, fallbackProjectId, setActiveProjectId]);


if (!hasInitialized) {
    return <Loading />;
  }

  if (!deviceInfo?.name) {
    return (
        {createOnboardingScreens({intl: formatMessage})}
    );
  }

  // instead of checking for list of project just check for active project id/
  if (!activeProjectId) {
    return (
        {createProjectOnboardingScreens({intl: formatMessage})}
    );
  }

  return (
    createDefaultScreenGroup({intl: formatMessage})
  );

```

Then in the `OnboardingCreateProject` screen DONT set the active project id when the project is created, instead navigate to the success screen and pass the projectID to the success screen. And when the user presses the `Done` on the success screen, we can set the active project id.

cimigree:
This is what I said about lack of a Success Screen above in the original PR comments.

> Calls useCreateProject, sets active project on success, and then because there is now a project, the MapScreen and regular app stack populates automatically.
> I have no information in the ticket about which screens I should add next, though there is UX and info from design for the future (I think?). So I was not sure what to do here.

Your PR with the Project Stats opt in is supposed to be the next screen (see Figma), so I didn't think it made a lot of sense to wire up something when your PR would be what is used.

But I can wire up a success screen solution to use in case the Project Stats opt in isn't ready for a while as you say.

ErikSin:
> But I can wire up a success screen solution to use in case the Project Stats opt in isn't ready for a while as you say.

Yes, i think it makes sense to wire up the success screen because even with the project stats, it will have to follow a similar architecture...

src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/NameSoloProject.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1334/commits/c22ce39b1f5c5b7f56fc81851e3029c1235255aa

src/frontend/screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/NameSoloProject.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1334/commits/c22ce39b1f5c5b7f56fc81851e3029c1235255aa

cimigree:
I changed the Root Stack to use the project id. I did not set the project ID for the onboarding when the project is created. I added a new success screen to use temporarily as I did not want to mess with the existing project created success screen (it is a bit complicated already) since the project created success in onboarding will be changing anyway with future work on the project statistics opt in (and it was not at all in scope for this Issue/ PR).

Screenshots:
projectsintroscreen
startnewproject
createproject
Success
gotomap
joinaproject

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -38,13 +46,22 @@ export const RootStackNavigator = () => {
...
+  const [hasInitialized, setHasInitialized] = React.useState(false);
 
   React.useEffect(() => {
-    if (!storedActiveProjectId && fallbackProjectId) {
-      setActiveProjectId(fallbackProjectId);
+    if (!hasInitialized) {
+      if (!storedActiveProjectId && fallbackProjectId) {
+        setActiveProjectId(fallbackProjectId);
+      }
+      setHasInitialized(true);
```
ErikSin: You need a loading state. Otherwise there could be a case where there is a `fallbackPorjectId` but no `storedActiveProjectId`, in which case `createProjectOnboardingScreens` will initially render, but then this use effect will be called, and then storedActiveProjectId will be set, causing the `createDefaultScreenGroup` to render.

ErikSin (approved):
Just need to add a loading state, but othewise looks good. You can merge this into your other PR. That other PR can then be merged into develop, but can you make sure to check the other PR after this is merged so that nothing got lost in the merge

Issue #1287, opened 2025-07-23
Project Stats / Name or Create my project


ximenabb:
**Description**
 Data Collection Statistics (Project: because data is collected in the context of a project, it’s important that all project participants are aware that this information is being shared with AD. This data would include:
    - Number of observations and tracks collected in a project, possibly summarized by user (observations/user) and time (observations/day)
    - Area covered by observations / tracks Again by user and time
    - Stats about track distance and time.
    - Stats about attachments (photos, audio)
    - Number of participants in a project broken down by time.

Goal : Introduce a project-wide opt-in. This is activated by the project creator, and anyone can opt-out via Project Stats screen https://github.com/digidem/comapeo-mobile/issues/1288

**Feature spec**
https://www.notion.so/digidem/Data-Metrics-Project-Statistics-265cc821e6184588bd5a8cadbab3b8eb?source=copy_link

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6964-3690&t=sALYvYEh8zOP6sjV-0

**ToDo**

User has access to the Name or Create my project in 3 ways:
- App Onboarding -  Tapping "Start a New Project" - see Projects Intro screen https://github.com/digidem/comapeo-mobile/issues/1290
- On All Project - tapping on "Start new project"
- On My Solo Project - tapping "+Invite"

## Share Project Statistics 

<img width="268" height="481" alt="Image" src="https://github.com/user-attachments/assets/b967477b-9383-457a-8750-7aa089bf67c9" />

- [ ] Integrate screen as designed
- [ ] Do not add the "See Example" link for now
- [ ] Tapping "No, Skip for Now" user goes to Success: New Project Created! / NO
- [ ] Tapping "Yes, Share Stats" user goes to Success: New Project Created! / YES 

## Success: New Project Created! / YES 

<img width="265" height="475" alt="Image" src="https://github.com/user-attachments/assets/5fe7b469-4f67-4abf-ba69-48e7659a0454" />

- [ ] Update screen as designed
- [ ] Display "Project Name"
- [ ] Display "Project statistics are being shared"
- [ ] Tapping "Done" user goes to Map view
- [ ] Tapping "Invite a Device" user goes to invite device flow

## Success: New Project Created! / NO
- [ ] Update screen as designed in **Success: New Project Created! / YES** 
- [ ] Display "Project Name"
- [ ] Do NOT Display "Project statistics are being shared"
[...]

PR #1339, opened 2025-08-13
feat: project stats info screen

ErikSin:
towards #1287

image

Creates the Shared Project Stats Screen. Currently it is just the one screen and updated navigation, but outside of that it functionally does nothing

## What is missing
- The API to turn on and off project stats has not been created yet
- The API to show that project stats has been turned on or off for a project


tests/e2e/specs/project/rename-project-from-drawer.test.ts
```diff
@@ -16,18 +16,23 @@ describe('Project - Rename Project from Drawer', () => {
     const nameInput = await $(byResourceId('PROJECT.name-inp'));
     await nameInput.setValue(output.names.project);
     await createBtn.click();
+  });
+
+  require('./project-usage-stats.test');
```
cimigree: This doesn't seem to work given the output in the error:
```
Error:  Unable to load spec files quite likely because they rely on `browser` object that is not fully initialized.
`browser` object has only `capabilities` and some flags like `isMobile`.
Helper files that use other `browser` commands have to be moved to `before` hook.
Spec file(s): file:///home/runner/work/comapeo-mobile/comapeo-mobile/tests/e2e/specs/flow.test.ts
Error: Error: Cannot find module './project-usage-stats.test'
```

src/frontend/screens/ShareProjectStats.tsx
```diff
@@ -0,0 +1,137 @@
...
+    id: 'screens.ShareProjectStats.infoAnonymized',
+    defaultMessage: 'Shared information will be completely anonymized',
+  },
+  fullyEncrypted: {
+    id: 'screens.ShareProjectStats.fullyEncrypted',
+    defaultMessage: 'Your project will stay fully encrypted',
+  },
+  skip: {
+    id: 'screens.ShareProjectStats.skip',
+    defaultMessage: 'No, Skip for now',
```
cimigree: Now should be capitalized

tests/e2e/specs/multiple-projects/create-and-switch.test.ts
```diff
@@ -13,8 +13,15 @@ describe('Multiple Projects - Create and Switch Between Projects', () => {
     const nameInput = await $(byResourceId('PROJECT.name-inp'));
     await nameInput.setValue(output.names.secondProject);
     await $(byResourceId('PROJECT.create-btn')).click();
+  });
+
+  require('../project/project-usage-stats.test');
+
+  require('./project-usage-stats.test');
```
cimigree: Or maybe it is this one that is causing the error. Whichever. I don't think it is working the way you expect.

cimigree (changes requested):
I think you should use ScreenContentWithDock for this screen.
On a small screen, the buttons are getting cut off and I think that is what ScreenContentWithDock is for.
ShareProjectStats

I also think you need to revisit your strategy for testing this component because the way you have it just seems to be erroring out.

ErikSin:
closing this for now until the api is supported

cimigree:
@ErikSin I finished the ticket by amending the success screen and adding the blurb on the invite received as the ticket specified. I also fixed up the tests and added a little test. I am having trouble getting myself off this and requesting you though as a reviewer.
Since there are no longer two screens for a solo project named and a new project, I simplified that success screen. Also, though, I think all of this stuff is going to be changing so I figure this is good in the meantime and matches the Figma as far as I can tell.

Success screen when opting in:
image

Success screen when skipping the opt in:
image

And here is the invite when Stats are shared.
image

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -62,6 +67,7 @@ export const InviteReceived = ({
   const {isTracking} = useTracking();
 
   const projectColor = invite.projectColor;
+  const showSharedStats = invite.sendStats;
```
ErikSin: show shared stats implies the user can show or not show the stats. Can we rename this to "statsShared" or something along the line that doesn't imply they have a choice in the matter

ErikSin (commented):
Approved-I just can't approve it on github because it is technically my PR. So you can approve it from your end and merge it

cimigree (approved):

PR #1350, opened 2025-08-20
chore: parallel testing

### LINKED ISSUE #1338: Optimize e2e for local testing
Currently e2e testing has a a few pain points that slow down the development cycle:
1. A new build is required every time app code changes, and build take an upward of 15mins
2. e2e tests are designed to run in a certain order, meaning that often testing one feature requires many other tests to be run first, which is often quite time consuming
3. Order of the tests are fragile. Many of the tests are in a certain order as they require the tests before in order to work, while other tests don't require a certain order. Knowing the order requires looking at code of the tests.
4. Appium will use an exisiting APK if the variant already exists on your phone. For example if there is a CoMapeo-RC and appium uses an RC build, it will not update that APK on your phone, leading to frustrating behaviour.

### Propose Solutions
- [ ] Run e2e test on the dev build. On some initial research it looks like it is possible to run appium with the expo dev build. Since the dev build updates automatically, it will mean we do not need to rebuild the apk on every code change. 
- [ ] Comment which test relies on each in the flow.test.ts file.
- [ ] Determine which test are common setup test(eg, device naming, creating a project), and write minimal test that allow for those setup test to run quickly, so that we can run those test instead of running the entire test suite

ErikSin:
Closes #1338

**E2E test are now under 12 mins!**

We are now leveraging parallel testing in browserstack where we can run up to 5 devices at the same time. In order to take advantage of this I broke up all the test folders into isolated test. This means all the test can and should be run in isolation, and does not require another test folder (eg, Audio tests should not require the observations test to create an observation before the audio tests are run). 

## What was done
- Each test folder now has an `{folder}/index.test.ts` file. This file runs all the other files on the desired order. This is similar to the flow.test.ts file, but scoped to only the files in that folder
- Created `onboarding/helper/minimal-onboarding-setup.test.ts` file. This allows for test to quickly run the onboarding flow. Since all the tests are run in isolation, all of the tests need to get past the onboarding flow to start the actual tests
- Created `minimal-project-creation.test.ts` file. similarily as above, some test require a project to be created and this allows that to be done
- In the `wdio.ci.config.js` I increased `maxInstances` to 5 which allows us to run 5 instances of the device at once (5 is the max according to our plan)
- In the `wdio.ci.config.js` specs, I added all the tests files (`{folder}/index.test.ts`) that need to be tested
- Removed the `flow.test.ts`
- Removed several test that were being unused
- Moved around several tests into a more appropriate folders
- Updated tests that previously relied on other tests. Most notable several tests relied on the existence of a observation. Updated the test to create the needed observations directly.

## How to move forward with e2e tests
- e2e test should be written in isolation, and should not require other flows. 
- Any new feature should have an e2e folder created, and that folder should have an `index.test.ts`, which is added to the `wdio.ci.config.js` specs


cimigree (changes requested):
This will be great to have the tests grouped and individualized even if it does take longer to run them all locally.
Please add back the remote tests. Also please consider adding back the map test. And I am not sure why I got rid of the gps test. Do you remember?  Is that something that can be added back?
Please update all of the documentation. I know it is a pain, but it will be really helpful going forward. We should just always do that. And you can include info on running the tests locally.
Would it be possible since we have 5 instances to use 5 different devices? I think that could be really beneficial.
Sorry I didn't do this sooner.

ErikSin:
> Also please consider adding back the map test. And I am not sure why I got rid of the gps test. Do you remember? Is that something that can be added back?

The map tests are being covered by other tests, so they are redundant. 
The gps tests were not being used in the original flow.test, so thats why I took them out (I assumed this had something to do with browserstack?)

ErikSin:
Docs have been updated!

cimigree (approved):
Just a few small things and then ready to merge (and then I will have to update all my open PRs. Yikes!)
1. Can you please update the wdio.local.config.js.template with the new specs?
2. It looks like this info in the Observations.md is no longer accurate based on your elimination of testing. Can you remove it?
- Checks the “You’re a coordinator on this project.” or “You’re mapping on your own.” text is displayed at the top of the ObservationsList.
3. You moved rename project from drawer into the solo-project folder (from Main). But you deleted the Main document. Can you just copy out info on that one test into the SoloProject.md document so it is not gone for good?
4. What do you think about creating an issue to make it so that we run the tests on 5 different devices on Browserstack since we now have max instances of 5? I feel like this could make the E2E tests even more useful and helpful for debugging in advance.

ErikSin:
> What do you think about creating an issue to make it so that we run the tests on 5 different devices on Browserstack since we now have max instances of 5? I feel like this could make the E2E tests even more useful and helpful for debugging in advance.

There is actually an annoying amount of complication to this. IMO this is a flaw with their API. As it is now If you only have one device set, but max instances set to 5, it will run 5 instances of that device, and then run different tests on different devices. 

If you set 2 devices, it will run ALL the tests on each device, so each test well be run twice. In order to circumvent that we need to manually set each test to each device. not ideal...

PR #1354, opened 2025-08-21
fix: scrollview onboarding success

### LINKED ISSUE #1352: Onboarding/ button cut on small devices
https://www.notion.so/digidem/Onboarding-button-cut-on-small-devices-2541b08162d580dcb0dbceccc156f173

cimigree:
Closes #1352 
Adds a scrollview around the content on the success to onboarding screen so that the button is available on all screen sizes.
I tested on the Emulator titled "small phone" with a 4.65" screen. I also changed the language to Spanish and increased the font size just to be sure.
I also tested on a regular phone (Pixel 8) to make sure it still behaved as expected on a "normal" screen size.
I had to leave the button out of the scroll view so that it would still be fixed to the bottom of the page but that seemed to work fine as you can see from the video.

https://github.com/user-attachments/assets/8bde6e4d-4061-455d-bb48-a11917127f83

### DIFF
--- a/src/frontend/screens/Onboarding/Success.tsx
+++ b/src/frontend/screens/Onboarding/Success.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {StyleSheet, View} from 'react-native';
+import {ScrollView, StyleSheet, View} from 'react-native';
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
Thanks!

PR #1356, opened 2025-08-21
fix: failing language test

cimigree:
Checks for spanish translation or english translation of settings. This is being translated correctly in the RC so the test is actually failing. Yay?
But locally and on develop, it is sometimes not translated. So this fix checks for either version.

### DIFF
--- a/tests/e2e/specs/settings/language.test.ts
+++ b/tests/e2e/specs/settings/language.test.ts
@@ -43,11 +43,14 @@ describe('Settings - Language Settings Flow', () => {
       await drawerIcon.click();
     }
 
-    //const settingsInSpanish = await $(byTextMatches('Ajustes de la'));
-    // commenting out until translation done
-    const settingsInEnglish = await $(byTextMatches('Settings'));
-    await settingsInEnglish.click();
+    const english = $(byTextMatches('Settings'));
+    const spanish = $(byTextMatches('Ajustes de la'));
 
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/settings/language.test.ts
```diff
@@ -43,11 +43,14 @@ describe('Settings - Language Settings Flow', () => {
...
-    const settingsInEnglish = await $(byTextMatches('Settings'));
-    await settingsInEnglish.click();
+    const english = $(byTextMatches('Settings'));
+    const spanish = $(byTextMatches('Ajustes de la'));
 
+    if (await spanish.isDisplayed()) {
+      await spanish.click();
+    } else {
+      await english.click();
+    }
```
ErikSin: is this necessary if we know that there are spanish translations?

cimigree: As I said, "But locally and on develop, it is sometimes not translated. So this fix checks for either version."
I was not seeing the translation when I was running this locally.

ErikSin: But now that we have merged the translation into develop, if its not on your local branch, it means that translations are not properly synced to your local branch and so it should be failing. The other possibility is that it was taken out of the spanish translations, which should also be a failure because that shouldn't be happening

ErikSin (approved):
Thanks!

Issue #1323, opened 2025-08-06
Data export umbrella issue


gmaclennan:
This is an umbrella issue for improvements to data export. Backend fixes are being tracked in https://github.com/digidem/comapeo-core/issues/1091

**Essential**

- [ ] Indicate that export was successful. This could be as simple as closing the export bottom sheet when export is successful. Currently after the data is shared, the user is returned to the app with the export sheet still open, which appears as if the share was unsuccessful. If the user cancels the share (e.g. goes back to the app without sharing) then I think the current behaviour of keeping the bottom sheet open is ok. Validate with design team.
- [x] Should set the mimetype when sharing - currently when sharing geojson, the system does not identify the url as a file, and offers apps which cannot handle file sharing as targets. This is currently not a problem when sharing the zip, because the system is auto-detecting the mimetype from the extension I think. Safest option is to always set the correct mimetype.

**Possible** (pending co-design feedback)

- [x] Sharing does not provide a way to save the exported file to phone storage - the "files" app does not show as a share target. I _think_ this could be an issue for users, but we should validate. We can provide [a way to save to storage](https://github.com/digidem/comapeo-mobile/pull/1322), but that cannot be included in the share sheet. It's possible to create [custom actions](https://developer.android.com/training/sharing/send#custom-actions) on the share sheet, but only for Android 14 or higher. Therefore I think we have two options if saving to phone storage is needed:
    1. Only support saving to storage - don't show the share sheet at all.
    2. Add two actions in the app: one for share and one for save to storage.

ximenabb:
Decision landed from Ximena and Gregor, for now, opt for "Only support saving to storage - don't show the share sheet at all."

PR #1357, opened 2025-08-21
feat: export observation as geojson

### LINKED ISSUE #1341: Data Export / Support saving to storage
Sharing does not provide a way to save the exported file to phone storage - the "files" app does not show as a share target. I think this could be an issue for users, but we should validate. We can provide https://github.com/digidem/comapeo-mobile/pull/1322, but that cannot be included in the share sheet. It's possible to create [custom actions](https://developer.android.com/training/sharing/send#custom-actions) on the share sheet, but only for Android 14 or higher. Therefore I think we have two options if saving to phone storage is needed:

- Only support saving to storage - don't show the share sheet at all.
- Add two actions in the app: one for share and one for save to storage.

For now, we will do the following :
Only support saving to storage - don't show the share sheet at all.

ErikSin:
closes #1341

Uses [react native document picker](https://github.com/react-native-documents/document-picker) to download the geojson file. Previoulsy we were using react native share which asks the user to share the file instead of download it.

https://github.com/user-attachments/assets/f1d5686d-5d10-4d71-8e9f-1347e02edee7


tests/e2e/specs/observations/export-observations.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  it('should open Export Observations bottom screen', async () => {
+    const downloadObsBtn = await $(byText('Download Observations'));
+    await downloadObsBtn.click();
+    await expect($(byText('All Observations'))).toBeDisplayed();
+    await expect($(byText('All Observations with Media'))).toBeDisplayed();
+  });
+
+  it('should open show error if no option is chosed', async () => {
```
cimigree: can you change to chosen?

cimigree (changes requested):
This is not a change you made here, but I am just wondering why this hook is in the projects file instead of the observations one.

Also, it looks like there was a request that there be some kind of success, or at least the closing of the modal/ "export sheet". In [here: ](https://github.com/digidem/comapeo-mobile/issues/1323) which I did not see happening. 

> Essential
> 
> Indicate that export was successful. This could be as simple as closing the export bottom sheet when export is successful. Currently after the data is shared, the user is returned to the app with the export sheet still open, which appears as if the share was unsuccessful. If the user cancels the share (e.g. goes back to the app without sharing) then I think the current behaviour of keeping the bottom sheet open is ok. Validate with design team.

I did not validate the name of the geo/json file or that it had the correct stuff in it. Hope that is ok.

ErikSin:
> Also, it looks like there was a request that there be some kind of success, or at least the closing of the modal/ "export sheet". In [here: ](https://github.com/digidem/comapeo-mobile/issues/1323) which I did not see happening.

This PR does not close #1323. Its closes #1341. There are currently no designs for a success sheet, and it is out of scope of what is being done here.

ErikSin:
> I did not validate the name of the geo/json file or that it had the correct stuff in it. Hope that is ok.

I think that is a semi important thing to test. Not necessarily that is it has the correct information, but making sure that it is a geojson file or zip file

src/frontend/hooks/server/projects.ts
```diff
@@ -103,28 +102,39 @@ export function useExportObservationsAndShare({
...
+          .mutateAsync({
+            path: normalizeFilePath(exportDir),
+            exportOptions: {
+              lang,
+              observations: exportType === 'Observation',
+              tracks: exportType === 'Tracks',
+            },
+          })
+          .then(path => {
+            saveDocuments({
```
cimigree: this isn't returned or awaited. So the mutation resolves before the user finishes saving. Please return saveDocuments(...) or await. I know you are not supposed to design a success but at least that makes it usable going forward.
There is an "isPending" in the ExportObservations screen, so it resolves and disappears before it is actually resolved.

src/frontend/hooks/server/projects.ts
```diff
@@ -103,28 +102,39 @@ export function useExportObservationsAndShare({
...
             lang,
-            observations: exportType === 'Observation',
-            tracks: exportType === 'Tracks',
+            observations: true,
+            tracks: false,
+            attachments: true,
           },
+        })
+        .then(path => {
+          saveDocuments({
```
cimigree: See all the comments above like about return/ await, cleaning up temporary files, and passing a suggested file name.

src/frontend/hooks/server/projects.ts
```diff
@@ -103,28 +102,39 @@ export function useExportObservationsAndShare({
...
+            path: normalizeFilePath(exportDir),
+            exportOptions: {
+              lang,
+              observations: exportType === 'Observation',
+              tracks: exportType === 'Tracks',
+            },
+          })
+          .then(path => {
+            saveDocuments({
+              sourceUris: [`file://${path}`],
```
cimigree: Can we clean up this temporary file? Or is it cleaned up and I am just missing it?

src/frontend/hooks/server/projects.ts
```diff
@@ -103,28 +102,39 @@ export function useExportObservationsAndShare({
...
+              mimeType: 'application/geo+json',
```
cimigree, line 130: I think you can pass a filename here so the extension shows up and it isn't "invalid" if a name isn't typed in.

cimigree (changes requested):
The Save dialog opens with an empty filename, so the file gets saved extension-less or as (invalid) if a user doesn't add a name. 

Please pass a suggested fileName so users get a prefilled name and then that way there will be the right extension (.geojson / .zip). 

This also directly addresses the “is it a geojson or zip” file that you asked me to check. Which I cannot check now as is.

Right now it comes through completely extension less as I said above. I was able to open the file with a json reader and it looks like json but I am not sure how to tell if it is geojson. If you want me to test the geojson, I would appreciate guidance on how to do that.
Consider renaming the file from useExport...and Share to Save or something since it no longer Shares.

ErikSin:
> If you want me to test the geojson, I would appreciate guidance on how to do that. 

I was just checking based on whether it had a json extension, and followed the [geojson specs](https://geojson.org/) when opened

src/frontend/hooks/server/projects.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1357/commits/e52a55d7c5cd113d8597574c07f9e70c56c9632d

src/frontend/hooks/server/projects.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1357/commits/06e4b9bd8de3e75c41633ec73a5146817a764753

src/frontend/hooks/server/projects.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1357/commits/06e4b9bd8de3e75c41633ec73a5146817a764753

src/frontend/hooks/server/projects.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1357/commits/e52a55d7c5cd113d8597574c07f9e70c56c9632d

cimigree:
@ErikSin, The E2E test is failing because you can't go back from the Save (downloads) screen without seeing the "Something Went Wrong" error screen. If you log it, this is what it says:
Error: user canceled the document picker
    at promiseMethodWrapper (http://10.0.0.98:8081/index.bundle//&platform=android&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:3994:45)
    at ?anon_0_ (http://10.0.0.98:8081/index.bundle//&platform=android&dev=true&hot=false&transform.engine=hermes&transform.bytecode=1&transform.routerRoot=app&unstable_transformProfile=hermes-stable:370853:98)
    at next (native)


Also, when downloading just Observations, there is no extension.
But it does look like geojson I think.
image

For the with media, the media comes separately. Is that right? And there is no indication which observation the media goes with. But then the document comes through with the .geojson extension. Yay! But it is CoMapeo_undefined_Obsvns_2025_08_26.geojson.
image

ErikSin:
@cimigree this is ready for re-review!

cimigree (approved):
Still not getting file extension for geojson but it is there for the zip file. I don't know what the specs were, but this is downloading stuff that looks like geojson to the files of the phone, so I am approving.

ErikSin:
> Still not getting file extension for geojson

I dont think it would be opening json file if th extension was not there. Its most likely that file extension are just hidden on the device you are opening it with

PR #1359, opened 2025-08-25
feat: low storage prompt

### LINKED ISSUE #1254: Low Storage Prompt
On CoMapeo v3, when a device reaches low storage, user encounters different issues, like not being able to save an audio recording. CoMapeo doesn’t expose a better error state than “Something Went Wrong”

See details : https://www.notion.so/digidem/Low-Storage-Notification-2021b08162d5806f8455d508e4bb1d6c

See designs : https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6210-3403&t=pQONWcogYiO7wqAq-0

### To Do

**Updated Map - Coordinator/Participant**

Image

- [x] When device storage is low (500mb or less) show dismissible alert on the map 
- [x] When device storage is low (500mb or less) show the alert indicator on the menu icon.
- [x] Display dismissible alert and indicator as designed. See figma 

**Multiple Projects Menu - Solo**

Image

- [x]  When user opens the menu they will see this alert first. 
- [ ] [Free up Space] should direct them to Free up space screen (see below)
- [x] This message is NOT dismissible and only disappears once device is has more storage.
- [x] this alert will cause the menu to need to scroll. Make sure scrollbar is present.
- [x] Display alert as designed. See figma 

**Free Up Space Instructions**

Image

- [ ] Display alert as designed. See figma 
- [ ] Tapping "go back" goes back to the projects menu

**Updated Observations List**

Image

- [x] Make the alert present across the app, as designed. See figma

cimigree:
closes #1254 
### 1) Storage state +  hook

* `src/frontend/contexts/LowStorageBannerContext.ts`
  * Zustand store with just whether or not the banner is dismissed and dismissing the banner
  * I didn't use valibot because it was so small and simple but can redo if needed
* `src/frontend/hooks/useStorageReadingQuery.ts`
  *  uses Tanstack Query to fetch the free bytes and total bytes in one hook. I can separate if that is preferred.
  * a hook to determine if storage is low
  * A method to invalidate the query
  * `LOW_THRESHOLD_BYTES = 500MB`. I tested this out by changing that setting to 100000 and then the low storage UI popped up on my device just for the purposes of seeing things for this PR. You could do the same...

### 2) Invalidating of the query
* When participating in an exchange
* When Creating an observation with attachments
* When Editing an obervation with attachments

### 3) UI: banner, menu alert, header badge

* **Map banner** (dismissible for the session)
* **Menu alert** (not dismissible)
* **Indicator on Device Icon** (10×10 orange circle with `!`)
 _Note_: I intentionally placed the badge logic directly in both headers (duplicated) to avoid adding components and nesting, because I know you don't like that.

---

## Tests
***Integration tests***
* **Header badge**

  * `src/frontend/sharedComponents/HomeHeader.lowStorage.test.tsx`
  * checks whether the badge is shown or not based on storage

* **Map banner**

  * `src/frontend/screens/MapScreen/MapScreen.lowStorage.test.tsx`

    * Banner shows when storage is low
    * Banner hides when dismissed
    * Dismissal resets after storage recovers (`isLow=false`) and re-shows on next low
_This was difficult, so I stopped here. I could add more tests for the other UI if you think it is necessary._
---
## Notes

_No E2E_: I wasn't able to figure out how to simulate low storage on Browserstack, so just using integration tests for now.

## Screenshot
image


src/frontend/sharedComponents/Storage/LowStorageBanner.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {View, StyleSheet, Pressable} from 'react-native';
+import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
+import {defineMessages, useIntl} from 'react-intl';
+
+import {BodyText} from '../Text/BodyText';
+import {BLUE_GREY, DARK_ORANGE, NEW_DARK_GREY} from '../../lib/styles';
+
+type Props = {
+  onDismiss: () => void;
+  testID?: string;
```
ErikSin: if testID is being used, lets not make it optional

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function useStorageReadingQuery() {
+  return useQuery({
+    queryKey: STORAGE_QUERY_KEY,
+    queryFn: getReading,
+  });
+}
+
+export function isLowStorage(
+  free: number | null,
```
ErikSin: isLowStorage shouldn't have to worry about null. The consuming compent should worry about it

eg:
```ts
const lowStorage = freeBytes===null ? false : isLowStorage(freeBytes)

```

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export function useStorageReadingQuery() {
+  return useQuery({
+    queryKey: STORAGE_QUERY_KEY,
+    queryFn: getReading,
+  });
+}
+
+export function isLowStorage(
```
ErikSin: Also, this should live in the lib folder

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+  free: number | null,
+  threshold: number = LOW_THRESHOLD_BYTES,
+) {
+  return (free ?? Infinity) <= threshold;
+}
+
+export function invalidateStorageReading(qc: QueryClient) {
+  qc.invalidateQueries({queryKey: STORAGE_QUERY_KEY});
+}
```
ErikSin: to follow the previous pattern, and for slight simplicity, i think we should just export the querykey and not this function. And then wherever we call this function just call `queryClient.invalidateQuery({exporedQueryKey})`

src/frontend/screens/MenuScreen.tsx
```diff
@@ -65,9 +70,29 @@ export function MenuScreen() {
   const displayTitle =
     role === 'solo' ? projectInfo.projectHeader : projectInfo.projectName;
 
+  const {data} = useStorageReadingQuery();
+  const isLow = isLowStorage(data?.freeBytes ?? null);
+  const freeBytes = data?.freeBytes ?? null;
+  const totalBytes = data?.totalBytes ?? null;
```
ErikSin: the `?? null` part is unnecessary.

for this part:

```ts
isLow && (
          <MenuLowStorageAlert
            // we know free bytes will have a value because otherwise isLow would be false
            freeBytes={freeBytes!}
            percentUsed={calcUsedPct(freeBytes, totalBytes)}
          />
        )

```

src/frontend/sharedComponents/HomeHeader.tsx
```diff
@@ -81,12 +87,23 @@ export function HomeHeader({
...
+              accessibilityLabel="Open Menu"
+            />
+            {isLow && (
+              <View
+                testID="low-storage-badge"
+                accessibilityLabel="Low storage alert"
+                style={styles.badge}
+                pointerEvents="none">
+                <Text style={styles.mark}>!</Text>
+              </View>
```
ErikSin: IMO this should be a shared component since the logic is fully handled by the parent and it is purely a visual component. At the very lease it should have shared styles...

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+      DeviceInfo.getTotalDiskCapacity(),
+    ]);
+    return {freeBytes, totalBytes};
+  } catch {
+    return {
+      freeBytes: null as number | null,
+      totalBytes: null as number | null,
+    };
+  }
+}
```
ErikSin: I dont think we should return a value on error

src/frontend/screens/MenuScreen.tsx
```diff
@@ -65,9 +70,29 @@ export function MenuScreen() {
...
+  function calcUsedPct(
+    freeBytes: number | null | undefined,
+    totalBytes: number | null | undefined,
+  ): number {
+    if (!totalBytes || totalBytes <= 0) return 0;
+    const free = Math.max(0, freeBytes ?? 0);
+    const used = Math.max(0, totalBytes - free);
+    return Math.round((used / totalBytes) * 100);
+  }
```
ErikSin: i think we should seperate this an write a unit test for it

ErikSin: Also, "pct" is a quite amibigous, like it doesnt read as "percentage" to me. I think we should just use "calcUsedPercentage"

src/frontend/sharedComponents/Storage/MenuLowStorageAlert.tsx
```diff
@@ -0,0 +1,124 @@
...
+    defaultMessage: 'Free up space',
+  },
+});
+
+export function MenuLowStorageAlert({
+  freeBytes,
+  percentUsed,
+}: MenuLowStorageAlertProps) {
+  const {formatMessage} = useIntl();
+  const pct = Math.min(100, Math.max(0, Math.round(percentUsed)));
```
ErikSin: same comment as above, pct doesn't read as percentage to me

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function useStorageReadingQuery() {
+  return useQuery({
```
ErikSin: Making this a suspense query will simplify your code

ErikSin (changes requested):
The overall architecture is good, and my requests for changes are not structural, but have to do with some formatting.

functions should be scoped and not have to deal with null/undefined values. `isLowStoarge` and `calcUsePct` should only take number (not null/undefined). If the values are null and undefined than do an early return eg: `const percentage  = (freeBytes===undefined || totalBytes===undefined)?0:calcUsedPct(freeBytes, totalBytes)`.

Also the insertion of null in the catch statement of getReading() is adding alot of verbosity, and i think we should just get rid of it completely. If there is an error, it should throw an error (which useQuery already does). And that way freeBytes and totalBytes will always be a number. Also using useSuspensQuery instead will simplify your code. 

Lastly, you have a very simple but very verbose styled badge (the exclamation mark that is used in the menu and home header). The parent already deals with the logic, of whether it shows or not. So i think it should just be a reusable component so that you are not copying and pasting the quite verbose styles.

ErikSin:
Also your test setup must have taken you a long time to do...impressive that you had the patience to determine all those mocks!

src/frontend/sharedComponents/Storage/LowStorageBanner.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/screens/MenuScreen.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/sharedComponents/HomeHeader.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/screens/MenuScreen.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/sharedComponents/Storage/MenuLowStorageAlert.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/fe25440ecde96c660f1934cb899ed03aa80cbeeb

src/frontend/sharedComponents/ExclamationBadge.tsx
ErikSin, line 1: i think this should go in the `sharedComponent/Storage` folder

src/frontend/hooks/useStorageReadingQuery.ts
```diff
@@ -0,0 +1,19 @@
+import DeviceInfo from 'react-native-device-info';
+import {useSuspenseQuery} from '@tanstack/react-query';
+
+export const STORAGE_QUERY_KEY = ['device', 'storage', 'reading'] as const;
```
ErikSin: sorry, i didnt look at this earlier. This doesnt need to have 3 strings. The order of strings determine a heiracrchy. So if there are other querys that are related, you build this heiracrchy so it is easier to invalidate. In this case it is unnecesssary as there are no related queries, so you can just call it  `['storageReading'] as const` or something like that

src/frontend/sharedComponents/Storage/MenuLowStorageAlert.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <View style={styles.detailsColumn}>
+        <BodyText variant="tinyMeta" style={styles.onlyLeft}>
+          {formatMessage(m.onlyLeft, {free})}
+        </BodyText>
+
+        <View
+          accessible
+          accessibilityRole="progressbar"
+          accessibilityValue={{now: percentage, min: 0, max: 100}}
+          accessibilityLabel={formatMessage(m.onlyLeft, {free})}
```
ErikSin, line 71: this deviates from how we are using other accessibilty labels. its not necessarily a problem, and you have been the one doing most of the accessbility stuff. but if we want to use these for accessibility reasons, then we should start translating them like you have done here. But that is going to cause a lot of problems for out e2e test. 

Just flagging it as something we need to consider.

src/frontend/sharedComponents/HomeHeader.lowStorage.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+  return {
+    __esModule: true,
+    useStorageReadingQuery: () => ({
+      data:
+        mockFreeBytes == null || mockTotalBytes == null
+          ? null
+          : {freeBytes: mockFreeBytes, totalBytes: mockTotalBytes},
+    }),
+    isLowStorage: (free: number | null, threshold: number = LOW) =>
+      (free ?? Infinity) <= threshold,
```
ErikSin: I dont think this needs to be mocked anymore (but i could be wrong)

ErikSin (approved):
whoops, sorry i meant to approve it!

src/frontend/sharedComponents/HomeHeader.lowStorage.test.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1359/commits/9ef711f428a9d74df0c98a09c2b2a45736706f9e
You were right! Good catch!

src/frontend/hooks/useStorageReadingQuery.ts
cimigree: OK fixed. Thanks for the explanation. I didn't really understand this before.
https://github.com/digidem/comapeo-mobile/pull/1359/commits/9ef711f428a9d74df0c98a09c2b2a45736706f9e

src/frontend/sharedComponents/ExclamationBadge.tsx
cimigree: Yes! Sorry I forgot about that folder.
https://github.com/digidem/comapeo-mobile/pull/1359/commits/9ef711f428a9d74df0c98a09c2b2a45736706f9e

src/frontend/sharedComponents/Storage/MenuLowStorageAlert.tsx
cimigree: Yes good point. Maybe for all future E2E we should rely on test ids instead of the accessibility Labels and make them translatable. Do we know anything about if/ how many people use them (the screen readers, that is)?

PR #1360, opened 2025-08-26
chore: Change app version name to major.minor

### LINKED ISSUE #1276: Change version naming to `${minor}.${patch}` to align with our own naming
Currently the app version is named `${major}.${minor}.${patch}`, e.g. `v1.2.0`. We have agreed not to use this naming because the confusion that three components to a version cause, and the fact that the `major` version does not have a significance for an app.

At the very least, the version shown to the user on the about page should be `${minor}.${patch}`, but to avoid inconsistencies with what a user would see on the play store or in the Android app installation list, we should just change the naming of the version to `${minor}.${patch}`. Pre-releases should retain their suffix, but also drop the major version.

gmaclennan:
Also adds tests to verify app name is correct for various inputs

Cleans up code for type safety

Allows version name to be overridden with environment variable (necessary for later support of incrementing pre-release versions, because it can't be done in this config file, which is called multiple times during build)

Fixes #1276


ErikSin (changes requested):
Sorry, we should document this, but e are intentionally keeping all config files as js files (using jsdoc for ts checking) to avoid any compiling issues

gmaclennan:
> Sorry, we should document this, but e are intentionally keeping all config files as js files (using jsdoc for ts checking) to avoid any compiling issues

Ok, I will change, although there are no compile issues with this file, because TS is supported for app.config.ts. What compile issues are you trying to avoid?

app.config.js
```diff
@@ -1,81 +1,169 @@
...
+}
+
+/**
+ * Validates that the version name, if provided, is of the format:
+ * <minor>.<patch>[-<pre-release>[.<pre-release-patch>]]
+ * where:
+ * - <minor> is a non-negative integer
+ * - <patch> is a non-negative integer
+ * - <pre-release> is one of: dev, rc, pre
+ * - <pre-release-patch> is a non-negative integer
```
ErikSin, line 120: `<minor>`, `<patch>`, `<pre-release>`, `<pre-release-patch>` need to be in back-ticks in order to show up properly

ErikSin (approved):
I small non blocking comment, but otherwise looks good

PR #1361, opened 2025-08-26
fix: projects long load

### LINKED ISSUE #1345: Long loading time / All projects
https://www.notion.so/digidem/Loading-time-long-on-All-Projects-screens-24d1b08162d580499252d1906439738e

cimigree:
towards #1345

### Speed up “All Projects” & “Exchange” coming from Menu.

**Changes**

#### 1) All Projects

* Add per-project Suspense  + basic card so the list renders immediately and each card fills as its own data comes in instead of having a whole screen loader.
* Uses useCallback for  `handlePressId` so that a new function isn't created on every render
* Use props to help the flatlist be as efficient as possible: `keyExtractor`, `initialNumToRender`, `windowSize`, `removeClippedSubviews`.

#### 3) Exchange (SyncScreen)

* Restructure into three tiny components/ gates so that the components can call their own hooks and so other unnecessary components aren't mounted and unnecessary calls aren't made. Sorry for the nesting, but this seemed like the best way to make the screen load faster.

Before change was made with 8 projects:
https://github.com/user-attachments/assets/0b1df658-8f8d-4762-b4fe-66b96dbdf9ae

After changes in this PR with same projects:
https://github.com/user-attachments/assets/a5d0238a-9e0c-4f94-8238-34be2584f323


src/frontend/screens/Exchange/index.tsx
ErikSin, line 1: this highlights what I think you've been talking about me not wanting nesting. Its not nesting that is the issue, its that nesting is often a symptom of architectural patterns that over complicate.

The first thing is that there is a bunch of prop forwarding going on:

the `SyncScreen` is passing the `projectId`, `wifiStatus`,`hasInternetAccess`, `onGoBack`, to the `soloExchangeChecker`. And then that is passing all those to `ExchangeWifiSwitch`. If the point is to break up the hooks, why not pass those hooks directly into the `ExchangeWifiSwitch`,  so that the `SyncScreen`, `soloExchangeChecker`, and `ExchangeSoloScreen` don't have to worry about loading those hooks. It also means you can get rid of the `soloExchangeChecker` completely, as you can just do that logic in the `SyncScreen` since there are no props that need to be passed.

So instead you would have
```tsx
export const SyncScreen = () => {

  //only 1 hook called at the top level
  const projectSettings = useProjectSettings();

  if (projectSettings.data?.name === undefined) {
    return <ExchangeSoloScreen />;
  }

  // all the other hooks are called inside this component and the sync screen doesnt need to worry about the loading state of those hooks until it is needed
  return <ExchangeWifiSwitch/>;
};
```
Now there is no need for `<SoloExchangeChecker/>` which was just prop forwarding and adding an unnecssary react component that we had to maintain.

So having the nested component makes sense, expecially `ExchangeWifiSwitch`, because it can encapsulate its own hooks, and the parents dont have to worry about the loading until it is actually being used.

src/frontend/screens/Exchange/index.tsx
```diff
@@ -1,64 +1,102 @@
...
 });
 
+const ExchangeLoader = () => <LoaderCard />;
+const PeersLoader = () => <LoaderCard />;
+const LoaderCard = () => (
+  <View
+    style={{height: 120, margin: 20, borderRadius: 12, backgroundColor: '#eee'}}
+  />
+);
+
```
ErikSin: Why not just use `<LoaderCard>` in place of `ExchangeLoader` and `PeersLoader`? Now we have to maintain ` ExchangeLoader` and `PeersLoader`, even though it is just forwarding the `LoaderCard`. Another example of where I think nesting shows a symptom of unnecessary complication.

src/frontend/screens/MenuScreen.tsx
```diff
@@ -177,6 +214,11 @@ export function MenuScreen() {
             </View>
           </MainMenuItemWrapper>
         </View>
+        <React.Suspense fallback={null}>
+          <PrefetchProjectsList />
+          <PrefetchAllProjectRoles />
+          <PrefetchExchangeQueries />
+        </React.Suspense>
```
ErikSin: This is really hacky, and I think a better approach would be to [prefetch](https://tanstack.com/query/latest/docs/framework/react/guides/prefetching) the queries via [usePrefetchQuery](https://tanstack.com/query/latest/docs/framework/react/reference/usePrefetchQuery). And then call those prefetch queries when the [menu is focused](https://reactnavigation.org/docs/navigation-events/#focus).

This would require us to expose the prefetch queries in core-react, but I think it is a cleaner alternative to creating these null react components.

ErikSin (changes requested):
The way of prefetching that you have is quite hacky and I don't think we shoud use react component for non react purposes. Its a little confusing, and I think will lead to subtle bugs down the line. 

I think we should write prefetch function inside core react and expose them as function that can be used by mobile. And then we can tie those prefetch functions to navigation life cycle

src/frontend/screens/MenuScreen.tsx
cimigree: Do you want me to do that or should I open a ticket in core react?

src/frontend/screens/Exchange/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1361/commits/e85de73921e71799b30472e2dcbf17bf139e5ef7

src/frontend/screens/Exchange/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1361/commits/e85de73921e71799b30472e2dcbf17bf139e5ef7

src/frontend/screens/MenuScreen.tsx
ErikSin: I think it would be quite simple to just do. So i would say attempt it, but time box it to a few hours, and if it becomes complicated for some unforeseen reason, we can move write an issue and attempt it next sprint

src/frontend/screens/AllProjects.tsx
```diff
@@ -43,63 +75,58 @@ export const AllProjects: NativeNavigationComponent<'AllProjects'> = () => {
...
+  <View
+    style={{
+      height: 112,
+      borderRadius: 12,
+      backgroundColor: '#eee',
+      marginBottom: 20,
+    }}
+  />
+);
+
```
ErikSin: While i think this looks cool, I think we need approval from design before we can introduce this

src/frontend/screens/Exchange/index.tsx
```diff
@@ -1,64 +1,57 @@
...
+  exchangeTitle: {id: 'screens.Sync.headerTitle', defaultMessage: 'Exchange'},
 });
 
-export const SyncScreen = ({navigation}: NativeRootNavigationProps<'Sync'>) => {
-  const wifiStatus = useLocalDiscoveryState(state => state.wifiStatus);
+const LoaderCard = () => (
+  <View
+    style={{height: 120, margin: 20, borderRadius: 12, backgroundColor: '#eee'}}
+  />
+);
```
ErikSin: Same comment, I think we need approval from design before we can introduce this

ErikSin (commented):
Those loader cards look really cool, but we need to get some sort of input from design before we introduce. It might be worth asking in the design channel.

On a more architectural level, what you introduced does make the code overall better (so worth merging), it doesn't really fix the fundamental problem. Prefetching will help, but i think we should dig into what is actually causing such a delay, and figure out if we can fix that via the backend or through core-react

ErikSin:
[Decided in slack](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1757001895440029?thread_ts=1756998744.728429&cid=C6ZMBFATZ) that we are adding loading ui to `AllProjects` screen.

ErikSin:
Can we take out all the stuff related to exchange for now? We can address that at a later time (if the backend fixes have not already fixed the issue

cimigree:
@ErikSin I removed the changes from the exchange screen.

src/frontend/screens/AllProjects.tsx
```diff
@@ -43,63 +75,58 @@ export const AllProjects: NativeNavigationComponent<'AllProjects'> = () => {
...
+        initialNumToRender={6}
+        windowSize={5}
+        removeClippedSubviews
+        renderItem={renderItem}
       />
     </View>
   );
 };
 
+export const ProjectCardLoader = () => (
```
ErikSin: this export can be removed

src/frontend/screens/AllProjects.tsx
```diff
@@ -25,13 +25,45 @@ const m = defineMessages({
   },
 });
 
+type ProjectListItem = ReturnType<typeof useManyProjects>['data'][number];
```
ErikSin, line 28: interesting, i didn't know you could use "number" as opposed to 0!

ErikSin (approved):
Looks good, just one small non blocking comment

PR #1365, opened 2025-08-27
chore: hardcoded release name on about page

### LINKED ISSUE #1277: Add named release name to the about screen.
The "About" screen should also show the name of the release, e.g. the name we assign to each "named release". What we put there prior to the first "named release" has not yet been decided, but could be something like "Preview" or "Beta".

cimigree:
closes #1277 

Adds a hardcoded release name to the about page. I called the label release name. Hope that makes sense!
Added new labels to E2E test also. I actually got rid of the colon (:) after emojis after I took the screen shot though.

about

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -448,11 +448,19 @@
   },
   "screens.AboutSettings.emojiSource": {
     "description": "Label for the source of emojis used in CoMapeo",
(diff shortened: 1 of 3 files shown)


ErikSin (approved):
Thanks!

PR #1367, opened 2025-08-28
fix: don't stop and start media server

### LINKED ISSUE #1364: Don't stop and start fastify server when app is backgrounded
Currently we use the FastifyController from Core to stop the fastify server from listening when the app is backgrounded, and we start listening again when the app comes into the foreground. We currently start listening on a random port each time. This creates several issues:

1. URLs for assets from the server (photos, icons, audio files) become stale after the app returns from the background, which means the assets won't load unless we re-request the URL (with the new port).
2. The server will not close while any connections are still open. Sometimes clients can connect with keep-alive, which means that the connection remains open with a timeout, even after a request. The result is that when the user backgrounds the app, then returns to it, the server has not completed closing. The server start is queued, and any requests for updated URLs are queued too, awaiting the server start. This results in icons, images etc. taking a long time to load (until the keep-alive connection timeouts). It's unclear what this is on mobile - as I understand it the underlying library for loading images is okHttp, which uses a connection pool with keep-alive somewhere between 10 seconds and 5 minutes, and fastify also has a default keep-alive of 72 seconds. The result is going to be random delays with closing connections, which results in assets not loading if the user returns to the app quickly.

Issue (2) is likely a major cause of `getIconUrl()` requests and other `getXUrl()` requests taking excessive amounts of time in the Sentry-reported spans.

Issue (1) can be solved by re-using the same port number, if it remains available (it would be very unlikely for a randomly-selected port to be used by a different app between a user backgrounding the app and returning to it), however that will not solve issue (2).

The historical reason that we are stopping and starting the server in the background was because of a hypothesis that leaving an open listening port in the background resulted in Android power management shutting down the node process, which was a cause of back-end crashes in earlier versions of Mapeo Legacy. My understanding of how Android power management works and how nodejs-mobile works has come a long way since then and I think this is not actually an issue that we should be concerned about. nodejs-mobile runs an embedded version of node in a thread in the same process as the app MainActivity (there is only one process for the whole app). If Android does decide to shut down the app because of power management (e.g. it is using too much CPU or memory in the background) then it will shut down the entire process. This has minimal consequences for the user - we persist all data and it will just result in a slower load-time when the user returns to the app. In addition, I have not seen anything since that suggests that Android OS would be more likely to terminate the app process because it is listening on a port, although the specifics of how Android power management works is very mysterious and also varies between manufacturers.

TL;DR, I don't think we should stop servers when the app is backgrounded, because
- It causes the issue (2) above because closing the server waits for open connections to terminate.
- There is no evidence that it is the cause of Android terminating the process earlier while the app is in the background.
- If Android OS does terminate the app process while it's in the background, it is largely transparent to the user, other than a slower app-open time when they return to the app.

We can start gathering some metrics on app cold-starts vs warm-starts to see whether there is any issue, and if we think we could reduce cold starts by minimizing memory or CPU usage in the background, then we can re-visit stopping servers in the background if we have some evidence that this is actually an issue. Otherwise, this is causing more problems than it is solving.

gmaclennan:
fixes #1364

Removes FastifyController import, which is no longer needed.

### DIFF
--- a/src/backend/src/app.js
+++ b/src/backend/src/app.js
@@ -2,7 +2,7 @@ import debug from 'debug'
 import { join } from 'path'
 import { mkdirSync } from 'fs'
 import { createRequire } from 'module'
-import { MapeoManager, FastifyController } from '@comapeo/core'
+import { MapeoManager } from '@comapeo/core'
 import { createMapeoServer } from '@comapeo/ipc/server.js'
 import Fastify from 'fastify'
 import * as Sentry from '@sentry/node'
@@ -76,7 +76,6 @@ export async function init({
   mkdirSync(customMapsDir, { recursive: true })
 
   const fastify = Fastify()
-  const fastifyController = new FastifyController({ fastify })
 
   const manager = new MapeoManager({
     rootKey,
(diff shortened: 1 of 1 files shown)


RangerMauve (approved):
LGTM!

PR #1370, opened 2025-09-02
chore: integrate post hog provider

ErikSin:
This is a proof of concept and not a finalized version. I would just like some eyes on whether this is the right approach, or if you think there is a better way to do this.

Challenges:
- The Post Hog Provider is the best way to autocapture events (except for screen events-see below). Without using the provider we have to manually capture each thing we want to keep track of and send it to posthog.
- Unfortunately [react navigation v7 does not allow for us to manually capture each screen](https://posthog.com/docs/libraries/react-native#with-react-navigationnative-and-autocapture) and we need to manually use postHog in the navigation container.
- IMO This a flaw with the API => In order to capture the screen information, the `NavigationContainer` needs access to `postHog`. Typically the `postHog` object is accessed via `usePostHog`, which is retrieved from the `PostHogProvider.` BUT the `PostHogProvider` must be a child of the NavigationContainer, so the NavigationContainer does NOT have access to `usePostHog`. So to circumvent this, we need to [instantiate the postHog object](https://posthog.com/docs/libraries/react-native#without-the-posthogprovider), pass that object the Navigation Container and also pass that to the PostHogProvider.
- PostHog [Internally keeps track of whether it is opted in or out](https://posthog.com/docs/libraries/react-native#opt-inout). This is persisted. So we did not want to have 2 sources of truth, and also persist whether the user has opted in to post hog. So the persisted store only sets WHEN the value was set to persisted. When the user opt in to postHog, we call `postHog.optIn()`, AND set the persisted `optedInStartDate`.


gmaclennan:
Thanks Erik, this is indeed a tricky tangle of different moving pieces.

- One important point to note, that is not documented anywhere in the posthog docs, is that persisted props (like `posthog.optedOut`) are not guaranteed to be loaded when you access them. By default posthog [uses expo filesystem](https://github.com/PostHog/posthog-js/blob/main/packages/react-native/src/native-deps.tsx#L121) to persist data, and loads it all async when you create the posthog instance, so `optedOut` will start as the the property passed to `defaultOptedOut`, then be set to the persisted value after `posthog.ready()` resolves (also not documented!). One way of simplifying this is to use mmkv for our storage - the [posthog storage class](https://github.com/PostHog/posthog-js/blob/main/packages/react-native/src/storage.ts) is designed to work with sync storage too, and would not have this issue.
- The posthog docs are misleading / wrong about the posthog provider needing to be a child of the NavigationContainer - this is only a requirement for autocapture to work (which is only supported for nav v6 and lower). If we are turning off autocapture of screens (as recommended), then posthog [doesn't try to access navigation context](https://github.com/PostHog/posthog-js/blob/00601e3725b249a0f9a9d7d69c583928e1bbfd5d/packages/react-native/src/PostHogProvider.tsx#L105), so it doesn't need to be a child of the NavigationContainer (the docs suggest that it should be in the example they provide, but then they say to use `usePosthog()`, so the docs are just wrong IMO)
- Your `onStateChange` prop for NavigationContainer looks good, just get the posthog instance with `usePosthog()` (which you can do now that you can put the posthog container outside)
- I think we should maintain our own opt-in/out state, not rely on the posthog one. Partly because of the async storage issues mentioned above, but mainly because it makes switching to something different really hard. I know that means we need to keep the two states in sync, but that would not be too hard. My recommendation for this is to pass the posthog instance to the createStore function and in the setOptIn(true|false) action, call `posthog.optIn/optOut`.
- Having two different contexts for the same thing, and AppUsageStats provider doesn't make sense to use as a provider because as far as I understand it does two things: (1) run a side-effect to opt-out after twelve months, and (2) create two wrapper functions to call posthog.optIn/Out and set the optInStartedAt state. Neither of these are stateful and belong in a context. For (1) I think the easiest option is to do this in App.js outside the component when you create the stores e.g.
    ```
    const { optedIn, optInStartedAt } = appUsageStatsPromptStore.instance.getState()
    if (optedIn && optInStartedAt < Date.now() - ONE_YEAR) {
      actions.optOut()
    }
    ```
    For (2), passing the posthog instance as a constructor option for `createAppUsageStatsPromptStore()` would you allow you to define an optIn/out action there, keep the state source-of-truth in that store, and call posthog optIn/out in the same action.
- In my view the conditional rendering of screen groups is not something we should do for determining which screen to show at startup, because it creates the kind of issues that you are dealing with in the two AppUsage screens and not being able to control navigation. Instead I would set `initialRouteName` dependent on `hasDoneOnboarding` and `shouldShowAppUsageOptIn`. We _should_ use conditional rendering for auth screens, although I see a difference between what's on `develop` (which I think is problematic because it conditionally renders the entire stack, not just the screens) and what is on this branch (which navigates to the auth-screen as a side-effect).
- Once we are managing our own opt-in state, we can pass it to the posthog provider as the `defaultOptIn` value, just to side-step the weirdness that might come from posthog's async persisted state loading.
- Remember that all the stores and actions can be accessed and used in App.js outside the component - it's super useful for setup and a lot easier than wiring things up in React.

ErikSin:
> * One important point to note, that is not documented anywhere in the posthog docs, is that persisted props (like `posthog.optedOut`) are not guaranteed to be loaded when you access them. By default posthog [uses expo filesystem](https://github.com/PostHog/posthog-js/blob/main/packages/react-native/src/native-deps.tsx#L121) to persist data, and loads it all async when you create the posthog instance, so `optedOut` will start as the the property passed to `defaultOptedOut`, then be set to the persisted value after `posthog.ready()` resolves (also not documented!). One way of simplifying this is to use mmkv for our storage - the [posthog storage class](https://github.com/PostHog/posthog-js/blob/main/packages/react-native/src/storage.ts) is designed to work with sync storage too, and would not have this issue.

Yes, lets just switch to mmkv

> The posthog docs are misleading / wrong about the posthog provider needing to be a child of the NavigationContainer - this is only a requirement for autocapture to work (which is only supported for nav v6 and lower). If we are turning off autocapture of screens (as recommended), then posthog [doesn't try to access navigation context](https://github.com/PostHog/posthog-js/blob/00601e3725b249a0f9a9d7d69c583928e1bbfd5d/packages/react-native/src/PostHogProvider.tsx#L105), so it doesn't need to be a child of the NavigationContainer (the docs suggest that it should be in the example they provide, but then they say to use usePosthog(), so the docs are just wrong IMO)

>Your onStateChange prop for NavigationContainer looks good, just get the posthog instance with usePosthog() (which you can do now that you can put the posthog container outside)

I will use the usePostHog Hook since the provider does not need to be a child of the navigation container

> I think we should maintain our own opt-in/out state, not rely on the posthog one. Partly because of the async storage issues mentioned above, but mainly because it makes switching to something different really hard. I know that means we need to keep the two states in sync, but that would not be too hard. My recommendation for this is to pass the posthog instance to the createStore function and in the setOptIn(true|false) action, call posthog.optIn/optOut.

>Remember that all the stores and actions can be accessed and used in App.js outside the component - it's super useful for setup and a lot easier than wiring things up in React.

I don't fully agree with this, but lets discuss it tomorrow at the meeting. I just dont love the idea of state syncing, but it might be the only reasonable thing to do here.

> Having two different contexts for the same thing, and AppUsageStats provider doesn't make sense to use as a provider because as far as I understand it does two things: (1) run a side-effect to opt-out after twelve months, and (2) create two wrapper functions to call posthog.optIn/Out and set the optInStartedAt state. Neither of these are stateful and belong in a context. For (1) I think the easiest option is to do this in App.js outside the component when you create the stores e.g. 

Yeah, i agree. The use of a provider here was an odd choice. I was just caught up in the posthog provider iniatilization that i think i just got confused

>For (2), passing the posthog instance as a constructor option for createAppUsageStatsPromptStore() would you allow you to define an optIn/out action there, keep the state source-of-truth in that store, and call posthog optIn/out in the same action.

>Remember that all the stores and actions can be accessed and used in App.js outside the component - it's super useful for setup and a lot easier than wiring things up in React.

Yes, and... We have discussed this as a front end team, and we are trying to keep the persisted state providers scoped to persisted state. By scoping, it separates the concern, and allows for easier testing. 

Secondly, if we are using the post hog provider, then we would have to access the postHog instance via the usePostHog hook. This means, the createAppUsageStatsPromptStore() would have to be instantiated inside react. Alternatively, we could create a singleton instance of postHog, and then pass that instance to the createAppUsageStatsPromptStore(), but then our custom optIn() and optOut() functions are changing the instance directly instead of using the {optIn, optOut} = usePostHog(). Which is not necessarily a problem, but it just think that code is a little confusing, having a react provider, but updating the instance outside of the provider. 

>Once we are managing our own opt-in state, we can pass it to the posthog provider as the defaultOptIn value, just to side-step the weirdness that might come from posthog's async persisted state loading.

Is this necessary if we are using MMKV?

gmaclennan:
> Secondly, if we are using the post hog provider, then we would have to access the postHog instance via the usePostHog hook. This means, the createAppUsageStatsPromptStore() would have to be instantiated inside react. Alternatively, we could create a singleton instance of postHog, and then pass that instance to the createAppUsageStatsPromptStore(), but then our custom optIn() and optOut() functions are changing the instance directly instead of using the {optIn, optOut} = usePostHog(). Which is not necessarily a problem, but it just think that code is a little confusing, having a react provider, but updating the instance outside of the provider.

When using the posthog provider, you pass an instance of posthog to it. That's the instance that you get back from the `usePosthog()` hook. You don't need to instantiate the store inside react. The {optIn, optOut} from the usePostHog() hook are exactly the same as the ones on the posthog instance you create outside of React and pass to the provider - it's the same instance:

```js
const posthost = new Posthog()
const appUsageProviderStore({ persist: true, posthog })

const App = () => {
  // ...
    <PosthogProvider client={posthog}>
      <AppUsageProvider store={appUsageProviderStore}>
```

ErikSin:
> When using the posthog provider, you pass an instance of posthog to it. That's the instance that you get back from the `usePosthog()` hook. You don't need to instantiate the store inside react. The {optIn, optOut} from the usePostHog() hook are exactly the same as the ones on the posthog instance you create outside of React and pass to the provider - it's the same instance:
> 
> ```js
> const posthost = new Posthog()
> const appUsageProviderStore({ persist: true, posthog })
> 
> const App = () => {
>   // ...
>     <PosthogProvider client={posthog}>
>       <AppUsageProvider store={appUsageProviderStore}>
> ```

Yes, I understood what you meant originally but I think your not fully understanding what im trying to convey. What i don't like about this architecture is there is slightly circular **react** logic. AppUsageProvider is updating the singleton instance of postHog. And that posthog instance is propogated via the provider. But the AppUsageProvider is also a child of that provider. In the react world, we should be updating the provider instance since its a child, but were circumventing it by updating the instance directly (which is the same instance as the provider instance). Its functionally fine to do it how your suggesting, it just feels a little circular.

cimigree:
Not sure if my comments are needed here, but since you asked... And also, I may be misunderstanding some of the back and forth, but...
- I think it makes sense to pass the PostHog instance into createAppUsageStatsPromptStore({ posthog }), and add actions.optIn/optOut that both update our state and call posthog.optIn/optOut. That way we have that info in case we decide to leave posthog.
- It would be great if you could make shouldShowAppUsagePrompt to a pure function again: shouldShowAppUsagePrompt(state, now?). It should not depend on PostHog or read optedOut. That will keep the tests simple and the function portable.
- That's great we can place the PostHogProvider outside the NavigationContainer (since we’re not using autocapture) and then use usePostHog() to track screen views.
- For the 12-month auto opt-out, probably good to not add another provider. Maybe we can read the store state once on startup and if optedIn && optInStartedAt < now - 365d, call actions.optOut(). 
- If we can set the  initialRouteName based on three flags: needs device onboarding, needs project onboarding, needs usage prompt; otherwise fall through to the app would that work? Not sure if we can do that but that is an idea if we don't want to render different navigation flows
- Using MMKV for PostHog storage to avoid async loading issues sounds good
- I don’t think we should delete the context test—not sure if that was an accident or just temporary

Hope this helps somewhat.

ErikSin:
I have an alternative:

1. Create a wrapper file that just exports plain js function

```ts
import PostHog from 'posthog-react-native';

let posthog: PostHog | null = null;

export function initAppUsageAnalytics(apiKey: string, options?: object) {
  posthog = new PostHog(apiKey, options);
}

export function optInAppUsageAnalytics() {
  if (!posthog) return;
  posthog.optIn();
}

export function optOutAppUsageAnalytics() {
  if (!posthog) return;
  posthog.optOut();
}

export postHogGetClient(){
 if(noPostHog){
  throw new Error("post hog client has not been initialized")
 }
 return posthog
}

```

2. the `appUsageProviderStore` imports the needed function directly (aka dont pass in an entire instance of postHog)

```ts
import optInAppUsageAnalytics, optOutAppUsageAnalytics} from "../path/to/analytics"

function createAppUsageStatsPromptStore({persist:boolean}){
 {...restOfStore}
 
 action={
  optIntoAppUsageStats:()=>{
   optInAppUsageAnalytics()
   store.setState({
    optedIn:true
   }),
  optOutAppUsageStats:()=>{
     optOutAppUsageAnalytics()
     store.setState({
      optedIn:false
     })
  }
 }
}

```


step 3: initialize the App Usage Analytics in App.tsx

```ts
import {initAppUsageAnalytics} from "../path/to/analytics"

// outside of react life cycle
initAppUsageAnalytics()

export const App = () => {
 return(<...EntireApp/>)
}

```

step 4: initialize post hog provider

```ts
import {postHogGetClient()} from "../path/to/analytics"
const posthog = postHogGetClient('aPi-KeY')

export const AppProviders(){
 (<PostHogProvider client={posthog}>
   <...otherProviders>
    ...
 )
}
```

We introduce a lightweight wrapper that exposes simple functions like optInAppUsageAnalytics() and initAppUsageAnalytics(). This wrapper encapsulates all PostHog-specific logic, keeping the rest of the app completely unaware of the underlying analytics provider. This will make it easier to switch out 

Post Hog requires its client instance to be passed to the provider, but since this is specific to PostHog, we isolate it within the wrapper and only expose it where absolutely necessary. This way, if we replace PostHog in the future, the change stays contained — no cascade effects or refactoring of other components required

Benefits:
- If we ever switch from PostHog to another tool, we only update the wrapper.
- No unnecesssary analytics SDK code leaks into React components.
- Easier testing: the tests dont rely on what specific sdk we are using


Also, i do agree with you that our store should keep its own persisted state about whether the user has opted in or not. It will make it easier if we switch out post hog

ErikSin:
Closing as we have come to a decision on how we do this

PR #1372, opened 2025-09-03
chore(deps): update prod deps

ErikSin:
closes #1371 

# What was skipped:
- mmkv: latest only compatible with new architecture)
- react-native-gesture-hander: the newer versions require newer versions of react-native.
- Expo related deps


cimigree (approved):
After a brief hiccup with me being on the wrong npm version, all went smoothly. 
I was concerned about react-native-reanimated, but I tested out what wasn't working before -- that is when tracks were active, I could not touch anything on the map (such as the add observation button) but that worked fine this time with the latest upgrade.
In addition to testing on my local devices (Pixel 8a on Android 16 and Samsung Galaxy A25 on Android 15, I tested on Browserstack on a Pixel 8 Pro on Android 14 and Pixel 10 on Android 16.

PR #1378, opened 2025-09-08
feat: success screen after data export

### LINKED ISSUE #1376: Data Export / Success screen
Description - Once an export is complete, user comes back to the Obserrvation list screen and sees the bottom sheet open.
There’s no success screen and the bottom sheet remains open.

See Figma files :
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=7726-1964&t=kpELh5NDebt6Je8w-1

This was designed without considering we are now saving to storage.

cimigree:
closes #1376 

This PR adds a success screen after successfully saving whichever kind of data.
The biggest changes I made were because of needing to type the export result.
I navigated to the regular error bottom sheet, but if we want to do something more like the import Map Error Bottom Sheet, I can amend the PR.

Sorry about the flash on the Observations List before going to the Success Screen. I don't know how to prevent that and would love to hear suggestions.

https://github.com/user-attachments/assets/77aea3c2-6f24-4681-939d-25f0a94f2355

### DIFF
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -111,6 +111,7 @@ import {sharedPhotoPreviewNavOptions} from '../../screens/PhotoPreviewModal/shar
 import {ConfirmPasscodeBottomSheet} from '../../screens/AppPasscode/ConfirmPasscodeSheet.tsx';
 import {CreateProjectScreen} from '../../screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/CreateProject.tsx';
 import {NameSoloProjectScreen} from '../../screens/Settings/CreateOrJoinProject/CreateOrNameSoloProject/NameSoloProject.tsx';
+import {ExportSuccess} from '../../screens/ExportSuccess.tsx';
 
 export const TAB_BAR_HEIGHT = 70;
 
@@ -445,6 +446,11 @@ export const createDefaultScreenGroup = ({
         name="DraftPhotoPreviewModal"
         component={DraftPhotoPreviewModal}
         options={DraftPhotoPreviewModalNavOptions({intl})}
+      />
+      <RootStack.Screen
+        name="ExportSuccess"
+        component={ExportSuccess}
+        options={{headerShown: false}}
       />
     </RootStack.Group>
     <RootStack.Group
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -493,9 +493,7 @@ export const createDefaultScreenGroup = ({
         name="ExportSuccess"
         component={ExportSuccess}
         options={{
-          headerShown: false,
           contentStyle: {backgroundColor: 'white'},
-          animation: 'none',
         }}
       />
     </RootStack.Group>
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -447,11 +447,6 @@ export const createDefaultScreenGroup = ({
         component={DraftPhotoPreviewModal}
         options={DraftPhotoPreviewModalNavOptions({intl})}
       />
-      <RootStack.Screen
-        name="ExportSuccess"
-        component={ExportSuccess}
-        options={{headerShown: false}}
-      />
     </RootStack.Group>
     <RootStack.Group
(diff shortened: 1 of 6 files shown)


ErikSin:
Why are creating and maintaining an export type? Would it not be simpler to navigate to the success screen on `onSuccess` regardless without injecting this manual type?

And then to deal with the "cancel" we can just do `if (isUserCancelled(err)) return` in the catch?

cimigree:
Because here is what I thought happens. onSuccess fires any time the mutation function resolves. If we return inside a catch for the cancel case, that still counts as a resolution, so onSuccess runs.

If we always navigate in onSuccess, we’ll show a Success screen even when the user cancels and nothing was saved. We also don’t want cancel to be an error/ show an error. So onSuccess needs something (an “outcome”/ return/ that then needs an export type) to tell saved vs cancelled, otherwise we can’t avoid the false success or unwanted error.

Or am I wrong about this?

ErikSin:
You are correct, but you could still bypass the creation of the type by exporting the  `isUserCancelled`, and using `isUserCancelled` directly in the onError. 

Eg:

```ts

// inside screen/exportObservation.tsx
exportAndShare.mutate(
      {exportType: typeToExport},
      {
        onSuccess: outcome => {
          if (outcome === 'saved') {
            navigation.replace('ExportSuccess', {exportType: typeToExport});
          }
        },
        onError: err => {
          if(isErrorCancelled(err)) return
          Sentry.captureException(err);
       }
)

```

cimigree:
@ErikSin Ok I put your code back as much as possible and everything seems to work fine. Thanks for the tips and guidance.

src/frontend/screens/ExportSuccess.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <BodyText
+            style={{
+              textAlign: 'center',
+              color: NEW_DARK_GREY,
+              paddingHorizontal: 20,
+            }}>
+            {t(m.exportBody)}
+          </BodyText>
+        </View>
+      </View>
```
ErikSin: this might not work for all translations. All langugage may not have the format of subject followed by the verb. I think we need to make the message:

exportBody={
   defaultMessage: `{export} have been downloaded to your device.`
}


I think there is a way to format directly inside default message. But if not, I think the priority should be translatable, and we change the design to follow that.

src/frontend/Navigation/Stack/AppScreens.tsx
```diff
@@ -488,6 +489,15 @@ export const createDefaultScreenGroup = ({
         name="ConfirmPasscodeSheet"
         component={ConfirmPasscodeBottomSheet}
       />
+      <RootStack.Screen
+        name="ExportSuccess"
+        component={ExportSuccess}
+        options={{
+          headerShown: false,
+          contentStyle: {backgroundColor: 'white'},
+          animation: 'none',
```
ErikSin: why did we turn off the animation?

ErikSin (approved):
Looks good, one small thing(not blocking, but dont merge until fixed). the message should be translatable. If we can't format it directly inside the default message then we should have design change it before your merge

src/frontend/Navigation/Stack/AppScreens.tsx
cimigree: I was trying to eliminate the flash when replacing the bottom sheet select with the success. And also, animation is none as the default...so I can probably get rid of that and the headerShown: false...I just wasn't sure as I am passing  in a new option but turns out setting it on the group does the trick.

src/frontend/screens/ExportSuccess.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1378/commits/dc0f06fd96e2a98b532ab31a766375e0c4c8aa04

What do you think?
image
image

src/frontend/Navigation/Stack/AppScreens.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1378/commits/dc0f06fd96e2a98b532ab31a766375e0c4c8aa04

cimigree:
@ErikSin Can you check the body message (now the whole thing is translated and formatted)? And let me know if it is ok to merge?

ErikSin:
Yeah. looks good to merge. Can you let Sabella know that you had to change her original format because it was not translatable please! Thanks.

PR #1384, opened 2025-09-09
chore: update comapeo dependencies

cimigree:
related to https://github.com/digidem/comapeo-core-react/pull/105
Updates comapeo/core-react and comapeo-core and comapeo/schema to enable ability to more easily change someone's role in a project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "@bam.tech/react-native-image-resizer": "3.0.11",
-    "@comapeo/core-react": "6.1.1",
+    "@comapeo/core-react": "6.2.0",
     "@comapeo/ipc": "5.0.0",
     "@expo-google-fonts/rubik": "0.4.1",
     "@formatjs/intl-getcanonicallocales": "2.5.5",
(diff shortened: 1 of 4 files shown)


package.json
achou11, line 1: need to update `@comapeo/schema` to `2.1.1` to match the version in core:

https://github.com/digidem/comapeo-core/blob/29a006c9038ca38403c81388dee738fa813e758f/package.json#L165

(unrelated, but core should be pinning to an exact version)

src/backend/package.json
achou11, line 1: same comment about updating `@comapeo/schema`

achou11 (changes requested):
Couple of blocking changes. Also reminds me to look into updating `@mapeo/mock-data` to account for latest schema changes. Unsure if that will also block this PR

achou11:
> Also reminds me to look into updating @mapeo/mock-data to account for latest schema changes. Unsure if that will also block this PR

Briefly looked into this and seems like no changes are needed there, so shouldn't be a blocker here 👍

src/backend/package.json
cimigree: Hmm. comapeo/schema is not an explicit dependency in the backend package.json. Should it be? @achou11

achou11: ah sorry no i don't think it should be - forgot that we don't make it explicit in the backend and rely on core defining the right version.

package.json
cimigree: https://github.com/digidem/comapeo-mobile/pull/1384/commits/8039c1c09125ab10860b1fce015c8bc148263a46

achou11 (approved):

PR #1385, opened 2025-09-10
fix: Fix inifinite loop in messagePort.removeEventListener()

gmaclennan:
I don't _think_ any of our code currently calls `removeEventListener` in the backend MessagePortLike, but calling it would have been bad.


achou11 (approved):
Wow good catch 😅

PR #1387, opened 2025-09-10
chore: formatting observation metadata sharing

### LINKED ISSUE #1381: Update 3rd party share - Obvs Metadata
Update:

- Lat-Lon- 5 decimals
- UTM-zero decimals
- All except for speed 0,
- speed 2 decimals points
- Add subject line

Add in subject :

- Add project, category, date

cimigree:
closes #1381 
Formats data in observation metadata sharing as desired.
I made some assumptions when making the regex(es), so I hope they are right but would like to know if otherwise. 
It does not support commas, it doesn't support numbers like this .5 (no digits before the decimal), no scientific notation. 
It looks to me like the subject is there, so I am not sure what the ask is. I reordered it and put commas in between instead of dashes. Is that all that is needed? Do I take out the "Comapeo data sent from"?

Screenshot:
image


src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -231,12 +231,38 @@ export const ObservationMetadata: NativeNavigationComponent<
...
+    });
+  }
+
+  function formatValueForShare(key: string, raw: number | string): string {
+    const decimals =
+      key === 'latitude' || key === 'longitude' ? 5 : key === 'speed' ? 2 : 0;
+
+    if (typeof raw === 'number') return raw.toFixed(decimals);
+    return handleNumberInString(raw, decimals);
+  }
```
ErikSin: this approach is a little fragile and requires implicit developer knowledge. `key` being any string, and transforming the value to zero decimals unless it is lat, lon, or speed is highly specific, and the introduction of any new values will automatically round that new value to 0, which may not be the intended purpose.

I think a better approach would be to update the values in the `listData` array directly using Number.toFixed(). This also would mean you would not have to use `handleNumberInString` as all the values are numbers before they are turned into strings

I know this would update the value in comapeo (before it is shared). But i think that is an ok compromise

ErikSin (changes requested):
Thanks for doing this. I understand why you did it the way you did here, and it makes alot of sense given the context that we want to change the share values but not the values in comapeo. But i think its adds a complexity that is not worth it, and so I am making the executive decision to just change the value directly so it is the same that is being shared and the same value in the app

cimigree:
Ok sounds good. Good job executiving!

I left the formatting for location alone as you didn't explicitly object to that.

Also, I fixed an oversight from before and changed the subject to have just the date (as was specified in the issue) and not the date and time as it used to.

image

---
As you can see from the picture below, we have the numbers as links, and since we can't really fix that without messing up the ability to copy/ paste, I just wanted you to be aware.

image

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -256,13 +273,17 @@ export const ObservationMetadata: NativeNavigationComponent<
...
         ? ''
         : `${formatMessage(m.location)}: ${formatCoords({lat, lon, format: coordinateFormat})}`;
 
+    const formattedLocation = baseLocation
+      ? roundUtmInLocationLine(baseLocation, coordinateFormat)
+      : '';
+
     await openShare.mutateAsync({
-      subject: `${projectName} - ${formatDate(createdAt, {format: 'long'})} - ${categoryName}`,
+      subject: `${projectName}, ${categoryName}, ${formatDate(createdAt, {dateStyle: 'long'})}`,
```
ErikSin, line 286: Can you tell me the difference between `{format: 'long'}` and `{dateStyle: 'long'}`? I am having trouble finding documentation for it.

ErikSin (approved):
Looks Good, thanks!

src/frontend/screens/ObservationMetadata.tsx
cimigree: Yes. format: long includes the time. That is what we have on observation sharing and it comes from what we defined in IntlContext.
```

export const formats: CustomFormats = {
  date: {
    long: {
      day: 'numeric',
      month: 'short',
      year: 'numeric',
      hour: '2-digit',
      minute: '2-digit',
    },
  },
};
```
I did not want the time included, so I used dateStyle (which you had also done in ObservationMetadata 5 months ago)
${formatDate(createdAt, {dateStyle: 'full'})}`;
I accidentally transposed long instead of full, but it was a happy accident because I believe it gives me what I want! It looks like from google searching that full gives you the day of the week also, which seems unnecessary in a subject.

This is a PR from when that was first added 10 years ago:
https://github.com/tc39/ecma402/issues/108#issue-183158971

Some docs:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

and 

https://cldr.unicode.org/translation/date-time/date-time-patterns#basic-date-formats

PR #1388, opened 2025-09-11
fix: fix query key for useGetCustomMapInfo

achou11:
`getMapsQueryKey` is a function that returns the base key as an array so in order to build other query keys on top of it, it needs to be called and flattened.


cimigree (approved):
Good catch! How did you realize?

achou11:
> Good catch! How did you realize?

Currently working on custom maps for desktop so was reviewing the implementation here 😄

PR #1390, opened 2025-09-11
chore(dep): Expo Upgrade

ErikSin:
All things considered, this was a pretty smooth transition to expo.

## Notable updates:
- The [newer version of expo breaks jest tests](https://github.com/expo/expo/issues/36831), a mock was added in the jest setup filed
- Simplified the passcode tests. The passcode e2e tests were being flaky and there was a lot of redundancy in the tests
- Some changes related to react 19 typings. There are more changes that can be made that are being caught by eslint (but they are all warnings).


app.json
```diff
@@ -19,7 +19,6 @@
         "@rnmapbox/maps",
         {
           "RNMapboxMapsImpl": "mapbox",
-          "RNMapboxMapsVersion": "11.1.0",
```
cimigree, line 22: Not sure if this is what caused the issue, but when I go to save a Track (after recording it, adding observations, having it show on a map), I get this error:
image

This happens on both of my devices and when I get to the Track categories screen after pressing Save Track.

image

I see the error, but once I dismiss it, I can still save the track and then it seems to show up on the map and be fine.

Can you verify?

Joining a project and exchange worked fine, and I checked other things from the E2ENoGos doc to make sure I was checking functionality and that tracks/ mapbox thing was the only issue I saw. Sharing an observation with a photo has the same issue it did before the fix I did in here:
https://github.com/digidem/comapeo-mobile/pull/1392

cimigree (changes requested):
Please check on Mapbox error. See comment for details.

app.json
ErikSin: If i add it back it causes this issue: 
image

ErikSin: But i am looking into the error you found now!

ErikSin: Ok, i figure it out [here](https://github.com/digidem/comapeo-mobile/pull/1390/commits/ff62eb1ae2aab03d01004cf7125ad671c534a0b9)

src/frontend/screens/MapScreen/index.tsx
```diff
@@ -172,12 +172,8 @@ export const MapScreen = ({
           {isFinishedLoadingStyle && authState !== 'obscured' && (
             <>
               <RemoteDetectionAlertsMapLayer />
-              {isTracking && (
-                <>
-                  <CurrentTrackMapLayer />
```
ErikSin, line 177: this was causing a race condition and causing [this error](https://github.com/digidem/comapeo-mobile/pull/1390#discussion_r2356680772):

Essentially endTracking was being called, which sets isTracking to be false. The location tries to update the layer, but react tears it down, causing mapbox to say " message=Source routeSource is not in style" aka does not exist. 

Since this map layer is based on persisted state of the location in tracks, which are cleared when there are no tracks, we dont need to conditionally render it with this check here

app.json
ErikSin: Reason here: https://github.com/digidem/comapeo-mobile/pull/1390#discussion_r2356968394

cimigree:
Sorry, @ErikSin But I am still getting that same error. It is just in a slightly different place. Now it appears when I return to the map after saving. Again, I can still see the Track and it seems to save just fine.

I uninstalled and reinstalled the app and cleared the android cache and it still happened on both my devices. Are you testing Tracks out and this is not happening for you?

image

image

cimigree (changes requested):
Sorry, @ErikSin But I am still getting that same error. It is just in a slightly different place. Now it appears when I return to the map after saving. Again, I can still see the Track and it seems to save just fine.

I uninstalled and reinstalled the app and cleared the android cache and it still happened on both my devices. Are you testing Tracks out and this is not happening for you?

image

image

ErikSin:
Ok, i figured it out. The CurrentTrackLayer was causing a race condition. When we cleared the track, we also stopped rendering it -both in the `<Map/>` which i fixed earlier, and directly in the CurrentTrackLayer with the `locationHistory.length > 1 && <ShapeSource.../>`

So it is now ALWAYS being rendered in the react tree, but it only shows the line when there are points.

I also realized the reason we are conditionally rendering it was because we wanted to always show the latest location in the current map track. The map uses the throttled location, but we always wanted the most precise location. having the most precise location is draining on the battery, hence why we only wanted to turned on when tracks was on.

But i fixed this by just turning on the most precise location in the map ONLY when tracking is set, and passing that location directly to the current map layer so that it does not need to be conditionally rendered

cimigree (approved):
@ErikSin I changed the appium version to use in Browserstack to match what I am using locally and it seems to work better?
 The Tracks are now working/ not crashing when saving although I do think we might need to refactor them at some point.
 Is this below something we should be concerned about or take note of? I see this warning when I do the expo prebuild.
` Using react-native@0.79.5 instead of recommended react-native@0.79.6.`
 And what does this mean? `android: EDGE_TO_EDGE_PLUGIN: No configuration found for `edgeToEdgeEnabled` field in the project app config, falling back to false. In Android 16+ (targetSdkVersion 36) it will no longer be possible to disable edge-to-edge. Learn more: https://expo.fyi/edge-to-edge-rollout`

PR #1392, opened 2025-09-11
chore: obs share whatsapp

### LINKED ISSUE #1382: Observation - Update 3rd party share pt 2
- Prioritize whats-app formatting: meaning email will show the markdown formatting (eg back ticks will be shown)
- Add date and time to subject

cimigree:
closes #1382 

This PR reverts the recent update to react-native-share. With the update, it no longer worked to share an observation that contained an image. It is documented in the repo here: https://github.com/react-native-share/react-native-share/issues/1687. They advise either a native reconfiguration or in the case of expo users, waiting for a fix.

This PR also adds the long form date (which contains the time) to the subject of the message. In What's App, there is no way to add a subject, so it is just the first line of the message.

I took the date out of the body because with the subject in the body, the message went date, date. It seemed like a mistake.

I also redid the formatting of the coordinates because Whatsapp automatically formats long numbers to be links for calling, which is not what we want at all! The fix involves adding zerowidth spaces to any numbers in the location greater than 5 digits long so that doesn't happen. There is no visible change in email.

No other formatting changes were specified, so I did not make any other changes.

Before the zero width spaces it looked like this:
image

Looks like this now:
image

And in email:
image


src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -198,18 +198,32 @@ export const ButtonFields = ({
         completedFields.push({label: field.label, value: displayedValue});
       }
 
-      const subject = `${t(m.comapeoData, {projectName: name || t(m.defaultProjectName)})} - _*${preset ? preset.name : t(m.fallbackCategoryName)}*_`;
-
       const date = formatDate(observation.createdAt, {format: 'long'});
 
-      const location =
+      // Insert a zero-width space after every 5 digits inside long numbers (>=6 digits) so it doesn't autolink
+      function breakLongNumbersForWhatsApp(s: string): string {
```
ErikSin: will this effect the number if it copy and pasted? I think if it does, than we should just deal with the autolink

ErikSin (commented):
I have one question (that im not sure is blocking or not)

cimigree:
@ErikSin Even though I was able to copy and paste with no difficulty or problems, my research says that those zero width spaces come along with the number and can break number parsers. Sorry about that!

So now we are back to:
image

ErikSin (approved):
Looks Good!

PR #1395, opened 2025-09-16
chore: Update core native deps

gmaclennan:
This updates the backend @comapeo/core dependencies to the latest versions:

- `fs-native-extensions@1.4.3`
- `simdle-native@1.3.7`
- `quickbit-native@2.4.8`
- `sodium-native@4.3.3`

These latest versions all switch from using node-gyp-build to using cmake and bare-addon. This required adding additional patches to nodejs-mobile-react-native build.gradle because bare-addon resolves addon modules to the prebuilds folder, where-as node-gyp-build resolved addons to the build/Release folder.

The prebuilds of fs-native-extensions and quickbit-native also required patches to the build in order to build successfully for Android, but these patches seem to work at runtime from testing in an emulator. See:

- https://github.com/digidem/fs-native-extensions-nodejs-mobile/commit/f4a58050fbfe117544c875cd3e91d100f2721a88
- https://github.com/digidem/quickbit-native-nodejs-mobile/commit/436033f1604992a310ab7149fdc920c6d881a3b1

PR #1398, opened 2025-09-16
chore: update root package lockfile

achou11:
Getting this diff when installing locally on latest develop (66fa9d55767fdb2c297889e01940d416ec59b9ff). Similar to what I observed in https://github.com/digidem/comapeo-mobile/pull/1395#discussion_r2352965605


cimigree (approved):
Looks good to me. Builds and runs fine.

PR #1399, opened 2025-09-16
fix: fix sqlite prebuilds for x86

gmaclennan:
This switches the prebuilds of better-sqlite3 to ones built with bare-make toolchain, instead of the previous node-gyp build. This seems to fix running the app on x86 architecture (previously the x86 builds of better-sqlite3 crashed the app).

The builds are generated by the updated workflow in https://github.com/digidem/better-sqlite3-nodejs-mobile/pull/5, which was run manually with v11.10.0 of better-sqlite3 and the `11.10.0-bare-make` release tag, so that the release would not conflict with the current node-gyp version we are using.

The question remains about how to actually build this for dev using an x86 emulator. Currently I have just enabled building APKs for all architectures, but this increases the APK size significantly. Looking for input on how to only enable x86_64 arch for local dev builds, and keep production apks to arm-only.

### DIFF
--- a/app.json
+++ b/app.json
@@ -81,7 +81,6 @@
         }
       ],
       ["./expo-config-plugins/removeExpoInputStyles.js"],
-      ["./expo-config-plugins/targetArmArchsOnly.js"],
       ["./expo-config-plugins/increaseJavaMemory.js"],
       ["./expo-config-plugins/flagSecure"],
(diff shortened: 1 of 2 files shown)


gmaclennan:
switching to bare-make build was addressed in #1517. x64 builds of the app can be addressed in a separate PR if needed.

PR #1415, opened 2025-09-24
feat: Adds Early access flag to the menu.

### LINKED ISSUE #1108: Updated Menu
**Description**
Separate CoMapeo between Early Access/Default Experience, thus creating a Dual Experience.
User can get access to new, not so perfect, features by opting in to the Early Access view. User can toggle this on or off.

**Feature spec**
https://www.notion.so/digidem/Dual-Experience-1811b08162d58000ab34f9ee02a8c991?pvs=4

**User requirements**
https://www.notion.so/digidem/Dual-Experience-1811b08162d58000ab34f9ee02a8c991?pvs=4#1891b08162d580c3b95efbd5a5be8f2a

**To Do**

Image

- [ ] In Menu/Settings, Introduce a **Early Access Mode** button 
- [ ] Under **Early Access Mode**  add dynamic text **Early Access is OFF** when Early Access is not activated
- [ ] Under **Early Access Mode**  add dynamic text **Early Access is ON** when Early Access is has been activated

## For testing

- [ ] Put the update to feature https://github.com/digidem/comapeo-mobile/issues/1286 under Early Access mode

**Design**
 https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=4897-2703&t=qWTxHCaQ67ERPDhL-1

cimigree:
closes #1108 
This PR adds a simple context/ store for the early access flag.
It also adds the item to the app settings menu.
It adds a todo (console log) for the actual screen to toggle it as that will be in another PR (it is a separate issue).
QUESTION: I was not sure where in the stack to put the Provider. Can you give me any advice about that?

image


src/frontend/contexts/EarlyAccessContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+  persist as createPersistedState,
+} from 'zustand/middleware';
+import {MMKVZustandStorage} from '../hooks/persistedState/createPersistedState';
+
+// NOTE: Do not change!
+export const STORAGE_KEY = 'EarlyAccessStore' as const;
+
+type EarlyAccessState = {
+  isEarlyAccessEnabled: boolean;
+};
```
ErikSin: There is a high chance that we are going to change this (I think we are going to persist what is being set as early access). So if you use valibot, migrating will be much easier down the line

src/frontend/contexts/EarlyAccessContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+  selector: (s: EarlyAccessState) => T,
+): T {
+  const {instance} = useEarlyAccessStoreContext();
+  return useStore(instance, selector);
+}
+
+export function useIsEarlyAccessEnabled(): boolean {
+  const {instance} = useEarlyAccessStoreContext();
+  return useStore(instance).isEarlyAccessEnabled;
+}
```
ErikSin: I don't think we should use this (or at least use it this way). This hook uses the entire store, and then returns a value from that store. While there is only one value in the store now, if other values were added, this hook would cause re-renders eveytime there is a change in state even if that state is unrelated to what this hook is returning

ErikSin (approved):
one semi blocking comment, but doesn't require a re-review when fixed

PR #1420, opened 2025-09-30
chore: order default presets

### LINKED ISSUE #1389: Use project-settings defaultPresets to render category chooser
The category chooser should use the [`defaultPresets` property of ProjectSettings](https://github.com/digidem/comapeo-schema/blob/main/schema/projectSettings/v1.json#L29) to determine the order of presets and which presets to show in the category chooser screen. The `point` array should be used for observations and the `line` array should be used for tracks. If the `defaultPresets` option is missing, or if any array is empty, fallback to display all presets in alphabetical order.

cimigree:
closes #1389 

This PR uses the default presets order set in the project settings for ordering the presets for Observations. Currently, they are just put in alphabetical order, but that is not the desired behavior. Also works for Tracks.

To me it made sense to do this in comapeo-mobile, not core, because 
1. Hooks and props should be agnostic; this ordering is UI/presentation logic.
2. It’s derived state (presets + settings → ordered list) and should be calculated when it is needed.
3. Keeps core’s data hooks lean instead of having the two-query coupling
But I could not remember if it was imperative to do this work in comapeo/core-react

It is used in the category choosers.
It has a fallback for Alpha sorting.

I added a jest test for various scenarios as this cannot be tested with E2E.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,5 +73,5 @@ jobs:
         run: npm run test:jest
         env:
           MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
-          COMAPEO_METRICS_URL: ${{ secrets.COMAPEO_METRICS_URL }}
+          COMAPEO_METRICS_URL: ${{ vars.COMAPEO_METRICS_URL }}
           COMAPEO_METRICS_API_KEY: ${{ secrets.COMAPEO_METRICS_API_KEY }}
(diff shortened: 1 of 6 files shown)


ErikSin:
> To me it made sense to do this in comapeo-mobile, not core, because
> 
>     1. Hooks and props should be agnostic; this ordering is UI/presentation logic.
> 
>     2. It’s derived state (presets + settings → ordered list) and should be calculated when it is needed.
> 
>     3. Keeps core’s data hooks lean instead of having the two-query coupling
>        But I could not remember if it was imperative to do this work in comapeo/core-react
> 
> 
> It is used in the category choosers. It has a fallback for Alpha sorting.
> 

The main reason we want this in core-react is because we want this to be shared by desktop.

I also think that the array that maintain the order and what presets are seen are an integral and universal part of how presets are seen and that there should be no variation outside of what is dictated by the array. In otherwords, I don't think it is the responsibility of the front end to determine the order. 

Lastly, I just think it will make it much easier to use. The ordering logic will be encapsulated and the front end won't have to deal with any of that logic.

cimigree:
@ErikSin Oh yeah. For desktop, too. That makes sense. I will close this and attempt the change in comapeo/core-react.

PR #1427, opened 2025-10-03
chore: posthog integration

ErikSin:
intergrates the work done in #1312 

Adds the post hog provider (missing the API key right now).

I also took out the `optedIn:boolean` because we can derive that value from optInStartedAt. In otherwords if optInStartedAt is null, we know that the user is not opted in, otherwise they are opted in

Lastly, I changed mmkvZustandStorage function name to  MMKVStoreInitializer. The initializer is a general function that can be used by posthog as well, so i took zustand out of the name. Alot of files are using this, so this pr looks like it touches a lot of files, but many of the files are just the name change


package.json
```diff
@@ -59,6 +59,7 @@
     "eventemitter3": "5.0.1",
     "expo": "53.0.22",
     "expo-application": "6.1.5",
+    "expo-asset": "11.1.7",
```
ErikSin, line 62: I had to add this because several of the tests required expo font, and expo font required this...im not sure why these tests were failing in the pr...

src/frontend/contexts/AppUsageStatsContext.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const actions = {
+    recordCompleteOnboarding: () => {
+      store.setState({completedOnboardingAt: Date.now()});
+    },
+    setOptedIn: (optedIn: boolean) => {
+      const now = Date.now();
+      const state = store.getState();
+
+      appUsageMetricsOptIn();
```
cimigree: Even if setOptedIn is false, it looks like you are calling `appUsageMetricsOptIn`, right? That doesn't seem right to me. Because it is this, 
```
appUsageMetricsOptIn: () => {
    postHog.optIn();
  }, 
```
right?

cimigree (changes requested):
Just some questions and one blocking issue see inline in code.
How can we test the auto-reset after 12 months? Can you add a test with mocked time?
Any thoughts on how to handle the prompt logic navigation wise? 
It's good that Posthog is disabled in dev/ tests. Good call!

src/frontend/contexts/AppUsageStatsContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1427/commits/d770cbae17ada809838ec224e849bad60b26afdc

cimigree (approved):

PR #1428, opened 2025-10-04
Chore drawer layout navigator

### LINKED ISSUE #1437: Add Drawer Menu
Replace the setting menu with the drawer menu

To Do
- [ ] Add drawer using [React Native Drawer Layout](https://reactnavigation.org/docs/drawer-layout/)
- [ ] Remove Menu Screen
- [ ] Add menu screen as the [renderDrawerContent
](https://reactnavigation.org/docs/drawer-layout/#renderdrawercontent)

ErikSin:
closes #1437 

I figured out a way to add a drawer without using the drawer navigator. React Navigation has the [Drawer-Layout](https://reactnavigation.org/docs/drawer-layout), which is just the drawer component and animation abstracted out of the navigator. So it allows us to create the drawer with the same animation without adding an entire navigator.

~~There are some complications to using it, so i just want to confirm that this actually the better library to be using~~

https://github.com/user-attachments/assets/97aba4d7-cb6b-40c9-ad70-06898599b76f

## Stipulation to using drawer layout over drawer navigator
- The android back button does not automatically close it, so we need to maintain a subscriber to listen for the back button being pressed - We would not need this is the drawer navigator
- ~~The menu does not close automatically when being navigated away from: Similar to above we need to maintain a subscriber to listen for navigation state changes, and close the menu manually when the navigation state changes - We would not need this is the drawer navigator~~

## Why I think we should still use the drawer layout over drawer navigator
- We are not actually using the drawer navigator as a navigator. We simply are using the drawer, but we dont actually have nested screens inside the navigator. The way we would be using it is just passing the `RootStackNavigator` inside of it as the only navigation child. Aka we are just using it for the drawer animation.
- Because we are going to list all the projects as a bottom sheet OVER the drawer, the navigation structure will become nested and overcomplicated. This requires the bottom sheet to be a sibling of the drawer, because if it is a child of the drawer, the bottom sheet will show up behind the drawer. This unnecessarily complicates the navigation as now we need a stack navigator as the parent. The navigation structure would look like
```ts
<ParentStack>
 <ParentStack.screen name="AllProjectBottomSheet"/>
 <ParentStack.screen name="Drawer" component = {
   <DrawerNavigator>
     <DrawerNavigator.Screen component = {RootStackNavigator}/>
   </DrawerNavigator> } 
 />
</ParentStack>

```
~~If we just use the drawer layout, we don't have to change our navigation structure at all. BUT with the drawer layout, we are manually closing the drawer when the user navigates away. Since our bottom sheets use react navigation, it is technically being navigated away when a bottom sheet opens. So it requires us to intercept the navigation event and NOT close it when the bottom sheet is the `AllProjectsBottomSheet`. [See example here](https://github.com/digidem/comapeo-mobile/pull/1428/files#diff-24bd48cb580c6b94aa80e2759bbbf2860bd57015a2b81df5fbc88bd1d514d85aR41)~~

Based on [this comment](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1759533316388469), i originally misunderstood the requirements. We do not want the menu to close when the user navigates away. Based on that it is **quite clear that using the Drawer Layout is the right approach** as it is the simplest AND the drawer navigator DOES NOT allow for us to keep the drawer open when navigating away


ErikSin:
@cimigree and @gmaclennan let me know what you guys think

ErikSin:
> @cimigree and @gmaclennan let me know what you guys think

I misunderstood the requirements initially. Im going to go ahead with using the drawer layout and do not require your input anymore :)

cimigree (approved):
This is a cool look. It is so much simpler than what we were talking about. No wonder you wanted me to take the drawer out of my onboarding PR!
It confuses me a little that the drawer is the wrapper? for the Tabs, but I guess it works, so that is the main thing.
And like I said, this works, and I was able to get the tests to pass.
This is a stepping stone and further details will need to be addressed I think? I feel like you have a plan for how everything is broken up and this piece (there is no issue connected to it) is necessary background. 
I assume you will be ok with passing the toggleDrawer prop through to the DrawerMenu so the behavior Aldo asks for can occur ("to go back to a main tab either **through the project card**, sliver of the tab showing still, or **gather observations** (which intentionally goes to the Map view) in some future PR.
The tests were time consuming to redo. I hope they are right for a little while!

PR #1429, opened 2025-10-04
feat: toggle app usage

### LINKED ISSUE #1285: App Usage Stats / New Module on Data & Privacy screen
**Description**
App Usage Statistics: how an individual user is using the app, eg how long a user had the app open, what screens and features they use (eg number of shares), but not information about data collected (because this is always in the context of a project which other participants must know about). This data can be very detailed eg which screens a user is visiting.
Goal :  integrate a passive opt-in for a fixed period of time to track app usage behavior

**Feature spec**
https://www.notion.so/digidem/Data-Metrics-App-Usage-Statistics-1f11b08162d580bb9cd8cb757f864ac3

**Interactive prototype**

https://www.figma.com/proto/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6296-3475&t=N69UmW1JW4FtjHC9-1&scaling=min-zoom&content-scaling=fixed&page-id=8%3A3&starting-point-node-id=6296%3A3475&show-proto-sidebar=1

**Figma**
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?t=MO12q80amltNyXCm-0

**To Do**

Image

- [ ] Integrate screen as designed in **Data & Privacy** , see figma for latest version
- [ ] If user did not allow sharing stats at app startup - leave box unchecked
- [ ] If user allowed sharing stats at app startup - leave box checked
- [ ] User can check or uncheck the box

ErikSin:
closes #1285

stacked on top of #1427 

image


cimigree (commented):
This looks like it will be good to go once you make the change on the parent branch that doesn't opt everyone in to Posthog sharing no matter what...

cimigree (approved):

PR #1432, opened 2025-10-05
chore: update project info on menu

### LINKED ISSUE #1424: Update ProjectInfoCard
**Description**
We received a set of feedback from our recent Brazil field trip. Several critical UX/UI elements created less than satisfactory results. 
Taking recent feedback into account, see what front-end UX/UI updates are possible that could help us start addressing critical issues without heavily impacting our Named Release timeline.

**Feature spec**
https://www.notion.so/digidem/UX-UI-Updates-for-Named-Release-2601b08162d580678348d79a2d33cdf0?source=copy_link

Image

Image

Image

Image

Image

## To Do
- [ ] remove "onPress" prop
- [ ] Update Solo Icon
- [ ] remove "style" prop
- [ ] add `useManyProjectsQuery` Into Card
- [ ] take out use of ProjectInfoCard for all projects
- [ ] Display device name when no project name, otherwise display project name
- [ ] Switch Project is visible only when other projects exist.
- [ ] Current Project Card displays Full Project Name, Device Role in this Project, and Short Description when available.
- [ ] Project Roles displayed under Project Name: Just You, Coordinator, and Participant.

ErikSin:
Closes #1424

stacked ontop of #1428 

image

image

image

I know we originally talked about updating the ProjectDescriptionCard. But i realized that it is only used in the menu, so it felt unnecessary to create its own component. BUT i did create a ColorCard, which is a reuseable component that allows us to change the background color, and simply takes a "child" component


src/frontend/sharedComponents/DrawerMenu.tsx
```diff
@@ -54,23 +55,40 @@ const m = defineMessages({
...
+  },
+  coordinator: {
+    id: 'Navigation.Menu.coordinator',
+    defaultMessage: 'Coordinator',
+  },
+  participant: {
+    id: 'Navigation.Menu.participant',
+    defaultMessage: 'Participant',
+  },
+  justYou: {
```
cimigree, line 70: Should be "Just You" (capital Y)

src/frontend/sharedComponents/DrawerMenu.tsx
```diff
@@ -54,23 +55,40 @@ const m = defineMessages({
     id: 'Navigation.Menu.allProjects',
     defaultMessage: 'All Projects',
   },
+  mappingOnOwn: {
+    id: 'Navigation.Menu.mappingOnOwn',
+    defaultMessage: 'You are mapping on your own.',
```
cimigree: You're (not You are)

src/frontend/sharedComponents/DrawerMenu.tsx
```diff
@@ -88,58 +106,72 @@ export function MenuScreen() {
...
+          <ColorCard backgroundColor={projectColor}>
+            <View style={{padding: 20, gap: 12}}>
+              <HeaderText variant="header2">
+                {role === 'solo'
+                  ? projectDetails.projectHeader
+                  : projectDetails.projectName}
+              </HeaderText>
+              <View
+                style={{
+                  flexDirection: 'row',
```
cimigree, line 118: I guess it is a question of style, but I usually prefer when the styles are more than 1 or 2 that they be a named variable. Is there a reason sometimes you name them and sometimes you don't?

src/frontend/sharedComponents/DrawerMenu.tsx
```diff
@@ -88,58 +106,72 @@ export function MenuScreen() {
...
-          </View>
+              {(role === 'solo' || projectDescription) && (
+                <BodyText style={{color: NEW_DARK_GREY}}>
+                  {role === 'solo'
+                    ? formatMessage(m.mappingOnOwn)
+                    : projectDescription}
+                </BodyText>
+              )}
+              {allProjects.length > 1 && (
+                // This button deviates from the standard SecondaryButton (the icon is aligned flex-end) and so instead of changing that component, I just copied the styles here, and created a custom button for this use case.
```
cimigree, line 149: This button is so long it kind of clutters up the file. Any chance it could be its own component? And that might make it easier to find in case the Buttons as a whole get updated and we need to update this.

cimigree (approved):
Fix typos please.
Two non blocking suggestions.
And I think it would be great if we had a unit test for useProjectRoleAndDetails.ts
Something in particular for participant vs coordinator since we can't have partcipants in the E2E and also, we can't check colors in E2E, so that would be helpful too. 
  
  But maybe

src/frontend/sharedComponents/DrawerMenu.tsx
ErikSin: I personally think it is better to not turn it into a reusable component. We do not want it being used anywhere else (this is an exception to all the other buttons). And i don't think verbose code is necessarily a problem. Do you feel strongly about making it its own component?

src/frontend/sharedComponents/DrawerMenu.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1432/commits/976a35b72680c993accbf3ad8f7aec6bd21eace0

src/frontend/sharedComponents/DrawerMenu.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1432/commits/976a35b72680c993accbf3ad8f7aec6bd21eace0

src/frontend/sharedComponents/DrawerMenu.tsx
ErikSin: This is a style preference—unless we want to create specific style guide rule about it (because it effect consistency or clarity in a major way) I dont think we need to change it right now 😄

PR #1434, opened 2025-10-07
fix: e2e alert acceptance

cimigree:
Makes tests work better
Make a helper to handle the Alert
Adjusts alert handling to look for Save button
Which works on Samsung, which is what this changes to use
Had to make a couple of other small changes to keyboard handling for Samsung use also


ErikSin (approved):
Test are passing! Thanks for doing that

Issue #1430, opened 2025-10-04
Fix e2e


ErikSin:
e2e on CI is having a success rate of less than 50%, due to flaky tests that are not representative of actual bug in our code. We should either remove or update the flaky tests, as they are wasting alot of our time.

PR #1435, opened 2025-10-08
chore: refactor create observation screen

ErikSin:
Towards #1430 

With the e2e test we noticed that the "No GPS" alert was being prompted when there clearly was a GPS signal. This Pr attempts to address that

First, we removed the useEffect that was updating the save button. On every change of location (which is often), we were running a use effect to update the save button because the save button was found in the header-the header is only able to updated via a use effect. Instead the save button is just accessing the values directly from the persisted store

Secondly, I simplified the location hooks. Previously the useMostAccurateLocationforObservation hook was returning the location being used as well as updating the location in the persisted store. We are now just using the location from the store everwhere.

Lastly, I removed the `Editor` component from the observation create. Instead i took the individual pieces from the editor component and put the directly in Observation Create so that the code would not have deeply nested logic that made it difficult to debug.

### DIFF
--- a/src/frontend/screens/ObservationCreate/LiveLocationView.tsx
+++ b/src/frontend/screens/ObservationCreate/LiveLocationView.tsx
@@ -1,0 +1,22 @@
+import {usePersistedDraftObservation} from '../../hooks/persistedState/usePersistedDraftObservation';
+import {Divider} from '../../sharedComponents/Divider';
+import {LocationView} from '../../sharedComponents/Editor/LocationView';
+import {useMostAccurateLocationForObservation} from './useMostAccurateLocationForObservation';
+
+export const LiveLocationView = () => {
+  useMostAccurateLocationForObservation();
+  const position = usePersistedDraftObservation(
+    store => store.value?.metadata?.position,
+  );
+
+  return position ? (
+    <>
+      <Divider />
+      <LocationView
+        lat={position?.coords?.latitude}
+        lon={position?.coords?.longitude}
+        accuracy={position?.coords?.accuracy}
+      />
+    </>
+  ) : null;
+};
--- a/src/frontend/screens/ObservationCreate/LiveLocationView.tsx
+++ b/src/frontend/screens/ObservationCreate/LiveLocationView.tsx
@@ -5,18 +5,16 @@ import {useMostAccurateLocationForObservation} from './useMostAccurateLocationFo
 
 export const LiveLocationView = () => {
   useMostAccurateLocationForObservation();
-  const position = usePersistedDraftObservation(
-    store => store.value?.metadata?.position,
+  const lat = usePersistedDraftObservation(store => store.value?.lat);
+  const lon = usePersistedDraftObservation(store => store.value?.lon);
+  const accuracy = usePersistedDraftObservation(
+    store => store.value?.metadata?.position?.coords.accuracy,
   );
 
-  return position ? (
+  return lat && lon ? (
     <>
       <Divider />
-      <LocationView
-        lat={position?.coords?.latitude}
-        lon={position?.coords?.longitude}
-        accuracy={position?.coords?.accuracy}
-      />
+      <LocationView lat={lat} lon={lon} accuracy={accuracy} />
     </>
   ) : null;
 };
--- a/src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
+++ b/src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
@@ -1,0 +1,259 @@
+import {useCreateDocument} from '@comapeo/core-react';
+import {Observation} from '@comapeo/schema';
+import {useAuthContext} from '../../contexts/AuthContext';
+import {
+  usePersistedDraftObservation,
+  usePreset,
+} from '../../hooks/persistedState/usePersistedDraftObservation';
+import {
+  useCreatePhotoAttachment,
+  useCreateAudioAttachment,
+} from '../../hooks/server/media';
+import {useDraftObservation} from '../../hooks/useDraftObservation';
+import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
+import {STORAGE_QUERY_KEY} from '../../hooks/useStorageReadingQuery';
+import SaveCheck from '../../images/CheckMark.svg';
+import {
+  isProcessedDraftPhoto,
+  isUnsavedAudio,
+} from '../../lib/attachmentTypeChecks';
+import {IconButton} from '../../sharedComponents/IconButton';
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
+import {
+  Attachment,
+  ClientGeneratedObservation,
+  Position,
+} from '../../sharedTypes';
+import * as Sentry from '@sentry/react-native';
+import {useTrackActions, useTrackState} from '../../contexts/TrackStoreContext';
+import {Alert, AlertButton, View} from 'react-native';
+import {UIActivityIndicator} from 'react-native-indicators';
+import {useQueryClient} from '@tanstack/react-query';
+import {defineMessages, useIntl} from 'react-intl';
+
+const MAXIMUM_ACCURACY = 10;
+
+const m = defineMessages({
+  noGpsTitle: {
+    id: 'screens.ObservationCreate.SaveButton.noGpsTitle',
+    defaultMessage: 'No GPS signal',
+    description: 'Title of dialog when trying to save with no GPS coords',
+  },
+  noGpsDesc: {
+    id: 'screens.ObservationCreate.SaveButton.noGpsDesc',
+    defaultMessage:
+      'This observation does not have a location. You can continue waiting for a GPS signal, save the observation without a location, or enter coordinates manually',
+    description: 'Description in dialog when trying to save with no GPS coords',
+  },
+  weakGpsTitle: {
+    id: 'screens.ObservationCreate.SaveButton.weakGpsTitle',
+    defaultMessage: 'Weak GPS signal',
+    description: 'Title of dialog when trying to save with low GPS accuracy.',
+  },
+  weakGpsDesc: {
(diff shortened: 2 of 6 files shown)


ErikSin:
I merged you e2e updates. But before i merge those into this PR, i would like to see if this work, independent of your e2e updates, help with the tests

ErikSin:
well...something completely un related is failing...

src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts
```diff
@@ -33,44 +32,40 @@ export function useMostAccurateLocationForObservation() {
     updateObservationPosition({position: undefined, manualLocation: false});
   }
 
-  useFocusEffect(
-    useCallback(() => {
-      if (!permissions || !permissions.granted || isLocationManuallySet) return;
+  useEffect(() => {
```
cimigree: Why not useFocusEffect? I had an issue where if I try to add an observation without granting location permission, when I go back to the app, after adding permission, nothing changed in the observation.
Wouldn't a useFocusEffect rerun when returning to the app?
Or maybe add permissions?.granted as a dependency?
I feel like useFocusEffect would also save battery and code energy.
If I turn off permission mid creation of an observation, it still has a location from before I revoked permission. Is that intended behavior?

src/frontend/screens/ObservationCreate/LiveLocationView.tsx
```diff
@@ -0,0 +1,22 @@
+import {usePersistedDraftObservation} from '../../hooks/persistedState/usePersistedDraftObservation';
+import {Divider} from '../../sharedComponents/Divider';
+import {LocationView} from '../../sharedComponents/Editor/LocationView';
+import {useMostAccurateLocationForObservation} from './useMostAccurateLocationForObservation';
+
+export const LiveLocationView = () => {
```
cimigree, line 6: When I enter manual coordinates they don't show in here. You got rid of:
```
const position = value?.metadata?.manualLocation
    ? {
        coords: {
          latitude: value.lat,
          longitude: value.lon,
        }
      }
    : value?.metadata?.position;
```
Is that intended?

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  },
+  keepWaiting: {
+    id: 'screens.ObservationCreate.SaveButton.keepWaiting',
+    defaultMessage: 'Continue waiting',
+    description: 'Button to cancel save and continue waiting for GPS',
+  },
+});
+
+export const ObservationCreateSaveButton = () => {
+  const value = usePersistedDraftObservation(store => store.value);
```
cimigree, line 78: QUESTION: Is this violating "Server State vs App State" principle?
The position in value.metadata.position comes from GPS (external/server state), but it's being stored in Zustand and reading it here.
Should we be reading current GPS state directly instead of from the store?
What happens if GPS signal is lost after position was stored but before save?

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  function handlePressSave() {
+    if (!value) throw new Error('No observation saved in persisted state');
+
+    if (value.metadata?.manualLocation) {
+      saveObservation(value);
+      return;
+    }
+
+    const accuracy = checkAccuracyAndLocation(value.metadata?.position);
```
cimigree, line 208: Again, this checks position from the store, not current GPS state.
Old code checked `liveLocation` returned from the hook.
If GPS signal is lost after position was stored, it won't be detected. Does that matter?

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+        },
+      ]);
+    }
+    addNewTrackObservation({
+      docId: observation.docId,
+      versionId: observation.versionId,
+    });
+  };
+
+  function checkAccuracyAndLocation(
```
cimigree, line 137: The old code checked if liveLocation exists (active GPS signal).
The new code only checks if position exists in the store but it could be stale.
This means we won't detect if GPS signal is lost between when position
was stored and when user presses save. Does that matter?

src/frontend/screens/ObservationCreate/index.tsx
```diff
@@ -49,314 +55,149 @@ const m = defineMessages({
...
-  return (
-    <Editor
-      presetName={presetName}
-      PresetIcon={
-        <PresetCircleIcon
-          size="medium"
-          iconId={preset?.iconRef?.docId}
-          testID={`OBS.${preset?.name}-icon`}
-          color={preset?.color}
+        <LiveLocationView />
```
cimigree, line 115: This Component has no props and pulls everything from hooks/stores internally, but, per our architecture guidelines, I thought: "Props should be agnostic" 
 Would it make sense to do this? `<LiveLocationView position={position} />`
That might make this dependency explicit and the component easier to test in isolation?

cimigree (changes requested):
I think switching out the Editor is a good idea. So much less confusing. I mean it had 11 props I think! I like the idea of pulling out the button and that file is much easier to understand and read through.
Same for the LiveLocationView. 
ObservationCreate is now much flatter and easier to follow. You can see exactly what's being rendered without digging through all the prop forwarding.
And I think you have definitely reduced the rerenders because of removing that useEffect/ useCalback logic
Thanks for tackling this monster.

Generally, saving observations seems to work great, but I found what I think might be some bugs and I mentioned them above also near the code I think might be causing the issue.
1. Manual coordinates don't display and I think they should or they used to
2. Permission changes not detected when permission is granted mid-observation and I think it should be
3. Potential stale GPS data? Not sure if this is an issue. More like a call...

There are some architecture guideline violations, I think, which you can take or leave...
1. Server State vs App State violation (mentioned above)
2. Missing derived state (manual location -- see above)
3. Props not agnostic for LiveLocationView  (see above)

src/frontend/screens/ObservationCreate/index.tsx
ErikSin: Generally I would agree, but this is actually an exception to avoid alot of unnecessary re-renders. Essentially the location is constantly changing. So if we had the location at the top level of ObservationCreate component, everytime the location changed EVERYTHING in the component would get re-rendered. So putting the hook directly in the Location component, it only re-renders that location component and not the entire screen.

Arguable we could do this will all hooks, and make sure they are only scoped to the one component to avoid unnecessary re-render. But i think its usually not worth the trade off because of what you mentioned, it makes it harder to test the components in isolation. A couple extra re-renders feels fine for that reason. But when something is causing many re-renders constantly, we should scope it.

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
ErikSin: fun fact, `liveLocation` actually just returned the [value from the store](https://github.com/digidem/comapeo-mobile/blob/2c4e2063573cafb59d1eaf717d29c03c0b6af580/src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts#L73). That's why i decided to get rid of it. I originally designed it as a convenience thing, but it made the code confusing and misleading...

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
ErikSin: see comment above

src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
ErikSin: > QUESTION: Is this violating "Server State vs App State" principle? 

I can see why you think that, but `value.metadata.position` is techincally app state. The reason is that when the user creates an observation, we want to save the position of observation. But since we are not saving it to the db right away (because the user has to fill in the observation create forms), we are saving it to app state. 

As an optimization we are updating the position when the accuracy get better, because we know that usually over time accuracy gets better. So the logic for updating the `value.metadata.position` only updates when the accuracy gets better. It does NOT update on every change => therefore it NOT server state

src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts
ErikSin: Yeah, your right we should use useFocusEffect here.

The original reason I changed it to a useEffect is because it felt unnecessary to start and restart the subscription every time the screen came into focus. With the useEffect, the subscription started once, and the got torn down when the user navigated away. But this is causing the problem that when the permission changes, it does not catch that.

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1435/commits/05353be0e58b7261e4501c51162b051ebefa5ca7

src/frontend/screens/ObservationCreate/LiveLocationView.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1435/commits/fcae2e7c749bb408e062dae1aa5cfa91386d8e1a

cimigree (approved):
Your recent commits addressed the issues I found:
Back to useFocusEffect
`LiveLocationView` now reads from `value.lat`/`value.lon` - manual coords display correctly
Removed the location-clearing logic

**Working well:**

- Manual coordinate entry displays correctly
- Location persists when permission is revoked mid-observation 
- Basic observation creation flow works smoothly
- No GPS prompt for manual entry works as expected

**One issue found:**
- Permission changes mid-observation don't take effect without force-closing the app probably because the Expo useForegroundPermissions doesn't rerun without a force close, so the effect never runs ... Just a guess. Not sure how to fix it or if it is even worth addressing. Your call

PR #1442, opened 2025-10-09
fix: strange loading indicator behavior

cimigree:
Fixes replacement of screen with UI indicator by using flex 0

Screen recording
https://github.com/user-attachments/assets/a15678c7-1d01-482d-b663-9697808cdd6b


ErikSin (changes requested):
there is just one more minor (but blocking) navigation issue
- when the button are in a loading state the user is able to press the back button. I think we should[ prevent remove](https://reactnavigation.org/docs/preventing-going-back/) while loading

cimigree:
@ErikSin 
I added the prevention of back button during loading for the two screens.
I also used the hook to prevent back button use on the success screen because I figured we didn't want people going back to rename the device once they already did? Is that true?

ErikSin:
I think we should use the `usePreventRemove` hook instead. If you look at the "previous approach" section in [react navigation docs](https://reactnavigation.org/docs/preventing-going-back/), it gives reason as to why this is a better hook.

Its a little more versatile, and will work for iOS as well

src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -44,6 +45,8 @@ export const Success = ({
   const {data: deviceInfo} = useOwnDeviceInfo();
   const deviceName = deviceInfo.name || '';
 
+  usePreventAndroidBackButton();
+
```
ErikSin: I think the better strategy would be to take the other screens completely out of the stack using replace.

From the into screen, instead of doing `navigation.navigate('DataPrivacy')`, do `navigation.REPLACE('DataPrivacy')`.

Then from the Data Privacy screen do `navigation.replace('DeviceNaming')`.

Then from the Device Naming do `navigation.replace('Success')`.

Then when the user is on the success screen, there is not screen to go back to, and we dont have to worry about intercepting and cancelling this back action

cimigree:
I this what you mean? I wasn't sure how to get the screen to advance after/ during saving without that state variable. Feel free to take a crack at it if you have a better idea, but we need the flex

ErikSin:
> I this what you mean? I wasn't sure how to get the screen to advance after/ during saving without that state variable. Feel free to take a crack at it if you have a better idea, but we need the flex

Could you not just do:
`usePreventRemove(status === 'pending' ,()=>{});`?

cimigree:
> > I this what you mean? I wasn't sure how to get the screen to advance after/ during saving without that state variable. Feel free to take a crack at it if you have a better idea, but we need the flex
> 
> Could you not just do: `usePreventRemove(status === 'pending' ,()=>{});`?

No I could not. That is why I had to add the state variable. The screen would not advance after saving.

src/frontend/screens/Onboarding/DeviceNaming.tsx
```diff
@@ -80,7 +85,8 @@ export const DeviceNaming = ({
       },
       {
         onSuccess: () => {
-          navigation.navigate('Success');
+          setAllowNavigation(true);
+          navigation.replace('Success');
```
ErikSin: We actually don't need the usePreventRemove logic anymore. If the user presses the back button it puts the app in the background, but when they reopen the app, the information will be set. So they will just open the app again to the correct page (success screen). Previously when we weren't using replace, it would put them back a screen and it would be really confusing behaviour, that won't happen anymore.

src/frontend/screens/Onboarding/MapOnYourOwnIntro.tsx
```diff
@@ -55,13 +56,19 @@ export const MapOnYourOwnIntro = ({
   const {formatMessage: t} = useIntl();
   const {mutate: createProject, status} = useCreateProject();
   const {setActiveProjectId} = useActiveProjectIdActions();
+  const [allowNavigation, setAllowNavigation] = React.useState(false);
+
```
ErikSin, line 59: You don't actually need this. Since there is conditional navigation logic happening you can just use `usePreventRemove(status === 'pending', () => {})` and it should work

cimigree: https://github.com/digidem/comapeo-mobile/pull/1442/commits/f149680647471b41391abb61b8767e1fad217b4a

src/frontend/screens/Onboarding/DeviceNaming.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1442/commits/f149680647471b41391abb61b8767e1fad217b4a

ErikSin (approved):
great! That looks really good

PR #1445, opened 2025-10-13
WIP: use SVG preset icons

gmaclennan:
Proof-of-concept in-progress using SVG preset icons, with fallback to PNG. This required a patch to existing config import code which was throwing an error when importing SVG icons, but this will be fixed with the new category archive format once https://github.com/digidem/comapeo-core/pull/1135 lands.

To actually test with SVG icons, you will need to load this custom version of the default config: [mapeo-default-config-v5.0.0.zip](https://github.com/user-attachments/files/22892635/mapeo-default-config-v5.0.0.zip) (rename it to `.comapeocat` - Github won't allow attachments with that extension).

It does work and renders the icons, but for projects with only PNG icons this will be a bit inefficient, since it tries to load SVG first, then falls back to PNG. This should be addressed with https://github.com/digidem/comapeo-core/issues/1114 which will remove the need to attempt to load SVG first. However the extra request to get the new URL should not have a big overhead due to improvements in @comapeo/core-react to URL get functions.

I don't think this requires additional work now, until https://github.com/digidem/comapeo-core/pull/1135 is merged and released along with a new default categories file with SVG icons only. The goal of this PR is to validate how much work will be needed to use SVG icons once we have them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "expo-document-picker": "13.1.6",
         "expo-file-system": "18.1.11",
         "expo-font": "13.3.2",
-        "expo-image": "2.4.0",
+        "expo-image": "~2.4.1",
         "expo-image-manipulator": "13.1.7",
         "expo-localization": "16.1.6",
         "expo-location": "18.1.6",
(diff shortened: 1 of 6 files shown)


gmaclennan:
addressed in #1499

PR #1449, opened 2025-10-15
feat: uses presets sorted by default order hook

### LINKED ISSUE #1389: Use project-settings defaultPresets to render category chooser

cimigree:
closes #1389 
Uses the new comapeo/core-react hook as cerated in here: https://github.com/digidem/comapeo-core-react/pull/129
I did not choose to use the server/presets file as it seemed extremely simple to just use the hook from comapeo/core-react and then I didn't have to pass any props.
Adds use of preset chooser to the observation and tracks category choosers. 
No longer sorts them by alpha because all sorting is handled in the backend however all presets are included so they do need to be filtered by line or point.

**Tracks category chooser now**
image

**Observations category chooser now**
https://github.com/user-attachments/assets/612ac4dd-6141-466c-8717-a2a0dfac7c6d

### DIFF
--- a/src/frontend/screens/PresetChooser/ObservationCategoryChooser.tsx
+++ b/src/frontend/screens/PresetChooser/ObservationCategoryChooser.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {View, StyleSheet} from 'react-native';
 import {defineMessages} from 'react-intl';
-import {usePresetsQuery} from '../../hooks/server/presets';
+import {usePresetsSelection} from '@comapeo/core-react';
 import {useDraftObservation} from '../../hooks/useDraftObservation';
 import {usePersistedDraftObservation} from '../../hooks/persistedState/usePersistedDraftObservation';
 import {CategoryGrid} from './CategoryGrid';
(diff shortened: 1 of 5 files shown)


src/frontend/screens/PresetChooser/ObservationCategoryChooser.tsx
```diff
@@ -21,16 +22,20 @@ const m = defineMessages({
 export const ObservationCategoryChooser: NativeNavigationComponent<
   'ObservationCategoryChooser'
 > = ({navigation}) => {
-  const {data: presets} = usePresetsQuery();
+  const {projectId} = useActiveProject();
+  const presets = usePresetsSelection({
+    projectId: projectId,
+    dataType: 'observation',
+  });
```
achou11, line 32: shouldn't the `lang` be specified?

src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
```diff
@@ -20,11 +21,15 @@ const m = defineMessages({
 export const TrackCategoryChooser: NativeNavigationComponent<
   'TrackCategoryChooser'
 > = ({navigation}) => {
-  const {data: presets} = usePresetsQuery();
+  const {projectId} = useActiveProject();
+  const presets = usePresetsSelection({
+    projectId: projectId,
+    dataType: 'observation',
+  });
```
achou11, line 31: shouldn't the `lang` be specified?

achou11 (commented):
Potentially blocking comments but otherwise the general changes make sense to me. Haven't tested on device

achou11 (approved):

PR #1456, opened 2025-10-20
chore: update name project

### LINKED ISSUE #1426: Update Name Project Action
Naming a default (solo) project should also update the project to be named and create another nameless project in the background. Do no switch to this new project, the user should remain on the original (now newly named project)

## To do
- [ ] Create project after renaming project

cimigree:
closes #1426 
Creates an unnamed project when a user successfully renames a project. 
Doesn't switch to that project.
Updates the end to end test for a solo project to check for it. 
That test isn't currently running but it will when the menu PR is merged in.
Screen recording attached so you can see it in action.

https://github.com/user-attachments/assets/6f1cfa9b-e021-4b37-b339-5107fa952cb8


ErikSin (approved):
Looks Good, thanks!

PR #1462, opened 2025-10-21
fix: uses correct device type

### LINKED ISSUE #568: handle device type enum values from backend more thoroughly
I probably should've caught this in #449, although the code of interest isn't technically erroring out

There are certain parts of the app that are assuming that the only device types that may be surfaced are `mobile` and `desktop`. However, depending on the source, a device type could be expressed as a variety of other values, as is the case with the [`DeviceInfo['deviceType']`](https://github.com/digidem/mapeo-schema/blob/37bc87d396700e49eba74ff66f2de4c9357610cd/proto/deviceInfo/v1.proto#L16-L21), which is most relevant for listing out project members and receiving project invite information.

The most apparent examples are:

https://github.com/digidem/comapeo-mobile/blob/0384b1a534b4c98051cdb8b23a9ca4f0f9a210cb/src/frontend/sharedTypes/index.ts#L7

https://github.com/digidem/comapeo-mobile/blob/0384b1a534b4c98051cdb8b23a9ca4f0f9a210cb/src/frontend/sharedComponents/DeviceNameWithIcon.tsx#L63-L67

https://github.com/digidem/comapeo-mobile/blob/0384b1a534b4c98051cdb8b23a9ca4f0f9a210cb/src/frontend/screens/Settings/ProjectSettings/YourTeam/index.tsx#L175

These affect the following contexts:

- Invites flow
- Listing members in the `Your Team` screen within the Project Settings

In terms of the actual bugs that exist (or will exist) if left unchanged:

1. When viewing the project's team members, all of the icons display a mobile device icon instead of the one that matches the member's actual device type

2. When going through the flow of inviting a new member, the icon that's displayed may not match the member's actual device type (will show desktop icon even if the member's device type is either unspecified or a tablet).

### LINKED ISSUE #1346: Display correct device type icon in invitable devices selection screen
Currently working on desktop discovery and have reached a point where the mobile app can detect my desktop app. However, in the mobile app it's showing the inviteable desktop device as a mobile device despite when it should instead be showing as a desktop device i.e. using a desktop icon.

Image

Looks like there's a todo I left from a while ago to use the actual device type provided by core. Right now it's hardcoded as `'mobile'` 😅 :

https://github.com/digidem/comapeo-mobile/blob/6649868b39536d8f632b74e6709ec95c315adacf/src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx#L78-L79

cimigree:
closes #1346
closes #568 (maybe just starts on...)

- Uses actual device type when selecting a device for inviting instead of just hardcoding to mobile. 
- Tries to handle other types from the enum.
- I made it a function to try to use the DeviceType from types and just fall back to mobile for unrecognized/ undefined etc. Not sure if that is right...Thoughts?
- Updated Select Device to use our current text components (and made the choice to use the header font for the header sentence)
- We didn't have an image for tablet so I just use mobile for tablet. Maybe I will ask design? Is it ok for people to use a tablet?
Works well though for the internal desktop app version I downloaded though.
image


src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -73,12 +91,9 @@ function InvitableDevicesList() {
   return (
     <View style={styles.deviceListContainer}>
       {invitableDevices.map(device => {
-        const {deviceId, status, name} = device;
-
-        // TODO: Use `device.deviceType`
-        const deviceType = 'mobile';
+        const {deviceId, status, name, deviceType: rawDeviceType} = device;
+        const deviceType = normalizeDeviceType(rawDeviceType);
```
achou11: think it'd be clearer to do the narrowing where it's needed in the UI as opposed to here.

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -73,12 +91,9 @@ function InvitableDevicesList() {
...
 
-        // TODO: Update DeviceCard component to better handle potentially undefined fields, deviceType enum
```
achou11, line 81: think it's worth addressing this more thoroughly in this PR, even if there's some slight uncertainties about design questions e.g. which icon(s) to use. Can always do follow-ups to address those issues later if needed.

See [desktop implementation](https://github.com/digidem/comapeo-desktop/blob/c0ea014f6414fae04dbb77d53a017afa1f5a4e24/src/renderer/src/routes/app/projects/-shared/device-icon.tsx) as an example.

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
```diff
@@ -31,6 +33,19 @@ const m = defineMessages({
...
+): NonNullable<DeviceType> {
+  if (
+    !deviceType ||
+    deviceType === 'UNRECOGNIZED' ||
+    deviceType === 'device_type_unspecified'
+  ) {
+    return 'mobile';
+  }
+  return deviceType;
+}
```
achou11: related to other comments, but this feels incomplete and misplaced. Think this should be happening in the DeviceCard component to handle the calculation of the kind of icon to use, as opposed to redefining the "raw" device type

src/frontend/screens/Settings/ProjectSettings/YourTeam/SelectDevice.tsx
achou11: redefining the device type here feels misleading and prone to issues later on. would suggest deriving the UI-related piece of information based on the device type, as opposed to narrowing the device type.

achou11 (changes requested):
While this has the minimum changes to fix the original issue, I think more should be done to prevent similar issues in the future.

> We didn't have an image for tablet so I just use mobile for tablet. Maybe I will ask design? Is it ok for people to use a tablet?

Desktop uses the `tablet-android` icon from Material Icons, which should be available as a vector icon already.

cimigree:
@achou11 Is this what you had in mind? It didn't quite seem to make sense to me to use the DeviceCard so I used the end UI element instead

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
```diff
@@ -100,6 +97,35 @@ export const DeviceNameWithIcon = ({
...
+    case 'mobile':
+      return <DeviceMobile width={size} height={size} />;
+    case 'tablet':
+      return (
+        <MaterialIcons name="tablet-android" size={size} color={DARK_GREY} />
+      );
+    case 'desktop':
+      return <DeviceDesktop width={size} height={size} />;
+    case 'selfHostedServer':
+      return <DeviceArchive />;
```
achou11: non blocking: is it possible to use the vector icons instead of the svgs for these?

achou11 (approved):
Changes look good 👍 

> @achou11 Is this what you had in mind? It didn't quite seem to make sense to me to use the DeviceCard so I used the end UI element instead

yep! sorry I didn't look into where the actual changes should occur, but where you have them makes sense to me.

cimigree:
@achou11 I had to make some more than quick changes and am hoping you can look it over again real quick...
It looks the same as far as I can tell?

This is on the develop branch without the changes:

image

---
And here is the new version. Not much of a difference as far as I can tell, but what do you think?

image

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
```diff
@@ -100,23 +95,59 @@ export const DeviceNameWithIcon = ({
...
+        borderRadius: size / 2,
+        width: size,
+        height: size,
       }}>
-      <ShieldIcon
-        width={50}
-        height={50}
-        style={{position: 'absolute', top: 15, left: 15}}
-      />
+      <IconComponent name={icon} size={size * 0.5} color={DARK_GREY} />
```
achou11: non-blocking and somewhat stylistic, but the complicated props interface for this component is usually a sign to consider composition:

```tsx
// Could argue that having this defined as a component isn't even necessary...
const DeviceIconBackground = ({ children, size }) => {
  return (
    <View style={{alignItems: 'center', backgroundColor: LIGHT_GREY, borderRadius: size / 2, width: size, height: size}}>
      {children}
    </View>
  )
})

// In the consuming component
const DeviceIcon = ({size}) => {
  // example of the self hosted server device type case...
  return (
    <DeviceIconBackground size={size}>
      <Octicons name="shield-lock" size={size * 0.5} color={DARK_GREY} />
    </DeviceIconBackground>
  )
}
```

achou11 (approved):
> Not much of a difference as far as I can tell, but what do you think?

Yeah not much of a difference although the shield icon is slightly different. Guess that variant isn't available in the icon library, which is probably why a custom svg was being used 🥲 Don't have strong feelings over reverting that specific one or not but will leave it up to you.

src/frontend/sharedComponents/DeviceNameWithIcon.tsx
cimigree: Ok just for the record, yes it makes total sense to use a wrapper for the background style. Thanks for the tip. I did that and also put back in the svg.

PR #1464, opened 2025-10-21
chore: updates language E2E test

### LINKED ISSUE #1068: Update language settings e2e test to account for "Follow system preferences" option
Extracted from https://github.com/digidem/comapeo-mobile/pull/1063#issuecomment-2759094191 for tracking.

At the very least, we should:

- add an assertion that the `Follow system preferences` option is initially selected on a fresh install
- check that switching from a specific language back to the `Follow system preferences` option updates the UI appropriately

cimigree:
closes #1068 
- Adds a test id on the list item to be able to tell if it is selected. (using the accessibility props just didn't work)
- Checks that Follow system preferences is checked for language on first load.
- Checks that Spanish is selected
- Updates the deprecated TouchableNativeFeedback which did not take a test id to Pressable which does take a test id.


src/frontend/sharedComponents/SelectOne.tsx
```diff
@@ -23,32 +23,39 @@ export const SelectOne = <T,>({
...
-              ? 'radio-button-checked'
-              : 'radio-button-unchecked'
+    {options.map((item, index) => {
+      const isSelected = item.value === value;
+      return (
+        <ListItem
+          key={
+            typeof item.value === 'string' || typeof item.value === 'number'
+              ? item.value
+              : index
```
ErikSin: could we simplify this and just make it `item.label + index`?

ErikSin (approved):
one small optimization, but looks good

PR #1468, opened 2025-10-22
chore: per package icon modules

### LINKED ISSUE #1386: Migrate to per-package icon modules instead of `react-native-vector-icons`
Noticed this warning when doing an install:

> npm warn deprecated react-native-vector-icons@10.3.0: react-native-vector-icons package has moved to a new model of per-icon-family packages. See the https://github.com/oblador/react-native-vector-icons/blob/master/MIGRATION.md on how to migrate

Migrating should help significantly with installation size (probably no/minimal difference in bundle + apk size).

Migration docs: https://github.com/oblador/react-native-vector-icons/blob/master/MIGRATION.md

cimigree:
closes #1386 

- Installed new per package icon modules.

    `@react-native-vector-icons/material-icons`
`@react-native-vector-icons/material-design-icons (for MaterialCommunityIcons)`
`@react-native-vector-icons/ionicons`
`@react-native-vector-icons/octicons`
`@react-native-vector-icons/fontawesome`
`@react-native-vector-icons/ant-design`

- Replaced all the imports, but stuck with use of default export/ import to keep changes minimal (it's totally supported)
- Fixed typescript types.
- Removed old packages.
- Verified that all is looking and acting fine.


achou11 (approved):
lgtm - thanks for doing this!

achou11:
> Replaced all the imports, but stuck with use of default export/ import to keep changes minimal (it's totally supported)

Non-blocking but probably would've gone with using the named exports since it wouldn't result in a massively different diff 😄 Might be worth checking if there's any difference in the bundled JS size (sometimes the import styles can affect that, unfortunately)

cimigree:
I don't think it makes a difference in bundle size so I am going to pass on those changes this time around.

PR #1478, opened 2025-10-23
chore: adds secondary destructive button

### LINKED ISSUE #1471: Create Secondary Destructive Button
Add opinionated button called SecondaryDestuctiveButton.

Image

## To Do
- [ ] Text should be red
- [ ] Outline should be grey
- [ ] Match Props of other buttons

cimigree:
closes #1471 

- I hope you don't mind, but I thought it might make sense to refactor the button file to be more DRY rather than repeat almost all of the same code for a 4th time. 
- Adds secondary destructive button.
- I just temporarily changed a secondary button to a secondary destructive button so you could see it for the screen shot and changed it back

image

### DIFF
--- a/src/frontend/sharedComponents/Buttons.tsx
+++ b/src/frontend/sharedComponents/Buttons.tsx
@@ -14,43 +14,14 @@ type ButtonProps = {
   style?: ViewStyleProp;
 };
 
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
Looks Good!

PR #1479, opened 2025-10-23
feat: updated team member card

### LINKED ISSUE #1469: Create Team Member Card
Create card which displays user on a team.

Image

```ts
type TeamMemberCardProps = {
 deviceType:DeviceType,
 name:string,
 onPress?:()=>void,
thisDevice:boolean
}

```

## To Do
- [ ] Style Card
- [ ] Add Props
- [ ] if no onPress, disable TouchableOpacity

### To reference
deviceNameWithIcon

cimigree:
closes #1469

- Adds the team member card
- Uses it in a temporary way on the team member page so you can play around with it if you wish.
- Includes props for deviceType, name, onPress, and thisDevice. 
- Creates its own components and styling rather than reusing existing ones because reusing would add too many complications and conditionals.  I can clean up those subcomponents when I do the team member page.
- I made the decision to allow the text to grow to two lines for long device names or people who are using large display and fonts
- I had to add the line height and font padding to get it so that the two text components would line up correctly.
- Does the Chevron look big enough? I wasn't sure. I also made it so the chevron is not there if there is no On Press. Sound good?

Some screen shots:

Largest text and large display
---
image

regular size text
---
image

### DIFF
--- a/src/frontend/screens/YourTeam/TeamMemberCard.tsx
+++ b/src/frontend/screens/YourTeam/TeamMemberCard.tsx
@@ -1,0 +1,135 @@
+import React from 'react';
+import {StyleSheet, View, TouchableOpacity} from 'react-native';
+import {
+  BLUE_GREY,
+  BLACK,
+  LIGHT_GREY,
+  DARK_GREY,
+  MEDIUM_GREY,
+} from '../../lib/styles';
+import MaterialIcons from '@react-native-vector-icons/material-icons';
+import type {DeviceType} from '../../sharedTypes';
+import {HeaderText} from '../../sharedComponents/Text/HeaderText';
+import {BodyText} from '../../sharedComponents/Text/BodyText';
+import {defineMessages, useIntl} from 'react-intl';
+import ShieldIcon from '../../images/BlackShield.svg';
+
+const m = defineMessages({
+  thisDevice: {
+    id: 'screens.YourTeam.TeamMemberCard.thisDevice',
+    defaultMessage: 'This device',
+  },
+});
+
+type TeamMemberCardProps = {
+  deviceType: DeviceType;
+  name: string;
+  onPress?: () => void;
+  thisDevice: boolean;
+};
+
+export const TeamMemberCard = ({
+  deviceType,
+  name,
+  onPress,
+  thisDevice,
+}: TeamMemberCardProps) => {
+  const {formatMessage} = useIntl();
+
+  return (
+    <TouchableOpacity
+      disabled={!onPress}
+      onPress={onPress}
+      style={styles.container}>
+      <DeviceIcon deviceType={deviceType} />
+      <View style={styles.textContainer}>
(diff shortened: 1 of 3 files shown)


src/frontend/screens/YourTeam/TeamMemberCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+  thisDevice: {
+    id: 'screens.YourTeam.TeamMemberCard.thisDevice',
+    defaultMessage: 'This device',
+  },
+});
+
+type TeamMemberCardProps = {
+  deviceType: DeviceType;
+  name: string;
+  onPress?: () => void;
```
ErikSin: lets actually just make the onPress required. Originally when we made the decision for it to be optional we didn't know if they can always be pressed, but now we know that they do. So if in the future we need to make it opitional, we can change it then, but lets keep it simple.

We can also get rid of the `disabled={!onPress}` in the `TouchableOpacity`

src/frontend/screens/YourTeam/TeamMemberCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+        return <ShieldIcon width={15} height={21} />;
+      case 'UNRECOGNIZED':
+      case 'device_type_unspecified':
+      case undefined:
+      default:
+        return (
+          <MaterialIcons name="help-outline" size={15} color={DARK_GREY} />
+        );
+    }
+  };
```
ErikSin, line 88: Can you actually make this a reusable component (in the sharedComponent folder) and allow it to just take a "size" prop as well? This icon is used a few places, so it will be good to have it be reusable

src/frontend/screens/YourTeam/index.tsx
ErikSin, line 1: lets remove these changes for now (ive tested them so no longer necessary)

ErikSin (approved):
Some minor changes before you merge, but othewise looks good

src/frontend/screens/YourTeam/TeamMemberCard.tsx
cimigree: I actually did this already in my team page PR (which I am about to put up). So maybe you can look at that next and I can not make the change in here? I did pretty much exactly what you were thinking in that PR.

src/frontend/screens/YourTeam/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1479/commits/11556dd898878ff31fa2d5d687e7b5df127de41d

src/frontend/screens/YourTeam/TeamMemberCard.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1479/commits/11556dd898878ff31fa2d5d687e7b5df127de41d

PR #1481, opened 2025-10-23
feat: team page

### LINKED ISSUE #1470: Update Team Screen
Image

### Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8642-3725&t=3zQJ9qBlD2aOoaLJ-4

## To Do
- [ ] Update Text
- [ ] Add in previous participants at the bottom
- [ ] render [TeamCards](https://github.com/digidem/comapeo-mobile/issues/1469)
- [ ] Conditionally render InviteDevice Button if user is coordinator or creator

cimigree:
closes #1470 

- Updates the Team Screen.
- Makes sets with the roles and defines them at the module level to make them easier to filter (and so they aren't redefined each render).
- Makes the device icon its own component and uses it in three places. The team member card, the past collaborator card, and the DeviceNameWithIcon. I had to pass a color and a size to it since it is a different color for the past collaborators.
- The old DeviceCard is now only used in one place and could be simplified (it used to also be used on the team screen) so I have simplified it and inlined it in the one file it is used in now. I feel like it is easier to make these changes while they are happening rather than trying to remember to go back and clean up later.
- I added a fake little array so you could see the past collaborators. Question. Should we not show the past collaborators section if there are none? I wasn't sure and am treating it like participants and coordinators where they always show.

Screen Recordings
---

https://github.com/user-attachments/assets/29cdfbf7-f4e7-4fe5-a4cb-730c7efb8b2e

https://github.com/user-attachments/assets/86e262a3-635f-4ea1-bf9b-7277e4c7c4b1


src/frontend/screens/YourTeam/index.tsx
```diff
@@ -37,41 +43,76 @@ const m = defineMessages({
...
-    : membersQuery.data.filter(
-        member =>
-          member.role.roleId === COORDINATOR_ROLE_ID ||
-          member.role.roleId === CREATOR_ROLE_ID,
-      );
-  const participants = !membersQuery.data
-    ? []
-    : membersQuery.data.filter(member => member.role.roleId === MEMBER_ROLE_ID);
+
+  const members = membersQuery.data || [];
```
ErikSin: i don't think we need the `|| []`

src/frontend/screens/YourTeam/index.tsx
```diff
@@ -89,85 +130,105 @@ export const YourTeam: NativeNavigationComponent<'YourTeam'> = ({
...
+          </View>
+        ))}
+      </View>
     </ScrollView>
   );
 };
 
 YourTeam.navTitle = m.title;
 
 const IconHeader = ({
```
ErikSin: I think it would be better to just get rid of iconHeader completely. My original reasoning for it was that it was being used twice and the both were material icon headers with the same styling. But looking back at that, there is such minimal code being shared that i don't think it was ever necessary in the first plave. And now that were adding an svg, i dont think it makes sense to change the prop. So I would suggest just get rid of this component completely and just use the svg and Material icon in the parent

src/frontend/screens/YourTeam/TeamMemberCard.tsx
```diff
@@ -38,7 +32,7 @@ export const TeamMemberCard = ({
 
   return (
     <TouchableOpacity onPress={onPress} style={styles.container}>
-      <DeviceIcon deviceType={deviceType} />
+      <DeviceIcon deviceType={deviceType} size={30} />
```
ErikSin: this is becoming a very deeply nested component, which will make refactoring difficult. I think it would be better to just have a prop `{Icon: ReactNode}` and then just pass `<DeviceIcon deviceType={deviceType} size={30} />` into the component. 

eg

```ts
<TeamMemberCard
   Icon={<DeviceIcon deviceType={deviceType} size={30} />}
   {...rest}
/>

```

src/frontend/sharedComponents/DeviceIcon.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import {View, StyleSheet} from 'react-native';
+import MaterialIcons from '@react-native-vector-icons/material-icons';
+import ShieldIcon from '../images/BlackShield.svg';
+import type {DeviceType} from '../sharedTypes';
+import {DARK_GREY, BLUE_GREY} from '../lib/styles';
+
+type DeviceIconProps = {
+  deviceType: DeviceType | undefined;
+  size?: number;
```
ErikSin: `size` is always being used, so i don't think it should be optional

src/frontend/screens/YourTeam/index.tsx
```diff
@@ -37,41 +43,76 @@ const m = defineMessages({
...
+  //     deviceType: 'tablet' as const,
+  //     role: {roleId: BLOCKED_ROLE_ID},
+  //   },
+  //   {
+  //     deviceId: 'fake-3',
+  //     name: 'Palo Verde',
+  //     deviceType: 'desktop' as const,
+  //     role: {roleId: NO_ROLE_ID},
+  //   },
+  // ];
```
ErikSin: you can get rid of this

src/frontend/screens/YourTeam/index.tsx
```diff
@@ -37,41 +43,76 @@ const m = defineMessages({
...
+    id: 'screens.Setting.ProjectSettings.YourTeam.pastCollaboratorsDescription',
+    defaultMessage: 'Devices no longer contributing to this project.',
   },
 });
+
+const COORDINATOR_ROLES = new Set([COORDINATOR_ROLE_ID, CREATOR_ROLE_ID]);
+const PAST_COLLABORATOR_ROLES = new Set([
+  BLOCKED_ROLE_ID,
+  LEFT_ROLE_ID,
+  NO_ROLE_ID,
```
ErikSin: have you confirmed with anyone if we should include `NO_ROLE_ID`? I feel like we shouldn't but I really don't know and I am not sure if you have more context

src/frontend/screens/YourTeam/index.tsx
```diff
@@ -177,4 +238,26 @@ const styles = StyleSheet.create({
...
+  pastCollaboratorCard: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    paddingVertical: 15,
+    gap: 15,
+    height: 60,
+  },
+  pastCollaboratorText: {
+    flex: 1,
+    minWidth: 0,
```
ErikSin: is minWidth:0 necessary?

ErikSin (approved):
Some smaller comments, but nothing blocking

src/frontend/screens/YourTeam/index.tsx
cimigree: No. No context. Just a guess. I will take it out.

src/frontend/screens/YourTeam/index.tsx
cimigree: No I actually needed to update some other things to get it to actually work as I wanted it to. Good call.

src/frontend/screens/YourTeam/TeamMemberCard.tsx
cimigree: Okay I think I did it as you asked.

cimigree: https://github.com/digidem/comapeo-mobile/pull/1481/commits/fb5cf1f088e1be686d6917654b16d8928c6b1ed6

src/frontend/screens/YourTeam/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1481/commits/fb5cf1f088e1be686d6917654b16d8928c6b1ed6

src/frontend/screens/YourTeam/index.tsx
ErikSin: no, thats not what i was suggesting. Lets ask the backend team before we take it out

ErikSin: @RangerMauve in what case would someone have `NO_ROLE_ID`?

RangerMauve: Happens when we've just gotten a connection from a peer but we haven't synced a role record for them yet. Should happen for short blips when we connect to a new user for the first time.

ErikSin: ok, so then yes Cindy, we should **not** include it here.

PR #1482, opened 2025-10-23
Feat/collab info screen

### LINKED ISSUE #1475: Collaborator Info Screen
Create Screen which shows user's info

Image

Image

## To do
- [ ] pass nav props {deviceId:string, isOwnDevice:boolean}
- [ ] in screen use `useSingleMember`
- [ ] style screen
- [ ] if self, show leave project button (unless user is coordinator)
- [ ] elif user is coordinator, show remove device Project Button
- [ ] elif user is particpant show no button

ErikSin:
closes #1475 

image

image

image

Creates the screen, with conditionally rendered buttons. The actions are not hooked up yet, and currently the onPress of the DeviceInfoCard navigates these users to the screen. When the DevceInfoCard gets updated via #1479 , the navigate action should take the user to this page


src/frontend/sharedTypes/navigation.ts
```diff
@@ -163,6 +163,11 @@ export type RootStackParamsList = {
   ProjectStatsTurnedOff: undefined;
   EarlyAccess: undefined;
   EarlyAccessOff: undefined;
+  CollaboratorInfo: {
+    deviceId: string;
+    isOwnDevice: boolean;
+    deviceType: DeviceType;
```
ErikSin, line 171: deviceType does not get returned from the api `useSingleMember`, so the simplest way was to just pass it from the parent.

src/frontend/screens/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <DeviceIcon deviceType={route.params.deviceType} size={80} />
+        {name && <HeaderText style={{textAlign: 'center'}}>{name}</HeaderText>}
+        <View
+          style={{
+            flexDirection: 'row',
+            alignItems: 'center',
+          }}>
+          <MaterialIcon
+            color={BLACK}
+            size={32}
```
cimigree: I think this is supposed to be 24 px

src/frontend/screens/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const ownRoleIsCoordinator =
+    ownRole.roleId === COORDINATOR_ROLE_ID ||
+    ownRole.roleId === CREATOR_ROLE_ID;
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.innerBox}>
+        <DeviceIcon deviceType={route.params.deviceType} size={80} />
+        {name && <HeaderText style={{textAlign: 'center'}}>{name}</HeaderText>}
```
cimigree: I think this is supposed to be variant="header2"

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+            flexDirection: 'row',
+            alignItems: 'center',
+          }}>
+          <MaterialIcon
+            color={BLACK}
+            size={32}
+            name={isCoordinator ? 'manage-accounts' : 'people'}
+            style={{marginRight: 10}}
+          />
+          <HeaderText variant="header4">
```
cimigree, line 98: I think this is supposed to be variant="header5"? But maybe not because coordinator is "header4"

src/frontend/screens/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+              year: 'numeric',
+              month: 'long',
+              day: 'numeric',
+            }),
+          })}
+        </BodyText>
+      </View>
+      {!isCoordinator &&
+        (route.params.isOwnDevice ? (
+          //this should be SecondaryDestructiveButton
```
cimigree: This comment can be removed

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+              // To Do: navigate to remove device success screen
+              console.log('Remove Device pressed');
+            }}
+          />
+        ) : null)}
+    </View>
+  );
+};
+
+CollaboratorInfo.navTitle = m.navTitle;
```
cimigree, line 145: From Figma, it looks like the screen title is "This Device" if it is this device... https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8642-4321&t=7yHig05YsAsL2Nf1-4

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -0,0 +1,141 @@
...
+          {formatMessage(m.addedOn, {
+            date: formatDate(joinedAt, {
+              year: 'numeric',
+              month: 'long',
+              day: 'numeric',
+            }),
+          })}
+        </BodyText>
+      </View>
+      {!isCoordinator &&
```
cimigree, line 112: This logic isn't right, I don't think but am not sure where things ended up. Can a collaborator remove another coordinator? If so, the logic is wrong.

I think should be:
```
{route.params.isOwnDevice && !isCoordinator ? (
  // Participant viewing themselves - show Leave Project button
  <SecondaryDestructiveButton
    text={formatMessage(m.leaveProject)}
    fullSize={true}
    style={styles.buttonStyle}
    renderIcon={({size, color}) => (
      <MaterialDesignIcons size={size} color={color} name="export" />
    )}
    onPress={() => {
      // To Do: Navigate to Leave Project Screen
      console.log('Leave Project Screen pressed');
    }}
  />
) : !route.params.isOwnDevice && ownRoleIsCoordinator ? (
  // Coordinator viewing anyone else - show Remove Device button
  <SecondaryDestructiveButton
    text={formatMessage(m.removeDevice)}
    fullSize={true}
    style={styles.buttonStyle}
    renderIcon={({size, color}) => (
      <MaterialIcon size={size} color={color} name="person-remove" />
    )}
    onPress={() => {
      // To Do: navigate to remove device success screen
      console.log('Remove Device pressed');
    }}
  />
) : null}
```
if a coordinator can remove a coordinator

cimigree (changes requested):
1. This will need to be updated from the TeamPage before it is merged in to 1, use the onPress on the team page and 2, use the DeviceIcon component
2. Supposed to say This Device as a title if it is 
3. Some styling things for sizing (and some spacing/ gaps/ margins I did not mention but which also don't match, but assume you eyed and chose on purpose)
4. Not sure about logic for buttons showing up. 
5. A question, what about servers? Should that be able to be removed from a project just like a regular device?
6. Shouldn't you fix the title of this PR?
7. Why is this not in the YourTeam folder?
8. Is it ok that the icon for Leave Project is slightly different from the designs (since it is coming from r-n-vector icons)? They say we can "Look at" it but not sure what that means?

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
ErikSin: Right now a coordinator can never be removed! In the future we will be able to, but for now I'm intentionally checking that the user being viewed is NOT a coordinator

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
ErikSin: oh yeah... thats confusing. I think i will leave it as "header4" because in 2 of 3 of the mockup in figma it is size 18

ErikSin:
>     3. Some styling things for sizing (and some spacing/ gaps/ margins I did not mention but which also don't match, but assume you eyed and chose on purpose)

No, not intentional. The only thing that i took out was the "Device Model" because we don't have that. Otherwise, i think i have a bad eye for spacing/sizing etc. So if there are things that I may have missed lmk!

>     5. A question, what about servers? Should that be able to be removed from a project just like a regular device?

No, they should not be. I will add logic to make sure that doesn't happen here

>     7. Why is this not in the YourTeam folder?

Im not sure... I think i was trying to not nest any page, but i guess it makes sense to put it in this folder.

>     8. Is it ok that the icon for Leave Project is slightly different from the designs (since it is coming from r-n-vector icons)?

I actually would like to just use the `<DeviceIcon />` component that you built. So ill wait for that to be merged and just switch to that so we are consistent throughtout

cimigree:
> I actually would like to just use the `<DeviceIcon />` component that you built. So ill wait for that to be merged and just switch to that so we are consistent throughtout

It's merged!

src/frontend/screens/CollaboratorInfo.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1482/commits/5e0906eeed5b0b31485d7498f76183f83cdd92a0

src/frontend/screens/CollaboratorInfo.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1482/commits/48676910e74a5bc5df1a144b3af9849a7b3fe931

src/frontend/screens/CollaboratorInfo.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1482/commits/02e32c679a118a97ac52c31addc81ec8985d37f0

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1482/commits/5c3eb7891532b6bdb372105b5e126983e4d5b25f

cimigree:
> > 8. Is it ok that the icon for Leave Project is slightly different from the designs (since it is coming from r-n-vector icons)?
> > ```
> 
> I actually would like to just use the `<DeviceIcon />` component that you built. So ill wait for that to be merged and just switch to that so we are consistent throughtout

I meant the one in the button. This one is the coded one aka your choice.
image
vs. the design in Figma,
Screenshot 2025-10-30 at 3 46 53 PM


---

The remove device icon is also different than the designs. It Figma, it is outlined. And yours is not.
Your code/ your choice
image

vs. the design in Figma
Screenshot 2025-10-30 at 3 56 13 PM

There aren't any rn vector icons to match these so I am wondering (in general, and for my PRs too) if we should use exported from Figma SVG icons.

ErikSin:
> I meant the one in the button. 

Whoops sorry! yeah, i got Icon approved by Aldo, good eye!

cimigree (approved):
Good to get the clarity on the icons in the buttons. Thanks.
I think the padding/ spacing is fine. The only thing I noticed is the spacing between the name and the Participant/ Coordinator looks a little bigger than the design, but it is not much of a difference.
I would love it, however, if you could move the screen into the YourTeam folder because as you agreed, it  logically belongs there and would only be nested 1 deep.

PR #1488, opened 2025-10-27
fix: observation edit with photo category change

### LINKED ISSUE #1486: Observation w photo/ change category/ app crash
https://www.notion.so/digidem/Observation-w-photo-change-category-app-crash-2981b08162d5801b89cdd44492e5ff82?source=copy_link

- Create an observation with a photo
- Re-open observation
- Edit the category
- Tap on Save
- App crash
- App can re-open

### LINKED ISSUE #1492: Active Track+ Edit category of observation with photo / app crash
https://www.notion.so/digidem/Active-track-change-category-on-an-observation-with-photo-app-crash-29a1b08162d580559514cbe378f7cf2e

**Case 1**
- Start a track
- While track is active, open an observation with a photo (created before activating the track)
- Edit that observation by changing the category
- Tap save
- App crash and cannot re-open

**Case 2**
- Start a track
- While track is active, collect an observation with a photo
- Edit that observation by changing the category
- Tap save
- App crash and cannot re-open

cimigree:
closes #1486 and 
closes #1492
The problem was that after saving, the category chooser was still in the navigation stack and it rerenders when clearDraft() is called, which clears the observationId from persisted state. This causes both CategoryChooser and Editor to re-render with observationId: undefined before navigation completes.
So, to fix this issue of temporary missing observation id, Show a loader during navigation transitions when it might be temporarily undefined
Handles the root cause, which is the components rendering with temporarily undefined state
Works for any scenario where the Editor renders during state transitions (not just category changes)

https://github.com/user-attachments/assets/fac6d847-4b6b-4688-a8ee-8d3084228dd5

Alternatively, we could instead change the Observation Category Chooser in here:
```
const handleSelect = (preset: Preset) => {
    updatePreset(preset);
    if (observationId) {
       navigation.goBack()
    // get rid of the below
    //  navigation.navigate('ObservationEdit', {observationId});
    } else {
      navigation.navigate('ObservationCreate');
    }
  };
```
I didn't test this way very much, so if you prefer that method, one of us should do some thorough testing.
Whatever you prefer.

### DIFF
--- a/src/frontend/sharedComponents/Editor/index.tsx
+++ b/src/frontend/sharedComponents/Editor/index.tsx
@@ -143,10 +143,7 @@ export const Editor = ({
                   );
                 }
                 // observationId must exist if there is a saved photo
-                if (isSavedPhoto(att)) {
-                  if (!observationId) {
-                    return <ThumbnailLoader size={size} key={att.hash} />;
-                  }
+                if (isSavedPhoto(att) && observationId) {
                   return (
                     <React.Suspense
                       key={att.driveDiscoveryId + att.hash + att.type}
@@ -165,10 +162,7 @@ export const Editor = ({
                   );
                 }
 
-                if (isAudioAttachment(att)) {
-                  if (!observationId) {
-                    return <ThumbnailLoader size={size} key={att.name} />;
-                  }
+                if (isAudioAttachment(att) && observationId) {
                   return (
                     <AudioSavedThumbnail
                       size={size}
--- a/src/frontend/sharedComponents/Editor/index.tsx
+++ b/src/frontend/sharedComponents/Editor/index.tsx
@@ -144,10 +144,9 @@ export const Editor = ({
                 }
                 // observationId must exist if there is a saved photo
                 if (isSavedPhoto(att)) {
-                  if (!observationId)
-                    throw new Error(
-                      'Observation ID is required for saved photos',
-                    );
+                  if (!observationId) {
+                    return <ThumbnailLoader size={size} key={att.hash} />;
+                  }
                   return (
                     <React.Suspense
                       key={att.driveDiscoveryId + att.hash + att.type}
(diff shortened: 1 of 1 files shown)


src/frontend/sharedComponents/Editor/index.tsx
```diff
@@ -144,10 +144,9 @@ export const Editor = ({
                 }
                 // observationId must exist if there is a saved photo
                 if (isSavedPhoto(att)) {
```
gmaclennan: For now I think remove the check below, and just do this the same as for audio (before this change)

```suggestion
                if (isSavedPhoto(att) && observationId) {
```

gmaclennan (changes requested):
Good investigative work, this seems to fix things.

I think for now maybe just a simpler fix - adding a check on the observationId to the `if (isSavedPhoto())` condition. 

The reason we're seeing this is because `attachments` is coming from the draft observation, but we're reading `observationId` from the draft observation store in this component, and `attachments` in the parent, so after the store changes, the child component is updating first I think. Since `attachments` and `observationId` are related (both properties of a draft observation) we should keep them together, so in a follow up we should probably pass `observationId` as a property to this component, however it might be better to spend time cleaning up this draft observation code (there is a refactor from a year ago) rather than more patches.

So, I don't think you need any change to the audio attachments check, and you can just add the suggested check and remove the nested condition check.

cimigree:
@gmaclennan Ok fixed per your suggestion.
Also closes #1492 

re 1492, that crash is much worse because after you try to come back to the app you get this crash too:
 ```
ERROR  Warning: TypeError: start.getTime is not a function (it is undefined)

This error is located at:

  21 |           <View style={styles.wrapper}>
  22 |             <View>
> 23 |               <Text style={styles.text}>{totalDistance.toFixed(2)}km</Text>
     |                                        ^
  24 |             </View>
  25 |             <View style={styles.separator} />
  26 |             <View>

Call Stack
  UserTooltipMarker (src/frontend/screens/MapScreen/CurrentTrack/UserTooltipMarker.tsx:23:40)
  RNMBXMapView (<anonymous>)
  MapScreen (src/frontend/screens/MapScreen/index.tsx:72:18)
```
  
  Which comes from `src/frontend/hooks/useFormattedTimeSince.ts`
  
  Not sure if this needs to be addressed since it is not an issue once the category switching bug is fixed...

gmaclennan (approved):
Confirming that with this fix I am no longer able to reproduce the bug.

PR #1489, opened 2025-10-27
fix: edit track category

### LINKED ISSUE #1487: Can’t edit category on tracks
- Start a track
- Save it
- Re-open track in the observation list
- Edit category of the track
- Save it
    - You end up seeing the same category you had initially saved the track with

https://www.notion.so/digidem/Can-t-edit-category-on-tracks-2981b08162d580198139c966a0ae1a2c?source=copy_link

cimigree:
closes #1487 

- The TrackEdit screen has a useEffect that loads the preset from the saved track on mount. However, this useEffect was running after the user selected a new category in CategoryChooser, overwriting their selection with the track's original preset.

So this fix only loads the  preset if we don't already have one.

- This preserves the preset selected in CategoryChooser while still loading the preset on initial mount.
- Also, the mutation status checks prevent loading stale data during the brief moment after save when clearCurrentTrack() clears the preset state.
- Use goBack() in TrackCategoryChooser to preserve the screen state.

Testing

- Create and save a track with category "Boundary" (or whatever)
- Edit the track and change category to "Stream" (or whatever)
- Save
- Track should now show "Stream" category.
- Go to edit it again. It should still be "Stream" (before I added the mutation checks this was not the case)

### DIFF
--- a/src/frontend/screens/TrackEdit/index.tsx
+++ b/src/frontend/screens/TrackEdit/index.tsx
@@ -64,12 +64,13 @@ export const TrackEdit: NativeNavigationComponent<'TrackEdit'> = ({
 
   useEffect(() => {
     if (
+      !preset &&
       track?.presetRef?.docId &&
       foundPreset?.docId === track.presetRef.docId
     ) {
       setTrackPreset(foundPreset);
     }
-  }, [track, foundPreset, setTrackPreset]);
+  }, [track, foundPreset, setTrackPreset, preset]);
 
   useEffect(() => {
     if (trackId) {
--- a/src/frontend/screens/TrackEdit/index.tsx
+++ b/src/frontend/screens/TrackEdit/index.tsx
@@ -66,11 +66,13 @@ export const TrackEdit: NativeNavigationComponent<'TrackEdit'> = ({
     if (
       !preset &&
       track?.presetRef?.docId &&
-      foundPreset?.docId === track.presetRef.docId
+      foundPreset?.docId === track.presetRef.docId &&
+      editTrackMutation.status !== 'pending' &&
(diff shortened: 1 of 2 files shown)


src/frontend/screens/TrackEdit/index.tsx
```diff
@@ -64,12 +64,15 @@ export const TrackEdit: NativeNavigationComponent<'TrackEdit'> = ({
...
       track?.presetRef?.docId &&
-      foundPreset?.docId === track.presetRef.docId
+      foundPreset?.docId === track.presetRef.docId &&
+      editTrackMutation.status !== 'pending' &&
+      editTrackMutation.status !== 'success'
     ) {
       setTrackPreset(foundPreset);
     }
-  }, [track, foundPreset, setTrackPreset]);
+  }, [track, foundPreset, setTrackPreset, preset, editTrackMutation.status]);
```
gmaclennan, line 75: Any useEffect that sets state is a risk for bugs, particularly an infinite loop (state change results in an effect re-run), and adding more dependencies does increase this risk. I don't know enough about all the hooks here, but we need to go through this with a fine-tooth comb to be really sure that `setTrackPreset()` will not change the values of any of `[track, foundPreset, setTrackPreset, preset, editTrackMutation.status]` - if it does, then this will break (it will keep running in a loop).

How can we quickly fix the bug, then prioritize addressing this as a follow-up?

ErikSin: IMO i think it will not take that long to just refactor this component away from the < Editor >, which is why we need all these useEffects.

I refactored the Observation Create to not use the editor and went fairly smooth. So I think im going to put the work into that today instead of patch fixing it like we are here

PR #1490, opened 2025-10-27
fix: language E2E test

cimigree:
Checks for Spanish language that is up to date with the new text in the app.
Our settings button label changed, so we have to change what the spanish language check is looking for 
We went from App Settings to Comapeo Settings so the Spanish language text finding has to change a smidge.

### DIFF
--- a/tests/e2e/specs/settings/language.test.ts
+++ b/tests/e2e/specs/settings/language.test.ts
@@ -48,7 +48,7 @@ describe('Settings - Language Settings Flow', () => {
       await drawerIcon.click();
     }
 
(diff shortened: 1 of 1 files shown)


ErikSin:
~~This PR circumvents the problem but doesn't actually fix the root issue. The original test is correct as the text in the app should be "Ajustes de la CoMapeo". The fact that it is not updating in crowdin is what should be addressed~~

OHHH nvm, i understand the issue. Crowdin is taking the new translation, while develop was using the old translation.

ErikSin (approved):

PR #1491, opened 2025-10-27
feat: leave project flow

### LINKED ISSUE #1476: Leave Project Flow
Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8642-4286&t=3zQJ9qBlD2aOoaLJ-4

Image

## To Do
- [ ] use `useProjectSettings` to determine project name (projectId comes from context)
- [ ] use `useLeaveProject` to leave projectm and should update project to default project (on success) on press of `Yes, Leave` button
- [ ] `cancel` navigates back
- [ ] `open [deviceName]` should `navigate.popToTop()`
- [ ] Coordinators CANNOT Leave Project
- [ ] Style Screens

cimigree:
Closes #1476 

- Adds screens for leaving a project.
- When navigating to the Confirmation screen, we navigate first so there isn't an app flash when the project reloads
- Prevents use of the Android back button on the Confirmation Screen
- I did not add end to end tests because we cannot have or create participants on Browserstack since we can only run one device

### **Questions**
1. I wasn't sure of the best way to get the default project (on the left project confirmation)... I chose the one that has name undefined, but I am wondering about existing users who have renamed a solo project. Is that a possibility?
2. The icons are slightly different and I used the vector icon library. But should I just import and use Svgs from Figma?

Screen Shots
image

image


src/frontend/screens/YourTeam/LeaveProject.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export const LeaveProject = ({
+  navigation,
+}: NativeRootNavigationProps<'LeaveProject'>) => {
+  const {formatMessage} = useIntl();
+  const {projectId} = useActiveProject();
+  const {data: projectSettings} = useProjectSettings();
+  const leaveProject = useLeaveProject();
+  const {setActiveProjectId} = useActiveProjectIdActions();
+  const {listProjects} = useClientApi();
```
ErikSin: We should not use the clientApi directly unless we need access to an async (non-suspensful) api. Instead we should use `useManyProjects` from core-react

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -120,8 +120,7 @@ export const CollaboratorInfo: NativeNavigationComponent<
               <MaterialDesignIcons size={size} color={color} name="export" />
             )}
             onPress={() => {
-              // To Do: Navigate to Leave Project Screen
-              console.log('Leave Project Screen pressed');
+              navigation.replace('LeaveProject');
```
ErikSin: I don't think we should use "replace" here. The user should be able to go back to this screen, unless there is some context that i am missing?

ErikSin (approved):
2 non blocking comments.

PR #1499, opened 2025-10-28
feat!: Update core: new comapeocat format & svg icons

gmaclennan:
BREAKING CHANGE: The new format for comapeocat files is incompatible with the previous one. This version of the app will not be able to import `.comapeocat` files that were not created with the new `npx comapeocat` tool.

This updates @mapeo/default-categories to v6, which are the same categories (with some minor fixes) but in the new `comapeocat` format. The new format only includes SVG icons, and only of one size (currently exposed as `medium` via the core API).

The only place we use category icons, I think, is in `PresetIcon.tsx`. For backwards-compatibility (existing projects which only have PNG icons) this will fall-back to trying to load a PNG icon if loading an SVG icon fails.

The new format and the core updates fix various issues with category imports and translations.

Also from the core update:

- fix: Rework initial settings storage on invitee v2 by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1125
- chore: update drizzle to v1-beta.1 by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1140
- fix: Don't use getAddress in start/stop of local-discovery by @RangerMauve in https://github.com/digidem/comapeo-core/pull/1136
- fix: NO_ROLE should not sync config core by @RangerMauve in https://github.com/digidem/comapeo-core/pull/1143
- chore: cleanup drizzle types by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1141
- fix: Blocked members only expect auth core for initial sync by @RangerMauve in https://github.com/digidem/comapeo-core/pull/1138
- chore: release workflow fixes by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1151
- chore: remove branch filter on release workflow by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1153
- feat!: new comapeocat format by @gmaclennan in https://github.com/digidem/comapeo-core/pull/1135


ErikSin:
I did some stress testing since this was a breaking change. The app never broke which is great!

There was some odd behaviour:

If a device 1 (on this version) adds device 2(on older version) to the project, device 2 only try to pull pngs (which is expected). Should device 2 be able to access the pngs? Currently it is coming up as blank:

image

achou11:
> There was some odd behaviour:
> 
> If a device 1 (on this version) adds device 2(on older version) to the project, device 2 only try to pull pngs (which is expected). Should device 2 be able to access the pngs? Currently it is coming up as blank:
> image

What you're seeing is expected behavior in this scenario. The default categories file that Device 1 is using no longer includes PNGs and only has SVGs. Since Device 2 has joined Device 1's project, it's using the categories based on Device 1, meaning that as Device 2 tries to fetch PNGs, it will come up empty, resulting in the lack of the icons in your screenshot.

gmaclennan:
Yes, unfortunately nothing we can do about this other than also include PNGs in the new format, which we want to avoid. The solution is to update device2 to the version that supports SVG icons, and it should just work. We should make sure this info is included in the release notes.

ErikSin (approved):
Looks good, thanks for doing that!

PR #1503, opened 2025-10-28
fix: prevent back button track category chooser

### LINKED ISSUE #1501: Back press on track category chooser problematic
**The bug**
Participants often got confused during the “Select a Category” step when creating a track. Some mistakenly tapped the “Back” button, which caused the loss of the entire track recording.

**To Reproduce**
1. Start and stop recording a track.
2. Reach the category selection screen.
3. Tap “Back” instead of choosing a category.
4. The app discards the recorded track without confirmation.

**Expected behavior**
If the user taps “Back” during category selection, the app should show a confirmation dialog (“Do you want to discard this track?”)

**Additional context**
Reported by @luandro from Wayamu training

cimigree:
closes #1501 

- Adds prevent back button to track category chooser.
- I tried to use usePreventRemove but that does not work because we do need to navigate away from the screen of course.
- I also thought it was more in keeping with our behavior to just prevent the back button instead of showing the dialogue as was suggested in the issue by Lu. We do show the discard dialogue if someone presses the x.
- I just use the prevent back button all the time and navigation away can be handled with the top left x/ check mark (where it calls the dialogue on a new track or goes back to the previous screen depending on if it is a new track or pre-existing track).

### DIFF
--- a/src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
+++ b/src/frontend/screens/PresetChooser/TrackCategoryChooser.tsx
@@ -11,6 +11,7 @@ import {CustomHeaderLeft} from '../../sharedComponents/CustomHeaderLeft';
 import {usePresetsSelection} from '@comapeo/core-react';
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 import {useAppLanguageTag} from '../../hooks/useAppLanguageTag';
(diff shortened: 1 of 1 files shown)


ErikSin:
I think this should follow the same behaviour as the observations, where if you press the back button, the bottom sheet pops up. Also #1509 introduces some changes that we should wait for before we do that.

PR #1505, opened 2025-10-28
fix: return to ongoing track

### LINKED ISSUE #1504: Force quit app during track/choosing category - loss of track
**Describe the bug**
Track recording is active, without being active. Not sure how to reproduce. 
In sum, when someone stops recording a track, and they are no longer nestled safely in the save flow (the track category chooser or the save track screen) because they have closed the app or they have accidentally pressed the back button or for whatever reason, they are stuck and can no longer get back to the app.

**To Reproduce**
1. Record a track
2. End the track and get taken to the category chooser
3. Force quit the app
4. Relaunch app
5. Track is not active
6. If you zoom in on the map, the dotted line shows the track is active but theres no way to save it

**Screenshots**

**Additional context**
The back button behavior is being ended and there is a PR for it, but I am sure there are other ways to cause this bug.

cimigree:
closes #1504 

- When someone navigates away from an ongoing track that is in the saving flow, they need a way to get back to it.
- This PR adds a focus effect for the map screen for an in-progress track.
- Adds automatic navigation back to the track save flow when users have stopped recording a track but haven't completed the save process
- Prevents users from being stuck with an unsaved track after accidentally navigating away from the save flow
- I used this check `isTracking === false but locationHistory.length > 0` as we discussed
- Follows the same pattern as useCheckDraftObservationAndNavigate for consistency
- I repeated the check for authState is obscured, which I think is necessary for each useEffect? But I am not sure...

To test:

- [ ]  Record a track and stop recording
- [ ]  Press back button from category chooser → verify you're taken back to category chooser, or force close the app and reopen it and should be on the category chooser screen
- [ ]  Choose a category, and reload the app from the save screen →  you should be taken to the save track screen
- [ ]  Complete the save flow successfully → verify you can navigate normally
- [ ]  Discard a track → verify you can navigate normally
- [ ]  Close app during save flow → verify on app reopen you're taken back to the save flow

### DIFF
--- a/src/frontend/screens/MapScreen/index.tsx
+++ b/src/frontend/screens/MapScreen/index.tsx
@@ -42,6 +42,7 @@ import {isLowStorage} from '../../lib/storage';
 import {LowStorageBanner} from '../../sharedComponents/Storage/LowStorageBanner';
 import {useAppUsageStatsStore} from '../../contexts/AppUsageStatsContext';
 import {useShouldShowAppUsagePrompt} from '../../hooks/useShouldShowAppUsagePrompt';
+import {useTrackState} from '../../contexts/TrackStoreContext';
 
 // This is the default zoom used when the map first loads, and also the zoom
 // that the map will zoom to if the user clicks the "Locate" button and the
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
looks good

PR #1508, opened 2025-10-29
feat: remove device

### LINKED ISSUE #1472: Remove Device Flow
Create the screens for removing a device from a project.

### Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8642-4252&t=Apjef5Ynq1BBGptp-4

### Screen Shots:

Image

# To Do
- [ ] Create Remove Device Screen
- [ ] Create Device Removed Success Screen
- [ ] use `useChangeMemberRole` from `@comapeo/core-react`
- [ ] should get deviceId from previoud screen (aka pass deviceId as a nav param)
- [ ] should change role to [Blocked_Role_id](https://github.com/digidem/comapeo-core/blob/82bc5e1dc32b9c151e1ac41def5f977574ca1a60/src/roles.js#L13)

### Caveats
- the Api will probably change, but i think its simple enough to use useChangeMember role for now as the api will most likely be very similar
- the PR can be stacked on #1482

cimigree:
closes #1472 
Adds screens for the removal of a device from the coordinator side
Not sure if this is the right icon to use. The - is on the wrong side. Should I just import the SVG from the Figma?
Screenshot 2025-10-30 at 1 28 44 PM

Screen recording:
https://github.com/user-attachments/assets/e50665b3-e778-47ba-b434-45f7f12ccae1

Per [Slack discussion with design](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1761858140914399), limiting reason characters to 100


src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -123,7 +123,7 @@ export const CollaboratorInfo: NativeNavigationComponent<
               navigation.navigate('LeaveProject');
             }}
           />
-        ) : ownRoleIsCoordinator ? (
+        ) : ownRoleIsCoordinator && name ? (
```
ErikSin: the frontend always validates so there always is a name. But in the slight chance that something happens and there is no name, I don't think we should block the user. So instead of doing this check we can just pass an empty string:

```ts

onPress={() => {
              navigation.replace('RemoveDevice', {
                deviceId: route.params.deviceId,
                deviceName: name || "",
              });
            }}
```

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -132,8 +132,10 @@ export const CollaboratorInfo: NativeNavigationComponent<
               <MaterialIcon size={size} color={color} name="person-remove" />
             )}
             onPress={() => {
-              // To Do: navigate to remove device success screen
-              console.log('Remove Device pressed');
+              navigation.replace('RemoveDevice', {
```
ErikSin: i don't think we should use "replace" here

src/frontend/screens/YourTeam/RemoveDevice.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <TextInput
+          style={[
+            styles.reasonInput,
+            {borderColor: errorTimeout ? RED : BLUE_GREY},
+          ]}
+          value={reason}
+          onChangeText={setReasonWithValidation}
+          placeholderTextColor={NEW_DARK_GREY}
+          multiline
+        />
```
ErikSin: We should be using the shared component `<HookFormTextInput/>`. This will remove the need to use `useTemporaryError`.

See `ProjectSettings/EditProjectDetails` for an example

ErikSin (changes requested):
I small but blocking change, We should use the `HookFormTextInput` whenever we have an input that requires validation

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1508/commits/8729979a74f9c7d3721440b33df1e8a6b801c128

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1508/commits/8729979a74f9c7d3721440b33df1e8a6b801c128

src/frontend/screens/YourTeam/RemoveDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1508/commits/8729979a74f9c7d3721440b33df1e8a6b801c128

ErikSin (approved):
Looks good!

PR #1509, opened 2025-10-29
fix: track edit refactor to change categories

### LINKED ISSUE #1487: Can’t edit category on tracks

ErikSin:
closes #1487 

The original issue was due a useEffect setting the preset in persisted state. This preset in persisted state was then used when saving the value. Because the value of the preset came from the db, we had to do this state syncing via a useEffect. This useEffect did not handle when a the preset was changed. 

Through discussion with Cindy, we decided to get rid of the persisted state entirely. By getting rid of the persisted state we do not need the useEffects. 

The state does not need to be persisted because we were not using it. For an observation we were persisting the state because we navigated back to the observation edit screen if the user had closed the app. We were not doing that with editting a track, so there was no reason to persist the state.

The bigger problem that this work uncovered was that editing a track was OVERWRITING the info on a current track. Both the current track AND edited track were using the same stores. So we were unintentionally editing a current track. 

I also refactored the Track Edit Screen to not use the `Editor`, which simplified the component

And i refactored the "Discard Track Edit" bottom sheet to NOT use gorham-bottom sheet, and to use react navigation.

https://github.com/user-attachments/assets/021fcfb0-b92c-4cf7-9c79-fc3b63282088


cimigree (changes requested):
This is looking really great and will definitely fix a problem that I am sure would have been a nightmare for a user. However, there is one fix that is needed.
The notes section is too small -- only one line maybe? Not sure what needs to be done, but it is not right as it is. See screen recording.


https://github.com/user-attachments/assets/2caabe90-6693-4950-bed6-290abe24e793

See the screenshot of the create track for the difference...

image

ErikSin:
> The notes section is too small -- only one line maybe? Not sure what needs to be done, but it is not right as it is. See screen recording.

hmmm. this is really weird. Im not sure why, but putting the entire screen in a scroll view seems to be the only fix that works. I tried flexing the view to 1 but that made no difference https://github.com/digidem/comapeo-mobile/pull/1509/commits/196212f93dd7889bff91142f3f9a42f952b0aa52

cimigree:
Good catch on the Param.
It definitely works better.
I think if you could update the DescriptionField you can avoid the ScrollView but it is used elsewhere and not sure how that will affect other screens
This is not blocking but I hope you can fix it. The top of my text is cut off slightly. One way is by wrapping the Controller in a view and adding paddingTop or a marginTop. Anyway... I think it would make this more polished looking if you could make that change before you merge.
image

cimigree (approved):

PR #1512, opened 2025-10-30
chore: warn track discard on press of back button

### LINKED ISSUE #1501: Back press on track category chooser problematic

ErikSin:
closes #1501.

Also closes #1503 as a an alternative pr.

This Pr prompts the user to confirm that they want to navigate away from the tracks preset page (and discard their track).

This behaviour mimics the behaviour seen in the observation preset page

### DIFF
--- a/src/frontend/screens/SaveTrack/HeaderLeft.tsx
+++ b/src/frontend/screens/SaveTrack/HeaderLeft.tsx
@@ -77,13 +77,7 @@ export const HeaderLeft = ({headerBackButtonProps}: HeaderLeftProps) => {
         onPress={openSheet}
         headerBackButtonProps={headerBackButtonProps}
       />
-      <BottomSheetModal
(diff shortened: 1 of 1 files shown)


cimigree (approved):
Thanks for doing that. Much simpler and I did not realize how. And will be more communicative to the users.

PR #1513, opened 2025-10-30
chore: update project hook

### LINKED ISSUE #1485: update useProjectRoleAndDetails header
Currently the useProjectRoleAndDetails optionally returns a header- as the default project does not have a name, but still returns the user name - aka the header. Due to the new designs, the `MainScreenHeader`, and `MenuHeader`, and the `AllProjects` list the project name AND the project role (eg "Runway - Coordinator"). Becuase of this, the useProjectRoleAndDetails should ALWAYS return a header, where header is a derived value from either the user name/project name and the users role

eg

```ts
function useProjectRoleAndDetails (){

 ...rest
 return({
  ...rest,
  projectHeader:projectName? `${projectName} - ${projectRole}` : {userName}
 })
}

```ts

# To do
- [ ] update useProjectRoleAndDetails
- [ ] update useage of hook. There are many places where we check whether the hook returns a header, but with use the header or use the projectName if there is no header. Update accordingly where most places can be replaced with just header.

### PR DESCRIPTION
**Pull request by cimigree**
closes #1485 

Changed the useProjectRoleandDetails to always return a header.
For a solo user on a default project, it hasn't changed. For other situations, Coordinator or Participant is added behind the project title.
Updates the use of the header throughout the app. See screenshots.

---
**Drawer Menu:**
image

**All Projects**
image

**Project Settings**
image

**Observation Create**
image

Editor
image

Edit Track
image

### DIFF
--- a/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
+++ b/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
@@ -53,7 +53,7 @@ export const EditProjectDetails: NativeNavigationComponent<
   }>({
     defaultValues: {
       color: projectDetails.projectColor,
-      projectName: projectDetails.projectName,
+      projectName: projectDetails.projectName!,
       projectDescription: projectDetails.projectDescription,
     },
   });
--- a/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
+++ b/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
@@ -53,7 +53,7 @@ export const EditProjectDetails: NativeNavigationComponent<
   }>({
     defaultValues: {
       color: projectDetails.projectColor,
-      projectName: projectDetails.projectName!,
+      projectName: projectDetails.projectName,
       projectDescription: projectDetails.projectDescription,
     },
   });
--- a/src/frontend/screens/ProjectSettings/index.tsx
+++ b/src/frontend/screens/ProjectSettings/index.tsx
@@ -115,17 +115,13 @@ export const ProjectSettings = () => {
   const participantWithRemote =
     projectInfo.role === 'participant' && remoteArchiveOn;
 
-  const displayTitle = isSolo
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: src/frontend/screens/ProjectSettings/index.tsx [hunk 1]
```diff
@@ -163,7 +159,7 @@ export const ProjectSettings = () => {
           }
           buttonText={formatMessage(m.viewTeam)}
           onPress={() =>
-            navigate('YourTeam', {projectName: projectInfo.projectName})
+            navigate('YourTeam', {projectName: projectInfo.projectName!})
```
ErikSin: projectName is actually not used in the YourTeam Page, so we can take out that nav param completely

src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
```diff
@@ -53,7 +53,7 @@ export const EditProjectDetails: NativeNavigationComponent<
   }>({
     defaultValues: {
       color: projectDetails.projectColor,
-      projectName: projectDetails.projectName,
+      projectName: projectDetails.projectName!,
```
ErikSin: the "!" is not needed

ErikSin (approved):
Just need to update the "YourTeam" nav props and update the nav actions that are related, but otherwise looks good

src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1513/commits/a35a2c74488f2b8f7fed82fea2a71a33d9b40035

src/frontend/screens/ProjectSettings/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1513/commits/a35a2c74488f2b8f7fed82fea2a71a33d9b40035

PR #1515, opened 2025-11-03
fix: invite available during onboarding

### LINKED ISSUE #1533: Can't receive invite during onboarding
**Describe the bug**
When someone is in the middle of the onboarding process, they cannot successfully receive an invite. See the screenshots section for the error that appears to developers

**To Reproduce**
1. Go through the onboarding process
2. Have someone send an invitation to a project
3. See screenshot
4. Not sure what happens on a non developer version, but I am sure the invitation won't be received by the invitee

**Expected behavior**
After someone has created a device name and read through some of the onboarding, they should be able to receive an invitation to a project

**Screenshots**

Image

## Notion:
https://www.notion.so/digidem/Onboard-Joint-a-project-dont-get-the-invite-2b61b08162d580758190cd0c32eff0ad?v=c58a7b022e084c2bae2c4379bf0c67e7

cimigree:
closes #1533 
Adds invite screens to onboarding.
Previously the screens were not available during onboarding, so if someone tried to receive an invite at that time, this error would appear:
image

- Because the EDITING_SCREEN_NAMES include most of the onboarding screens, **the invite will only show up after the success screen**, which is _I think_ what we want, but please let me know if not.
- I changed the InviteReceived so that if someone doesn't have an active project, that is they are in the onboarding flow, then they **will have a default project also created** (but not switched to it). This was a guess as to desired behavior based on everything else we have set up in regards to naming a project and leaving a project etc.
- I tried it out locally and it seemed to work fine.
- Is this the right way to achieve this fix?


src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -82,6 +88,15 @@ export const InviteReceived = ({
...
         onSuccess: projectId => {
           setActiveProjectId(projectId);
+
+          if (!activeProjectId) {
+            createProject.mutate(undefined, {
+              onError: err => {
+                Sentry.captureException(err);
+              },
+            });
+          }
```
ErikSin: I understand the logic behind checking for an `activeProjectId`, but I think the less fragile check would be to check all the projects, and if there is NOT and a project without a name (aka a solo project) then create a solo project

```ts

const allProjects = useManyProjects()

const hasDefaultProject = allProjects.some(proj=>!proj.name)

...rest
if(!hasDefaultProject) => createDefaultProject
```

src/frontend/Navigation/Stack/OnboardingScreens.tsx
ErikSin, line 1: Since were introducing 3 bottom sheet, we should just use a `<RootStack.group>` so they can all share the same `options` instead of copying and pasting 4 times

ErikSin (changes requested):
A couple small, but blocking comments

src/frontend/screens/Invites/InviteReceived.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1515/commits/a881df72477ecde4fa82e696ebe09a4f7f5f520d

src/frontend/Navigation/Stack/OnboardingScreens.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1515/commits/a881df72477ecde4fa82e696ebe09a4f7f5f520d

ErikSin (approved):

PR #1517, opened 2025-11-03
fix: use better-sqlite3 build with 16kb page-size support

gmaclennan:
The builds of better-sqlite3 using our new build infra (using bare-make) with support for 16kb page sizes are published with a release suffix `-bare-make`. This updates the downloadPrebuilds function to download this version.

For the other nodejs-mobile native modules, the releases with 16kb page size support overwrote the previous version, so there are no changes needed in mobile - the download function will just download the newer builds. If the CI on this PR passes then we should be all ok.

### DIFF
--- a/scripts/download-prebuilds.mjs
+++ b/scripts/download-prebuilds.mjs
@@ -21,10 +21,6 @@ export async function downloadPrebuilds(modules, {verbose} = {verbose: false}) {
 
   return Promise.all(
     modules.map(async ({name, usesNapi, version}) => {
-      if (name === 'better-sqlite3') {
-        // We currently use this release suffix for better-sqlite3 prebuilds
-        version += '-bare-make';
-      }
       if (verbose) {
         console.log(`${name}: prebuilds start (${version})`);
       }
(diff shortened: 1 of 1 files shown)


scripts/download-prebuilds.mjs
```diff
@@ -56,12 +60,6 @@ export async function downloadPrebuilds(modules, {verbose} = {verbose: false}) {
 
           fs.unlinkSync(path.join(targetDir, artifactInfo.name));
 
-          // better-sqlite3 includes an additional native module for testing purposes
-          // removing since it's not needed and also causes issues with nodejs-mobile-react-native
-          if (name === 'better-sqlite3') {
-            fs.unlinkSync(path.join(targetDir, 'test_extension.node'));
-          }
```
achou11, line 63: are you sure this isn't needed anymore?

achou11: oh i guess the downloaded artifact from our builds repo has different assets now?

gmaclennan: yes, the bare-make build workflow does not create this artifact.

gmaclennan: Although builds are failing, so I'll need to look into that tomorrow

achou11 (approved):

PR #1518, opened 2025-11-04
chore: e2e for team

cimigree:
1. Updates a mistake in the setting of the title for the collaborator screen if it is the person's device.
2. Adds tests in a team folder for the Team screens
3. Adds those tests to the config and the config template
4. Adds documentation for the tests
5. Adds and updates documentation for other tests just so it is current and up to date
6. We will be able to add on to these when a collaborator can leave a project, which will be good to have.


ErikSin (approved):
Given that the tests are passing im not going to do an in-depth review.

PR #1525, opened 2025-11-06
feat: notified of removal via bottom sheet

### LINKED ISSUE #1473: Notify Device on Project Removal
Notify a user when they have been removed or blocked from a project. When a user is blocked from a project they are NOT removed from a project, they front end still need to remove the from the project via leaveProject(). This is intentional so that the front end is not suddenly and unexpectedly not in a project while the user is actively interacting with the project. By having the frontend chose when the leaving of the project happens, we can reset the navigation state first and make sure there are no intermediate state that aren't accounted for.

There are 2 ways we should check.
1. When a user first opens a project the first thing we should so is check the role in the project, if the role is BLOCKED, we should call leaveProject(), switch them to the default project, and open the removed from project bottom sheet
2. If a user is interacting with a project, initialize the `useProjectOwnRoleChangeListener`. if the role changes to BLOCKED, navigate the user to home, call leaveProject(), switch them to the default project, and open the removed from project bottom sheet.

### Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8642-4355&t=DqKVEh9tG6NQ4xFL-4

### Screenshots:

Image

## To do
- [ ] Create Bottom Sheet. 
- [ ] The onPress of the close button should handle the leaveProject(). Therefore the button should show the UI activity indicator when the mutation is happening, and close the bottom sheet on complete
- [ ] Create `<ProjectRemovalListener/>` 
- [ ] Initialize ProjectRemovalListener in the `Layout` of [this Rootstack.Navigator](https://github.com/digidem/comapeo-mobile/blob/efdbd277386d9f03b6223299e516403b2fa50d31/src/frontend/Navigation/Stack/index.tsx#L114)

ErikSin:
closes #1473 

stacked on #1508 

This is heavily based off #1521. There were clarification about the API that were made that caused this refactor.

First, when a device blocked, the front end still has to manually remove the user from the project (via leaveProject()). 

On first open of a project, we initially check to see if the user is BLOCKED, if so, we open the bottom sheet notifying the user that they have been blocked, and remove them on closure of the bottom sheet

If a user has a project open we use `useProjectOwnRoleChangeListener` to listen for any changes in the role. If the users role gets change to BLOCKED, we move navigate them to the bottom sheet and follow the same steps as above.

The logic for both these can be found in the `ProjectRemovalListener`


src/frontend/screens/AllProjects.tsx
```diff
@@ -36,6 +36,10 @@ export const AllProjects = () => {
   const {formatMessage} = useIntl();
   const {isTracking} = useTracking();
 
+  console.log(data);
```
cimigree: Please remove console logs.

src/frontend/sharedComponents/ProjectRemovalListener.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const ProjectRemovalListener = ({
+  activeProjectId,
+  currentRouteName,
+  navigation,
+}: {
+  activeProjectId: string;
+  currentRouteName: string | undefined;
+  navigation: NavigationHelpers<AppStackParamsList>;
+}) => {
```
cimigree: This approach only monitors the activeProjectId. If a user is removed from a project they're NOT currently viewing, no notification appears. They have to manually switch to the blocked project (which shows as their device name in AllProjects) to see the bottom sheet. Is this the desired behavior? It seems sub optimal to me.

src/frontend/sharedComponents/ProjectRemovalListener.tsx
```diff
@@ -0,0 +1,62 @@
...
+            name: 'RemovedFromProjectBottomSheet',
+            params: {projectId: activeProjectId},
+          },
+        ],
+      }),
+    );
+  }, [activeProjectId, navigation]);
+
+  //checks on first open of project that the user has not been blocked
+  React.useEffect(() => {
```
cimigree, line 39: This effect correctly shows the bottom sheet when opening a project where the user is already blocked. However, it only runs when activeProjectId changes (when user switches projects). If removed from Project A while viewing Project B, the user won't see any notification until they manually switch to Project A. Is that what you wanted?

src/frontend/screens/RemovedFromProjectBottomSheet.tsx
```diff
@@ -0,0 +1,151 @@
...
+            </HeaderText>
+            {reason && (
+              <HeaderText variant="header5">
+                {formatMessage(m.reasonLabel, {reason})}
+              </HeaderText>
+            )}
+          </View>
+
+          <View style={styles.buttonContainer}>
+            {leaveProject.status === 'pending' ? (
```
cimigree, line 95: I still don't love that the leave project only happens on button press. It is good that if you force close the app and reopen you still see the bottom sheet, but I imagine that would be kind of a frustrating and maybe unnecessary user experience. Couldn't the leave project just be run in a useEffect when the sheet mounts?

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -111,7 +112,27 @@ export const RootStackNavigator = () => {
...
   return (
     <ActiveProjectProvider activeProjectId={activeProjectId}>
-      <RootStack.Navigator {...commonNavigatorProps}>
+      <RootStack.Navigator
+        {...commonNavigatorProps}
+        layout={({navigation, state, children}) => {
+          const currentRouteName = state.routes[state.index]?.name;
+          return (
+            <React.Suspense fallback={<Loading />}>
+              <PendingInvitesListener
```
cimigree: Why is the duplication of PendingInvitesListener necessary? Isn't it in the CommonNavigatorProps? That is still in here.

cimigree (changes requested):
So, the notification only works for an active project:  Removals from inactive projects go unnoticed until the user  switches to them.
Blocked projects show as device names in the All Projects menu because useProjectRoleAndDetails returns projectName: undefined for blocked projects, they appear in AllProjects with the device name, creating confusion.
Manual discovery required: Users must manually discover they've been removed by noticing a strange device name project duplication in their list and tapping on it.
In general, I am not crazy about how the logic to do all the things is split between the listener and the bottom sheet. I feel like it should all be in one place or another if possible. Can you explain the reasoning for me why the split?

https://github.com/user-attachments/assets/9656e49a-8b14-429b-900a-9b7af92429f6

src/frontend/sharedComponents/ProjectRemovalListener.tsx
ErikSin: Yes, this is the desired behaviour!

src/frontend/screens/RemovedFromProjectBottomSheet.tsx
ErikSin: I'm not sure I fully understand why you think it would be frustrating. Is it frustrating that the user has to press the button to interact with any other part of the app (aka they are completely blocked)?

I personally think it would be worse user behaviour to automatically remove the user from the project. Lets say I was removed from a project, i open the project, and then I close the app suddenly for some reason before i was able to interact with the bottom sheet. If we use a useEffect, they will have been removed from the project, and when I open the app, they will no longer be able to see the project/reason for removal. IMO I think this would be really confusing.

I'm open to changing to this behaviour if there is something that I am overlooking and not considering.

src/frontend/Navigation/Stack/index.tsx
ErikSin: because of the way the spread operator works,  since Im adding a new `layout` prop, it completely overwrites what was in the CommonNavigatorProps

src/frontend/screens/AllProjects.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1525/commits/4746fe518a44a7e6b70ac16f658c13592371abd3

ErikSin:
> So, the notification only works for an active project: Removals from inactive projects go unnoticed until the user switches to them. Blocked projects show as device names in the All Projects menu because useProjectRoleAndDetails returns projectName: undefined for blocked projects, they appear in AllProjects with the device name, creating confusion. 

The problem was the useProjectRoleAndDetails hook. since the role was was "BLOCKED" the hook was returning a solo project (as it did that by default) which was showing the incorrect UI. It has been updated, so this problem shouldn't exist anymore

cimigree (changes requested):
The removal notification is working fine.
BUT...
Something is wrong but I don't know what. The removal notification works, but I get a blank drawer menu after accepting an invite. I have no idea why.
I tested over and over and over and this is not happening on the develop branch. Here is a video. 
It happens for an invited participant or coordinator. It does not seem to affect any of the other screens.

https://github.com/user-attachments/assets/f3917506-a6bc-4725-a92f-5efd03e36115

ErikSin:
> The removal notification is working fine. BUT... Something is wrong but I don't know what. The removal notification works, but I get a blank drawer menu after accepting an invite. I have no idea why. I tested over and over and over and this is not happening on the develop branch. Here is a video. It happens for an invited participant or coordinator. It does not seem to affect any of the other screens.

I fixed this BUT i don't actually know why it was happening in the first place. Its seems like the introduction of the `ProjectRemovalListener` in the navigation Layout did this for some reason. So i just moved the Listener to the tabs screen and it seemed to fix it.

cimigree (approved):
I wish you knew why the change in placement works.
I do see a blank map screen when I accept an invite but I guess it resolves?
The notification works fine.
What about this slack convo and the project removal? I am unclear on what we are supposed to be doing...
https://awanadigital.slack.com/archives/C01GJ3V8D62/p1762957540182379
Does the change in the core dependency in a separate PR affect this at all?

PR #1534, opened 2025-11-10
fix: new project invite back flow

### LINKED ISSUE #1519: “Start new project” or “Invite Devices”/ Select Device - tap back goes back to creating a project
Description :
Tapping Back at "select device to invite"  take you back to the project creation flow, not success screen
see screenshots : https://www.notion.so/digidem/Start-new-project-or-Invite-Devices-Select-Device-tap-back-goes-back-to-creating-a-project-2a11b08162d58072be58cc2ea708f756?source=copy_link

cimigree:
closes #1519
Updates navigation actions to make project creation flow handle back arrows and back button correctly.
This PR fixes two things.
1. This was caused by some behavior that got changed with the new project creation flow that we didn't update correctly. We had done this -- replacing many of the screens when navigating:
https://github.com/digidem/comapeo-mobile/pull/1259/files
But with the change in flow it was no longer necessary to replace the invite screen when going there from the Success screen. Now, with straight navigation, we can go back to the success screen from the select devices to invite screen.
Since the back button is disabled on the success screen, we can't go back from there, which makes sense and now everything works as expected. Simple and effective fix!

2. Also, while I was debugging, I noticed you can go back with the Android back button from the Project Stats sharing screen, which is not desirable as it takes you to the create project screen and you can end up creating two projects. So this PR also prevents the Android back button on the Project Stats sharing screen.

### DIFF
--- a/src/frontend/screens/ProjectCreation/ShareProjectStats.tsx
+++ b/src/frontend/screens/ProjectCreation/ShareProjectStats.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {StyleSheet, View, BackHandler} from 'react-native';
+import {StyleSheet, View} from 'react-native';
 import MaterialIcons from '@react-native-vector-icons/material-icons';
 import {defineMessages, useIntl} from 'react-intl';
 import * as Sentry from '@sentry/react-native';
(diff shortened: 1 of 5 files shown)


src/frontend/screens/ProjectCreation/ShareProjectStats.tsx
```diff
@@ -53,6 +54,18 @@ export const ShareProjectStats = ({
...
+    React.useCallback(() => {
+      const subscription = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => true,
+      );
+
+      return () => subscription.remove();
+    }, []),
+  );
+
```
ErikSin: Can you use `usePreventAndroidBackButton` instead? Also I noticed `usePreventAndroidBackButton` returns the useFocusEffect, and it should not return it-it should just be in line. So can you update it (in otherwords just take out "return")?

ErikSin (approved):
I small thing, but non blocking

PR #1535, opened 2025-11-10
fix: not create new project when track active

### LINKED ISSUE #1520: Can create new project with active track
**Describe the bug**
- Activate track on Project A
- Create new project - Project B
- User is taken to new project , without warning of active track
- Project B has active track, can save to project

Goal:
Should display warning when active track/creating or naming project. Same warning as switching project

cimigree:
closes #1520 
Blocks new project from being created when a track active. 
Uses it the same way it is done in the All Projects screen.
Do not block if just renaming a project.
When saving it, go to category chooser instead of Save Track screen.

I am not sure how to go back to the Create new project screen after saving without adding a lot of complication. Is that essential that we go back to that screen? See recording.

Screen recording:

https://github.com/user-attachments/assets/4ad47548-a058-4741-b192-24e154c9b5fc


ErikSin:
[I asked in design](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1762799761309069) when we should show this warning, and the concluded that we should show this warning much earlier and that we should show this warning when the tap on 'New Collaboration' from the menu

cimigree:
Ok I [put a comment in Slack for design](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1763034670930319?thread_ts=1762799761.309069&cid=C6ZMBFATZ), and [got a response](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1763048183002499?thread_ts=1762799761.309069&cid=C6ZMBFATZ). 

They were good to change it from "You're currently recording a track. To join project, stop recording" to "Tracks is recording. Stop Tracks to continue."

They were also ok with adding the modal pop up when someone on a default/ solo/ unnamed project presses Start New Project in addition to the New Collaboration (which works for someone on a named project).

ErikSin (approved):

PR #1542, opened 2025-11-10
fix: auth screen improperly navigating when authenticated

### LINKED ISSUE #1541: App Passcode doesnt work when App went in bg mode
**Describe the bug**
- Activate App Passcode
- Click to make App go in bg mode
- Open it again
- Add passcode , doesnt seem to work no matter how many times

- If force quit and relaunch, app passcode works

https://www.notion.so/digidem/App-Passcode-doesnt-work-when-App-went-in-bg-mode-2a71b08162d580bfa5d7e265890d5a47?source=copy_link

ErikSin:
closes #1541

The Auth had remnants of the old architecture that was causing navigation problems when the user became authenticated. Namely, we introduced conditional rendering for the auth screen, but it still had a `useEffect` that tried manually navigate back after the user was authenticated. This [goes against react navigation's suggested architecture](https://reactnavigation.org/docs/auth-flow/#dont-manually-navigate-when-conditionally-rendering-screens).

I also took the Auth Screen outside of the createAppScreens. Since it is being conditionally rendered, the other screens do not need access to the Auth screen. And when user was being authenticated, it was keeping the user on the Auth screen because the Auth screen existed inside the createAppScreens.

### DIFF
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -11,7 +11,6 @@ import {EnterPassToTurnOff} from '../../screens/AppPasscode/EnterPassToTurnOff';
 import {SetPasscode} from '../../screens/AppPasscode/SetPasscode';
 import {TurnOffPasscode} from '../../screens/AppPasscode/TurnOffPasscode';
 import {Security} from '../../screens/Security';
-import {AuthScreen} from '../../screens/AuthScreen';
 import {ObscurePasscode} from '../../screens/ObscurePasscode';
 import {ObservationCategoryChooser} from '../../screens/PresetChooser/ObservationCategoryChooser.tsx';
 import {TrackCategoryChooser} from '../../screens/PresetChooser/TrackCategoryChooser.tsx';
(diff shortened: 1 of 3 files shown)


cimigree (approved):
Two questions I hope you will think about and possibly address.

This sort of works and seems architecturally sound, but is it the expected behavior that no matter where you are in the app, after you come back from backgrounding you end up on the map screen? I imagine it is because When authState toggles, the code returns completely different Navigator instances, destroying all navigation history. Do you think that is ok? You still come back to a observation being edited or created or Track being created so at least that works. But if you are anywhere else at all, you come back to the map.

Also, it would have been great if we had an E2E to catch this issue. Think it is worth adding one?

ErikSin:
> But is it the expected behavior that no matter where you are in the app, after you come back from backgrounding you end up on the map screen? I imagine it is because When authState toggles, the code returns completely different Navigator instances, destroying all navigation history. Do you think that is ok? You still come back to a observation being edited or created or Track being created so at least that works. But if you are anywhere else at all, you come back to the map.

You are correct, since we are conditionally rendering the Auth screen, the main navigator gets destroyed. In order to save the navigation state we would have to save the navigation state to persisted state, which i think adds too much complication.

I think the tradeoff here of losing the nav state to have simpler code is worth it IMO, expecially since the user will be navigated back to creating an observation if they were in the middle of doing that.

PR #1548, opened 2025-11-11
chore: s3 release uploads

### LINKED ISSUE #1405: Automate publishing APK to comapeo.app
For the website we need a fixed URL that will download the latest release of the APK.

**Proposal**

1. https://downloads.comapeo.app/android/latest does a 302 redirect to download the latest version.
2. Versions available to download via https://downloads.comapeo.app/android/$RELEASE_NAME e.g. https://downloads.comapeo.app/android/v3.1
3. Store releases in a cloudflare R2 object store bucket.
4. Cloudflare worker to implement `/latest` redirect and proxy http requests to the R2 bucket.

**Technical details**

1. CoMapeo Mobile release process should upload the APK to R2 (R2 is S3 compatible) with a suitable consistent name that includes the version name, e.g. `/android/comapeo-v3.1.apk`
2. CoMapeo Mobile release process should publish/update a file with enough data to construct a URL to the APK. It could be that there is already a public file from EAS that we could use - they must have some public endpoint for OTA updates that could have the info we need, or we could generate a JSON.
3. Cloudflare worker, similar to [mapeo-releases-worker](https://github.com/digidem/mapeo-releases-worker/blob/main/index.js) should respond to `GET /android/latest` by reading the file from (2) to construct the URL, and pass it with a 302 redirect.
4. Cloudflare worker proxies requests to `GET /android/:versionName` to the corresponding file in the R2 bucket

cimigree:
closes #1405 
Adds a release upload script to run after a successful eas build to work with the new downloads worker on cloudflare that serves APKs from `downloads.comapeo.app`.

This works with the downloads worker deployed at https://downloads.comapeo.app which provides stable URLs for latest releases:
- https://downloads.comapeo.app/android/release/latest
- https://downloads.comapeo.app/android/rc/latest
- https://downloads.comapeo.app/android/test/latest

The worker reads `latest.json` files from R2 and redirects to the actual APK file path.


cimigree:
@gmaclennan I added the new way to get the version as you suggested. I also attempted to get the signing credentials.
Got some info from EAS.
https://docs.expo.dev/build-reference/android-builds/

> The most important part is the release signing config. It's configured to read the keystore and passwords from the credentials.json file at the project root. Even though you're not required to create this file on your own, it's created and populated with your credentials by EAS Build before running the build.

https://docs.expo.dev/app-signing/local-credentials/

cimigree:
@gmaclennan Added some conditional logic based on whether or not it is a production build.

gmaclennan:
Ok I ran some tests by running some builds on EAS from this branch with the pre-release profile.

1. Turns out the `aws` CLI is not installed on EAS build servers, so I switched to [`mc`](https://github.com/minio/mc) which is more lightweight and easy to install.
2. I changed the folder for pre-release builds to `/android/pre` because I found we already have a profile called `test` for the builds we use for e2e testing, and I feel that calling pre-release builds "test" builds can confuse them with the builds used for e2e tests.

This is working now at least for pre-releases. On the next RC and play store release we can see if this works for those once this is merged, and we can fix as needed. As this is a post-build hook, if it fails, the APK/AAB is still built and can be used.

gmaclennan (commented):
Looks good - see my fixes. Good to merge, although I am confused about the Socket warnings and all the changed files - we aren't changing any deps in this PR right?

cimigree:
No, there are no dependency changes. I have absolutely no idea what is happening with socket and why and was wondering the same thing. The only thing is that we are technically changing the package.json, by adding the script, so maybe that is why?

Glad you discovered mc and yes, using pre makes sense as we use test for tests!

gmaclennan (approved):

PR #1554, opened 2025-11-12
chore: update core dep

ErikSin:
update comapeo-core to @latest


cimigree (approved):
Do you know what this change involves? It's hard to know how to test without that. But at least we have E2E and they pass...

ErikSin:
> Do you know what this change involves? It's hard to know how to test without that. But at least we have E2E and they pass...

core follow semver versioning. Since this is a patch upgrade, there should be no change in apis and all the fixes should be internal to core

PR #1555, opened 2025-11-13
feat: hooks for last on project or last coordinator

### LINKED ISSUE #1544: Create Hook useIsLastCoordinator and useIsLastMember
Create 2 reusable hooks to determine if a user is the last coordinator of a project or the last member of a project.

### Technical Proposals:
```ts
function useIsLastCoordinator({deviceId}){
 const projectId = useActiveProjectId()
 const members = useManyMembers({projectId})
 const coordinators = membersQuery.data.filter(member =>
    COORDINATOR_ROLES.has(member.role.roleId),
  );
return coordinator.length===1 && coordinators.includes(member=>member.deviceId===deviceId_
}

```

```ts
function useIsLastMember({deviceId}){
 const projectId = useActiveProjectId()
 const members = useManyMembers({projectId})
 const membersOnProject = membersQuery.data.filter(member =>
    COORDINATOR_ROLES.has(member.role.roleId) || member.role.roleId === MEMBER_ROLE_ID
  );
return membersOnProjec.length===1 && membersOnProjec.includes(member=>member.deviceId===deviceId_
}

```

cimigree:
closes #1544 
Adds hooks in the projects file for checking if a device is the last member and/ or last coordinator on a project.


ErikSin (changes requested):
The hooks look good and I think are doing what they should be doing.

I do think they should not be in the `hooks/server` folder. I suppose there is a little ambiguity as to what lives in the server folder, but to me the main operation feels less like a server side operation and more of a client side sorting operation. So i think we should just move them into their own files in `hooks`

I also think we should attempt to write integration tests for these. Lmk if you want to try and work on those together

cimigree:
Ok hope this is what you meant by integration tests. As this PR is just about the hooks, I just added tests to test the hook...

ErikSin (approved):
TBH that is not what i was expecting for the tests. You could have created a project and added members to that project instead of mocking everything, but these tests do test the correct things....

PR #1559, opened 2025-11-17
chore: update prod deps

### LINKED ISSUE #1528: Update Deps

ErikSin:
closes #1528

No code changes were necessary for this upgrade.

I did have to change the minimum node version from 20  to 20.19 as the vector icons require 20.19 or greater (see photo below):

image

## What I did NOT upgrade
- mmkv: mmkv is being upgraded in #1560, as that upgrade is necessary for the new play console 16kb requirement, so it being done seperately so we can merge it in sooner if necessary
- react-native-reanimated: New version requires the new react-native-architecture
- @rnmapbox/maps: this was a minor upgrade, but it was causing errors that were causing the app to not build properly. It is in the roadmap to switch to map-libre soon so i just opted to no not do this upgrade.


package.json
```diff
@@ -36,32 +36,32 @@
     "@bam.tech/react-native-image-resizer": "3.0.11",
     "@comapeo/core-react": "7.2.0",
     "@comapeo/ipc": "6.0.1",
-    "@expo-google-fonts/rubik": "0.4.1",
-    "@formatjs/intl-getcanonicallocales": "2.5.5",
-    "@formatjs/intl-locale": "4.2.11",
-    "@formatjs/intl-pluralrules": "5.4.4",
-    "@formatjs/intl-relativetimeformat": "11.4.11",
-    "@gorhom/bottom-sheet": "5.2.4",
+    "@expo-google-fonts/rubik": "0.4.2",
```
ErikSin, line 39: This namespacing was a little confusing to me. I'm still not sure if `@expo`(as opposed to `expo`- with no "@") is managed by the expo upgrades.

achou11: Based on https://github.com/expo/google-fonts/tree/main/font-packages/rubik#usage, seems like this should be able to be managed via Expo CLI.

.nvmrc
```diff
@@ -1 +1 @@
-20
+20.19.5
```
achou11: Probably should update the docs in the README as well as any CI workflows that set up node to align with this version but don't read from this file.

achou11 (approved):
Overall nothing in the changelogs stands out so happy to rely on CI passing as a good enough measure of ability to merge.

> @rnmapbox/maps: this was a minor upgrade, but it was causing errors that were causing the app to not build properly.

If you remember or have it logged somewhere, curious to know what the errors were.

package.json
ErikSin: ok! i will take it out of this PR

achou11:
Forgot to ask this in my initial review: is updating the backend prod deps usually done as a separate task? Mostly just asking because I can't remember 😄 - would be reasonable if so though.

ErikSin:
> Forgot to ask this in my initial review: is updating the backend prod deps usually done as a separate task? Mostly just asking because I can't remember 😄 - would be reasonable if so though.

i think the real question here is, who is responsible for the frontend's backend?!? 💁🏽 I think we should hire a brontend/frackend (both of these sound really dirty) developer 🤣

.nvmrc
ErikSin: So i pinned most of the CI workflows to the nvmrc version. You created most of these workflows, can you just confirm that it was ok to do that?

.github/workflows/e2e-appium-browserstack.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '20'
+          node-version: '20.19'
```
achou11:
```suggestion
          node-version-file: '.nvmrc'
```

eas.json
```diff
@@ -8,11 +8,11 @@
     "base": {
       "android": {
         "image": "latest",
-        "node": "20.17.0"
+        "node": "20.19.0"
       },
       "ios": {
         "image": "latest",
-        "node": "20.17.0"
+        "node": "20.19.0"
```
achou11: Should match exact version used in dev + CI

```suggestion
        "node": "20.19.5"
},
"ios": {
"image": "latest",
        "node": "20.19.5"
```

achou11 (changes requested):
Some small adjustments needed

eas.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1559/commits/a7e85be5e4ab910297b4983ef840b711f53fb20a

.github/workflows/e2e-appium-browserstack.yml
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1559/commits/a7e85be5e4ab910297b4983ef840b711f53fb20a

achou11 (approved):
Probably worth updating with latest develop branch just in case lockfile changes need to be redone. Otherwise think ready for merge!

package.json
```diff
@@ -36,32 +36,32 @@
...
+    "@react-native-vector-icons/octicons": "20.4.0",
+    "@react-navigation/bottom-tabs": "7.8.5",
+    "@react-navigation/native": "7.1.20",
+    "@react-navigation/stack": "7.6.4",
     "@rnmapbox/maps": "10.1.42",
-    "@sentry/react-native": "6.14.0",
-    "@tanstack/react-query": "5.85.9",
-    "@turf/helpers": "7.2.0",
-    "@turf/random": "7.2.0",
+    "@sentry/react-native": "7.6.0",
```
achou11, line 61: Noting that this makes use of their [JS SDK](https://github.com/getsentry/sentry-javascript/releases/tag/10.24.0) version that introduces a breaking change where a previously experimental opt-in feature is now non-experimental and enabled by default. Would recommend disabling it explicitly for now unless it's already known that we want to have this feature enabled (I'm not aware of any decisions on this).

PR #1560, opened 2025-11-18
chore: update mmkv - 16kb requirement

ErikSin:
Upgrades react-native-mmkv to version 4. Version 4 uses nitro modules, which are compatible with both the old and new architecture. This version also satisfies the [16kb page size requirement](https://android-developers.googleblog.com/2025/05/prepare-play-apps-for-devices-with-16kb-page-size.html)


achou11 (approved):
glad the migration was straightforward! I'm inclined to believe that this fixes page size requirement but might be worth confirming that.

PR #1561, opened 2025-11-19
fix: show an error when there is an error importing categories

gmaclennan:
The `importConfig()` function used by `useImportProjectConfig()` returns an array of errors rather than throwing an error, and the code in Config.tsx ignores the return value. This method and hook are now deprecated, and the new `useImportProjectCategories()` hook will throw an error as expected if there is an error when importing.


ErikSin (approved):
Thanks

PR #1563, opened 2025-11-19
chore: new default config

ErikSin:
Updates `@mapeo/default-config` to `@comapeo/default-categories`. This includes updating the build scripts in the backend to point to the new categories.

The majority of the files changed are e2e test which create observations that use the new default categories

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -10,7 +10,7 @@ const MIGRATIONS_FOLDER_PATH = new URL(
 ).pathname
 
 const DEFAULT_CONFIG_PATH = new URL(
-  './node_modules/@mapeo/default-config/dist/mapeo-default-config.comapeocat',
+  'node_modules/@comapeo/default-categories/dist/comapeo-default-categories.comapeocat',
   import.meta.url,
 ).pathname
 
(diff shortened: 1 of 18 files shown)


src/backend/index.js
```diff
@@ -10,7 +10,7 @@ const MIGRATIONS_FOLDER_PATH = new URL(
 ).pathname
 
 const DEFAULT_CONFIG_PATH = new URL(
-  './node_modules/@mapeo/default-config/dist/mapeo-default-config.comapeocat',
+  'node_modules/@comapeo/default-categories/dist/comapeo-default-categories.comapeocat',
```
achou11: think this works fine but might be safer to explicitly write this as a relative path
```suggestion
  './node_modules/@comapeo/default-categories/dist/comapeo-default-categories.comapeocat',
```

achou11 (approved):
Didn't thoroughly check the changes in e2e tests since I trust you and CI more than myself. 1 non-blocking suggestion but otherwise lgtm!

src/backend/index.js
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1563/commits/8b8e2d87f03795683b4ca68c9cb01c539d988d2e

PR #1573, opened 2025-11-24
feat: update leave project screen

### LINKED ISSUE #1545: Update LeaveProject Screen
The leave project screen should be updated to have variable text based on whether the user leaving is a coordinator or participant

## Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8885-4207&t=j5UjthP4OUQvYvup-4

https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8885-4664&t=j5UjthP4OUQvYvup-4

## Screenshots

Image

Image

## To do
- [ ] change text based on role

cimigree:
closes #1545 

Fetches role in the Leave Project screen.
I chose to do this instead of passing it as a param since there might be an interstitial in between the Collaborator screen and this one.
Changes the text based on whether the person leaving is a coordinator or not.

**Coordinator:**
image

**Participant:**
image


src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -161,7 +161,7 @@ export const CollaboratorInfo: NativeNavigationComponent<
               navigation.navigate('LeaveProject');
             }}
           />
-        ) : ownRoleIsCoordinator ? (
+        ) : ownRoleIsCoordinator && !isOwnDevice ? (
```
ErikSin: `!isOwnDevice` is redundant

ErikSin (commented):
I would like your opinion as I am having some concerns that the check for a coordinator/participant might be a little fragile.

You did not introduce the original checks in this PR, i think we introduced them when we were first creating these features, but now that i have more of a holistic picture i am wondering if we should make some changes.

Currently in the collaborator info, leave project, (and I think the interstitial screens), we check if the user is a coordinator based on their role id - which I think is the correct way to check if they are a coordinator. But if they are not a coordinator, then we assume they are a particpant, and I think this might be a problem. There are several other roles outside of participant and coordinator, including being block, having left, and having no role at all.

Do you think this is something we should be concerned about? What are the chances that someone not a particpant and we accidently end up on the screen? Am i over thinking this?

cimigree:
> Do you think this is something we should be concerned about? What are the chances that someone not a particpant and we accidently end up on the screen? Am i over thinking this?

We are filtering on the team screen (src/frontend/screens/YourTeam/index.tsx) so only folks with coordinator type roles and specifically participant roles (```const participants = membersQuery.data.filter(
    member => member.role.roleId === MEMBER_ROLE_ID```
  );
  will end up being clickable, and then will end up on the subsequent screens.  The filtering there is very specific, so I think adding additional checks on the screens that are gotten to by clicking on the team screen are not necessary.

What kind of scenario are you imagining? I am having trouble seeing why an extra check is necessary, but it is easy enough to make something like const `isParticipant = role.roleId === MEMBER_ROLE_ID;` instead of `!isCoordinator` if it would make you feel better.

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1573/commits/d38d290a7bfc1c1fbe4ebe77009b8fedbd7ed387

ErikSin:
> We are filtering on the team screen (src/frontend/screens/YourTeam/index.tsx) so only folks with coordinator type roles and specifically participant roles (`const participants = membersQuery.data.filter( member => member.role.roleId === MEMBER_ROLE_ID` );

This is what I am worried about, and has given me clarity on why i think we should pass a nav param. We are relying on another page **indirectly** to enforce being a participant/coordinator. But a developer can use this page for other purposes ( for example we might want to show this page for past collaborators) and we are doing a really naive check because it was only navigated to from the particpants or coordinators list. It requires **implicit** developer knowledge to know that only participants and coordinators can access this screen. 

I think we need to pass a nav param instead of doing that naive check. That way we know that we can only navigate to this page as a particpant and coordinator, and if we do change this component in the future, it will be explicit that this page was only designed for particpant and coordinator.

src/frontend/sharedTypes/navigation.ts
```diff
@@ -168,12 +168,13 @@ export type RootStackParamsList = {
   ProjectStatsTurnedOff: undefined;
   EarlyAccess: undefined;
   EarlyAccessOff: undefined;
-  LeaveProject: undefined;
+  LeaveProject: {memberType: 'coordinator' | 'participant'};
   LeftProjectConfirmation: undefined;
   CollaboratorInfo: {
     deviceId: string;
     isOwnDevice: boolean;
     deviceType: DeviceType;
```
ErikSin: i don't think we should be passing deviceType. In `CollaboratorInfo` we are using `useSingleMember` which has access to deviceType

ErikSin (approved):
1 non blocking comment

PR #1581, opened 2025-11-26
Removes role belonging to device in project from all project headers.

### LINKED ISSUE #1578: Remove device role from project header
Device role should NOT be part of the project header in `useProjectRoleAndDetails()` - this was a was a miscommunication in design handoff.

### Notion:
https://www.notion.so/digidem/Remove-device-role-from-Project-Name-2b61b08162d5808fa72dec2f9c6cacbb?v=c58a7b022e084c2bae2c4379bf0c67e7

### To do:
- [ ] Remove device role in `projectHeader` in useProjectRoleAndDetails()

cimigree:
closes #1578 Which reverts https://github.com/digidem/comapeo-mobile/pull/1513
This removes the device role (participant, coordinator) from the headers on projects from all places where those headers live.
Makes changes in tests also.
I am not sure if that is right. It is now not anywhere where it used to be? Which doesn't seem to match Figma, but I guess that is why were confused in the first place 
    ¯\_(ツ)_/¯

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -2203,11 +2203,5 @@
   },
   "soloProject.startNewProject.uniqueProject": {
     "message": "Each project is unique and separate."
-  },
-  "useProjectRoleAndDetails.coordinator": {
-    "message": "Coordinator"
-  },
-  "useProjectRoleAndDetails.participant": {
-    "message": "Participant"
   }
 }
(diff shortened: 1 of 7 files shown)


ErikSin (approved):
Thanks

PR #1584, opened 2025-11-26
Fix export observation file extension

### LINKED ISSUE #1583: Data export/  Saved file doesnt end in geojson extension

**Additional context**

ErikSin:
closes #1583 

Also there was a dev only error related to `<HeaderText />`. Because it uses "Chunks" its need a key or else the react unique key error gets thrown. This PR also addresses that

### DIFF
--- a/src/frontend/screens/ExportSuccess.tsx
+++ b/src/frontend/screens/ExportSuccess.tsx
@@ -65,7 +65,9 @@ export const ExportSuccess = ({
             {...bodyMsg}
             values={{
               bold: (chunks: React.ReactNode) => (
-                <HeaderText variant="header5">{chunks}</HeaderText>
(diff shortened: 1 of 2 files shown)


src/frontend/screens/ExportSuccess.tsx
```diff
@@ -65,7 +65,9 @@ export const ExportSuccess = ({
             {...bodyMsg}
             values={{
               bold: (chunks: React.ReactNode) => (
-                <HeaderText variant="header5">{chunks}</HeaderText>
+                <HeaderText key={chunks?.toString()} variant="header5">
```
achou11, line 68: could be misremembering but think this could be a static value e.g. `"key"`. it's an issue in react-intl so no need to have a super proper solution.

achou11 (approved):
tested this on device and has the expected behavior now 👍

PR #1589, opened 2025-11-27
chore: remove Gorham Bottom sheets

### LINKED ISSUE #1579: Remove uses of Gorhom Bottom Sheet Modal
Remove all uses of the [Gorhom bottom sheet modal](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/sharedComponents/BottomSheetModal/index.tsx) and replace with the bottom sheet modal that utilizes the React Navigation and Bottom Sheet Wrapper

EDIT: work was done in #1589. It introduces bugs and weird behaviour in turn off Passcode and Observation Edit. Because of that I think we should incrementally add each bottom sheet. So I created a sub issue for each bottom sheet. Lets create a PR for each one so we can track any bugs that may be introduced.

## ~~To Do~~
- [x] ~~Replace bottom sheet in `TurnOffPasscode` ~~
- [x] ~~Replace bottom sheet in `BackgroundMaps`~~
- [x] ~~Replace bottom sheet in `ObservationCreate/HeaderLeft`~~
- [x] ~~Replace bottom sheet in `ObservationEdit/HeaderLeft`~~
- [x] ~~Replace bottom sheet in `AddRemoteArchive`~~
- [x] ~~Replace bottom sheet in `SaveTrack/HeaderLeft`~~
- [x] ~~Replace bottom sheet in `SharedComponents/CustomHeaderLeftClose`~~
- [x] Delete `sharedComponent/BottomSheetModal` folder
- [x] Uninstall `@gorhom/bottom-sheet` and remove any associated code (eg. BottomSheetModalProvider)
- [x] Remove any relevant jest configuration

cimigree:
closes #1579 
Replaces all uses of Gorham Bottom Sheet with App Screens that use the Bottom Sheet Wrapper and are navigated to.
A couple of small changes also came up.
I was not able to get from the turn off passcode bottom sheet to security without funny things happening with the back button without using pop(4). If you know of another (replace and popTo did not work)
I had to do some fiddling with the bold message text in two places because I was getting an error around "Each child in a list should have a unique key prop."  So a key needed to be added...

I could not screenshot the app passcode turning off (not permitted in that state). But here are other screenshots for comparison. It is not an exact match, but it is close!

### Custom Map Added:
| Before | After |
|--------|-------|
| Custom Map Added - Before | Custom Map Added - After |

### Custom Map Removed:
| Before | After |
|--------|-------|
| Custom Map Removed - Before | Custom Map Removed - After |

### Discard Observation Created:
| Before | After |
|--------|--------|
| image | image |

### Discard Observation Edited:
| Before | After |
|--------|--------|
| image |  image |

### Remote Archive What's Included:
| Before | After |
|--------|--------|
|  image | image |

### Discard Track Confirmation:
| Before | After |
|--------|--------|
|  image | image |


ErikSin:
I haven't had time to do a full review but I found a problem with the edit observation flow:

https://github.com/user-attachments/assets/3b3c6bae-8ad4-4f07-abfd-c5b586185099

cimigree:
> I haven't had time to do a full review but I found a problem with the edit observation flow:
> 
I spent more than 4 hours on this and https://github.com/digidem/comapeo-mobile/pull/1589/commits/f47b9ed1674420ee5b3a510675b2be77a0bb9ba9 is the best I could come up with.
The ObservationEdit screen stays in the stack and reloads the old observation during the handle discard AFTER the draft has been cleared.
Another thing that worked that I am guessing is no good, was putting the clearDraft in a set timeout after the navigation. So I did not submit that for consideration but thought you should know it is some kind of race condition/ bug with popTo.
If you have any other ideas, I welcome them! I can tell you the 10000 things I tried that didn't work in that case though

ErikSin:
I originally thought this would be a trivial change, since we were just replacing UI. But due to the bugs we have found, i think we need to do a more thorough review of each change. I think the best bet would be to close this PR, and create separate PRs for each of the bottom sheets that go replaced, and then copy and paste those change over. That way, I can more thoroughty review each feature, and if a bug does get introduces we won't have to undo ALL the changes. 

I create sub tasks in [this ticket](https://github.com/orgs/digidem/projects/39/views/13?pane=issue&itemId=141447743&issue=digidem%7Ccomapeo-mobile%7C1579) for each bottom sheet.

PR #1602, opened 2025-12-02
chore: invite integration test

### LINKED ISSUE #1580: Create integration tests for invites
Currently invites are not covered by e2e tests as it requires 2 or more devices to test. Similar to the integration [tests in sync](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/Exchange/index.test.tsx), we should be able to write tests that show 2 "devices" - created in the jest environment- that can receive invites. See [inviteToProjects](https://github.com/digidem/comapeo-mobile/blob/7204da4e9fef68e08e23ab688f033bb937bab0a3/tests/integration/helpers/core.ts#L145) to get a sense of how we can do that in a jest environment.

## To do
- [ ] Create test for receiving an invite
- [ ] Create test for accepting an invite (both for the invitee and invitor)
- [ ] Create test for reject an invite (both for the invitee and invitor)
- [ ] Create the above test while user in onboarding flow
- [ ] Create tests for being removed from a project (both for the invitee and invitor)

cimigree:
closes #1580 
Adds a test for some invite functionality testing.

### DIFF
--- a/src/frontend/screens/Invites/invites.test.tsx
+++ b/src/frontend/screens/Invites/invites.test.tsx
@@ -1,0 +1,279 @@
+import type {MapeoManager} from '@comapeo/core';
+import type {MapeoClientApi} from '@comapeo/ipc';
+import {randomBytes} from 'node:crypto';
+import {pEvent} from 'p-event';
+import {MEMBER_ROLE_ID} from '../../sharedTypes';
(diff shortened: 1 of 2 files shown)


ErikSin:
closing as we are going to replace with e2e tests

PR #1603, opened 2025-12-02
chore: can't add any device that has already been on project

ErikSin:
Removes the ability to re-add a device that has already been on a project


achou11 (approved):
Based on what was discussed internally, this achieves the goal of removing the ability to re-invite members who either left the project or were removed from it.

PR #1604, opened 2025-12-03
fix: enable re-inviting of removed and left members

gmaclennan:
This reverts https://github.com/digidem/comapeo-mobile/commit/f560bd07e1c15bfa012d469b92bd415706b56c56 and updates @comapeo/ipc which includes [this fix](https://github.com/digidem/comapeo-ipc/pull/49) which fixes issues with re-inviting members who have left or been removed.

### DIFF
--- a/docs/EndToEndTests/Team.md
+++ b/docs/EndToEndTests/Team.md
@@ -8,6 +8,9 @@
 - [Collaborator Info as Coordinator](#collaborator-info-as-coordinator)
   - [Test Objectives](#test-objectives-1)
   - [Special Considerations](#special-considerations-1)
+- [Leave Project Warning Screen](#leave-project-warning-screen)
+  - [Test Objectives](#test-objectives-2)
+  - [Special Considerations](#special-considerations-2)
 
 ## Team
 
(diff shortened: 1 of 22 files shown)


gmaclennan:
Pre-release build for this PR: https://downloads.comapeo.app/android/pre/comapeo-v9.0-pre+ec3b9ba.apk

gmaclennan:
@SocketSecurity ignore npm/@browserstack/ai-sdk-node@1.5.17

This is not great that this dependency has obfuscated code and does not have the source code available on Github as far as I can find. It is a dependency of `@wdio/browserstack-service`, but we do not actually use the features from this dependency. It has been published for over a year, and it has >150,000 weekly downloads, so I think _this version_ is low risk, however if it were to update we should be very careful if it remains obfuscated.

gmaclennan:
@ErikSin After discussing with Ximena, plan is to target this for v9, rather than a hotfix on v8. I feel like re-inviting removed and left members is a bit enough change that it needs a thorough QA as part of the next release cycle.

ErikSin (approved):
Tested on 2 devices and seems to be working!

PR #1605, opened 2025-12-03
fix: Fix translations of category names in obs list

gmaclennan:
The ObservationsList component was not using translated presets, resulting in the English names showing up.

### DIFF
--- a/src/frontend/hooks/server/track.ts
+++ b/src/frontend/hooks/server/track.ts
@@ -8,6 +8,7 @@ import {
 
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 import {useAppLanguageTag} from '../useAppLanguageTag';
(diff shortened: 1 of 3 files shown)


cimigree:
@gmaclennan, It looks like the tracks file is also not using that query. Both in the server hook and in the TrackListItem. I assume that should also be fixed?

gmaclennan:
> @gmaclennan, It looks like the tracks file is also not using that query. Both in the server hook and in the TrackListItem. I assume that should also be fixed?

Yes, could you maybe add the fix as an additional commit on this branch?

gmaclennan (commented):
@cimigree your changes look good to me

cimigree (approved):
One end to end test is failing, but if it can be merged anyway, I looked at the video of the test and it is just a Browserstack flake out.

PR #1607, opened 2025-12-03
feat: map share select device

### LINKED ISSUE #1569: Select Device Screen
Re-purpose the select device to be used for both inviting a user to a project (which is already implemented) and for sending a map to a user. The one screen should be registered to 2 navigation stack screens, being differentiated by name. The filtering of devices should be changed based on which navigation screen is being shown

## Screen Shots:

Image

## Figma:
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8999-4208&t=wtavZShoLeOJGpJm-4

## To do
- [ ] Move the [SelectDevice](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/YourTeam/SelectDevice.tsx) screen outside of 'YourTeam' folder
- [ ] Remove the [InvitableDevicesList](https://github.com/digidem/comapeo-mobile/blob/77d87deb56b8ab9aaaecaffbb4f6e9728f5cc86c/src/frontend/screens/YourTeam/SelectDevice.tsx#L61) component, and copy and paste the logic and react code directly into the parent (To avoid an unncessary nested component)
- [ ] create function that returns the list of devices (see ts example below). this function should return the list of [invitableDevices](https://github.com/digidem/comapeo-mobile/blob/77d87deb56b8ab9aaaecaffbb4f6e9728f5cc86c/src/frontend/screens/YourTeam/SelectDevice.tsx#L67) (which should be renamed SelectableDevices)
- [ ] Write unit tests for this function (Feeding project members and peers directly into the function, instead of using the react-query hooks in the function)
- [ ] Register the SelectDevice screen to another Nav Screen

Selectable Devices Function
```ts
getSelectableDevicesProps = {
 peers:PublicPeerInfo[],
 projectMembers:MemberInfo[],
 type: "invites" | "sendMap" // maybe a better name then "type" if you can think of something more meaningful
}

// in `SelectDevice` Screen
const selectableDevices = getSelectableDevices({peers, projectMembers, type: route.name === 'SelectMapShareDevice' ? "sendMap" : invites


```

cimigree:
closes #1569 

### Re-purposes SelectDevice for invites and map sharing.

- Moved SelectDevice to sharedComponents
- Registered it to two routes
- Removed InvitableDevicesList component, inlined into parent
- Created getSelectableDevices() function so that it takes inputs to make testing easier
- Adds a test for testing whether the correct devices show up for that function.
- Added a temporary "Share Map" button in BackgroundMaps for testing until we have built the proper UI

**Note: Map sharing currently only shows current project members. Do we need to actually show the people who share any project with the sharer? How do we do that?**


src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -0,0 +1,226 @@
...
+type SelectionMode = 'invites' | 'shareMap';
+
+export const SelectDevice = ({
+  navigation,
+  route,
+}: NativeRootNavigationProps<'SelectDevice' | 'SelectMapShareDevice'>) => {
+  const ssid = useLocalDiscoveryState(state => state.ssid);
+  const {formatMessage: t} = useIntl();
+
+  const devices = useInitiallyConnectedPeers();
```
ErikSin: I know you didn't originally name this, but can we rename devices to "availablePeers"

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -0,0 +1,226 @@
...
+  projectMembers,
+  selectionMode,
+}: GetSelectableDevicesParams): PublicPeerInfo[] {
+  if (selectionMode === 'shareMap') {
+    // For map sharing, return peers that are members of the current project
+    // TODO: possibly update this to include members from all projects the user is part of
+    return peers.filter(device => {
+      const isProjectMember = projectMembers.some(
+        member => member.deviceId === device.deviceId,
+      );
```
ErikSin: people who have been blocked or who have left will still return as true in this case. So we need to check that they are not of role `blocked` or `left` here.

ErikSin (changes requested):
One blocking comment. getSelectedDevice needs to filter those who are of role `blocked` or `left`. I think we should also create tests for those cases

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1607/commits/f7bc7402143717ada91a2558499c1a041e0ac6ce

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1607/commits/f7bc7402143717ada91a2558499c1a041e0ac6ce

ErikSin (changes requested):
its looking good and everything you built is good to go. The only reason i am not approving it is because the logic for selecting a device (in invites) has changed again. So you will need to update from develop, and copy and paste that logic over onto your function.

cimigree:
@ErikSin I added the new logic and adjusted and added tests for same.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -175,6 +176,15 @@ export function BackgroundMapsScreen() {
...
 
+        {/* TODO: Temporary dev button for map sharing - will be replaced with proper UI */}
+        <SecondaryButton
+          fullSize
+          text="Share Map"
+          onPress={() => {
+            navigate('SelectMapShareDevice');
+          }}
+        />
+
```
ErikSin: Lets get rid of this for now (so its not in the main branch)

ErikSin (approved):
Looks good! Just one small comment to get rid of the temp UI, but otherwise can merge! Good Job

PR #1612, opened 2025-12-04
fix: remove useEffect causing loop

ErikSin:
Fixes this issue: https://www.notion.so/digidem/From-v6-to-v8-Infinite-solo-project-creation-2bf1b08162d580cea0c9d7425e0bbb63

**Problem:** 

1. User A has to be on v6 and rename there default project (In v6 when renamed the default project we didnt create another one)
2. User B (doesn't matter what version) adds user A to a project 
    1.  then removes user A to a project
3. User A upgrades to v8. User A is notified that it has been removed from a project. When it gets removed from a project it tries to open the default project. The default project doesn't exist so it tries to create. But its getting stuck in an infinite loop and is constantly creating default projects until the user closes the app.
4. When you open the app again, there are tons of new default project because it got stuck in that infinite loop

- The key to recreating is that there is NO default projects when the user gets removed from a project
- the user can be on v7 and this issue can happen.
    - BUT the user has to be on v6 first, rename their default project, and the upgrade to v7. Since there is no default project v7 doesnt try to create one. So it is possible to see this happen with a v7 device, it just had to have been upgraded from v6 with NO default project

### DIFF
--- a/src/frontend/screens/RemovedFromProjectBottomSheet.tsx
+++ b/src/frontend/screens/RemovedFromProjectBottomSheet.tsx
@@ -13,8 +13,6 @@ import {
   useProjectSettings,
 } from '@comapeo/core-react';
 import {useActiveProjectIdActions} from '../contexts/ActiveProjectIdStoreContext';
-import {useEffect, useState} from 'react';
-import {Loading} from '../sharedComponents/Loading';
 import {UIActivityIndicator} from 'react-native-indicators';
 import {ColorCard} from '../sharedComponents/ColorCard';
 import {DEFAULT_PROJECT_COLOR} from '../constants';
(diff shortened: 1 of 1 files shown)


ErikSin:
This fix introduced a [new minor problem](https://github.com/digidem/comapeo-mobile/issues/1613). This problem happens not that often, and can be solved by simply closing the app and re-opening it.

I made the executive decision to introduce this fix and create a [hotfix for it](https://github.com/digidem/comapeo-mobile/pull/1611) even though it introduces a this minor bug.

My reasons are:
1. This fixes a [catastrophic bug](https://www.notion.so/digidem/From-v6-to-v8-Infinite-solo-project-creation-2bf1b08162d580cea0c9d7425e0bbb63) that effects data integrity.
2. The minor bug is very inconsistent and does not happen that often
3. The minor bug is solvable by user, all they have to do is close and reopen the app.
4. The announcement for Abare is planned for early next week and we do not have time to fix the minor bug as well, so Introducing the a new minor bug and fixing a catastrophic bug seems like the best strategy

ErikSin:
I tested this by [copying these changes](https://github.com/digidem/comapeo-mobile/pull/1611/commits/051421e695ed501f23e41b14563ded58caf676f2) onto the v8.2 branch and then created an RC. By doing this I was able to easily switch between v6 and v8 (this would have been extremely tedious with a dev version). Switching between v6 and v8 is necessary to replicate the original bug.

PR #1617, opened 2025-12-08
feat: new background map screen

### LINKED ISSUE #1609: Background Map screen with downloaded map
The background map screen should now a different UI when there is a background map (aka the user is not using the default map)

## Screen shots

Image

## Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8999-4165&t=4IXPJ9rFV7CynyRR-4

## To do
- [ ] Create the new background map UI
- [ ] Conditionally render it when there is a map available.

cimigree:
closes #1609  
Adds new UI for when there is a background map.
I moved the custom map details into the same file. Usually you prefer that instead of having a separate file. But I did keep it as a separate component because there is a lot of UI. Hope that is ok.
I also switched out an old button.

image


ErikSin (changes requested):
Having it as a separate component is not a problem, and i think often can make the logic a lot clearer. But the separation of your components is a little unclear to me (and maybe i am just missing some context).

It seems like to me the logic should be:

If no background map => show this entire screen:

Image

Otherwise show this entire screen:

Image

But your current logic is:
If no background map, show this text:

Image

And then ALWAYS render `<CustomMapInfoSection / >`

And then inside CustomMapInfoSection:

If no background map => render the `Choose File` button

else => show the CustomMapInfo Page.

Its a little confusing to follow. I know that there is also a bunch of conditional rendering that Andrew introduced, so perhaps there is some context that I am missing as to why you nested your conditional rendering?

cimigree:
I changed the conditional logic, as what you said made sense. Sorry I was very overwhelmed by this file. 

But in so doing, I am now showing if there is an error state, which for some reason we didn't do before? Hope this is ok. Any idea why we wouldn't have shown an error before?

So in E2E, that causes the error to render in the map screen and the regular background map screen with Choose File not to render. 

In E2E, the code is trying to fetch custom map info using the Mapbox fallback URL (which doesn't support custom maps), causing the query to error. 

So I adjusted the E2E test to check for that error screen and then back out. That seemed better than trying to make a bunch of changes in our actual code just for E2E tests.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -122,82 +149,79 @@ export function BackgroundMapsScreen() {
...
+            }}
+          />
+        ) : customMapInfoQuery.isLoading || customMapInfoQuery.isPending ? (
+          <Loading size={10} />
+        ) : hasBackgroundMap ? (
+          <MapInfoScreen
+            customMapInfo={customMapInfoQuery.data!}
+            onRemoveMap={handleRemoveMap}
+          />
+        ) : (
```
ErikSin: this can be slightly simplified.

isPending covers isLoading, so you get rid of isLoading.

And then isPending, isError, and data covers the 3 potential states of tanstack query, so you don't need to create the hasBackgroundMap const.

So you can do this:

```ts
        {customMapInfoQuery.status === 'error' ? (
          <ErrorScreen
            onRemove={() => {
              removeCustomMapMutation.mutate(undefined, {
                onError: err => {
                  Sentry.captureException(err);
                  navigate('ErrorBottomSheet');
                },
              });
            }}
          />
        ) : customMapInfoQuery.isPending ? (
          <Loading size={10} />
        ) : customMapInfoQuery.data ? (
          <MapInfoScreen
            customMapInfo={customMapInfoQuery.data}
            onRemoveMap={handleRemoveMap}
          />
        ) : ....rest}

```

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -302,6 +402,10 @@ const styles = StyleSheet.create({
     paddingHorizontal: 20,
     paddingVertical: 40,
     gap: 36,
+    flexGrow: 1,
+  },
+  containerWithMap: {
+    paddingVertical: 20,
```
ErikSin: can you just make this a `marginVertical:20` and apply it to the `styles.hasMapContainer` and then get rid of the conditional style?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -265,35 +289,111 @@ export function BackgroundMapsScreen() {
...
+      </BodyText>
+      <DestructiveButton
+        fullSize
+        text={t(m.removeMapFile)}
+        onPress={onRemove}
+        renderIcon={({color, size}) => (
+          <MaterialIcon name="delete" size={size} color={color} />
+        )}
+      />
+    </>
```
ErikSin: The error screen use to look like this:

Image

it seems like you got rid of this top part, and are only showing the bottom part, which is why the test are suddenly failing.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -265,35 +289,111 @@ export function BackgroundMapsScreen() {
...
+        )}
       />
-    );
-  }
+    </View>
+  );
+}
 
-  return customMapInfoQuery.isFetching ? (
-    <Loading size={10} />
```
ErikSin, line 294: the loading has also changed, it use to look like this:

Image

But with the new screen, the way you did it is actually the only way it will work (so dont change it back to how it was), im just mentioning it for posterity (also don't resolve this comment).

ErikSin (changes requested):
Wow this component complicated before you started working on it. I thought it was simple, but I realize now we should have discussed it before hand, sorry for the oversight.

We need to change the logic a bit, because the error screen is designed to allow the user to add a new map without removing the map first

So i think the logic should be

```tsx
// isLoading is encapsulated in isPending, so no need to have both
     customMapInfoQuery.isPending ? (
          <Loading size={10} />
        ) : customMapInfoQuery.data ? (
          <MapInfoScreen
            customMapInfo={customMapInfoQuery.data}
            onRemoveMap={handleRemoveMap}
          />
        ) : (
          <NoMapScreen error={customMapInfoQuery.error} onChooseFile={handleChooseFile} />
        )}

```

And the the NoMapScreen should incorporate the error like it did beforehand

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1617/commits/3e68054d3f754385fbea3e085507640779652a07

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1617/commits/3e68054d3f754385fbea3e085507640779652a07

This is from the E2E test. This is what it looks like now.
Screenshot 2025-12-10 at 2 54 54 PM

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -265,35 +282,115 @@ export function BackgroundMapsScreen() {
...
+          </BodyText>
+          <DestructiveButton
+            fullSize
+            text={t(m.removeMapFile)}
+            onPress={onRemoveMapFile}
+            renderIcon={({color, size}) => (
+              <MaterialIcon name="delete" size={size} color={color} />
+            )}
+          />
+        </>
```
ErikSin: this should be centered on the screen. Also, should we use the secondary destructive button so it was closer to how it looked before?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -265,35 +282,115 @@ export function BackgroundMapsScreen() {
...
+}) {
+  const {formatMessage: t} = useIntl();
+
+  return (
+    <>
+      <View style={styles.descriptionContainer}>
+        <BodyText>{t(m.description1)}</BodyText>
+        <BodyText>{t(m.description2)}</BodyText>
+      </View>
+      <ChooseMapFile onChooseFile={onChooseFile} />
```
ErikSin: Were just prop forwarding here. I think we can just get rid of this `ChooseMapFile` component and copy and past it onto this file.

If your worried about how large this file is (which i personally don't think is a big deal) we could seperate the `NoMapScreen` and the `MapInfoScreen` into seperate files

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -265,35 +282,115 @@ export function BackgroundMapsScreen() {
...
+        <SecondaryButton
+          fullSize
+          text={t(m.sendMap)}
+          onPress={() => {
+            navigate('SelectMapShareDevice');
+          }}
+          renderIcon={({color, size}) => (
+            <MaterialIcon name="send" size={size} color={color} />
+          )}
+        />
```
ErikSin, line 408: Can you comment this out for now until we have the functionality to actually share maps

ErikSin (approved):
Small non blocking things.

I think we also need to adjust the padding of the MapInfoScreen. It should be a padding of 20 all around. Its definitely more than that:

Image

PR #1621, opened 2025-12-08
feat: adds waiting for map share accept

### LINKED ISSUE #1570: Create new Waiting for Map Accept
Create screen `WaitingForMapToAccept` screen. Can copy the UI directly from [WaitingForInviteAccept](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/YourTeam/ReviewAndInvite/WaitingForInviteAccept.tsx) screen.

## Screen Shots

Image

## Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8999-4275&t=wtavZShoLeOJGpJm-4

## To Do
- [ ] Create WaitingForMapToAccept screen\
- [ ] Call `useSendMapShare` in useEffect of screen
- [ ] Register screen to nav stack

cimigree:
closes #1570 
Adds a waiting for map accept screen. 
Uses a packed version of comapeo/core-react for development purposes
for the text, in the Figma it says "Invite Sent" but I am guessing it is supposed to say Map Sent instead, right?
The frontend CI test is failing because of that packed file in our package.json. It is looking for pinning to a certain version number but the tgz file is not meeting the requirements.

---

image

---
Next steps:
- Remove the tgz and switch to the published version of @comapeo/core-react


ErikSin (commented):
Have you tried it with the [stubbed out api](https://github.com/digidem/comapeo-core-react/pull/143) instead of manually writing mapShare.ts?

Gregor included a `tgz` file in the stubbed out PR that you can use to try the new API.

You can also generate a new `tgz` file by running `npm pack` if gregor's latest updates aren't in the the file he generated.

If the instructions don't make sense on how to do that, feel free to also ask gregor to help you 🙂

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const {deviceId, mapId} = route.params;
+
+  const [time, setTime] = React.useState(0);
+  const sendMapShareMutation = useSendMapShare({projectId});
+  const requestCancelMapShareMutation = useRequestCancelMapShare({projectId});
+  const hasSentMapShareRef = React.useRef(false);
+
+  React.useLayoutEffect(() => {
+    navigation.setOptions({headerShown: false});
+  }, [navigation]);
```
ErikSin: Isn't this already set in AppScreens.tsx?

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,156 @@
...
+        ) => {
+          console.log('Map share result:', result);
+          // TODO: Handle different decisions and navigate accordingly
+          // - If decision is 'ACCEPT', the map is being downloaded
+          // - If decision is 'REJECT', check the reason (could be 'ALREADY' if they have it)
+          navigation.popTo('BackgroundMaps');
+        },
+      },
+    );
+  }, [deviceId, mapId, navigation, sendMapShareMutation]);
```
ErikSin: You can get rid of the hasSentMapShare ref completely (and also not check for an error) by destructuring sendMapShareMutation

eg

```ts
const {mutate} = useSendMapShare({projectId})

useEffect(()=>{
 mutate({deviceId, mapId}, {...rest})


},[mutate, deviceId, mapId])

```

Since mutate is a stable function, it wont rerun if the other properties of useSendMapShare change. Also the mutate deals with error, so you don't need to deal with that error check at the beginning

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,156 @@
...
+    if (hasSentMapShareRef.current) return;
+
+    hasSentMapShareRef.current = true;
+
+    sendMapShareMutation.mutate(
+      {deviceId, mapId},
+      {
+        onError: (err: Error) => {
+          Sentry.captureException(err);
+          navigation.navigate('ErrorBottomSheet');
```
ErikSin: this should actually be `navigation.replace`. We don't want the user to come back to this screen after the error. They should go back to the previous screen so they can try again

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,156 @@
...
+  usePreventAndroidBackButton();
+
+  const cancelShare = React.useCallback(() => {
+    if (requestCancelMapShareMutation.status === 'error') {
+      navigation.goBack();
+      return;
+    }
+
+    // TODO: Replace with real shareId once we track it from the send operation
+    const TEMP_FAKE_SHARE_ID = 'fake-share-id-for-testing';
```
ErikSin: I don't fully understand the api yet, where should this be coming from?

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -106,8 +108,13 @@ export const SelectDevice = ({
 
           const handlePress = () => {
             if (selectionMode === 'shareMap') {
-              // TODO: Navigate to map sharing in subsequent PR
-              console.log('Share map with device:', deviceId);
+              if (!mapId) {
+                return;
+              }
```
ErikSin: For situation like this, where the code should never go here, we should navigate to the error screen and send an error to sentry. Just so we are notified if the code does ever go here

ErikSin (commented):
Non blocking comments, but we don't want to introduce the tarball to the main branch, so I think we just need to maintain this PR as a PR for now

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1621/commits/8b92832277e4324298eaead2452d40e85716778f

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: Maybe set it in state from the mutation action? Not sure how else to do it. 
https://github.com/digidem/comapeo-mobile/pull/1621/commits/8b92832277e4324298eaead2452d40e85716778f

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1621/commits/8b92832277e4324298eaead2452d40e85716778f

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1621/commits/8b92832277e4324298eaead2452d40e85716778f

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: Removed:
https://github.com/digidem/comapeo-mobile/pull/1621/commits/8b92832277e4324298eaead2452d40e85716778f

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
ErikSin: Im not criticizing how you are doing it. I just am trying to understand where this is coming from. How is this shareId generated? You are creating a fake one right now. But when the API is ready, does this come from another api call? Again, not a criticism, and I am not asking you to change it. Im just trying to understand the API.

cimigree: The shareId comes from the response when useSendMapShare resolves. Since useSendMapShare is a long-running Promise that waits for the recipient to respond (accept/reject/already-has-it), we need to call it in the WaitingForMapAccept screen where we can show the waiting UI while it's pending. 
The flow is:
User selects device in SelectDevice → Navigate to WaitingForMapAccept with {deviceId, mapId}
WaitingForMapAccept mounts → Call useSendMapShare({deviceId, mapId})
Screen shows 'Waiting for Device to Accept Map' while Promise is pending
When recipient responds → Promise resolves with {shareId, decision} and then we navigate based on the decision


### BUT...
One limitation with the current API: the Cancel button doesn't actually work while waiting because we don't have the shareId yet (it only comes when the Promise resolves). So, this is another thing to make sure the API provides that I am pretty sure it doesn't yet have! (is this possible to resolve twice? I feel like we talked about this but I don't remember the answer). I am making a list...

ErikSin: >One limitation with the current API: the Cancel button doesn't actually work while waiting because we don't have the shareId yet (it only comes when the Promise resolves)

Yeah this is definitely a problem and should be addressed as soon as possible

> I am making a list...

Are there other things related to the API that you have concerns abou? If you do, I don't think we should wait, i think we should comment those concerns in [this PR](https://github.com/digidem/comapeo-core-react/pull/143) as soon as possible so they can be addressed as Gregor is building it. 

I would also suggest when you comment them, also ping gregor in slack so he is aware of them. At the very least please let gregor know about the limitiations with the cancel API

cimigree:
@ErikSin updated after call with Gregor. There is one little hiccup with the Typescript errors, but other than that, I think you will feel much better about this now!

ErikSin:
Closing as we will be hooking up to the api and can be re-opened when that is complete

PR #1623, opened 2025-12-10
feat: map decline screens

### LINKED ISSUE #1571: Map decline screens
Create one screen (with 2 different states) that notfies the user sending the map the map has been declined (for a simple decline, or declined due to lack of space).

## Screen shots

Image

Image

## Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8999-4299&t=wtavZShoLeOJGpJm-4

# To Do
- [ ] Create Screen
- [ ] Reason should be passed as a nav param
- [ ] change message on screen based on reason the map was declined

cimigree:
closes #1571 
Stacked on top of #1621 

I added some temporary buttons in the Select Device screen to navigate to the Decline screens so I and you could see them, but otherwise, I also tried to wire up the navigation from the Waiting For Device Accept screen as well.

Screen shots:

---

image

image


ErikSin:
Closing as it will be re-opened after being hooked up to the api.

PR #1635, opened 2025-12-11
chore: refactor observation edit

### LINKED ISSUE #1630: Refactor observation edit
Refactor observation edit to NOT use the `<Editor />`, and to be more similar to the [Observation Create Screen](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/ObservationCreate/index.tsx), where each UI piece is imported into the parent screen.

We also want to take out the [useEffect that holds all the logic](https://github.com/digidem/comapeo-mobile/blob/857f9c373940fc6f25ba4b213900161464d9458a/src/frontend/screens/ObservationEdit/index.tsx#L84), and put that logic inside the header. The header can access all the necessary info to save the observation, as that information is in the Observation Store. We do not need to use a useEffect to feed the information into the header like it currently is.

# To do
- [ ] Remove Editor
- [ ] Replace Editor with individual UI pieces
- [ ] Create ObservationEditSaveButton, which is used as the HeaderRight. Move the saving/useEffect logic into this header

ErikSin:
closes #1630


src/frontend/screens/ObservationEdit/HeaderLeft.tsx
```diff
@@ -69,7 +65,7 @@ export const HeaderLeft = ({
   function handleDiscard() {
     clearDraft();
     closeSheet();
-    navigation.popTo('Observation', {observationId});
+    navigation.goBack();
```
cimigree, line 68: I am not sure if this is a big deal, but if I am in editing an observation and force close, and go back to the editing of the observation, which all works fine, when I discard, I am back on the map screen

src/frontend/screens/ObservationEdit/ObservationEditSaveButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+    } catch (err) {
+      Sentry.captureException(err);
+      navigation.navigate('ErrorBottomSheet');
+      return;
+    }
+  }
+
+  function handleSaveSuccess() {
+    clearDraft();
+    navigation.goBack();
```
cimigree, line 143: Same is true here as in the HeaderLeft, Again. Not sure if it is a big deal but if I go back to an observation editing after force close, when I save I go to the map screen.

cimigree (approved):
There is a small UI change with navigation from reloading but it is not a big deal and I think I understand why the change was made.
I tried to test it out as best as I could and the UI seems to look and act the same except for the two minor changes I noted.
The code is a little dense, but I think I can figure it all out and it will probably be easier to make updates in the future.
I like that the save logic is split out just to make the files a bit more readable
Is the SaveTrackScreen next up? And then the Editor can be put to rest?
Do you think we can deal with the BottomSheet now? Or will that still be an issue?

ErikSin:
> Is the SaveTrackScreen next up? And then the Editor can be put to rest? 

That's the plan!

> Do you think we can deal with the BottomSheet now? Or will that still be an issue?

Yeah, we should be able to deal with the bottom sheet now. It was the useEffect that no longer exists that was causing the problem

src/frontend/screens/ObservationEdit/HeaderLeft.tsx
ErikSin: hmm, that not what I intended. ill spend some time looking at this again

src/frontend/screens/ObservationEdit/ObservationEditSaveButton.tsx
ErikSin: Same comment as above, not my intention....

src/frontend/screens/ObservationEdit/HeaderLeft.tsx
ErikSin: Ohh I misunderstood what you were originally saying. Yeah, i think that's acceptable behaviour in this case. It just simplifies the component (which you know I love).

PR #1638, opened 2025-12-16
feat: receiving map bottom sheet

### LINKED ISSUE #1572: Map Received Bottom Sheet
The bottom sheet should appear notifying the user that someone is trying to send them a map. There should be a listener (not built yet) that navigates to the bottom sheet when a map is sent. That listner should live in the RootStackNavigator's `Layout` (similar to the invite listener).

## Screen shots

Image

Image

Image

## Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8979-3919&t=wtavZShoLeOJGpJm-4

## To do
- [ ] Run useManyMapShares() in a useEffect at top `Layout` prop of Rootstack Navigator
- [ ] Create Bottom Sheet
- [ ] `Accept` navigates to REplaceBAckgroundMapScreen and passes `shareId` as a nav prop
- [ ] `Decline` calls `useManyMapShares()` (with a `reason` which is not in the stubbed API at the moment), and navigates away from the bottom sheet.
- [ ] Determine how much free space is available on the phone, show warning if there is not enough space. This determines if the location warning shows up all
- [ ] convert [bounding box to polygon](https://turfjs.org/docs/api/bboxPolygon), and then [determine if current location is in polygon](https://turfjs.org/docs/api/booleanPointInPolygon). show warning if user is not inside bounding box

cimigree:
Closes #1572

Adds the map received bottom sheet (with conditional UI for various scenarios) and necessary logic.

### Disk Space Check
- Uses `useStorageReadingQuery` to compare available space vs map size
- Space warning takes priority over location warning

### Location Bounding Box Check
- Uses turf.js geospatial library
- Converts bbox to polygon using `bboxPolygon`
- Checks if user's **current location** is inside using `booleanPointInPolygon` (is this the best version and way to access current location to use?)
- Calculates distance to map edge when outside using Turf's `pointToPolygonDistance`.
- Shows distance in km and sets via an object if needed

### PendingMapSharesListener
- Follows the `PendingInvitesListener` pattern
- Respects editing screens and existing map share screens (doesn't interrupt)
- Right now, since the API isn't real (and no one is actually sending a map), the status just remains pending, which makes the bottom sheet pop up even after an action is taken. So I am resetting with the ref. That is temporary until the other end is wired up.

## Screenshots

**Regular map received:**

image

**Location warning:**

image

**No space warning:**

image

### What is left to do that I can do:
- Navigate to `ReplaceBackgroundMapScreen` (when created in next PR)

### What is currently missing from the API:

**1. Reject Map Share - Missing `reason` Parameter**
- Currently `rejectMapShareMutationOptions` only accepts `{ shareId }`
- But, we need to accept `{ shareId, reason }` where reason is `'DISK_SPACE' | 'USER_REJECTED'| 'LOCATION'` because we can't communicate why map was rejected to sender. Gregor knows this and will add it to the API.


ErikSin:
Closing. To be reopened when hooked up to the API

PR #1641, opened 2025-12-17
chore: switch to new observation store

### LINKED ISSUE #338: Remove useMostAccurateLocation from observation

### LINKED ISSUE #1331: Finish migration to new draft observation store
I noticed that the [new draft observation store](https://github.com/digidem/CoMapeo-mobile/blob/8d02e546750dd2cbeddeeb9021cde0b1d2640b64/src/frontend/contexts/PersistedStores/DraftObservationStore.ts) is not being used, although the [Provider is being used](https://github.com/digidem/comapeo-mobile/blob/fef201da8dcdf93e99c8a567d62ca4834b1a2b9b/src/frontend/contexts/AppProviders.tsx#L92). There does not seem to be an issue for tracking finishing migration to this new store, so I am creating this so this does not get forgotten.

There are some edge-case bugs in the existing [`useMostAccurateLocationForObservation()`](https://github.com/digidem/CoMapeo-mobile/blob/edaffc9025f9334260de5cd96442535a832308d0/src/frontend/screens/ObservationCreate/useMostAccurateLocationForObservation.ts) which are addressed in the new store, among them:

- An observation position only improves while the user is on the ObservationCreate screen, not when they are adding details or changing the category of an observation.
- There are no checks in `useMostAccurateLocationForObservation()` for whether the observation is newly created or being edited, it relies on the dev knowing that they must only use this hook in a component that only shows for newly created observations, otherwise an existing observation could have its location changed, corrupting data. It would be easy for a small refactor to introduce this bug.
- An observation position could get less accurate, because the `useMostAccurateLocationForObservation()` hook re-initializes the subscription when the user navigates to a different screen like the details screen, and returns. If the location now has a lower accuracy, it will be used in-place of the existing more accurate location.
- The location continues to update as the user moves away (see [new store logic](https://github.com/digidem/comapeo-mobile/blob/fef201da8dcdf93e99c8a567d62ca4834b1a2b9b/src/frontend/contexts/DraftObservationContext.tsx#L140)), resulting in an observation possibly being recorded far away from where the user first created it (e.g. a user is in a moving vehicle while recording the observation)

As an initial step I suggest adding a comment to the inclusion of the DraftObservationProvider to highlight to devs that it is not being used, despite being included in the context, since this could result in confusion about the existing of two separate draft observation stores.

ErikSin:
closes #1331 
closes #338 

Unfortunately there is no way to do this incrementally. Because we are replacing one store with another, we have to change all the parts it touches.


ErikSin:
@cimigree our e2e tests are pretty thorough for everything related to observations. So are as functionality goes, I don't think you need too spend much time on that while reviewing this PR.

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -39,11 +40,12 @@ export const DraftObservationProvider = ({
   children,
   draftObservationStore,
 }: DraftObservationProviderProps) => {
-  createDraftObservationLocationUpdator(draftObservationStore);
+  const locationStore = useLocationContext();
+  createDraftObservationLocationUpdator({draftObservationStore, locationStore});
```
cimigree: I think this might cause a memory leak. Shouldn't it be called in a useEffect?

Like maybe

```
React.useEffect(() => {
    const cleanup = createDraftObservationLocationUpdator({
      draftObservationStore,
      locationStore
    });
    return cleanup; 
  }, [draftObservationStore, locationStore])
```

I think this is a blocking issue.

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -74,28 +72,25 @@ const MOVED_AWAY_THRESHOLD_METERS = 100;
...
+  locationStore: ReturnType<typeof useLocationContext>;
+}) {
   let appState: AppStateStatus = AppState.currentState;
-  let isNewlyCreatedDraftInStore = isNewlyCreatedDraft(instance.getState());
+  let isNewlyCreatedDraftInStore = isNewlyCreatedDraft(
+    draftObservationStore.instance.getState(),
+  );
+  let locationStoreSubscriber: (() => void) | null = null;
 
   AppState.addEventListener('change', async nextAppState => {
```
cimigree: This listener is never cleaned up or removed... Shouldn't it be? See above...

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -74,28 +72,25 @@ const MOVED_AWAY_THRESHOLD_METERS = 100;
...
-    if (appState === 'active' && !locationPermissionGranted) {
-      locationPermissionGranted = (
-        await Location.getForegroundPermissionsAsync()
-      ).granted;
-    }
     watchPositionIfNeeded();
   });
 
-  instance.subscribe(storeState => {
+  draftObservationStore.instance.subscribe(storeState => {
```
cimigree: Shouldn't this listener be cleaned up? See above...

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -131,7 +128,7 @@ async function createDraftObservationLocationUpdator({
       Date.now() - location.timestamp > STALE_LOCATION_THRESHOLD_MS;
     if (isStale) return;
 
-    instance.setState(prev => {
+    draftObservationStore.instance.setState(prev => {
```
cimigree: I manually tested the moving away... My initial location was +/- 7 meters. I moved about 50 meters away, yet the location updated. I think, based on Gregor's point 4, (The location continues to update as the user moves away resulting in an observation possibly being recorded far away from where the user first created it (e.g. a user is in a moving vehicle while recording the observation)). As far as why this is happening, I have some ideas...

My guess is that 

`const initialAccuracy = initialPosition?.coords.accuracy;` is null on first update so the `hasMovedAway` logic is skipped

This seems like a blocking bug to me.

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
```diff
@@ -378,6 +392,7 @@ type UnsavedPhotoAttachment = {
   location?: LocationObject;
   timestamp: number;
   abortController: AbortController;
```
cimigree: When I create an observation, add a photo, and force close, the photo is not there when I reopen. Perhaps because: `abortController: AbortController;` cannot be serialized to JSON?

This seems like a blocking bug to me.

src/frontend/contexts/DraftObservationContext.tsx
```diff
@@ -74,28 +72,25 @@ const MOVED_AWAY_THRESHOLD_METERS = 100;
...
+  let locationStoreSubscriber: (() => void) | null = null;
 
   AppState.addEventListener('change', async nextAppState => {
     appState = nextAppState;
-    if (appState === 'active' && !locationPermissionGranted) {
-      locationPermissionGranted = (
-        await Location.getForegroundPermissionsAsync()
-      ).granted;
-    }
     watchPositionIfNeeded();
```
cimigree: I think there might be a race condition with permissions.

`watchPositionIfNeeded()` is called before the await completes...
The permission check is async. 
If the user grants permission after the app becomes active, the subscription won't start until the next AppState change.

cimigree (changes requested):
Wow! What a monumentous task! Overall, this is going to be a super update. But unfortunately, I think there are some things that need to be fixed first.
Our E2E tests do not handle attachments or location, so I manually tested these and found some issues and for some of them guessed where in the code the problem arose. There were four other things, a couple blocking IMO too that come from code review.

From Code review:

1. Memory leak (see code comments in `DraftObservationContext`)
2. Race condition with permissions (see code comments in `DraftObservationContext`)
3. Lack of tests. You deleted `src/frontend/hooks/useDraftObservation.test.tsx` but added nothing to replace it
4. No logic for migration...Is the idea that if someone is in the middle of creating an observation, they will not be updating the app?

From testing...

1. The moved away logic (that is, not updating when moving away) doesn't seem to be working? But I am not sure... I think this warrants a unit test or more manual testing/ verification (see code comment)
2. Photos do not persist when force closing and reopening (see code comment)
3. Another potential blocker or at least an unfortunate negative change, the photo orientation is not working for me. When I take photos in landscape mode, they show up upside down in portrait mode in both thumbnails and in full view. In production, they look perfect.

src/frontend/contexts/DraftObservationContext.tsx
ErikSin: I don't think I fully understand what you mean here. `appState = nextAppState` is synchronous (no await needed) and then  `watchPositionIfNeeded()` is called.

I think I don't fully understand where the race condition is happening.

src/frontend/contexts/DraftObservationContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1641/commits/bdc72866a26ff7ea8d28570f1633b536afcba14e

src/frontend/contexts/DraftObservationContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1641/commits/bdc72866a26ff7ea8d28570f1633b536afcba14e

src/frontend/contexts/DraftObservationContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1641/commits/bdc72866a26ff7ea8d28570f1633b536afcba14e

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
ErikSin: I didn't write most of this code, but is the abort actually ever called? Perhaps I am just not seeing it, but I think we are creating the abort controller but never calling it, and it should be called during clearDraft or delete photo

src/frontend/contexts/DraftObservationContext.tsx
cimigree: It looks like you already fixed this!
It is much cleaner now and avoids the possible race condition I was worried about.
Nice.

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
cimigree: I think you are right that it should be called.
In deleteUnsavedAttachment and also clearDraft
To explain further...
`JSON.stringify() `converts the `AbortController` instance to {} (empty object) because AbortController doesn't have a toJSON() method (It's in the attachment from here:
```
const newAttachment = createNewPhotoAttachment({
      id: nextAttachmentId++,
      metadata,
      picture,
    });
    const {
      abortController: {signal},
      id,
    } = newAttachment;
    _addAttachment(newAttachment);

```
When the app reopens and tries to recreate the persisted state, the type guard in src/frontend/lib/attachmentTypeChecks.ts (isUnsavedPhotoAttachment) fails because

`!isAbortController(candidate.abortController)` // Returns false because` {} `is not an AbortController instance
So the attachment is rejected and the photo disappears.

ErikSin: ok i took the abort controller out of persisted state: https://github.com/digidem/comapeo-mobile/pull/1641/commits/cc5ce181b8eac617ec72a0132f982b0f6a217372

I also called the abort controller in the delete and in the clear function

ErikSin:
> 4\. No logic for migration...Is the idea that if someone is in the middle of creating an observation, they will not be updating the app?

I kind of forgot about migration...That being said, I'm not sure if it worth writing the code. I think it will be quite complicated code to go from the previous store to this new store. And The chances that someone is in the middle of creating an observation, closes the app, updates the app, and reopens the app seems very low to me. As well, I don't think it would be that detrimental to lose one observation if that was to happen.

But maybe I will ask the rest of the app dev team on thursday and see what they think.

src/frontend/contexts/DraftObservationContext.tsx
ErikSin: But initial position eventually gets set, no? Maybe lets touch base tomorrow, im not sure i know why this is happening and it seems like you have some insight that might be helpful!

ErikSin: Found it. We were returning the distance in km not m, but doing all the measurement in m... fix here: https://github.com/digidem/comapeo-mobile/pull/1641/commits/b96d09be091d33e55a5a598942f910dfd6e10ed4

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
```diff
@@ -480,33 +561,33 @@ function reasonToError(reason: unknown): Error {
...
-      else rotation = 90;
-    }
-  } else if (x > -ACC_AT_45_DEG && x < ACC_AT_45_DEG) {
-    // camera is vertical
-    if (y <= 0) rotation = 180;
-    else rotation = 0;
-  } else {
-    // camera is horizontal
-    if (x >= 0) rotation = -90;
-    else rotation = 90;
```
ErikSin, line 507: @gmaclennan I took this out. The logic is correct, BUT the photos are already being rotated by expo camera.

in expo camera, we have [skipProcessing](https://docs.expo.dev/versions/latest/sdk/camera/#camerapictureoptions) as `false`. This means that it automatically rotates the photo. So what was happening was that expo was rotating the photo, and then we were rotating the photo AGAIN with this function.

In the case that it might accidentally fail to rotate, we can determine that based on the exif info. So instead of using the accelermetor, we can just use the exif info and only rotate it when the exif info shows that the rotation has not been done.

I know that older devices were having trouble with rotation, but i think we have that covered now. But can you confirm if there is any other context that i might be missing that would require us to use the accelerometer?

ErikSin:
> Another potential blocker or at least an unfortunate negative change, the photo orientation is not working for me. When I take photos in landscape mode, they show up upside down in portrait mode in both thumbnails and in full view. In production, they look perfect.

Fixed here: https://github.com/digidem/comapeo-mobile/pull/1641/commits/a42322c12b91abd162daf963896aa68d1a73a82c

[Some context on what was happening](https://github.com/digidem/comapeo-mobile/pull/1641#discussion_r2738423590).

ErikSin:
> 3\. Lack of tests. You deleted `src/frontend/hooks/useDraftObservation.test.tsx` but added nothing to replace it

I added tests: https://github.com/digidem/comapeo-mobile/pull/1641/commits/2294a79275c4e918966fb98ed57cfc64347e78e5

ErikSin:
@cimigree I think i addressed everything, exception for migration. I will confirm tomorrow, but I don't think we should do migration...

cimigree (approved):
Looks great!
I reviewed the code changes and tested various things manually including the missing photos after force close, the orientation issues, and the move away functionality.
Wow! What a triumph and a nice improvement this will be.

src/frontend/contexts/PersistedStores/DraftObservationStore.ts
```diff
@@ -46,11 +53,35 @@ export function convertPosition(
...
+  // We create an abortController in memory which allows us to abort photos being processed
+  const abortControllers = new Map<number, AbortController>();
   let instance: StoreApi<DraftState>;
 
   if (persist) {
-    instance = createStore(
-      createPersistedState(createEmptyStoreState, {name: '@MapeoDraft'}),
+    instance = createStore<DraftState>()(
+      createPersistedState(createEmptyStoreState as () => DraftState, {
+        name: '@MapeoDraftStore',
```
ErikSin, line 64: I updated the store name to `@MapeoDraftStore` so that the old store wouldn't be used and cause the app to crash (since we are not migrating)

PR #1648, opened 2026-01-06
chore: use new bottom sheet observation edit

### LINKED ISSUE #1643: Remove Gorhom Bottom Sheet Model from Observation Edit
Now that the `<Editor/>` has been removed from the Observation Edit, we can take out the Gorhom Bottom Sheet Modal that is used to confirm the user navigating away.

## To do
- [ ] Refactor the bottom sheet as a stack navigation page
- [ ] Remove the use of the gorhom bottom sheet

cimigree:
close #1643 
Uses our bottom sheet for handling the attempted discard of an observation being edited.

### Confirm Discard of an Observation while Editing it:
| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -3,17 +3,6 @@
     "description": "Button on dialog to keep editing (cancelling close action)",
     "message": "Continue editing"
   },
-  "AppContainer.EditHeader.discardChangesButton": {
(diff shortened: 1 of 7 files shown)


ErikSin (approved):
Looks good! on our way to getting rid of that bottom sheet dep!

PR #1649, opened 2026-01-06
chore: remove Gorhom bottom sheet from passcode

### LINKED ISSUE #1644: Remove Gorhom Bottom Sheet from Passcode
Remove the Gorhom Bottom Sheet from the `AppPasscode/TurnOffPasscode.tsx`

## To do
- [ ] Refactor the bottom sheet as a stack navigation page
- [ ] Remove the use of the gorhom bottom sheet

cimigree:
closes #1644 
Gets rid of the Gorhom bottom sheet when turning off the app passcode.
Adds a screen using BottomSheetWrapper
Navigates to it when turning off the passcode
When the passcode is turned off, we have to pop(4) to get back to security and remove unnecessary screens from the stack. I can figure out no other way to do this. popTo('Security') seems like it should work, but it does not.
I can't take screen shots while the app is in passcode mode, so can't take photos of the before and after. 

Since the other PR (#1648 ) was merged before this one, I added the deletion of the old Bottom Sheet to this PR.


src/frontend/screens/AppPasscode/TurnOffPasscodeBottomSheet.tsx
```diff
@@ -0,0 +1,86 @@
...
+export const TurnOffPasscodeBottomSheet = ({
+  navigation,
+}: NativeRootNavigationProps<'TurnOffPasscodeBottomSheet'>) => {
+  const {formatMessage: t} = useIntl();
+  const {setPasscode} = useSecurityActions();
+
+  async function unsetAppPasscode() {
+    await setPasscode(null);
+    // Popping 4 removes the bottom sheet, DisablePasscode, EnterPassToTurnOff, and lands on Security
+    navigation.pop(4);
```
ErikSin: I figured out why this wasn't working. It is because we are awaiting setPasscode. I don't know why that's causing navigation errors, but it is...

Regardless, we should be showing a loading state when the passcode is being set to null

```ts
const [isLoading, setIsLoading] = useState(false)

function unsetAppPasscode() {
  setIsLoading(true);
  setPasscode(null)
   .then(()=>{ 
      setIsLoading(false)
      navigation.popTo('Security')
   })
   .catch(//navigate to error bottome sheet here)

 
}

// show uiActivityIndicator in place of buttons while isLoading is set to false

```

ErikSin (changes requested):
Looks good. Just one blocking comment about loading state (and solves the popTo problem).

cimigree:
The change in code did not enable me to use popTo successfully. It navigates to Security (it always did) but the other screens stay in the stack, so when you go back from security, you go back through with the bottom sheet open and the Passcode screen. I am confused why you thought it worked. So I left it as pop(4), assuming we don't want users randomly navigating back through screens they should not be seeing.

I also added some code so we wouldn't get this error:
 ```
ERROR  Warning: Each child in a list should have a unique "key" prop.

Check the render method of `Text`. It was passed a child from AppPasscode. See https://react.dev/link/warning-keys for more information.

  26 |  */
  27 | export const BodyText = ({
> 28 |   children,
     |           ^
  29 |   style,
  30 |   variant,
  31 |   ...otherTextProps

Call Stack

```

cimigree:
I spent over an hour on this trying to figure out why a null passcode would cause popTo not to remove screens from the stack. Sorry to say I didn't figure anything out.
If I only pop 3 I end up on this screen:
image

So I left it at pop(4).

ErikSin (approved):

PR #1655, opened 2026-01-13
chore: upgrade expo 54

### LINKED ISSUE #1647: Upgrade to Expo 54

cimigree:
closes #1647 
stacked on top of #1651

Upgrades to Expo 54, the last Expo upgrade that does not force the new architecture.
### The biggest differences are related to 
1. using edge to edge on Android, so I had to add some SafeAreaViews using 'react-native-safe-area-context' (SafeAreaView from react-native is deprecated) and adjust some of the views that involve keyboards -- the password setting screens and the observation create screen. The status bar is now colored dark and has a white background. The navigation bar is now colored medium grey, which matches what it used to be and is more noticeable and useful on my Samsung (that has a navigation bar) than the Pixel 8 in most of the photos.
2. the expo/file-system being deprecated, so we have to use a legacy version, which we can do at least through Expo 55.
### Small Changes
1. Some other small changes were necessary, like upgrading the node version locally, but it was actually correct in our `.nvmrc ` file
2. I adjusted some of the tests to try to be less flakey by using different categories -- ones that are more centrally located on the first screen.
3. I used a useEffect in input passcode because I was getting warnings there from the react compiler
4. I changed the key in the keyboard accessory because using index is not recommended

### Screenshots of Google Pixel 8a
image
image
image
image
image
image
image

### Screenshot of Samsung A25
image

### FYI:
I tested on all my emulators in Android studio with various sizes of phone from Android 30 - 35.


package.json
```diff
@@ -166,31 +166,34 @@
     "@wdio/spec-reporter": "9.16.2",
     "@wdio/types": "9.16.2",
     "babel-jest": "30.0.4",
+    "babel-plugin-react-compiler": "1.0.0",
```
ErikSin, line 169: According to expo, [the compiler is already configured in expo 54](https://docs.expo.dev/guides/react-compiler/). So we shouldn't need this.

src/frontend/lib/file-system.ts
```diff
@@ -66,20 +66,18 @@ export async function selectFile({
 export async function getExpoImageStorageSize(
   imageURL: string,
 ): Promise<number> {
-  let fileInfo: FileSystem.FileInfo;
```
ErikSin, line 69: I think we should keep this type definition

src/frontend/hooks/useKeyboardListener.ts
```diff
@@ -4,16 +4,19 @@ import {Keyboard} from 'react-native';
 
 export function useKeyboardListener() {
   const [keyboardVisible, setKeyboardVisible] = useState(false);
+  const [keyboardHeight, setKeyboardHeight] = useState(0);
```
ErikSin, line 7: what is this change for?

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -57,15 +59,20 @@ export const RootStackNavigator = () => {
...
-        navigateToInviteScreen={inviteId =>
-          navigation.navigate('InviteReceived', {inviteId})
-        }
-      />
-      {children}
-    </React.Suspense>
+    <SafeAreaView
+      edges={['bottom']}
+      style={{flex: 1, backgroundColor: MEDIUM_GREY}}>
+      <StatusBar barStyle="dark-content" />
```
ErikSin, line 65: I think we should reconsider this. The push from Android is to have edge-to-edge-display. What your doing is adding a status bar to the edge-to-edge to make it look like it is not edge-to-edge.

I understand that this is probably the easiest way but I think we should spend time looking into how to properly implement edge to edge.

ErikSin (changes requested):
2 blocking comments.

1. On doing some reading, I think we need to do expo upgrade first before doing the compiler. It looks like the configuration is done for the compiler out of the box in expo 54.
2. We should try to properly implement edge to edge. If you don't mind, I can try to take this on (and do it as a PR ontop of this PR)

PR #1656, opened 2026-01-13
fix: fallback translations to user device settings language

gmaclennan:
Uses the locale from the users device settings as a fallback for translations. If the user has defined multiple locales, the fallbacks will be used in order. If no translation exists in any language, then English from the default messages is used.

For this to work it needs ~https://github.com/digidem/comapeo-mobile/pull/1557~ EDIT: https://github.com/digidem/comapeo-mobile/pull/1667 to be merged - that PR is a result of modifying the crowdin settings to not include untranslated strings in the translated files.

I noticed that some languages (e.g. api.json) include some strings that are marked as "translated" but are still in English. These strings will appear in English if the language is set to api. This needs to be fixed in Crowdin by deleting translated strings that are still actually in English.


ErikSin (dismissed):
This still is not working as intended unfortunately.

I set my language to system language (in android settings) to Portuguese, and it is still always reverting to english if the translation is not avaialable. I checked the language files, and chose languages where the strings were not available in English, but it is still reverting to english

gmaclennan:
> This still is not working as intended unfortunately.

This should be fixed in https://github.com/digidem/comapeo-mobile/pull/1656/commits/bcdd1ab95307cea3c9d24f990f2740a62e861b00. The issue was that the system language preference is normally returning the BCP-47 code with the regional variant (e.g. `pt-BR`), but our translations are currently only using the 2-letter code (e.g. `pt`), so this was not matching our translations. When we update our translations to support regional variants, we should probably make this code a bit more intelligent to also pull in regional variant translations from other regions with the same language code as fallbacks.

achou11 (approved):
Waiting for CI to pass but otherwise changes seem to work as expected.

PR #1662, opened 2026-01-21
chore: icon title description shareable component

cimigree:
towards #1653 
Adds a shared styled component and uses it in many files.
It is for an icon with a title beneath and an optional description below that.
It's pretty simple and I didn't use it where it might be too complicated. I know that there are more places it could be used for future development, like for the map sharing.
Screen shots of new screens where it is used below:
image
image
image
image
image
image
image
image
image
image


src/frontend/sharedComponents/IconTitleDescription.tsx
```diff
@@ -0,0 +1,65 @@
...
+ *   icon={<SuccessCheck />}
+ *   title={formatMessage(m.success)}
+ *   description={formatMessage(m.successMessage)}
+ * />
+ */
+export const IconTitleDescription = ({
+  icon,
+  title,
+  description,
+  headerVariant = 'header2',
```
ErikSin: Lets get rid of the `headerVariant` since its not being used anywhere. And i think these components should be as opinionated as possible

src/frontend/sharedComponents/IconTitleDescription.tsx
```diff
@@ -0,0 +1,65 @@
...
+  gap = 20,
```
ErikSin: I'm also wondering if we can get rid of `gap` and just have it fixed to 20. I know there were a couple components that had gap set to 10, but i don't think it would be detrimental to just allow the gap to be 20. Can we try it and see what it looks like?

ErikSin (changes requested):
Looks good, and I think this is a going to be really useful.

I have 2 small, blocking comments. I think we should try to make this as opinionated as possible.

src/frontend/sharedComponents/IconTitleDescription.tsx
cimigree: I think it looks fine. What do you think?
gpspermissionsback
didnotmovebs
image

ErikSin: yeah, i think that's fine. Lets just get rid of the gap prop and have it opinionated as set to 20

cimigree: I already did? https://github.com/digidem/comapeo-mobile/pull/1662/commits/dcb7ee43a3bc47f2378c95a83601df42fc2988cd

cimigree:
I have this change to a test file in here that I also have in the Expo upgrade... Sorry for the duplication. I really wanted to fix the annoying flaky test ASAP I guess. I can leave it here or take it out of here if you think it's best or I can make a separate PR with just test changes or... Whatever you prefer.

ErikSin (approved):
looks good, excited to have this!

PR #1671, opened 2026-01-28
chore: expo 54 upgrade

### LINKED ISSUE #1497: Fix flaky e2e tests

### LINKED ISSUE #1647: Upgrade to Expo 54

cimigree:
closes #1647 
closes #1497 

Upgrades to Expo 54, the last Expo upgrade that does not force the new architecture.
### The biggest differences are related to 
1. using edge to edge on Android, so I had to add some safe area insets to make sure the screen was below the status bar using 'react-native-safe-area-context' (SafeAreaView from react-native is deprecated) and **adjust some of the views that involve keyboards -- the password setting screens and the observation create screen**. That is why the height component of the keyboard listener. I mostly needed it with Observations and the actions row.

The navigation bar is now colored medium grey, which matches what it used to be and is more noticeable and useful on my Samsung (that has a navigation bar) than the Pixel 8 in most of the photos. 

I also had to adjust the status bar content color depending on if the screen was dark or light. For the default, since our app has mostly a white background, I used the dark content. For some others with dark backgrounds (see photos) I used light content. For the Camera screen, I had to use the StatusBar since the options couldn't be set in the tabs.

2. the expo/file-system being deprecated, so we have to use a legacy version, which we can do at least through Expo 55.

### Small Changes
1. Some other small changes were necessary, like upgrading the node version locally, but it was actually correct in our `.nvmrc ` file
2. I adjusted some of the tests to try to be less flakey by using different categories -- ones that are more centrally located on the first screen.
3. I used a useEffect in input passcode because I was getting warnings there from the react compiler
4. I changed the key in the keyboard accessory because using index is not recommended

image

image

image


src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import {StyleSheet, View} from 'react-native';
-import {Ionicons} from '@expo/vector-icons';
```
ErikSin, line 3: What was`@expo/vector-icons`? It was never in our package.json so how were we using it?

package.json
```diff
@@ -179,18 +180,18 @@
     "fastify": "4.29.1",
     "globals": "16.5.0",
     "husky": "9.1.7",
-    "jest": "30.2.0",
-    "jest-expo": "53.0.13",
+    "jest": "29.7.0",
```
ErikSin: is the downgrading of jest intentional?

package.json
```diff
@@ -130,6 +130,7 @@
     "@babel/core": "7.28.5",
     "@babel/preset-env": "7.28.5",
     "@babel/runtime": "7.28.4",
+    "babel-preset-expo": "54.0.9",
```
ErikSin, line 132: What is this being added for?

ErikSin:
For posteristy, as discussed during the meeting, we should attempt to upgrade our node environment to the lts node release (v24).

cimigree:
Also Upgraded to Node 24
Updated the .nvmrc
I also had to update the build-translations.mjs file because In Node 24, t[he path property was removed from Dirent objects](https://github.com/nodejs/node/pull/55548). This means file.path now returns undefined, which causes errors when trying to use it with path.join().

ErikSin:
> Also Upgraded to Node 24
> Updated the .nvmrc

Im pleasantly surprised that went without a hitch. I was expecting all the CI to break!

README.md
```diff
@@ -11,13 +11,10 @@ The next version of Mapeo Mobile
    ```
 
 2. Set up the development environment
-
    1. Install [Node](https://nodejs.org), ideally through a proper node version manager such as [nvm](https://github.com/nvm-sh/nvm), [fnm](https://github.com/Schniz/fnm), [asdf](https://asdf-vm.com/), or [mise](https://mise.jdx.dev/)
-
-      - This project uses **Node v20** with **npm v10**
+      - This project uses **Node v24+** with **npm v10** (LTS)
```
ErikSin: I think we should be using the npm version that corresponds with node version, which you can find [here](https://nodejs.org/en/about/previous-releases) - if you click "details".

based on that we should be using npm 11

.nvmrc
```diff
@@ -1 +1 @@
-20.19.5 
+24
```
ErikSin: I think we should be more specfic and set it to 24.13.0 (the latest)

ErikSin (changes requested):
Looks really good. the edge to edge is really nice. It also inadvertently fixes the aspect ratio on the camera!

Remote archive is being sketchy, but I think that isn't related to this upgrade. Just mentioning it for posterity.

I went through the entire manual testing guide. I did notice a weird error where if you switched the exchange settings the whole app crashed, but I CANNOT recreate it. So not a blocker, but if you don't mind trying it just to confirm.

2 blocking comment, which are pretty small
 
1. the track screen is a little off. Im not sure why this PR touched that, but it does not look like this in develop

Image

2. I think we should pin the node version to the highest node version in the nvmrc. Im hoping this is small thing....but i see that tests are breaking now and my (uneducated) guess is that it is due to the node version.

src/frontend/screens/Onboarding/Success.tsx
cimigree: https://docs.expo.dev/guides/icons/#expovector-icons

> The @expo/vector-icons library is installed by default on the template project using npx create-expo-app and is part of the expo package. It is built on top of [react-native-vector-icons](https://github.com/oblador/react-native-vector-icons) and uses a similar API. It includes popular icon sets you can browse at [icons.expo.fyi](https://icons.expo.fyi/).

cimigree:
Ok I think I fixed the track screen. It no longer seems to need the safe area context since we are dealing with that in the app as a whole. It was just doing extra unneeded spacing in here, I think, which we must have needed before.
image


Also, I tried switching the exchange settings and all was fine. Back and forth and back and forth again.

package.json
cimigree: It is necessary now for Expo 54. I didn't run the automatic upgrade and I missed adding it, thinking it was just for the compiler. So when I tried to run the code I got this error:
"Cannot find module 'babel-preset-expo'"

ErikSin (approved):
Looks Good, especially the edge to edge on the map screen!

PR #1672, opened 2026-02-02
chore: tests for coordinate formats

### LINKED ISSUE #1062: add tests for coordinate format helper functions

ErikSin:
closes #1062

### DIFF
--- a/src/frontend/lib/coordinateFormat.test.ts
+++ b/src/frontend/lib/coordinateFormat.test.ts
@@ -1,6 +1,13 @@
 import {is} from 'valibot';
 
-import {CoordinateFormatSchema, convertToUTM} from './coordinateFormat';
+import {
+  CoordinateFormatSchema,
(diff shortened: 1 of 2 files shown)


cimigree (approved):
Using NOAA as the authoritative source is a good foundation. Cool.
Good edge case handling.
Good to test going in either direction of conversion.
Helpful comments 

The only thing that you could add though it is not necessary as this is thorough as is, is  a test for UTM zone boundary transitions (for example, coordinates at 0 degrees vs 6 degrees longitude to show zone numbering changes).

PR #1673, opened 2026-02-02
chore: add error boundary at top level

### LINKED ISSUE #930: Set up error boundary at the very top level
At the moment, we don't have an error boundary set up to catch unhandled runtime errors that may occur in the application, which causes the app to crash without surfacing any useful information to the user.

We should at least have one set up near the top-level of the application, similar to what [Mapeo does](https://github.com/digidem/mapeo-mobile/blob/0c0ebbb9ef2261e21cd1d1c8bd5ab2fe42017ea3/src/frontend/App.js#L48).

One proposed solution for implementation is to use the error boundary provided by Sentry ([docs](https://docs.sentry.io/platforms/react-native/integrations/error-boundary/)), which I presume would have the functionality of automatically reporting errors to sentry, which is probably what we'd do anyways.

ErikSin:
closes #930

To test, I threw an error in `<Serverloading/>` and in `<LocaleStoreProvider/>`.  Note: In development mode, React will rethrow errors caught within an error boundary to the global error handler. This means the error will look like an uncaught error, but if you dismiss that error, you will see the error boundary fallback.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -43,6 +43,7 @@ import {Suspense} from 'react';
 import {Loading} from './sharedComponents/Loading.tsx';
 import {createEarlyAccessStore} from './contexts/EarlyAccessContext.tsx';
 import {FatalError} from './screens/FatalError.tsx';
+import {FatalErrorUntranslated} from './screens/FatalErrorUntranslated.tsx';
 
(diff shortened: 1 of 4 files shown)


cimigree (approved):
Very cool to have the two levels of Error Boundary. That should cover everything?!
I threw the errors and saw the nice screen.
I was not able to restart the app with the hard coded errors, obvi.

PR #1646, opened 2026-01-05
chore: upgrade dev deps

cimigree:
closes #1624 
## Dev Dependency Upgrade (January 2026)

This PR updates the development dependencies as part of our 6 month scheduled maintenance.
Used `npm run upgrade-deps:dev` and `expo-doctor`. 
I also manually bumped key supporting packages (like Expo SDK patch, React Native patch, Jest, TypeScript) and reviewed compatibility with Expo SDK **`53.0.25`** and React Native **`0.79.6`**. This quiets the warnings whenever I do `npx expo prebuild`.

### Upgraded Packages

All changes were either **minor or patch-level**, with the exception of two intentional major upgrades (`eslint-plugin-react-hooks` and `dotenv-cli`). All upgrades have been tested locally and in CI.

| Package | From | To |
|---------|------|----|
| `@babel/core`                        | 7.28.0   | 7.28.5   |
| `@babel/preset-env`                  | 7.28.0   | 7.28.5   |
| `@babel/runtime`                     | 7.27.6   | 7.28.4   |
| `@comapeo/cloud`                     | 0.1.0    | 0.3.0    |
| `@comapeo/schema`                    | 2.1.1    | 2.2.0    |
| `@eslint/eslintrc`                   | 3.3.1    | 3.3.3    |
| `@eslint/js`                         | 9.30.1   | 9.39.2   |
| `@formatjs/cli`                      | 6.7.2    | 6.8.2    |
| `@tanstack/eslint-plugin-query`      | 5.81.2   | 5.91.2   |
| `@turf/distance`                     | 7.2.0    | 7.3.1    |
| `@types/react`                       | 19.0.10  | 19.2.7   |
| `@types/semver`                      | 7.7.0    | 7.7.1    |
| `@wdio/browserstack-service`         | 9.16.2   | 9.21.0   |
| `@wdio/globals`                      | 9.16.2   | 9.17.0   |
| `@wdio/local-runner`                 | 9.16.2   | 9.21.0   |
| `@wdio/mocha-framework`              | 9.16.2   | 9.21.0   |
| `@wdio/spec-reporter`                | 9.16.2   | 9.20.0   |
| `@wdio/types`                        | 9.16.2   | 9.20.0   |
| `babel-jest`                         | 30.0.4   | 30.2.0   |
| `dotenv-cli`                         | 8.0.0    | 11.0.0   |
| `eslint`                             | 9.30.1   | 9.39.2   |
[...]

PR #1674, opened 2026-02-03
chore: eslint upgrade

### LINKED ISSUE #1665: Upgrade Eslint

cimigree:
closes #1665 

The recent dev dependencies upgrade (PR #1646) introduced a whole bunch of  eslint warnings (~68) from stricter linting rules. The warnings come from two main sources:

1.`eslint-plugin-react-hooks@7.0.1` - New stricter rules for React Hooks.
2.`@eslint-react/eslint-plugin@1.52.2` - React 19 best practices and patterns

Changes made and why.
#### 1. **Date.now() During Render** 
In `DateDistance.tsx` - Replaced `Date.now()` with `useCurrentTime(10000)` hook
In `AppUsageStatsContext.tsx` - Moved `Date.now()` check into `useEffect`

**Why**? Calling an impure function can produce unstable results that update unpredictably when the component happens to re-render.

#### 2. **setState in Effect**
In `useCurrentTime.ts` - Removed redundant `setResult()` call (initial state sufficient)
In `usePasscodeLockout.ts` - Derived `minutes` from `lockUntil` using `useMemo` instead of storing in state
In `ActiveProjectIdStoreContext.tsx` - Moved `setIsInitialized`with `Promise.resolve().then()`

**Why**? Synchronous setState in effect body causes cascading renders that hurt performance.

#### 3. **Ref Access During Render** 
In `InputPasscode.tsx` - Moved `inputRef.current?.focus()` and input clearing logic into `useEffect`

**Why**? Refs should only be accessed in effects or event handlers, not during render.

#### 4. **Array Index as Key** 
- `KeyboardAccessory.tsx`
- `SelectOne.tsx`

**Why**: I think we know this one!

I also disabled test mock warnings for test files only.

I think that's everything? LMK if you have questions.


src/frontend/contexts/ActiveProjectIdStoreContext.tsx
```diff
@@ -71,22 +71,19 @@ export const ActiveProjectIdStoreProvider = ({
...
-          setIsInitialized(true);
           return;
         }
 
         const fallbackProjectId = projects[0]?.projectId;
-
         if (fallbackProjectId) {
           store.actions.setActiveProjectId(fallbackProjectId);
-          setIsInitialized(true);
-          return;
```
ErikSin, line 89: why did we do this?

src/frontend/contexts/ActiveProjectIdStoreContext.tsx
```diff
@@ -71,22 +71,19 @@ export const ActiveProjectIdStoreProvider = ({
   useEffect(() => {
     if (isInitialized) return;
     if (store.instance.getState().projectId) {
-      setIsInitialized(true);
+      Promise.resolve().then(() => setIsInitialized(true));
```
ErikSin: What is this for?

src/frontend/contexts/AppUsageStatsContext.tsx
```diff
@@ -112,10 +112,13 @@ export const AppUsageStatsProvider = ({
...
-  const optInStartedAt = value.instance.getState().optInStartedAt;
-  if (optInStartedAt && Date.now() - optInStartedAt > ONE_YEAR_MS) {
-    value.actions.resetOptInAndTimer();
-  }
+  useEffect(() => {
+    const optInStartedAt = value.instance.getState().optInStartedAt;
+    if (optInStartedAt && Date.now() - optInStartedAt > ONE_YEAR_MS) {
+      value.actions.resetOptInAndTimer();
+    }
+  }, [value]);
```
ErikSin: `value` references a stable store that never changes. So this useEffect will never react to the optInStartedAt changing, and will only run once on initial mount. But running on initial mount is not a problem, and is probably the desired behaviour. So instead of running it in a useEffect, lets just run it when we create the store.

eg

```ts
function  createAppUsageStatsStore(params){
 ...create store
 ...create actions
 
 // Check if opt-in has expired (older than 1 year)
  const optInStartedAt = store.getState().optInStartedAt;
  if (optInStartedAt && Date.now() - optInStartedAt > ONE_YEAR_MS) {
    actions.resetOptInAndTimer();
  }

  return {instance: store, actions};
}

```

That way it is not tied to the react lifecycle at all

eslint.config.mjs
```diff
@@ -85,21 +85,21 @@ const frontendConfig = pluginTs.config(
...
       'react-hooks/rules-of-hooks': 'error',
-      // Downgrade new stricter rules to warnings for now (from react-hooks v7)
-      'react-hooks/set-state-in-effect': 'warn',
-      'react-hooks/purity': 'warn',
+      'react-hooks/set-state-in-effect': 'error',
+      'react-hooks/purity': 'error',
+      'react-hooks/refs': 'error',
+      'react-hooks/globals': 'error',
+      'react-hooks/immutability': 'error',
       'react-hooks/incompatible-library': 'warn',
```
ErikSin: can we just erase these completely? By default they are set to error

ErikSin (changes requested):
While were at it, lets also fix the `forwardRef` eslint warning and the unnecessary `<Context.Provider/>` warning as well!

src/frontend/contexts/ActiveProjectIdStoreContext.tsx
cimigree: I was trying to avoid synchronous setting of state. But I think I thought of a better way?

src/frontend/contexts/ActiveProjectIdStoreContext.tsx
cimigree: I thought (think?) that the finally code, which is settingInitialized(true) is running even if those "return". Right?
So I think we are doing setIsInitialized(true) two times more than needed. But do you think we don't hit the finally?

src/frontend/contexts/ActiveProjectIdStoreContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1674/commits/92867d2c9a62cf0c714883700431571f1f617d14

src/frontend/contexts/AppUsageStatsContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1674/commits/92867d2c9a62cf0c714883700431571f1f617d14

eslint.config.mjs
cimigree: https://github.com/digidem/comapeo-mobile/pull/1674/commits/92867d2c9a62cf0c714883700431571f1f617d14

ErikSin:
Nice! now we only have ONE eslint warning!!

eslint.config.mjs
```diff
@@ -85,21 +85,16 @@ const frontendConfig = pluginTs.config(
...
       'react-hooks/rules-of-hooks': 'error',
-      // Downgrade new stricter rules to warnings for now (from react-hooks v7)
-      'react-hooks/set-state-in-effect': 'warn',
-      'react-hooks/purity': 'warn',
       'react-hooks/incompatible-library': 'warn',
-      'react-hooks/immutability': 'warn',
-      'react-hooks/refs': 'warn',
-      'react-hooks/globals': 'warn',
+      // used in test mocks
       '@eslint-react/hooks-extra/no-useless-custom-hooks': 'warn',
```
ErikSin: this can be taken out now because you have turned it off in the tests.

eslint.config.mjs
```diff
@@ -85,21 +85,16 @@ const frontendConfig = pluginTs.config(
...
       // There are some cases in app code when it's needed
       '@typescript-eslint/no-require-imports': 'off',
       // We want to strictly adhere
       'react-hooks/exhaustive-deps': 'error',
       // We want to strictly adhere
       'react-hooks/rules-of-hooks': 'error',
-      // Downgrade new stricter rules to warnings for now (from react-hooks v7)
-      'react-hooks/set-state-in-effect': 'warn',
-      'react-hooks/purity': 'warn',
       'react-hooks/incompatible-library': 'warn',
```
ErikSin: this can also be taken out. I believe by default it is a warning. And if it is an error, its only happening in one place so you can turn it off in that one file

ErikSin:
can you update this from develop? now that the new `usePasscodeLockout` is merged

src/frontend/screens/Settings/CreateTestData.tsx
ErikSin, line 1: the main architectural thing was just to remove the forward ref? The other changes were just related to getting rid of the old button, correct?

src/frontend/contexts/ActiveProjectIdStoreContext.tsx
ErikSin: i was just asking your reasoning behind it, not criticizing it.

ErikSin (approved):

src/frontend/screens/Settings/CreateTestData.tsx
cimigree: Yes button and old text types as well.

PR #1680, opened 2026-02-04
fix: photo orientation

### LINKED ISSUE #1675: Observation Photos in Portrait Mode landscaped
**Describe the bug**
When I make an observation and take a photo in portrait mode, it is saving it in Landscape mode instead. FYI, photos taken in Landscape mode get orientated to look right.

**To Reproduce**
1. Make an observation
2. Take a photo with it in portrait mode
3. Photo will appear in observation in draft form, in saved form, in editing form, etc, all in landscape orientation.

**Expected behavior**
A photo taken with an observation should orient correctly.

**Screenshots**

Image

Image

Image

Image

**Additional context**
This is on the current develop branch.

cimigree:
closes #1675 
Skips double rotation of images through expo camera.
When taking a photo with the Expo Camera (skipProcessing: false), the camera automatically rotates the photo to the correct orientation.
However, the EXIF metadata still contains the original orientation value (before Expo's automatic rotation).
The code was then applying another rotation based on this EXIF orientation value, causing double rotation.
So, I removed the EXIF-based rotation logic (getRotationFromExifOrientation())
Set rotation to always be 0 degrees since Expo Camera already handles orientation.

### DIFF
--- a/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
+++ b/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
@@ -160,7 +160,7 @@ export function createDraftObservationStore({persist}: {persist: boolean}) {
    * Only processes steps that are not already complete.
    */
   async function processPhoto(attachment: UnsavedPhotoAttachment) {
-    const {id, raw, original, thumbnail, preview, photoExif} = attachment;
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
I'm confused how neither of us caught this initially, I know we both tested photos pretty pretty thoroughly

PR #1682, opened 2026-02-04
fix: security bypass for map upload

### LINKED ISSUE #1676: Map upload should not show security screen
Notion: https://www.notion.so/digidem/Security-screen-appears-when-uploading-map-2871b08162d580fd9f20f2eb792b9eb7

When choosing a new map file, the user is navigated away from comapeo. When security is enabled, the user is then prompted to the security screen when they return to the app. 

## To Do
- [ ] Stop Map Upload from opening auth screnn

cimigree:
closes #1676 
Bypasses security for file upload for either map upload or uploading of a config file.
I did it in a similar way that we check for if the [share dialogue is open](src/frontend/hooks/share.ts) that Andrew did a bit ago.
I added a function to check just to try to keep things clean in the body of the Auth Context, but I can take that out if you think it is unnecessary. I have no strong opinion about it.


achou11 (commented):
Still seem to be able to encounter the issue when doing the following:

1. Go to background maps settings screen.
2. Press choose file button
3. _Cancel_ the file selection i.e. using hardware back button or back gesture.

Seems it could be related to a race condition between expo-document-picker and react native's app state, though I'm not entirely sure. Adding a long enough timeout before resolving in the case of the cancellation [here](https://github.com/digidem/comapeo-mobile/blob/b98c555bbf7594a484ce2bd38101b8efcc2e31c7/src/frontend/lib/file-system.ts#L36) seems to address the issue, although I wouldn't suggest as a fix and more just proof of it being a timing issue 😅 

---

Was taking a closer look at the auth context code, especially this block:

https://github.com/digidem/comapeo-mobile/blob/b98c555bbf7594a484ce2bd38101b8efcc2e31c7/src/frontend/contexts/AuthContext.tsx#L64-L70

Any thoughts on why the `'active'` state is included in this case? Seems unnecessary and counterintuitive to "unauthenticate" when re-entering the app, as that ideally only happens when we background/leave the app. Could easily be missing a case that this accounts for though...

I ask because I _think_ a potentially helpful change would be something along the lines of:

```ts
if (passcode !== null) {
  if (nextAppState === 'background' || nextAppState === 'inactive') {
    const mutations = queryClient.getMutationCache().findAll({
      mutationKey: ['background'],
      status: 'pending',
    });

    if (mutations.length > 0) return;

    setAuthState('unauthenticated');
  }
}
```

Couple of notable changes here:

1. No longer handling the `'active'` app state case.

    At a glance, I don't see the need for it and it seems like something that would contribute to incorrect behavior.

2. _Non-reactively_ calculating the bypass when we actually need to. 

    There are various things about this code that could be prone to timing issues - mostly saying that as the reality of dealing with React, tanstack query, and async JS code in general 😄 Relying less on React's reactivity and calculating the auth bypass when we actually need it in the listener seems to solve the issue I mentioned, as well as handle the other cases (please do verify yourself if you try it though!). The example code requires updating the mutation keys of the relevant mutations to have a shared prefix (e.g. `['background', ...]` - not attached to this naming at all). Not a strict requirement but it's helpful to have so that future mutations could follow a similar pattern.

    Not entirely sure, but this could be a case where using [`useEffectEvent`](https://react.dev/reference/react/useEffectEvent) makes sense, but that's not available in the version of React we're using. Also not sure if that would actually solve the issue, but it could in theory!
    
---
    
EDIT: pressed submit too early!

Most of what I've written above is just me trying to reason about the issue more holistically and exploring ideas for fixes. I don't find the result particularly elegant or ideal by any means but I think any solution is going to be a bit icky just due to the (suspected) nature of it. Mostly deferring to you (and Erik) on what you think is a reasonable approach for addressing the general issue.

cimigree:
@achou11 I think I set it up how you envisioned, but let me know if your thinking was on different lines. I like it! I am not sure why the "active" was in there. It was in there since the feature was first added two years ago, according to the git history/ blame and Erik wrote the code at that time. I tested it out a bunch and all seemed to work as I expected.

src/frontend/contexts/AuthContext.tsx
```diff
@@ -58,22 +56,23 @@ export const AuthProvider = ({children}: {children: React.ReactNode}) => {
...
-            nextAppState === 'background' ||
-            nextAppState === 'inactive'
-          ) {
+          if (nextAppState === 'background' || nextAppState === 'inactive') {
+            const backgroundMutations = queryClient.getMutationCache().findAll({
+              mutationKey: ['background'],
+              status: 'pending',
+            });
+
+            if (backgroundMutations.length > 0) return;
```
achou11, line 68: maybe add a comment about what this code is doing (e.g. preventing unauthorization due to specific in-app actions)

achou11 (approved):
> I think I set it up how you envisioned, but let me know if your thinking was on different lines.

What you currently have is basically what I did locally 👍

Definitely not required, but could be worth documenting the "convention" of the `background` mutation key prefix and why it's a thing; not sure where though...Just one of those things that could easily be forgotten, but I also don't see a need to create some kind of abstraction/helper for it at the moment.

> I am not sure why the "active" was in there. It was in there since the feature was first added two years ago, according to the git history/ blame and Erik wrote the code at that time.

Yeah a lot of this code was ported from _Mapeo_ Mobile so there could be a lot of digging to get the context. As long as you also think that my reasoning for removing it makes some amount of sense, that's good enough for me. Just really can't think of a situation where it's needed...(famous last words)

PR #1684, opened 2026-02-05
chore: refactor use passcode lockout

ErikSin:
Simplifies the usePasscodeLockout hook. The refactor in #1665 exposed how complicated the code was. This simplifies the hooks (and respects the new eslint rules).

One thing to note. Previously when the timer ran out we would set `lockUntil` to 0. I took out that logic because lockUntil is a date. So we are checking the date relative to `date.Now()`. So a stale date doesn't matter since it is relative.

### DIFF
--- a/src/frontend/hooks/usePasscodeLockout.ts
+++ b/src/frontend/hooks/usePasscodeLockout.ts
@@ -1,42 +1,40 @@
-import {useEffect, useState} from 'react';
-import {
-  useSecurityActions,
-  useSecurityState,
-} from '../contexts/SecurityStoreContext';
-import {getRemainingLockoutMinutes} from '../lib/security';
-import {defineMessages, useIntl} from 'react-intl';
+import {useEffect, useRef, useState} from 'react';
 
-const m = defineMessages({
-  lockoutMessage: {
-    id: 'hooks.usePasscodeLockout.lockoutMessage',
-    defaultMessage:
-      'Try again in {minutes, plural, one {# minute} other {# minutes}}',
-  },
-});
+function calcMinutes(lockUntil: number): number {
+  const remaining = Math.ceil((lockUntil - Date.now()) / 60_000);
+  return remaining > 0 ? remaining : 0;
+}
 
-export function usePasscodeLockout() {
-  const {lockUntil} = useSecurityState();
-  const {setLockUntil} = useSecurityActions();
-  const [minutes, setMinutes] = useState(0);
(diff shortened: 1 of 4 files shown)


src/frontend/hooks/usePasscodeLockout.ts
```diff
@@ -1,42 +1,40 @@
...
+  lockUntil: number;
+  clearError: () => void;
+}): number {
+  const [minutes, setMinutes] = useState(() => calcMinutes(lockUntil));
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
 
   useEffect(() => {
-    let timeout: NodeJS.Timeout | null = null;
-    if (lockUntil) {
-      const calcMinutes = getRemainingLockoutMinutes(lockUntil);
```
cimigree, line 26: Can you please remove this function from `src/frontend/lib/security.ts` and the test for it in `src/frontend/lib/security.test.ts` since you are no longer using it?

src/frontend/hooks/usePasscodeLockout.ts
```diff
@@ -1,42 +1,40 @@
...
+        intervalRef.current = null;
+        clearError();
+      }
+    };
+
+    update();
+
+    if (!calcMinutes(lockUntil)) return;
 
+    intervalRef.current = setInterval(update, 15_000);
```
cimigree, line 31: It's a little bit of a bummer that a one minute lockout can take up to 75 seconds, but probably not a big deal. And I only noticed that happening when I backgrounded the app, so it should not be an issue for the E2E test, but in general this code is cleaner so it is worth it.

cimigree (approved):
Not blocking, but would make me feel better. Can you take out the function that you are no longer using? And the test for it?

src/frontend/hooks/usePasscodeLockout.ts
ErikSin: But i think this is an improvement from before, no? Previously the timer would only run if the app was in the foreground. So the 1 minute timer would always pause when the app went into the background. I dont fully know the nuances of backgrounding timers and interval though, so if the timer doesnt cause this, i would be willing to switch to a timer instead of an interval

cimigree: I think the interval is better for backgrounding because we are now comparing every 15 seconds to the absolute timestamp. The 75 seconds won't happen every time and it's only 15 seconds.

PR #1685, opened 2026-02-10
fix: backend node

cimigree:
[Updated] Removes the separate nvmrc for the backend, as it no longer needs to be kept different and distinct from the root nvmrc.


ErikSin:
Lets write a test to make sure that the nvmrc for both the backend and the frontend are the same! Something similar to [this test](https://github.com/digidem/comapeo-mobile/blob/543a2ab0e5add581232832995e1bb7ab5d345f0a/src/frontend/packageJson.test.ts#L30)

ErikSin (approved):
Thanks!

PR #1687, opened 2026-02-10
fix: observation edit attachments not persisting

### LINKED ISSUE #1678: Editing Saved Observation with attachements not working
When editing a saved observation with an attachment (photo or audio file), and app is force closed, the app reopens to the edit observation scree, without the attachments, and the user is unable to save (error bottom sheet comes up).

## To do:
- [ ] Asses and fix the error

ErikSin:
closes #1678 

The issue was that we were navigating to the wrong screen when re-opening the app. Instead of navigating to observationEdit, we were accidentally navigating to ObservationCreate. In order to determine which screen to navigate to, we look for the `docId`. If there is a `docId`, it is a saved observation, if there is no `docId`, it is a new, unsaved observation. With the new observation store we are saving the `docId`in a new place. So i updated to check to properly look for the `docId`.

I also updated the navigation from the `ConfirmDiscardObservationEditBottomSheet`. Previously it was 'popTo' to the `observation` screen. But when closing and re-opening the app, we navigate from the map screen to the observationEdit screen. This means that `popTo(Observation)` adds `Observation` to the stack because it is not there originally. So when the user clicks back from the observation screen, it does back to the observationEdit screen, which is not the desired behaviour.

### DIFF
--- a/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
+++ b/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
@@ -38,16 +38,13 @@ const m = defineMessages({
 
 export const ConfirmDiscardObservationEditBottomSheet = ({
   navigation,
-  route,
 }: NativeRootNavigationProps<'ConfirmDiscardObservationEditBottomSheet'>) => {
   const {formatMessage: t} = useIntl();
   const {clearDraft} = useDraftObservationActions();
 
   function handleDiscard() {
     clearDraft();
-    navigation.popTo('Observation', {
-      observationId: route.params.observationId,
-    });
+    navigation.pop(2);
   }
 
   return (
--- a/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
+++ b/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
@@ -65,6 +65,7 @@ export const ConfirmDiscardObservationEditBottomSheet = ({
         <View style={styles.buttonsContainer}>
           <DestructiveButton
             fullSize
+            testID="OBS.discard-obs-btn"
             text={t(m.discardObservationButton)}
             renderIcon={() => <DiscardIcon />}
             onPress={handleDiscard}
--- a/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
+++ b/src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
@@ -44,6 +44,7 @@ export const ConfirmDiscardObservationEditBottomSheet = ({
 
   function handleDiscard() {
     clearDraft();
+    // We are trying to just goBack(), but we need to close 2 screens, the bottom sheet and ObservationEdit.
     navigation.pop(2);
   }
 
--- a/src/frontend/screens/MapScreen/index.tsx
+++ b/src/frontend/screens/MapScreen/index.tsx
@@ -235,6 +235,7 @@ function useCheckDraftObservationAndNavigate({
 }) {
   const {navigate} = useNavigationFromHomeTabs();
   const existingObservation = useDraftObservationState(store => store.value);
+  const id = useDraftObservationState(store => store.id);
 
   useFocusEffect(
     React.useCallback(() => {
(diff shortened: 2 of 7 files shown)


src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
```diff
@@ -38,16 +38,13 @@ const m = defineMessages({
...
 }: NativeRootNavigationProps<'ConfirmDiscardObservationEditBottomSheet'>) => {
   const {formatMessage: t} = useIntl();
   const {clearDraft} = useDraftObservationActions();
 
   function handleDiscard() {
     clearDraft();
-    navigation.popTo('Observation', {
-      observationId: route.params.observationId,
-    });
+    navigation.pop(2);
```
cimigree, line 48: This is a pretty obtuse change. It took me a while to figure out why this was necessary and how it is working. Could you please add a brief comment in here to explain its necessity for future us/ developers?

cimigree (approved):
Lots of good catches/ debugging in here.

I think it would be worthwhile to add a couple/ few E2E tests for these situations...
Something like, 'it should navigate to ObservationEdit (not ObservationCreate) when reopening app during edit'... (with changes, changes are still there, can save... that kind of thing) Obviously we can't check for attachments but I think it would still be worthwhile. Maybe in here: tests/e2e/specs/observations/edit-observation.test.ts

And then maybe another E2E that would check the discard an observation being edited flow. Like, edit, make a change, discard, and check that we go to observation detail screen and then observation list and check the change isn't kept. Then also edit, make a change, force close, discard, and return to the map screen in that case.

What do you think?

src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
ErikSin: This actually was an [error introduced here](https://github.com/digidem/comapeo-mobile/pull/1648/changes#diff-782eb831c896d157898feca9ab13e889f69436eb28a35c9c756a994dc197d4cfL68). The change introduced in this previous pr added the observation screen to the stack instead of going back, which caused the error described in the description.

What kind of comment were you looking for? In otherwords what do you think would be good for future developers to know when they see this code?

cimigree: Maybe something about why it is necessary with the App restart scenario...
// Pop 2 screens to close both the bottom sheet and ObservationEdit.
// We use pop(2) instead of popTo('Observation') because the Observation
// screen may not exist in the stack (e.g., when force closing and reopening the app while editing an 
// Observation. If we used popTo() without an Observation in the stack, it would
// add Observation to the stack, causing a navigation loop.

That's kind of a long explanation but that's the idea... Explaining why pop(2) is necessary and we can't just go right to the Observation itself.

ErikSin:
> I think it would be worthwhile to add a couple/ few E2E tests for these situations... Something like, 'it should navigate to ObservationEdit (not ObservationCreate) when reopening app during edit'... (with changes, changes are still there, can save... that kind of thing) Obviously we can't check for attachments but I think it would still be worthwhile. Maybe in here: tests/e2e/specs/observations/edit-observation.test.ts
> 
> And then maybe another E2E that would check the discard an observation being edited flow. Like, edit, make a change, discard, and check that we go to observation detail screen and then observation list and check the change isn't kept. Then also edit, make a change, force close, discard, and return to the map screen in that case.

I created test for everything except for "edit, make a change, discard, and check that we go to observation detail screen and then observation list and check the change isn't kept. " That one is slightly out of scope for this PR, but i am creating a ticket for it now!

src/frontend/screens/ObservationEdit/ConfirmDiscardObservationEditBottomSheet.tsx
ErikSin: I think what I am a little confused about is why we actually made the change to `popTo('Observation')` in the first place.

Its seems pretty intuitive to me that we just go back when closing the screen. I think it is a little less intuitive that we have to go back 2 screens, but in the context of the bottom sheets being its own screen, its not hard to parse. 

What I am trying to get at is: is it confusing in THIS PR because we **see** the change from `popTo('observation')` to `pop(2)`? Or is the `pop(2)` by itself confusing.

The reason why i think that matters is that `popTo('observation')` was actually an error we didn't catch. So when someone sees this code in the future I don't think they would question why we didnt do `popTo('observation')`.

cimigree: No, not the change necessarily. I think it is good when there is a pop with a number to have a comment just because it is a little unclear which screen it is going to go to and why without a popTo or navigate with a screen listed or even a goBack.

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1687/commits/3867e8a3c76ed0b0303fbd6d22b3d5a4244e22df

ErikSin:
> I created test for everything except for "edit, make a change, discard, and check that we go to observation detail screen and then observation list and check the change isn't kept. " That one is slightly out of scope for this PR, but i am creating a ticket for it now!

Ticket: https://github.com/digidem/comapeo-mobile/issues/1695

PR #1688, opened 2026-02-11
chore: add react compiler

### LINKED ISSUE #1645: Install React Compiler
Since React 19, React introduced the [React Compiler](https://react.dev/learn/react-compiler). Attempt to Install the new compiler](https://docs.expo.dev/guides/react-compiler/). Timebox this to 2 days. And we can re-assess this work if it takes longer than that

## To Do
- [ ] Install New compiler

ErikSin:
Closes #1645.

Mostly minor changes.
1. I had to update the `TabBar` to use the bracket syntax to look like a jsx component in the Tab Navigator, otherwise the compiler was throwing an error because it thought it was NOT a jsx component trying to use a hook
2. installed the Eslint plugin for the React Compiler
3. there were some minor eslint changes. Most of the changes were that a few of the `MessageDescriptors` for translations has "use" at the beginning of the name which eslint was mistaking as a hook, so I just changed those names.


package.json
```diff
@@ -168,10 +167,12 @@
     "babel-jest": "30.2.0",
     "babel-plugin-transform-import-meta": "2.3.3",
     "babel-plugin-transform-inline-environment-variables": "0.4.4",
+    "babel-preset-expo": "54.0.9",
     "date-fns": "4.1.0",
     "dotenv-cli": "11.0.0",
     "eslint": "9.39.2",
     "eslint-plugin-jest": "29.5.0",
+    "eslint-plugin-react-compiler": "19.1.0-rc.2",
```
ErikSin, line 175: the `eslint-plugin-react-compiler` needs to be the same version as react. There is no stable 19.1.0 version, so we have to use the rc here

cimigree (approved):
Nice work!
I am impressed that you were able to figure this out. I could not!
All works smoothly

A non-blocking question:
I have the React Compiler Marker VSCode extension installed, and it shows that some of the components/ screens are not optimized. Such as `src/frontend/screens/AuthScreen.tsx` for example.
Should we attempt to optimize these? Make a plan? Or just ignore... Are there next steps? What do you think?

ErikSin:
> A non-blocking question: I have the React Compiler Marker VSCode extension installed, and it shows that some of the components/ screens are not optimized. Such as `src/frontend/screens/AuthScreen.tsx` for example. Should we attempt to optimize these? Make a plan? Or just ignore... Are there next steps? What do you think?

the `React Compiler Marker VSCode extension` requires me to open each file separately to see if the code is optimized, which is not ideal.

I researched other tools, and everything just lead me back to using eslint plugin...which we are using. 

That's long winded way of saying that I don't think we need to address those in this PR. And lets incrementally fix things when it comes up on the vscode extension.

Issue #143, opened 2023-12-20
User location and following buttons on map disappear when location services are turned off


achou11:
_Transferred from https://github.com/digidem/CoMapeo-mobile/pull/136#issuecomment-1865105011_

---

Just playing around with the location services behavior and noticing different behavior here than in Mapeo Mobile

Basic steps are:

1. Open app
2. Turn off location services
3. Return to app and wait for a few seconds
4. Turn on location services
5. Return to app and wait for a few seconds

---

- Mapeo:

https://github.com/digidem/CoMapeo-mobile/assets/18542095/ffaa2e95-91e8-44b3-aa09-40340c683db3

- CoMapeo:

https://github.com/digidem/CoMapeo-mobile/assets/18542095/f3146222-a18d-4cd9-b73c-f6c02bf82bcd

You'll notice the main differences are the related to the user location marker and following button after step 3. In Mapeo, they remain on the screen. In CoMapeo, they disappear, but return once the location services are on.

What's happening in CoMapeo seems less desirable, but I don't actually know what the desired behavior is.

achou11:
cc @gmaclennan maybe you have some additional context here?

achou11:
From Erik, about why what Mapeo is doing doesn't seem like desired behavior (https://github.com/digidem/CoMapeo-mobile/pull/136#issuecomment-1865143719):

> My reasoning for why I don't think it is desired behaviour is for 2 reasons:
>
>  1. it could mislead the user into thinking that there location service is on/confuse them as to where they currently are.
>  2. Security reasons. If they turn off there location services at a sensitive site, it will mark there last location at that sensitive site. So if someone (eg security guard) was to check their phone, there would be a record that they were at that sensitive site.

I replied with the following:



>>    it could mislead the user into thinking that there location service is on/confuse them as to where they currently are.
>
> In Mapeo, there's the gps pill in the map screen UI, which I think is supposed to address this concern. Currently not implemented in CoMapeo
>
>>    Security reasons. If they turn off there location services at a sensitive site, it will mark there last location at that sensitive site. So if someone (eg security guard) was to check their phone, there would be a record that they were at that sensitive site.
>
> yeah this is a good point! i was coming from the angle that location services being available is not always due to an explicit action. For example if you have a hardware issue i think that could cause this behavior. In that case, not sure I'd want things that were showing up a few seconds ago to disappear

achou11:
[...]

PR #1693, opened 2026-02-12
feat: map sharing

### LINKED ISSUE #1570: Create new Waiting for Map Accept

### LINKED ISSUE #1755: Map Black after receiving map file
As seen in the video, After receiving a bgmap, the map does not update. Furthermore after force closing the app, the map becomes black. 

## Video
https://github.com/user-attachments/assets/4765dd1f-3fc1-4814-9617-f6382dbe26f4

### APK Version
Both Devices (Also see beginning of video): 
Image

### Assets/Map used
`stamed-terrain-z8.smp` in this [folder](https://drive.google.com/drive/folders/19iZUWrlTQIBpqr7MnP0u5E5dmFudJYKt).

### LINKED ISSUE #1756: `ReplaceBackgroundMap` actions not working
As seen in video the user cannot decline or accept a map, as clicking them navigates the user to the error bottom sheet. This only happens when the sender cancels sending the map. 

## Steps to reproduce
1. Device A uploads a background map
2. Device B uploads a background map
3. Device A sends map to Device B
4. Device A cancels sending map
5. Device B presses "Accept" 
6. Device B is navigated to [Replace Background Map Screen](https://github.com/digidem/comapeo-mobile/pull/1704/changes#diff-4d3fc18a279c3165b3fb2c5719d9d00a0522fcaa0b3d6db9c10dce5632dda45a). When user presses either button they are navigated to the error bottom sheet.

## Video
https://github.com/user-attachments/assets/6459cf5b-1049-4db1-bd21-761cde120a8a

### Version
Also See Beginning of Video
Both devices:

Image

### LINKED ISSUE #1757: Map Reciever Cannot properly cancel a map being shared
As seen in video, if a receiver of map tries to cancel, after the sender has cancelled, the reciever is navigated to the error bottom sheet. They are also stuck on the page, and the user has to force close and re open the app.

## Video:
https://github.com/user-attachments/assets/e5798e4f-64ac-422a-8730-ed010f5591c0

## APK Version
See beginning of video as well
For both devices: 

Image

### LINKED ISSUE #1784: Upgrade Map work to latest core deps
When #1783 is merged into `develop`, update the map sharing work from `develop` and then upgrade `core` and `core-react` to the latest versions

## To do
- [ ] Update core to @latest
- [ ] Update core-react to @latest

cimigree:
closes #1570 
closes #1757 
closes #1755 
closes #1756 
closes #1784 

Adds the Map Sharing Feature.


src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -106,8 +112,23 @@ export const SelectDevice = ({
...
-              // TODO: Navigate to map sharing in subsequent PR
-              console.log('Share map with device:', deviceId);
+              sendMapShare(
+                {projectId, receiverDeviceId: deviceId, mapId: 'custom'},
+                {
+                  onError: (err: Error) => {
+                    Sentry.captureException(err);
+                    navigation.navigate('ErrorBottomSheet');
+                  },
+                  onSuccess: async (
```
cimigree: This is a promise, so I had to adjust the code accordingly. Not sure if this is the root of the problem.

src/frontend/hooks/server/maps.ts
```diff
@@ -132,15 +131,22 @@ export function useGetCustomMapInfo() {
...
+        );
       }
 
       return v.parse(CustomMapInfoSchema, await response.json());
     },
   });
 }
 
-async function fetchCustomMapInfo(baseUrl: string) {
+async function fetchCustomMapInfo(styleUrl: string) {
```
cimigree: I was not able to get the style url so had to adjust this. Not sure at all of this is the right thing to do.

src/frontend/hooks/server/maps.ts
```diff
@@ -23,8 +23,7 @@ const DEFAULT_CUSTOM_MAP_FILE_PATH = CUSTOM_MAPS_DIRECTORY + '/default.smp';
 
 const CustomMapInfoSchema = v.object({
   created: v.pipe(
-    v.string(),
-    v.isoTimestamp(),
+    v.number(),
```
cimigree: Does this change make sense?

src/frontend/contexts/AppProviders.tsx
```diff
@@ -83,6 +87,15 @@ export const AppProviders = ({
   earlyAccessStore,
   appUsageStatsStore,
 }: AppProvidersProps) => {
+  const mapServerListenPromise = React.useMemo(
```
cimigree: I had to memoize this because I was getting an error:
"Listen method has been called more than once without closing"
   - Happened on force close and restart
Because the appRpc.mapServer.listen() call was happening on every render of the AppProviders component.
It was also complaining about needing an argument (hence the empty object)

src/frontend/packageJson.test.ts
```diff
@@ -33,12 +33,30 @@ describe('frontend package.json', () => {
     )) {
       const backendVersion = backendMapeoDependencies[dependency];
       if (!backendVersion) continue;
+      // For file: dependencies, compare just the filename (paths may differ)
```
cimigree: Do you think it's ok to do this just for tests? That way more pressing, problematic test failures will be visible.

gmaclennan:
I've update the [core-react PR #143](https://github.com/digidem/comapeo-core-react/pull/143#top) with what I hope is a fix for this issue. Here is an updated tarball with the latest from commit https://github.com/digidem/comapeo-core-react/pull/143/commits/22f6d9507020de9c824558833242d575d8a3a62e

[comapeo-core-react-7.3.0-pre.3-MAP-SHARING.tgz](https://github.com/user-attachments/files/25298767/comapeo-core-react-7.3.0-pre.3-MAP-SHARING.tgz)

src/frontend/contexts/AppProviders.tsx
```diff
@@ -1,10 +1,20 @@
...
+
+// Use expo/fetch in production, global fetch in tests (Jest environment)
+// expo/fetch doesn't work in Jest because it uses native modules
+let fetchImpl: typeof fetch;
+if (process.env.NODE_ENV === 'test') {
+  fetchImpl = globalThis.fetch;
+} else {
+  const expoFetch = require('expo/fetch');
+  fetchImpl = expoFetch.fetch;
+}
```
gmaclennan: Would be better to mock expo/fetch with globalThis.fetch in the jest mocks, rather than have this in the code base. This works though, it's just a nitpick

src/frontend/contexts/AppProviders.tsx
gmaclennan: Good point, I'll have a think about improving the ergonomics of the api for this - it's too easy right now to use it in a way that causes errors.

src/frontend/hooks/server/maps.ts
gmaclennan: Yes sorry I changed the type on this. This should all be in the new hooks now from core react so you should be able to remove this from the app code

src/frontend/hooks/server/maps.ts
gmaclennan: Hopefully the new hooks solve this.

src/frontend/packageJson.test.ts
gmaclennan: You could also just comment out the expect during dev until it's ready - we shouldn't be using file dependencies by the time the pr is finished

gmaclennan: Or don't run these tests on draft prs

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -106,8 +108,22 @@ export const SelectDevice = ({
 
           const handlePress = () => {
             if (selectionMode === 'shareMap') {
-              // TODO: Navigate to map sharing in subsequent PR
-              console.log('Share map with device:', deviceId);
+              sendMapShare(
+                {projectId, receiverDeviceId: deviceId, mapId: 'custom'},
+                {
+                  onSuccess: result => {
+                    Promise.resolve(result).then(mapShare => {
```
gmaclennan: Something up with my code here - I'll take a look

cimigree:
@gmaclennan 
So sorry but getting more fetch/ http type errors.
For the useGetCustomMapInfo: [Error: fetch failed: [start] Cannot convert 'http://127.0.0.1:40461/maps/custom/info' to a Kotlin type.]

And for the sendMapShare: [Error: fetch failed: [start] Cannot convert 'http://127.0.0.1:45635/mapShares' to a Kotlin type.]

The bird thinks:
The fetch implementation in @comapeo/core-react is not compatible with React Native on Android.

But again, I can't figure anything else out so don't know if that is the actual case

cimigree:
@gmaclennan 
Mostly Good News!
I am using maps from here: https://drive.google.com/drive/folders/19iZUWrlTQIBpqr7MnP0u5E5dmFudJYKt (the QA testing assets)
The first map I uploaded and tried to share was this one: 
maplibre-demotiles.smp
It uploaded fine but when I tried to share it I got this error:
 [Error: Bounds at 3 must be within max of spherical mercator projection -180,-85.051129,180,85.051129]
 
 But then I tried a couple other maps and they worked with no error to get to the next step. I haven't yet gotten further than that, yet. Sorry.
 This works now for the two maps below but not for the first one (see above).
```
 const mapShare = await sendMapShare({
                  projectId,
                  receiverDeviceId: deviceId,
                  mapId: 'custom',
                });
                navigation.navigate('WaitingForMapAccept', {
                  shareId: mapShare.shareId,
                });
```
 demotiles-z2 (1).smp
 
 And this one:
 osm-bright-z6 (1).smp
 
 I am going to continue on to next steps after I clean this up a bit (UI in that background map screen isn't quite right), but I thought you wanted to know about where I was able to get so far.

gmaclennan:
That's great to hear Cindy. That error you are seeing is because the schema validation is a bit over-zealous in Core - that particular map has a floating point error in one of the bounds, which the core validation considers invalid (which is correct, in a way). Fix in core here: https://github.com/digidem/comapeo-core/pull/1231

gmaclennan:
@cimigree a few small changes here: 
- The issue with the types for `File.pickFileAsync()` are actually a bug in the expo types, not the types in core-react. I've added a cast, and I've submitted a bug report.
- I made a small refactor of the app rpc api, to avoid the need for the memoization of server.listen(), which we shouldn't be calling in the render of a component anyway.
- I fixed the types for mapserver.listen() to remove the need to pass an empty object
- I update map-server, which mainly fixes cleanup of temp files during map import.

cimigree:
@gmaclennan I can no longer build the app and am getting this error:
To clear this warning, run:
`watchman watch-del '/Users/cindygreen/awana/comapeo-mobile' ; watchman watch-project '/Users/cindygreen/awana/comapeo-mobile'`

Android Bundling failed 1387ms index.js (3208 modules)
Unable to resolve "@comapeo/map-server/constants.js" from "node_modules/@comapeo/core-react/dist/esm/hooks/maps.js"
> 1 | import { CUSTOM_MAP_ID } from '@comapeo/map-server/constants.js';
    |                                ^
  2 | import { errors } from '@comapeo/map-server/errors.js';
  3 | import { useMutation, useQuery, useQueryClient, useSuspenseQuery, } from '@tanstack/react-query';
  4 | import { useCallback } from 'react';

Import stack:

 node_modules/@comapeo/core-react/dist/esm/hooks/maps.js
 | import "@comapeo/map-server/constants.js"

 node_modules/@comapeo/core-react/dist/esm/index.js
 | import "./hooks/maps.js"

 src/frontend/contexts/ActiveProjectIdStoreContext.tsx
 | import "@comapeo/core-react"

 src/frontend/App.tsx
 | import "./contexts/ActiveProjectIdStoreContext"

 index.js
 | import "./src/frontend/App"


 | import "./index"

Does this need to be fixed on the core-react side or should I add the map server import to the root package.json?

cimigree:
@gmaclennan can you help update now that I have updated to using various comapeo actual packages instead of tarballs?
Getting this error:
 ERROR  [TypeError: 0, _comapeoIpcClientJs.createAppRpcClient is not a function (it is undefined)]. 
I am not sure the proper way now to create the mapServer.

ErikSin (changes requested):
Im just adding a blocking review so we don't merge into develop.

This PR is going to be the base PR that the subsequent map sending PRs are going to be merged into. once all those are merged into this PR, we can do a final review and merge this PR into develop

patches/expo+54.0.33.patch
ErikSin, line 1: Please add a description to the patches/readme describing what this patch does

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -239,25 +241,33 @@ function NoMapScreen({
...
-          </View>
-        </Button>
+          </Button>
+        )}
         <BodyText variant="smallMeta" style={styles.fileTypeText}>
           {t(m.acceptedFileTypes)}
         </BodyText>
       </View>
-      {error && (
+      {error && !error.message?.includes('Map not found') && (
```
gmaclennan:
```suggestion
      {error &&
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        (error as any).code !== 'MAP_NOT_FOUND'  && (
```

gmaclennan: I think better to check on error.code now that we have custom errors coming from map server hooks.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -144,68 +146,61 @@ export function BackgroundMapsScreen() {
...
+      }
+      await importCustomMapMutation.mutateAsync({file});
+    },
+  });
   const importCustomMapMutation = useImportCustomMapFile();
   const removeCustomMapMutation = useRemoveCustomMapFile();
-  const customMapInfoQuery = useGetCustomMapInfo();
+  const {data, isRefetching, error} = useGetCustomMapInfo();
+
+  const customMapInfo = data as CustomMapInfo | null | undefined;
```
gmaclennan: The fact this does not have a type is a bug in core-react. It's fixe in v9.0.2 which should be published by the time you read this.

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -147,22 +174,22 @@ type GetSelectableDevicesParams = {
...
 export function getSelectableDevices({
   peers,
   projectMembers,
   selectionMode,
+  syncState,
 }: GetSelectableDevicesParams): PublicPeerInfo[] {
   if (selectionMode === 'shareMap') {
+    if (!syncState) return [];
+
+    const connectedDeviceIds = Object.keys(syncState.remoteDeviceSyncState);
```
gmaclennan: This feels like an obscure way to get devices that are members of the same project. In the other branch below, we use the projectMembers list to determine project membership, which feels much more direct to me, rather than the "if a device is listed in the sync state then it is a member of the same project", which is a much looser guarantee. Also, for the share map, we only want to be able to share with active members, not with devices that have left or been blocked.

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -60,8 +68,10 @@ export const SelectDevice = ({
   const {formatMessage: t} = useIntl();
 
   const availablePeers = useInitiallyConnectedPeers();
-  const projectId = useActiveProject();
-  const projectMembersQuery = useManyMembers(projectId);
+  const {projectId} = useActiveProject();
+  const syncState = useSyncState({projectId});
+  const projectMembersQuery = useManyMembers({projectId});
+  const {mutateAsync: sendMapShare} = useSendMapShare({projectId});
```
gmaclennan: It seems like some of this work was done in a previous PR (e.g. the `selectionMode` from the route), but this feels like the wrong abstraction for this component. Having a whole screen serve different functions means unused hooks, misleading testID names (unless you branch each of those too) and increasing maintenance burden as additional features may be requested for one of the particular states. 

The abstraction which would make sense to me is a `<DeviceListItem deviceInfo={device} />` component, and then separate screens for each case (SelectMapShareDevice, SelectInviteDevice). It hardly seems worth sharing `getSelectableDevice` if it is essentially two different functions depending on the value of `selectionMode`.

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,124 @@
...
+    } else if (mapShare.status === 'declined') {
+      // TODO: Show the map decline UI with the reason for decline
+      navigation.popTo('BackgroundMaps');
+    } else if (mapShare.status === 'canceled') {
+      navigation.popTo('BackgroundMaps');
+    }
+  }, [mapShare, navigation]);
+
+  React.useEffect(() => {
+    const interval = setInterval(() => setTime(prev => prev + 1), 1000);
```
gmaclennan: This will drift and eventually be inaccurant, but maybe that's fine? I'm pretty certain we already have a hook and/or component for accurately showing time since a value.

gmaclennan (commented):
Overall it looks good. I've suggested some minor code improvements, and shared my suggestion/opinion about the architecture and abstraction choices, which I think has an impact on maintainability of the SelectDevice screen.

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: @gmaclennan 
The project members list was what I had, but instead what I need is two devices who are ON the same project -- have the same project open. Just finding out if they both are members is not sufficient, because if I try to share a map between two devices that do not have the same project open, I get this error: (PeerNotFoundError: PeerNotFoundError: Peer not found). 

Is there another way to check whether two devices have the same project open? I took this code from Exchange to check Sync, which checks that exact thing. If a person is left or blocked, they can't actually be IN the project, so that condition should be satisfied. If you can change the backend somehow to not have that error then I can revert back to the other way. I did think that was what was specified in the UI documents in notion -- that the two devices both be in the same project, not that the two devices have the same project open.

Otherwise, I would need an additional backend type check to check if two or more devices have the same project open. That is not something I can check from the frontend otherwise, I don't think.

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/4b806cbb77e4981d8a68feb02bfe21bfd3c120cd

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: Sort of... https://github.com/digidem/comapeo-mobile/pull/1693/commits/4b806cbb77e4981d8a68feb02bfe21bfd3c120cd
I am getting something different back from the API. Wondering if I need to update schema?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/4b806cbb77e4981d8a68feb02bfe21bfd3c120cd

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: As discussed in App Dev meeting on March 5, we will keep this code and the requirements for two devices to be on the same project for now but a change will be made in core and core-react so that two devices who are IN the same project but not ON the same project can share maps in the future.

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/23d474792e1b66fd098ce7f93509b3f85c164f4b

src/frontend/screens/BackgroundMaps/SelectMapShareDevice.tsx
```diff
@@ -0,0 +1,175 @@
...
+            case 'disconnected':
+              isDisconnected = true;
+              break;
+            default:
+              throw new ExhaustivenessError(status);
+          }
+
+          const handlePress = async () => {
+            try {
+              const mapShare = await sendMapShare({
```
ErikSin, line 114: Is the loading/await time neglible enough that we do not need to introduce a loading state here?

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
```diff
@@ -0,0 +1,123 @@
...
+      // TODO: Show the map decline UI with the reason for decline
+      navigation.popTo('BackgroundMaps');
+    } else if (mapShare.status === 'canceled') {
+      navigation.popTo('BackgroundMaps');
+    }
+  }, [mapShare, navigation]);
+
+  return (
+    <View style={styles.container}>
+      <InviteSent />
```
ErikSin: This should be renamed to something more approriate. Maybe `<SendingIcon>`?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -239,50 +239,70 @@ function NoMapScreen({
...
-            <View>
-              <HeaderText variant="header5" style={styles.buttonTextBase}>
-                {t(m.chooseFile)}
-                <HeaderText variant="header5" style={styles.asteriskText}>
-                  {' '}
-                  *
+        {isUploading ? (
+          <Loading size={12} />
+        ) : (
+          <Button
```
ErikSin: Why are we using the deprecated button here?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -239,50 +239,70 @@ function NoMapScreen({
...
-      )}
+        )}
     </View>
   );
 }
 
+// Note: The type definition in @comapeo/core-react v9.0.2 doesn't match the actual API response
+// API returns: {name, size, created} but types say: {mapName, estimatedSizeBytes, mapCreatedAt}
+// Using the actual API response structure here
+type CustomMapInfo = {
```
ErikSin: why is core-react not returning the api response? In otherwords is this an error in core-react or is it intentional?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -144,68 +146,59 @@ export function BackgroundMapsScreen() {
...
-        {customMapInfoQuery.isPending ? (
-          <Loading size={10} />
-        ) : customMapInfoQuery.data ? (
-          <MapInfoScreen
-            customMapInfo={customMapInfoQuery.data}
-            onRemoveMap={handleRemoveMap}
-          />
-        ) : (
+        {isRefetching ? (
+          <Loading size={20} />
```
ErikSin: is `useGetCustomMapInfo()` not a suspense query and that is why this `isRefetching` check is necessary?

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -144,68 +146,59 @@ export function BackgroundMapsScreen() {
...
+    mutationFn: async () => {
+      const result = await File.pickFileAsync(undefined, 'application/*');
+      // The return type of `File.pickFileAsync()` is incorrect. See https://github.com/expo/expo/issues/43201
+      const file = (Array.isArray(result) ? result[0] : result) as File;
+      if (!file) {
+        throw new Error('No file selected');
+      }
+      await importCustomMapMutation.mutateAsync({file});
+    },
+  });
```
ErikSin, line 160: `selectFileMutation` is doing 2 things: 1. selecting a file and 2. importing that file. We should separate the implementation and chain them in `handleChooseFile`

eg:
```ts
const selectFileMutation = useMutation({
    mutationKey: FILE_SELECT_MUTATION_KEY,
    mutationFn: async () => { //ONLY SELECTS FILE}
  });

function handleChooseFile(){
     selectFileMutation.mutate({...prop},{
       {onSuccess:file=>{
          importCustomMapMutation.mutate({file}
        })
}




```

src/frontend/screens/BackgroundMaps/SelectMapShareDevice.tsx
```diff
@@ -0,0 +1,175 @@
...
+  },
+  sameVersion: {
+    id: 'screen.Settings.ProjectSettings.BackgroundMaps.SelectDevice.sameVersion',
+    defaultMessage: 'Make sure both devices are on the same version of CoMapeo',
+  },
+  sameProject: {
+    id: 'screen.Settings.ProjectSettings.BackgroundMaps.SelectDevice.sameProject',
+    defaultMessage: 'Make sure both devices have the same project open',
+  },
+});
```
ErikSin, line 57: just a small thing but can we rename these ids. These are not in Setting/ProjectSettings.
 Eg. rename to: `screen.BackgroundMaps.SelectDevice.sameProject`

ErikSin (commented):
Some small cleanup things, but he architecture looks good

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: I was not clear on how to do the red asterisk with the new button, but then I noticed there is no asterisk in the Figma anymore, so I actually can use the new button and have done so now.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: It is an error in core-react, I think, or perhaps some other backend package feeding through core-react. Might need to check with @gmaclennan about that.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: No, it is not. It just uses useQuery so it won't be picked up by the suspense wrapper.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: I had them chained but Gregor changed to this implementation. I like Gregor's choice better than the previous way and what you have suggested. Here is why...
This is one operation and the way he has it allows us to keep together the loading and error states. 
If it were chained, we would need multiple error handlers and multiple loading states when it doesn't really matter because the actions always need to go together. 
I think it is cleaner with one mutate call rather than complex nesting.
Also, it allows the whole operation to succeed or fail. If it were chained, we would need additional state management for if the file is selected but import fails.
I don't think there is a benefit to chaining them. They are one add map file action. Perhaps a different name?
@gmaclennan Any thoughts? I remember you explained this to me (and I can see the logic now) but I can't find that explanation. You might have spoken it to me on a call, but I didn't take notes.

src/frontend/screens/BackgroundMaps/SelectMapShareDevice.tsx
cimigree: No it is actually a long time, but it is not handled in this PR. It is handled in subsequent PRs and a different screen that is not in this PR. We will show the next screen automatically and then that screen handles the transfer time, which can be very long, actually.

cimigree:
Here is the new Background Maps button -- updated to look like Figma.
image

src/frontend/screens/BackgroundMaps/SelectMapShareDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/efbafcff46c0274205b7f6454539c0983847551d

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/efbafcff46c0274205b7f6454539c0983847551d

src/frontend/screens/BackgroundMaps/WaitingForMapAccept.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/efbafcff46c0274205b7f6454539c0983847551d

patches/expo+54.0.33.patch
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/efbafcff46c0274205b7f6454539c0983847551d

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
gmaclennan: This was a bug in the types of core-react, fixed in https://github.com/digidem/comapeo-core-react/commit/f4e75b4aea8eb7519aa7f2459328147d539c44ce, but we haven't published a new release yet (Github was down when I merged this). @achou11 can publish a new release, and then you can update here and remove this code.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
gmaclennan: yes, pretty much what you said. I don't think "this action does two things, so it should be two actions" is a strong argument for changing things in a way that adds to code complexity - as Cindy says, this requires more state handling and error handling. You could decide to keep `useSelectFile` as an abstraction if you want to have a single implementation shared by various components, but I think it's an abstraction that adds more overhead than it solves. This code as-is seems easy-to-read and to maintain.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
gmaclennan: yes, this never was a suspense query before the changes, so I kept it as non-suspense in the new map server work. I think the reasoning for it not being suspense to begin with is because we handle an error in this hook as "no custom map exists", which is harder to implement around a suspense query. I think using a non-suspense query here makes sense. The refetching and error-handling code is tricky here also because of react-query gotcha's around cache invalidation and query errors - if we did make it suspense we would still need all the same code around refetching because when the custom map is removed it doesn't invalidate the cache and does not throw to the suspense boundary if there is a non-error value already in the cache.

cimigree:
@ErikSin Sorry for the messy and repeated pushes but I somehow messed up the node_modules/ package lock and had to go back and start from scratch.

- Added a keep alive for the map sharing screens so they don't go dark during map sharing, which would cancel the map sharing.

- Also updated/ upgraded all the backend comapeo packages to the latest versions.
   - This adds much faster map loading which closes #1755 
   - Doing so allows us to use the error handling which is a little less fragile than just checking for strings. See [this PR to comapeo/core-react for details.](https://github.com/digidem/comapeo-core-react/pull/163)
      - That closes #1757 
      - And closes #1756 

- We are now also able to share maps between two users who are both IN the same project even if they are ON different projects, so I changed the select device for map sharing accordingly.

- I also updated the map share cancelled sheet so that when it closes it just goes back instead of to the background maps screen as specified by design.
- I did not take on https://github.com/digidem/comapeo-mobile/pull/1704#issuecomment-4135370825 because I wasn't sure about it...        I will wait to hear your thoughts.
- Apk sent to you via Slack.

ErikSin:
Just starting to review, and just noting that my package lock is slightly different when i run `npm i`. Its just different by one line but we should figure out where the change is coming from. Can you run `npm i` again and see what happens.

I am using node-v24.13.0 and npm-v11.6.2

ErikSin:
> We are now also able to share maps between two users who are both IN the same project even if they are ON different projects, so I changed the select device for map sharing accordingly.

This is not working as expected

https://github.com/user-attachments/assets/bf969df5-b7bb-4042-a8fa-a3e30d6df975

It seems like from one device it doesnt matter if your actively on the project. But for the other device  it does matter?

I initially thought it was one of the devices was a particant and not a coordinator, but that doesn't effect the outcome.  

I don't know if it matters that much? but weird behaviour none the less

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,314 @@
...
+            <Loading size={12} />
+          ) : (
+            <SecondaryButton
+              fullSize
+              text={t(m.decline)}
+              onPress={handleDecline}
+            />
+          )}
+        </View>
+      </View>
```
ErikSin, line 257: This UI is a little off. 

1. for buttons we use a `<UIActivtyIndicator/>`, not `<Loading/>`
2. we need block both button if were going to show a loader, not just the decline button

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,314 @@
...
+        onError: (err: unknown) => {
+          const error = toError(err, 'Failed to start map download');
+          Sentry.captureException(error);
+          navigation.replace('ErrorBottomSheet', {error});
+        },
+      },
+    );
+  };
+
+  const handleDecline = () => {
```
ErikSin, line 163: This is taking more than 2 minutes to register on the other phone:

https://github.com/user-attachments/assets/a782a4d3-967c-414e-99bf-d10e998bb15e

src/frontend/screens/BackgroundMaps/SendingBackgroundMap.tsx
```diff
@@ -0,0 +1,330 @@
...
+  const currentTime = useCurrentTime(1000);
+  useKeepAwake();
+
+  usePreventAndroidBackButton();
+
+  const elapsedSeconds = mapShare
+    ? Math.floor((currentTime.getTime() - mapShare.mapShareCreatedAt) / 1000)
+    : 0;
+
+  const cancelShare = React.useCallback(() => {
```
ErikSin, line 96: This is taking several minutes for the other device to register:

https://github.com/user-attachments/assets/30dac581-a8fe-4244-bfc2-930576d16609

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+    if (mapShare.status === 'error') {
+      const error = toError(mapShare.error, 'Map download failed');
+      Sentry.captureException(mapShare.error);
+      navigation.replace('ErrorBottomSheet', {error});
+    }
+  }, [mapShare, navigation]);
+
+  const handleCancel = React.useCallback(() => {
```
ErikSin, line 60: There is some really weird behaviour. The progress bar of the receiving phone seems to be quite far behind. And once the sending phone has completed, I cannot press the cancel button on the receiver phone. In this video i am trying multiple times to press "cancel" and I cannot:

https://github.com/user-attachments/assets/26823b72-cfc0-43d5-94d4-80d17292cd2d

ErikSin (commented):
Its better, now that we have upgraded core and core react, but its still not working fully as expected. See the videos below, but it seems like my samsung Galaxy A03s is having a lot of troubles sending and recieving the cancel in a timely manner. It also is having trouble finding the other device.

Im not sure what the best way to move forward is. This feature doesn't feel ready to release as is.

cimigree:
> Just starting to review, and just noting that my package lock is slightly different when i run `npm i`. Its just different by one line but we should figure out where the change is coming from. Can you run `npm i` again and see what happens.
> 
> I am using node-v24.13.0 and npm-v11.6.2

yes actually there was a change! So sorry I have no idea why!
This is where it is:
"node_modules/bare-events": {
      "version": "2.8.2",
      "resolved": "https://registry.npmjs.org/bare-events/-/bare-events-2.8.2.tgz",
      "integrity": "sha512-riJjyv1/mHLIPX4RwiK+oW9/4c3TEUeORHKefKAKnZ5kyslbN+HXowtbaVEqt4IMUB7OXlfixcs6gsFeo/jhiQ==",
      "license": "Apache-2.0",
      **"peer": true,**
      "peerDependencies": {
        "bare-abort-controller": "*"
      },
      "peerDependenciesMeta": {
        "bare-abort-controller": {
          "optional": true
        }
      }
    },
    
   Adds the bolded line.

ErikSin:
> > Just starting to review, and just noting that my package lock is slightly different when i run `npm i`. Its just different by one line but we should figure out where the change is coming from. Can you run `npm i` again and see what happens.
> > I am using node-v24.13.0 and npm-v11.6.2
> 
> yes actually there was a change! So sorry I have no idea why! This is where it is: "node_modules/bare-events": { "version": "2.8.2", "resolved": "https://registry.npmjs.org/bare-events/-/bare-events-2.8.2.tgz", "integrity": "sha512-riJjyv1/mHLIPX4RwiK+oW9/4c3TEUeORHKefKAKnZ5kyslbN+HXowtbaVEqt4IMUB7OXlfixcs6gsFeo/jhiQ==", "license": "Apache-2.0", **"peer": true,** "peerDependencies": { "bare-abort-controller": "*" }, "peerDependenciesMeta": { "bare-abort-controller": { "optional": true } } },
> 
> Adds the bolded line.

Can you push these changes?

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1693/commits/2f5341baf216fa4eaa8b38243d2759cac29b3498

src/frontend/screens/BackgroundMaps/SendingBackgroundMap.tsx
```diff
@@ -0,0 +1,346 @@
...
+    cancelMapShare(
+      {shareId},
+      {
+        onSuccess: () => {
+          navigation.goBack();
+        },
```
ErikSin, line 102: Now that the state is being handled by the parent, you need to get rid of this `goBack()` i believe

src/frontend/screens/BackgroundMaps/SendingBackgroundMap.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+export function SendingBackgroundMap({
+  route,
+  navigation,
+}: NativeRootNavigationProps<'SendingBackgroundMap'>) {
+  const {formatMessage: t} = useIntl();
+  const {shareId} = route.params;
+
+  const mapShare = useSingleSentMapShare({shareId});
+  const {mutate: cancelMapShare} = useCancelSentMapShare();
```
ErikSin: I think we need catch the `status==='pending'` here, can show a loading state when that is true

ErikSin (commented):
I tested it with an APK, and things do seem to be working better...Im still not convinced that its in a state to release. But i am going to send an APK to QA and let them decide whether the frontend should pause on development (and wait for the backend team to update), or if we can go ahead with merging and releasing as is

src/frontend/screens/BackgroundMaps/SendingBackgroundMap.tsx
cimigree: If you leave take this out we just stay on the Waiting for device to Accept Map screen and don't go back when you hit cancel. So, still need to have that navigation go back here.

ErikSin: Im getting a flash of the waiting For Accept screen though. I thought it was due to this, but maybe its something else. Ill take a video

ErikSin: https://github.com/user-attachments/assets/36a7d4bd-e9a9-49ea-92b9-761ff37bb1a4

cimigree: Oh ok. I think it was the UI falling through. The status wasn't any of the above/ specified, so the main UI was showing, which is the waiting screen. But now that I have added the pending status, that will show the loader instead (at least that is what I am seeing).

gmaclennan:
Just to mention again, the flashing of screens and the need to catch different states that @ErikSin is catching in the review are mostly related to the architectural decision to split this into screens, which I pointed out in https://github.com/digidem/comapeo-mobile/pull/1704#issuecomment-4135370825 and in several conversations before that. It is possible to keep finding these edge-cases and fixing them as you have been, but changing the architecture would simplify this a lot and avoid these bugs appearing in the first place.

cimigree:
> Just to mention again, the flashing of screens and the need to catch different states that @ErikSin is catching in the review are mostly related to the architectural decision to split this into screens, which I pointed out in [#1704 (comment)](https://github.com/digidem/comapeo-mobile/pull/1704#issuecomment-4135370825) and in several conversations before that. It is possible to keep finding these edge-cases and fixing them as you have been, but changing the architecture would simplify this a lot and avoid these bugs appearing in the first place.

I changed it to the way you wanted it which is what was actually causing that flash that Erik mentioned in
https://github.com/digidem/comapeo-mobile/pull/1693/commits/2f5341baf216fa4eaa8b38243d2759cac29b3498

I left out the one screen that is arrived at through user interaction but combined all of the restl.

ErikSin (approved):
WOWWW That was several months that we can never get back...

Thanks for having the patience to finish this!

PR #1694, opened 2026-02-13
chore: pass errors to error bottom sheet

### LINKED ISSUE #1690: Pass Errors to Error Bottom Sheet
Currently the Error Bottom Sheet shows a generic error. Eventually the error bottom sheet will show an error. We need to go through all the errors and pass the error to the bottom sheet (as a nav prop). The nav prop will be unused right now, but this is setting us up for future use of it.

## To do
- [ ] Pass Errors To Bottom Sheet

cimigree:
closes #1690 
**Summary**

- Adds error navigation parameter to ErrorBottomSheet (not used yet but ready for future!)
- Creates a utility to convert errors where the type of error we are getting is unknown to Error instances
- Updates all uses of the ErrorBottomSheet

**Questions:**
- I wasn't really sure what to do/ say for the error messages, since I am not sure at all how the error is going to be handled in the error bottom sheet and how translations will work there, so I didn't make the error messages translatable at this point. I feel like they are just placeholders? But if you know more about this feature and think we will in fact use these messages, then I am happy to make them translatable.
- I made the error required in the Error Bottom Sheet. We had not discussed this, so could you let me know if you have another thought about that? 

**Note**
I touched the files that are not optimized for React Compiler here, but since I am touching so many files and this is such a big PR, I didn't think this was the time to optimize those components. But please let me know if you think it would in fact be a good idea to optimize them in here.

### DIFF
--- a/src/frontend/hooks/useTracking.ts
+++ b/src/frontend/hooks/useTracking.ts
@@ -8,6 +8,7 @@ import {LOCATION_TASK_NAME} from '../sharedTypes/location.ts';
 import {useNavigation} from '@react-navigation/native';
 import * as Sentry from '@sentry/react-native';
 import {useLocationContext} from '../contexts/LocationContext';
+import {toError} from '../utils/errors';
 
 export function useTracking() {
   const {setTracking, addNewLocations, clearCurrentTrack} = useTrackActions();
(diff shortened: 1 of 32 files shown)


ErikSin (approved):
Looks Good!

PR #1696, opened 2026-02-16
chore: update productions deps

### LINKED ISSUE #1689: Upgrade Production Deps
Update Production Deps

ErikSin:
closes #1689. I did a pretty thorough manual test ([following this doc](https://docs.google.com/document/d/1wU4oj5m7sHJvjLpAr-DgPIiaj_VBR8tkvy7IeecL00U/edit?tab=t.0)) and everything seems to be working as expected.

## Deps I did not upgrade:
- expo (and expo font), react, and react-native
- react-native-reanimated: requires the new architecture
- react-native-share: this was breaking the share functionality it previous attempts to upgrade
- react-native-nitro-modules: I upgraded but I did not upgrade to the latest. it is a peer dep of mmkv, so i used the [version in mmkv](https://github.com/mrousavy/react-native-mmkv/releases/tag/v4.1.1)
- react-native-zeroconf: There are patches that written for this library. I don't have enough context to understand the patches and update accordingly, and I would require help from Gregor. Gregor is quite busy right now, so i think it is ok to skip for this round.
- @rnmapbox/maps: it was causing errors that were causing the app to not build properly. this new version uses mapbox sdk v11, and [v11 has alot of breaking features](https://github.com/mapbox/mapbox-maps-android/releases/tag/v11.0.0). We are switching to map libre soon, so not worth the upgrade right now

## Changes in the code
react intl has switched common js to esm in this upgrade. I had to update the jest config and I had to change the script that searched the folders for the polyfill to account for the change to esm


package.json
cimigree: This is no longer used. Probably can be removed?
You took it out in https://github.com/digidem/comapeo-mobile/pull/1641/

package.json
```diff
@@ -87,43 +87,43 @@
     "expo-system-ui": "6.0.9",
     "expo-task-manager": "14.0.9",
     "filter-obj": "6.1.0",
-    "fs-extra": "11.3.2",
+    "fs-extra": "11.3.3",
     "geojson": "0.5.0",
     "geojson-geometries-lookup": "0.5.0",
     "lodash.isequal": "4.5.0",
     "map-obj": "6.0.0",
```
cimigree: This can also come out. Another one you removed in the Observation Store refactor.

package.json
```diff
@@ -87,43 +87,43 @@
...
     "nanoid": "5.1.6",
```
cimigree: And this one too was removed in that PR.

package.json
```diff
@@ -54,17 +54,17 @@
...
-    "@turf/random": "7.3.0",
+    "@sentry/react-native": "8.0.0",
+    "@tanstack/react-query": "5.90.21",
+    "@turf/helpers": "7.3.4",
+    "@turf/random": "7.3.4",
     "buffer": "6.0.3",
     "cheap-ruler": "4.0.0",
-    "eventemitter3": "5.0.1",
+    "eventemitter3": "5.0.4",
     "expo": "54.0.31",
```
cimigree: Can expo be bumped to the latest patch version? 54.0.33? Is there a specific reason you did not update?

package.json
cimigree: Same as question about expo, can we do the patch update? to 14.0.11? Or is there a specific reason not to.

cimigree (approved):
All of my physical testing passed. Non blocking comments, but I hope you will make these small changes.

I think it would be good to remove those three unused dependencies that you stopped using in the store update, see comments in package.json.

I would like to also know about the two little patch bumps and why we aren't doing them. Especially given the bugs Gregor keeps finding, it seems wise to update these patch versions in case there are errors. 

I ran npx devcheck and expo doctor to get that info.

package.json
ErikSin: I think im just a little confused by how expo upgrades work. If i run `npx expo install --fix` it actually downgrades a bunch of packages, which is part of the reason we have been doing expo separately. But let me do some more research on what we should do

cimigree: Yeah I wouldn't do the downgrades. But I thought you could just update the little patch versions and leave the rest alone. That is what I have done in the past.

ErikSin: ok! I just was unsure if that was "allowed" in expo world, but if you have done it before, than its clearly fine!

fixed: https://github.com/digidem/comapeo-mobile/pull/1696/commits/e350244369bd93c8ca38ba399f01542fd3ca571a

package.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1696/commits/e350244369bd93c8ca38ba399f01542fd3ca571a

package.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1696/commits/7d77a6337d6108c4213f7516f401f438194d924e

package.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1696/commits/7d77a6337d6108c4213f7516f401f438194d924e

package.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1696/commits/7d77a6337d6108c4213f7516f401f438194d924e

PR #1697, opened 2026-02-17
feat: decline map

### LINKED ISSUE #1571: Map decline screens

cimigree:
closes https://github.com/digidem/comapeo-mobile/issues/1571
Stacked on top of https://github.com/digidem/comapeo-mobile/pull/1693

Screen shots:
image
image


ErikSin:
Perhaps I misunderstood what we said in our conversation, but I thought we had decided that screens which could stand alone (eg did not require a useEffect at the top level to manage state from the previous screen) would be its own nav screen? Its fine if we want to do it this way, but I just want to clarify if I misunderstood our conversation or if a new architectural decision was made afterwords!

ErikSin (approved):
Looks good!

cimigree:
> Perhaps I misunderstood what we said in our conversation, but I thought we had decided that screens which could stand alone (eg did not require a useEffect at the top level to manage state from the previous screen) would be its own nav screen? Its fine if we want to do it this way, but I just want to clarify if I misunderstood our conversation or if a new architectural decision was made afterwords!

Hmm. @ErikSin, I thought I was doing what we decided. Which screen/ component/ UI do you think might be a candidate to be its own nav screen?
I believe all of these subscreens/ components/ UI that are now in the new version of WaitingForMapAccept (SendingBackgroundMap) are rendering here instead of being stand alone screens that would have to have a useEffect. They are all (and will be in subsequent PRs) alternate versions of what shows up for the mapShare status. The useEffect will be much simplified/ eliminated in subsequent PRs. Is that what you mean? I am confused. Might be better to do a synchronous catch up...

PR #1700, opened 2026-02-17
feat: receive map bottom sheet

### LINKED ISSUE #1572: Map Received Bottom Sheet

cimigree:
closes #1572 
Stacked on https://github.com/digidem/comapeo-mobile/pull/1693 
### Disk Space Check
- Uses `useStorageReadingQuery` to compare available space vs map size
- Space warning takes priority over location warning

### Location Bounding Box Check
- Uses turf.js geospatial library
- Converts bbox to polygon using `bboxPolygon`
- Checks if user's **current location** is inside using `booleanPointInPolygon` (is this the best version and way to access current location to use?)
- Calculates distance to map edge when outside using Turf's `pointToPolygonDistance`.
- Shows distance in km and sets via an object if needed

### PendingMapSharesListener
- Follows the `PendingInvitesListener` pattern
- Respects editing screens and existing map share screens (doesn't interrupt)

## Screenshots

**Regular map received:**
image

**Location warning:**
image

**No space warning:**
image

### What is left to do that I can do:
- Navigate to (or show) `ReplaceBackgroundMapScreen` (when created in next PR) and receiving map...

### Questions
- Thinking MB for the map should be to fixed 2? Or 3? 16 seems like too many! Thoughts?
- Accepting a map works and it updates on the receiver side, but I saw this error:
` ERROR  Mapbox [error] RNMBXMapView | Map load failed: {type=source, message=Source mapboxUserLocation is not in style, begin=1458804469759}`, which I was able to dismiss and all worked fine.


src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,308 @@
...
+  const {shareId} = route.params;
+  const mapShare = useSingleReceivedMapShare({shareId});
+
+  React.useEffect(() => {
+    if (
+      !mapShare ||
+      mapShare.status === 'canceled' ||
+      mapShare.status === 'error'
+    ) {
+      if (navigation.canGoBack()) {
```
ErikSin: in what situation would `navigation.canGoBack()` be false?

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,308 @@
...
+      {
+        onSuccess: () => {
+          // TODO: show different UI if downloading starts successfully but there is a warning (e.g. location not covered)
+          navigation.goBack();
+        },
+        onError: (err: unknown) => {
+          const error = toError(err, 'Failed to start map download');
+          Sentry.captureException(error);
+          navigation.goBack();
+          navigation.navigate('ErrorBottomSheet', {error});
```
ErikSin: I think you can just do `navigation.replace('ErrorBottomSheet', {error})`here.

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,308 @@
...
+    declineMapShare(
+      {shareId, reason},
+      {
+        onSuccess: () => {
+          navigation.goBack();
+        },
+        onError: (err: unknown) => {
+          Sentry.captureException(err);
+          if (navigation.canGoBack()) {
+            navigation.goBack();
```
ErikSin: same question as above, is there a situation that warrants this check?

ErikSin (commented):
a few small comments, but I would like to approve after the previous prs have been merged and this pr has been updated, so i don't miss anything

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: Because of the listener, this can be opened directly on app start up, before any user navigation. Before I put this check in, there were errors/ app crashes in that situation.

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: Same as above...

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1700/commits/364f74e25dd2c9e3aeb5a52f8cf111e4e5ff30f0

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
ErikSin: ~Right, that makes sense. So when the user can't go back, I don't think we should use `navigation.replace('BackgroundMaps')`. Just because the background maps screen has a back button at the top, and if the user was to press that, there would be no where in the navigation to go back to. I think we should use `navigation.replace('Home');` so that the navigation is reset to the default navigation (where home is the first screen)~

^scratch that. See my comment below, but we need to think of a better solution for this problem because it is going to cause problems in the next pr

ErikSin (changes requested):
This all looks good but i am blocking this because we need to think about how to deal with the navigation checks. On reviewing #1704, we are NOT doing consistently doing the `canGoBack` check. But the bigger problem is that this check needs to happen on every subsequent screen because we are using `replace` for most navigation action when navigating away from the bottom sheet. And because of this, the same problem exists: that there might not be a screen to go back to.

I think we need to figure a better solution than doing the canGoBack for every nav action in this screen and subsequent screens as it will be way too much overhead.

One solution is to possibly wait for there to be something in the nav stack before we navigate to the bottom screen. Im open to other solutions, but thats the only thing i can think of right now

cimigree:
> This all looks good but i am blocking this because we need to think about how to deal with the navigation checks. On reviewing #1704, we are NOT doing consistently doing the `canGoBack` check. But the bigger problem is that this check needs to happen on every subsequent screen because we are using `replace` for most navigation action when navigating away from the bottom sheet. And because of this, the same problem exists: that there might not be a screen to go back to.
> 
> I think we need to figure a better solution than doing the canGoBack for every nav action in this screen and subsequent screens as it will be way too much overhead.
> 
> One solution is to possibly wait for there to be something in the nav stack before we navigate to the bottom screen. Im open to other solutions, but thats the only thing i can think of right now

Yes, good catch. And I like the idea about only navigating if there is one screen in the stack. I added a check to the navigation. That is the only way that made sense to me. Otherwise we would have to add a special navigation if/ else check in many, many places.

On the other hand, I am not sure if someone can open the app from cold and have this receiver bottom sheet showing, because they can only be invited to share if their app is open. Once they close the app, the listener seems to go away.

I think I might have been seeing that error when I was double navigating back with the use effect and then through the handle accept or handle decline, which I have fixed in the subsequent PR.

So I don't know if the navigation check is even necessary.

ErikSin (approved):
Looks good!

PR #1702, opened 2026-02-18
fix: screen saver blocked on audio

### LINKED ISSUE #1691: Investigate Audio Recorder being backgrounded by screen saver
Notion: https://www.notion.so/digidem/Audio-records-from-mic-only-until-screen-saver-kicks-in-silence-in-recordings-1901b08162d58001a829f2af109d43d6

## To Do
- [ ] Determine if with the new Expo Audio if the app screen saver is kicking in while recording
- [ ] Determine potential fix

cimigree:
closes #1691 
Adds keep awake module from expo to prevent screen saver from coming up during audio recording. 

Optimizes audio recorder for react compiler.
-  removed the finally block 
- The React Compiler can't optimize try/finally blocks.
- [Here is a reference to the issue in react native](https://github.com/facebook/react/issues/34131)
- [And another issue referencing the same thing](https://github.com/facebook/react/issues/35644)

The screen saver does not come on during development (dev mode). If you want to see the screen saver pop up, you need to build a test or rc. But with this code, even when the screensaver does come on in all other situations, it will not come on during the recording of audio, thus preventing the audio from being stopped prematurely.

**Just FYI, there is now a way to record in the background using expo-audio in case we want to go that route.**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "expo-font": "14.0.10",
         "expo-image": "3.0.11",
         "expo-image-manipulator": "14.0.8",
+        "expo-keep-awake": "15.0.8",
         "expo-localization": "17.0.8",
         "expo-location": "19.0.8",
         "expo-secure-store": "15.0.8",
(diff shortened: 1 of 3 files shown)


ErikSin (approved):
Thanks!

PR #1703, opened 2026-02-18
chore: address circular deps

### LINKED ISSUE #345: address require cycle warnings related to navigation code
Similar to #323, but more focused on navigation app screens

See https://github.com/digidem/comapeo-mobile/pull/342#pullrequestreview-2053935095 for context

ErikSin:
closes #345 

# Circular deps addressed:
Locale Types: The `LocaleStoreContext` was defining the `type LocaleState`, which was being imported by `intl.ts`. Several function from `intl.ts` are being imported by`LocaleStoreContext`. I separated the definition of `type LocaleState` into its own file.
Files: 
- LocaleStoreContext.tsx
- intl.ts
- sharedTypes/locale.ts

DMS types: Previously `DMSForm` was exporting `DmsData` and `DmsUnit`, which was being imported by DMSInputGroup. I separated `DmsData` and `DmsUnit` into its own type folder
Files:
- src/frontend/screens/ManualGpsScreen/DmsForm/types.ts
- src/frontend/screens/ManualGpsScreen/DmsForm/index.tsx
- src/frontend/screens/ManualGpsScreen-/DmsForm/DmsInputGroup.tsx

ExportType: Previously the type `Export` was being exported by the `ExportObservation` screen. I seperated the exportType into the shared navigation types
Files:
- src/frontend/hooks/server/projects.ts
- src/frontend/screens/ExportObservations.tsx
- src/frontend/screens/ExportSuccess.tsx
- src/frontend/sharedTypes/navigation.ts

RootStack. The RootStack was being exported by Stack/index. I separated the RootStack into its own file
Files:
- src/frontend/Navigation/Stack/AppScreens.tsx
- src/frontend/Navigation/Stack/index.tsx
- src/frontend/Navigation/Stack/OnboardingScreens.tsx
- src/frontend/Navigation/Stack/RootStack.ts

PostHog: The Posthog instance was being created in App.tsx. App.tsx was importing AppNavigator. AppNavigator was importing the PostHog Instance. I seperated the PostHog instance into it own file. I also consolidated the AppVariant (which was being used by PostHog). The App Variant was being used in several places, but were checking for it on each instance instead of just having one instance of it. I also updated the tsconfig to properly include app.config.js in typescript checking.
Files:
- src/frontend/App.tsx
- src/frontend/AppNavigator.tsx
- app.config.js
- tsconfig.json

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -1,17 +1,15 @@
+// @ts-check
 const {execSync} = require('child_process');
 const semverParse = require('semver/functions/parse');
 
 /** @import {ExpoConfig} from '@expo/config-types' */
-/** @typedef {typeof VALID_APP_VARIANTS[number]} AppVariant */
+/** @import {AppVariant} from './src/frontend/lib/appVariant' */
 
 const EAS_PROJECT_ID = '2d5b8137-12ec-45aa-9c23-56b6a1c522b7';
 const EAS_UPDATES_URL = 'https://u.expo.dev/' + EAS_PROJECT_ID;
-const VALID_APP_VARIANTS = /** @type (const) */ ([
-  'development',
-  'production',
-  'releaseCandidate',
-  'preRelease',
-]);
+const VALID_APP_VARIANTS = /** @satisfies {readonly AppVariant[]} */ (
+  /** @type {const} */ (['development', 'production', 'rc', 'preRelease'])
+);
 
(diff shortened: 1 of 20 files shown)


cimigree (approved):
Is that the right issue to close? update production dependencies? Other than that, all great! Thanks for taking this on.

PR #1704, opened 2026-02-18
feat: replace background map

### LINKED ISSUE #1591: Replace Background Map Flow
## Screen Shots

Image

Image

Image

## Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=8979-3997&t=wtavZShoLeOJGpJm-4

## To Do
- [ ] Create Background Map Confirmation Screen
- [ ] Create Map Updating Screen
- [ ] Create Map Success Screen

cimigree:
closes #1591 
stacked on top of https://github.com/digidem/comapeo-mobile/pull/1693 
This PR adds the last of the flow of map sharing.
Adds logic for what happens after a receiver elects to accept a map shared.
Navigates to a replace the map sheet page if the person has a current custom map.
Otherwise navigates to a receiving map screen
From there shows a Background map updated screen (a component within the file).
The close from there goes to the background maps screen.

I added the receiving map flow to the screens that block the received map listener, otherwise it just stays open on top of the screens throughout the flow

I had to put the progress bar and the fetch of the mapShare in its own file/ component so that I would not hit an error I hit ("Maximum update depth exceeded" errors) unless it was structured this way. There is one in sending a map and one in receiving a map. Hence the two... I can consolidate into one file, but I like the maxim, repetition over complication. So, let me know if it would be better as one file.

See screen shots below.
image

image

image

I have also added some UI for if  either the receiver or sender cancels midway through sharing that is not as of yet in the Figma:
image
based on this [Slack conversation with Design ](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1772049186403659)


ErikSin:
I haven't done a review yet, I was just testing the functionality. And when i attempt to send a map, I am getting the error bottom sheet (on the sender side) and the following error: `Bounds at 3 must be within max of spherical mercator projection -180,-85.051129,180,85.051129`

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,312 @@
...
+  navigation,
+}: NativeRootNavigationProps<'MapReceivedBottomSheet'>) {
+  const {formatMessage: t} = useIntl();
+  const {shareId} = route.params;
+  const mapShare = useSingleReceivedMapShare({shareId});
+
+  React.useEffect(() => {
+    if (
+      !mapShare ||
+      mapShare.status === 'canceled' ||
```
ErikSin: Why are we not navigating to the `MapShareCanceledBottomSheet` if the mapShare.status is 'canceled'?

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,312 @@
...
+        onSuccess: () => {
+          navigation.goBack();
+        },
+        onError: (err: unknown) => {
+          Sentry.captureException(err);
+          if (navigation.canGoBack()) {
+            navigation.goBack();
+          } else {
+            navigation.replace('BackgroundMaps');
+          }
```
ErikSin: I think we have the same problem of having to check navigation.canGoBack() in the `onSuccess` like you do in `onError`. This problem is actually going to propagate to all the subsequent pages that are navigated to since most of the navigation actions use `replace`.

I think we need to think of a more elegant way of dealing with this. Perhaps we should figure out a way to only show this bottom sheet when there is at least one screen in the navigation stack? Im open to other ideas, but I think checking `navigation.canGoBack()` on this screen and all subsequent screens is going to create too much overhead.

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const {mutate: abortDownload} = useAbortReceivedMapShareDownload();
+  const mapShare = useSingleReceivedMapShare({shareId});
+
+  usePreventAndroidBackButton();
+
+  React.useEffect(() => {
+    if (!mapShare) {
+      navigation.goBack();
+      return;
+    }
```
ErikSin: I don't think this check is necessary. Just based on the type, mapShare is never undefined or null

ErikSin (changes requested):
Im blocking this right now because of the navigation problem of checking if the user can go back. I think we should try to solve that problem in #1700 first, and then update this pr from #1700 and then I can re-review

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
ErikSin: As mentioned in my comment, lets try to solve this in #1700 first and updating from 1700 should fix this

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: Upon further review, this whole check is unnecessary. We don't have a value for the map share until the receiver interacts with the bottom sheet (as I told you earlier, and I forget exactly why...) so it is much cleaner to handle this in the accept and decline buttons and then navigate from there to the canceled sheet, so I have done that. I previously thought you wanted to just close the map received sheet if the person declined the map share after it had been cancelled, but we discussed that before we had the canceled screen, so I think it does make more sense to show that sheet in the case of cancellation.

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -0,0 +1,302 @@
...
+          navigation.replace('ErrorBottomSheet', {error});
+        },
+      },
+    );
+  };
+
+  const handleDecline = () => {
+    if (!mapShare || mapShare.status === 'canceled') {
+      navigation.replace('MapShareCanceledBottomSheet');
+    } else {
```
ErikSin, line 159: I think we should just `goBack()` here. If the sender cancelled AND the receiver declines, the receiver doesn't need to know it was cancelled because they didn't want it anyways. I think it would be confusing to decline and then be navigated to the cancelled screen

src/frontend/screens/BackgroundMaps/MapShareCanceledBottomSheet.tsx
```diff
@@ -0,0 +1,60 @@
...
+  },
+});
+
+export function MapShareCanceledBottomSheet({
+  navigation,
+}: NativeRootNavigationProps<'MapShareCanceledBottomSheet'>) {
+  const {formatMessage: t} = useIntl();
+
+  const handleClose = () => {
+    navigation.popTo('BackgroundMaps');
```
ErikSin, line 31: If backgroundMaps is not in the stack (which in most cases I don't think it will be) this will cause problems. I think we need to `goBAck()` and insure that this screen was navigated to via `replace` OR we could `popToTop`.

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
```diff
@@ -0,0 +1,175 @@
...
+          const error = toError(err, 'Failed to cancel map download');
+          Sentry.captureException(error);
+          navigation.navigate('ErrorBottomSheet', {error});
+        },
+      },
+    );
+  };
+
+  const handleDone = () => {
+    navigation.popTo('BackgroundMaps');
```
ErikSin, line 81: same comment as above. The user will most likely not have BackgroundMaps in the stack

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -124,17 +125,24 @@ export function MapReceivedBottomSheet({
...
+  const {data: customMapInfo, error: customMapError} = useGetCustomMapInfo();
   const {mutate: declineMapShare} = useDeclineReceivedMapShare();
   const {mutate: downloadMapShare} = useDownloadReceivedMapShare();
 
   const handleAccept = () => {
-    // Add code here to navigate to canceled sheet
+    if (!mapShare || mapShare.status === 'canceled') {
+      navigation.replace('MapShareCanceledBottomSheet');
+      return;
+    }
```
ErikSin, line 136: this is not being caught:

https://github.com/user-attachments/assets/8f75a0d1-e0ba-4d4b-8336-feb9da081e9f

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
```diff
@@ -0,0 +1,175 @@
...
+          navigation.popTo('BackgroundMaps');
+        },
+        onError: (err: unknown) => {
+          const error = toError(err, 'Failed to cancel map download');
+          Sentry.captureException(error);
+          navigation.navigate('ErrorBottomSheet', {error});
+        },
+      },
+    );
+  };
```
ErikSin: the abort doesn't seem to be working either:

https://github.com/user-attachments/assets/be8f364e-6bb7-4d00-ad17-2c5b5ab7a2ea

ErikSin (changes requested):
There is really delayed behaviour. Its taking 2 minutes for the screens to even register that there is a map being sent. See video below, jump to 2:10 to see what i mean. 

The other videos i shared in the comments, they might also be a delayed response and I just didn't wait long enough?

@gmaclennan i think you should take a look at this as well

https://github.com/user-attachments/assets/d76f6bb7-8352-4fd5-8db0-b84ed7adfc7c

src/frontend/screens/BackgroundMaps/MapShareCanceledBottomSheet.tsx
cimigree: This does work fine for me without the BackgroundMaps in the stack. I tested several times. I tested this with the exact scenario you described - cancelled from sending device so MapShareCanceledBottomSheet shows up on receiving device with this stack: Home, MapReceivedBottomSheet → ReceivingBackgroundMap → MapShareCanceledBottomSheet  (No BackgroundMaps in the stack).

When I hit Close, `popTo('BackgroundMaps')` goes back to Home first, then navigates to BackgroundMaps. [This is actually the behavior of popTo -- to navigate to the screen if it is not in the stack... ](https://reactnavigation.org/docs/stack-actions/#popto). I thought it was nice to go to background maps since that is the feature we were dealing with (to kind of show there is no change if the sharing is cancelled or below, when a background map is received) but `goBack()` works fine too. I think it's about UX preference not functionality. Given that being true, would you still prefer I change it to `goBack()`?

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
cimigree: Same comment as above.

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: I am not seeing this. I get navigated to the cancelled sheet whenever I try it. Don't know what to do or why this is happening for you.

src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
cimigree: Same as above.

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1704/commits/199e332e93b83c7bda6e585be0b330e9bd992072

src/frontend/screens/BackgroundMaps/MapShareCanceledBottomSheet.tsx
ErikSin: My concern is actually about the unreliability of `popTo()`. Perhaps its not the case anymore because it is a more established api, but when we initially switched to react navigation v7, there were several times where `popTo()` was causing weird behaviour (eg removing gorham bottom from the passcode flow was the main example. But there were several other time where i opted not to use it because it wasn't doing the expected behaviour) But it seems to be working, so maybe lets not take my previous anxieties about this into consideration...

In regards to UX preference it does seem a little weird to cancel the map sharing and then be in the background map screen. But i also am maybe not the one to make those decision. I can ask design and see what they think!

gmaclennan:
Some feedback on the current pattern of navigating as a side-effect of map share state updates. I've mentioned before that I think this introduces potential bugs, but have been unable to be specific.

**What's happening now**
The map share flow uses separate navigation screens to represent each step in what's really a state machine. Sending has 2 screens after device selection (`SendingBackgroundMap`, `MapShareCanceledBottomSheet`). Receiving has 4 (`MapReceivedBottomSheet`, `ReplaceBackgroundMap`, `ReceivingBackgroundMap`, `MapShareCanceledBottomSheet`).

Screens transition between each other via `useEffect` watching the share status and calling `navigation.replace()`. But not consistently — within `SendingBackgroundMap`, `declined`/`downloading`/`completed` are handled as conditional renders, while `aborted`/`error` trigger navigations. There's no clear principle for which statuses get which treatment.

**Why change this?**
Effect-driven navigation races with rendering. When a status changes, the effect fires `navigation.replace(...)`, but the current screen keeps rendering with the new status until the navigation completes. You get a frame of the wrong UI — e.g. download progress showing briefly for an already-canceled share. Conditional rendering handles this synchronously.

Screens can mount into a state they don't expect. `ReplaceBackgroundMap` has an effect that navigates away on `canceled`/`aborted`, but effects run after the first render. This can lead to UI flicker, but it also means every screen needs a guard against being mounted with an unexpected state — it's easy to miss one and end up with a bug.

The navigation stack becomes part of the state machine, but it's not designed for that. `replace` mutates the back stack, so back-button behavior depends on which terminal state was reached. `popTo('BackgroundMaps')` assumes a specific stack shape. This works today but is fragile.

The same status-watching logic is duplicated across screens. "What to do when a share is canceled" is implemented independently in four places. Once [comapeo-core-react#163](https://github.com/digidem/comapeo-core-react/pull/163) lands and the share status becomes the single authoritative state (with typed errors and no optimistic updates), we can centralize the status→UI mapping rather than fixing it in four places.

**Proposed alternative**
Keep the map share received modal/full-screen boundary where it is, but consolidate everything after that boundary.

**Receiving flow — 2 screens instead of 4:**
`MapReceivedBottomSheet` stays as-is. One new full-screen (`ReceiveMapFlow`) replaces `ReplaceBackgroundMap`, `ReceivingBackgroundMap`, and `MapShareCanceledBottomSheet` (receiver side). It conditionally renders: replace confirmation → download progress → success/canceled/error, all driven by the share status plus a bit of local state for the replace confirmation step. It's not necessary to track state from the success of mutations because this will always be reflected in the updated map share status.

**Sending flow — 1 screen instead of 2:**
`SendingBackgroundMap` absorbs its terminal states (`MapShareCanceledBottomSheet` sender side). It already conditionally renders most states — just extend the pattern to cover `aborted`/`error` inline instead of navigating away.

`SelectMapShareDevice` stays separate — it's user-initiated navigation to pick a device, which is a natural screen boundary.

**Canceled/error/aborted bottom sheets**
This does mean moving some of the bottom sheets into a screen, which perhaps is a change from the UX designs. I don't this is a major UX issue, and since we are using `replace()` for these screens, I don't think we get the animation anyway? I think reliability of the feature is more important than the bottom sheet UX in this case, and the idea of bottom sheets is to maintain the context of the content they are appearing over, which is lost if we are replacing the screen when showing the bottom sheet anyway.

**Do you really need to do this?**
It's not essential. Erik's feedback has caught many of the edge-cases where the mutable navigation stack means care needed when navigating between screens, and you can fix missing guards for unexpected states when the status changes before a screen finishes rendering, but it is fragile, and I think if we don't change this we will continue to discover edge-case bugs down the line.

src/frontend/screens/BackgroundMaps/MapShareCanceledBottomSheet.tsx
ErikSin: I asked design, and they also said we should not go back to the background map screen: https://awanadigital.slack.com/archives/C6ZMBFATZ/p1774471443909769

cimigree: Yes I saw and updated this in the current PR: https://github.com/digidem/comapeo-mobile/pull/1693/commits/946e31a2c0d5a7ef682b25343278dd9211f60bec

cimigree:
> **Proposed alternative** Keep the map share received modal/full-screen boundary where it is, but consolidate everything after that boundary.
> 
> **Receiving flow — 2 screens instead of 4:** `MapReceivedBottomSheet` stays as-is. One new full-screen (`ReceiveMapFlow`) replaces `ReplaceBackgroundMap`, `ReceivingBackgroundMap`, and `MapShareCanceledBottomSheet` (receiver side). It conditionally renders: replace confirmation → download progress → success/canceled/error, all driven by the share status plus a bit of local state for the replace confirmation step. It's not necessary to track state from the success of mutations because this will always be reflected in the updated map share status.
> 
> **Sending flow — 1 screen instead of 2:** `SendingBackgroundMap` absorbs its terminal states (`MapShareCanceledBottomSheet` sender side). It already conditionally renders most states — just extend the pattern to cover `aborted`/`error` inline instead of navigating away.
> 
> `SelectMapShareDevice` stays separate — it's user-initiated navigation to pick a device, which is a natural screen boundary.
> 
I did most of this in https://github.com/digidem/comapeo-mobile/pull/1693/commits/2f5341baf216fa4eaa8b38243d2759cac29b3498 in https://github.com/digidem/comapeo-mobile/pull/1693

I wanted to keep the screen with a user interaction to navigate to the next screen separate, so `ReplaceBackgroundMap` navigates to  `ReceivingBackgroundMap` after user interaction. I incorporated the rest of what you said.

PR #1720, opened 2026-03-02
fix: add timezone to metadata

### LINKED ISSUE #1679: Metadata share should include timezone
Notion: https://www.notion.so/digidem/Metadata-missing-time-Zone-2a11b08162d5801f81def57ab1c6e08e

## To Do
- [ ] Add timezone

cimigree:
closes #1679 
Uses intl time formatting to extract and display short timezone on photo and observation metadata.
It displays the currently set timezone that the device is set to at the moment of displaying/ sharing that metadata.

I was not sure which format type to choose, but chose short. Here are all of the options. Would a different option be better?

-  // short: 12/16/2021, PST
-  // long: 12/16/2021, Pacific Standard Time
- // shortOffset: 12/16/2021, GMT-8
- // longOffset: 12/16/2021, GMT-08:00
- // shortGeneric: 12/16/2021, PT
- // longGeneric: 12/16/2021, Pacific Time
- From here: https://developer.mozilla.org/en-

US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

Used for the photo metadata:
image

Observation metadata:
image

And sharing the Observation metadata:
image

This one is one where the phone is set to Hawaii Time. What's App share...
image

### DIFF
--- a/src/frontend/screens/ObservationMetadata.tsx
+++ b/src/frontend/screens/ObservationMetadata.tsx
@@ -271,7 +271,7 @@ export const ObservationMetadata: NativeNavigationComponent<
 
     const date = `${formatMessage(m.date)}: ${formatDate(createdAt, {dateStyle: 'full'})}`;
 
-    const time = formatTime(createdAt, {timeStyle: 'medium'});
+    const time = formatTime(createdAt, {timeZoneName: 'short'});
 
     const baseLocation =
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks!

PR #1731, opened 2026-03-03
fix: text now visible on project edit

### LINKED ISSUE #1709: Invisible Text on Project Editing
**Describe the bug**
On the Edit Project Details screen, found from the Coordinator Tools, the placeholder text is not visible, and when you go to type, nothing is visible. BUT only on a Pixel.

**To Reproduce**
1. On a Pixel, open drawer menu.
2. Click on Coordinator Tools.
3. Click on Edit Info at the top.
4. See error as described above.

**Expected behavior**
The text should be visible both as a placeholder and when typing.

**Screenshots**

Image

 - Device: Google Pixel 8a
 - OS: Android 16
 - Version Abare, 8.3

**Additional context**
The other screens that use the HookFormTextInput seem fine.
This doesn't happen on develop but does happen in an RC and on the Prod version
This is not happening on my Samsung. JUST on the Pixel. Happening on Erik's Pixel also.

cimigree:
closes #1709 
The style prop on the Hook Form Text Input was in the wrong place and was getting overridden by {...RNInputProp} so we would lose the color (and font weight) we set there.
Now React Native's style array properly merges the styles.

Base styles are applied first then RNInputProp.style can come in from the parent to override if necessary.
So now, color: BLACK is preserved unless parent explicitly overrides it.

I also added the placeholder color.
I think that there are now defaults on Pixels of white text and/ or placeholders that there did not used to be and aren't on Samsungs. So it didn't used to matter that our style wasn't applied. But now it does!

image

To test you need to build an RC and install it. And it will work now!

### DIFF
--- a/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
+++ b/src/frontend/screens/ProjectSettings/EditProjectDetails.tsx
@@ -5,7 +5,7 @@ import {HookFormTextInput} from '../../sharedComponents/HookFormTextInput';
 import {useForm} from 'react-hook-form';
 import {NativeNavigationComponent} from '../../sharedTypes/navigation';
 import {BodyText} from '../../sharedComponents/Text/BodyText';
-import {BLACK, COMAPEO_BLUE} from '../../lib/styles';
+import {COMAPEO_BLUE} from '../../lib/styles';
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks!

PR #1732, opened 2026-03-04
chore: e2e for discarding edited observation

### LINKED ISSUE #1695: Create e2e to test discarding of edits made in ObservationEdit
Currently we have e2e that show we can successfully edit an observation, but we do not have a test showing that we when closing the edit screen, and the user chooses NOT to save those edit, that we do in fact NOT save those edit.

## To Do
- [ ] create test to edit, make a change, discard, and check the change isn't kept.

cimigree:
closes #1695 
Adds test to check to make sure discarded edits on an observation are discarded.


ErikSin (approved):
Looks good!

PR #1734, opened 2026-03-05
 feat: updated download observations screen

### LINKED ISSUE #1725: Download Observations - Updated
**Description**
Updating the observation list with https://github.com/digidem/comapeo-mobile/issues/1711 impacts how the user can now download observations

**Details**

- [ ] Tapping on the download button now opens its own screen
- [ ] Tapping on the back button opens up Observations List in the default state (View All).
- [ ] Tapping on  Download to download the selected version to their device.

**Design*

Image

**Feature Spec**
(https://www.notion.so/digidem/Filtered-data-view-MVP-17a1b08162d580f3837ee37b01dedc24)

**Core Strings**

cimigree:
closes #1725 
Switches download observations to full screen.
I left the text the way it was in the file. Not sure if the text in the screen shots and figma makes sense? 
image

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -939,6 +939,9 @@
   "screens.ExportObservations.download": {
     "message": "Download"
   },
+  "screens.ExportObservations.downloadObservations": {
+    "message": "Download Observations"
+  },
   "screens.ExportObservations.errorMessage": {
     "message": "Choose an option"
   },
--- a/messages/en.json
+++ b/messages/en.json
@@ -1409,9 +1409,6 @@
   "screens.ObservationList.TrackListItem.Track": {
     "message": "Track"
   },
-  "screens.ObservationList.downloadObservation": {
-    "message": "Download Observations"
-  },
   "screens.ObservationList.observationListTitle": {
     "description": "Title of screen with list of observations",
     "message": "Observations"
--- a/src/frontend/Navigation/Stack/AppScreens.tsx
+++ b/src/frontend/Navigation/Stack/AppScreens.tsx
@@ -538,6 +538,11 @@ export const createAppScreens = ({
         component={CollaboratorInfo}
         options={createCollaboratorInfoNavOptions({intl})}
       />
+      <RootStack.Screen
+        name="ExportObservations"
+        component={ExportObservations}
+        options={{headerTitle: intl(ExportObservations.navTitle)}}
+      />
     </RootStack.Group>
     <RootStack.Group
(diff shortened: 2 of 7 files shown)


ErikSin (changes requested):
Sorry this wasn't super clear on the github ticket, but you can see it in the figma. We also need to update the observation list screen. First we are getting rid of the Download Observation Button. And then we are adding a download icon on the header, and clicking on that icon now navigates the user to the download observation screen: 

Image

cimigree:
@ErikSin Not sure of the best way to do this -- the architecture is messy, I think. Do you have a better idea?
Here is what it looks like how I did it anyway.

Long named project and screen zoom and big font
---
image

Short named project
---
image

No observations (no change):
---
image

ErikSin:
> @ErikSin Not sure of the best way to do this -- the architecture is messy, I think. 

Yeah, i agree its a little messy. I think its because we are updating the entire header to account for the observation list. We are adding conditional statements to show the download icon, and we are setting those conditional statements to true only when the user is in the observation list screen.

An alternative way which I think would be slightly cleaner is to use `headerRight `in the `<Tab.Screen name='ObservationList' />`

the custom header we use overwrites the `headerRight`, but it has access to it via `options` so we can just render it

eg:
```tsx
// in tab/index.tsx
<Tab.Screen
 name="ObservationsList"
 headerRight: () => (<DownloadIcon onPress={...navAction}/>)
 header: props => (<HomeHeader />),
}}/>



// In SharedComponent/HomeHeader.tsx

export function HomeHeader({
  backgroundColor,
  showBottomBorder,
  onPress,
  options, // is already available via BottomTabHeaderProps, were just not using it
}: HomeHeaderProps) {

  return (
    <View>
      <View style={styles.headerRow}>
        {...rest of header that is already there}
        {options.headerRight?.({tintColor: DARK_GREY)}
      </View>
    </View>
  );
}

```

cimigree:
@ErikSin I tried again. Any better? Any ideas how to do the width differently so it can be less complicated?

ErikSin:
> @ErikSin I tried again. Any better? Any ideas how to do the width differently so it can be less complicated?

I pushed changes that simplify the code and leverages flex so we don't have to have the `shrinkTitle` prop. I also made the download button agnostic and just accept an onPress so you don't have to deal with the `ts-expect-error` code with `useNavigation`

ErikSin (approved):
Looks good!

PR #1738, opened 2026-03-11
chore: manually set up effect for handling 'own-role-change' events

achou11:
`useProjectOwnRoleChangeListener()` core-react unnecessarily exposes a listener callback and there are plans to remove this in the next release (which will introduce a variety of breaking changes). Setting up the listener is more appropriate to do in the app code in this case.

This PR does some early work to prepare for a new release of core-react, and the changes here do not affect functionality related to handling the relevant events.

---

https://github.com/user-attachments/assets/f9637cdc-dc96-42fa-9491-6ff9ec045aa8

### DIFF
--- a/src/frontend/sharedComponents/ProjectRemovalListener.tsx
+++ b/src/frontend/sharedComponents/ProjectRemovalListener.tsx
@@ -3,6 +3,7 @@ import {CommonActions, useNavigationState} from '@react-navigation/native';
 import {
   useOwnRoleInProject,
   useProjectOwnRoleChangeListener,
+  useSingleProject,
 } from '@comapeo/core-react';
 import type {RoleChangeEvent} from '@comapeo/core/dist/mapeo-project';
 import {BLOCKED_ROLE_ID} from '../sharedTypes';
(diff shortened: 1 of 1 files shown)


src/frontend/sharedComponents/ProjectRemovalListener.tsx
```diff
@@ -45,17 +46,23 @@ export const ProjectRemovalListener = () => {
...
+  useProjectOwnRoleChangeListener({projectId});
+
+  const {data: projectApi} = useSingleProject({projectId});
+
+  React.useEffect(() => {
+    function handleRoleChange(event: RoleChangeEvent) {
+      if (event.role.roleId === BLOCKED_ROLE_ID) {
+        dispatchToRemovedProjectBottomSheet();
+      }
+    }
```
achou11, line 59: ideally this is done via a [`useEffectEvent`](https://react.dev/reference/react/useEffectEvent) but that's only available in react 19.2. doesn't make a huge difference in this case since `dispatchToRemovedProjectBottomSheet` is relatively stable anyways.

ErikSin (approved):

PR #1739, opened 2026-03-11
chore: remove usages of `useDocumentCreatedBy()` hook from @comapeo/core-react

achou11:
`useDocumentCreatedBy()` is deprecated in favor of checking the `createdBy` field in a document and checking that against a device ID. I'm considering removing the hook completely in the next release so figured it could be helpful to update this codebase to be prepared for that change.

One of the outcomes is that a few custom hooks have been removed in favor of direct usage + composition: `useGetCreatedBy()`, `useIsMyDocument()`, and `useCanEditOrDelete()` (which was a wrapper around `useIsMyDocument()`). These felt unnecessary compared to making use of copying+pasting a few repeated lines of logic in a few places, but can restore them if there's a preference for that.


ErikSin:
For posterity, @achou11 and I discussed that I am going to take this over. Im going to close this for now, and reopen when I update!

PR #1740, opened 2026-03-12
chore: switch to react native vision library

### LINKED ISSUE #1718: Switch to react native vision camera library
We are switching from expo camera to [React Native Vision Camera](https://react-native-vision-camera.com/docs/guides/). This will allow for 

## To Do
- switch [permission hooks](https://react-native-vision-camera.com/docs/guides/#gettingrequesting-permissions)
- switch out [camera ](https://react-native-vision-camera.com/docs/guides/#use-the-camera-view)
- Make sure photos are being properly saved into the zustand observation store
- DO WE WANT ZOOM?

ErikSin:
closes #1718

### DIFF
--- a/app.json
+++ b/app.json
@@ -50,14 +50,6 @@
         }
       ],
       [
-        "expo-camera",
-        {
-          "cameraPermission": "Allow $(PRODUCT_NAME) to use the camera?",
-          "microphonePermission": "Allow $(PRODUCT_NAME) to access your microphone",
-          "recordAudioAndroid": true
-        }
-      ],
-      [
         "@sentry/react-native/expo",
         {
           "url": "https://sentry.io/",
--- a/app.json
+++ b/app.json
@@ -19,6 +19,7 @@
       "expo-localization",
       "expo-secure-store",
       "expo-document-picker",
+      "react-native-vision-camera",
       [
         "@rnmapbox/maps",
         {
--- a/jest.config.js
+++ b/jest.config.js
@@ -54,6 +54,7 @@ const NODE_MODULE_PATTERNS_TO_TRANSFORM = [
   'yocto-queue',
   'zip-stream-promise',
   'zip-stream',
+  '@lodev09/react-native-exify',
 ];
 
 /** @type {import('jest').Config} */
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -24,6 +24,10 @@ jest.mock('expo/src/winter/ImportMetaRegistry', () => ({
 
 jest.mock('react-native-nitro-modules', () => {});
 
+jest.mock('@lodev09/react-native-exify', () => ({
+  read: jest.fn(() => Promise.resolve(null)),
+}));
+
 jest.mock('@react-native-community/netinfo', () => mockNetInfo);
 
 jest.mock('react-native-safe-area-context', () => mockSafeAreaContext);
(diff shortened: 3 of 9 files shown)


cimigree:
@ErikSin, the frontend CI is really failing here. Could you figure that out before I review, please?

cimigree (approved):
In general everything looks great and the zoom is really fun and I am sure will be quite useful! Just some comments/ questions.

A small change I noticed is that if I push the + button rapidly (to take a bunch of pictures), I get navigated to the something went wrong screen. The first image still takes, but I need to dismiss the something went wrong screen to get back to it.
On the old version, the button just maybe lightens a little and the photo just navigates to the observation, which seems like a better UI response to me as the something went wrong screen is jarring. What do you think?

Also, the size of the photos is sometimes a little bigger now. We used to specify quality: 75 (or something like that) but now it is the default, which is max quality. 
Is there a reason for that change? Would it make sense/ is it possible to lower the quality a bit so the size is smaller per photo? Or is this the desired behavior?

Not a change in this PR, but I noticed that if you try to take a picture without permissions, there is no messaging about why the camera doesn't work.
And if I refuse permissions once, it doesn't come up again. Just something to think about.

ErikSin:
> A small change I noticed is that if I push the + button rapidly (to take a bunch of pictures), I get navigated to the something went wrong screen. The first image still takes, but I need to dismiss the something went wrong screen to get back to it. On the old version, the button just maybe lightens a little and the photo just navigates to the observation, which seems like a better UI response to me as the something went wrong screen is jarring. What do you think?

Fixed here: https://github.com/digidem/comapeo-mobile/pull/1740/commits/884a89c2c3cc7dbe87ca6c54149aba7a4b56515a.

ErikSin:
> Also, the size of the photos is sometimes a little bigger now. We used to specify quality: 75 (or something like that) but now it is the default, which is max quality. Is there a reason for that change? Would it make sense/ is it possible to lower the quality a bit so the size is smaller per photo? Or is this the desired behavior?

I just tested this by taking photos before and after adding react native vision camera  The pixel size of the the photos before and after were the same for me (4080 x 3072). And the size of the photo files were about same before and after (ranging from 1 - 1.4MB). Perhaps its because my phone is relatively new?

When you tested it out, was the photo and file size significantly different?

cimigree:
> > Also, the size of the photos is sometimes a little bigger now. We used to specify quality: 75 (or something like that) but now it is the default, which is max quality. Is there a reason for that change? Would it make sense/ is it possible to lower the quality a bit so the size is smaller per photo? Or is this the desired behavior?
> 
> I just tested this by taking photos before and after adding react native vision camera The pixel size of the the photos before and after were the same for me (4080 x 3072). And the size of the photo files were about same before and after (ranging from 1 - 1.4MB). Perhaps its because my phone is relatively new?
> 
> When you tested it out, was the photo and file size significantly different?

What I did was look at photos I had taken previously and compared them and I did see a decent size general difference (.6 mb to 1.3 mb). But what I didn't understand was how much the size of the photos can vary and why. 

I still don't, actually... 
To check because of what you said, I took specific photos with the package and then took the same photos with the develop branch and in that case while I do notice a difference there is no particular pattern to it and it is not that the rnvc is bigger in size. Sometimes it is smaller! I also tried with my two devices and found the same thing.
Sorry for the false alarm. 

I think the pictures using rnvc may not really be bigger, and it was just a fluke I had seen based on some old photos which had been taken in different places of different things being smaller photos and new ones being bigger for whatever reason...

PR #1749, opened 2026-03-23
feat: expose error details

### LINKED ISSUE #1710: Expose Error Details
**Description**
On CoMapeo Mobile, when user sees an error, the messaging doesn’t allow for easy debugging. This increases product support requests and reduce autonomy
https://www.notion.so/digidem/Expose-Error-Details-on-Mobile-1451b08162d580c4b828e7f1c2978b33

**Proposal**
For this MVP, we will follow Desktop, an display raw details of that error that could help with debugging.
- Coordinator and Participants can see this error state
- User can close the error and continue navigating the app

**Screens**
**Prompt with Advance Info Closed**

Image

**Prompt with Advance Info Open**

Image

Prototype : https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?m=dev

**Core Strings**
- Something went wrong
- Advanced
- Close

cimigree:
closes #1710 
Adds UI to the error bottom sheet to have the 'Advanced' toggle
Adds logic to display the errors that are being passed in to the Error Bottom Sheet
Adds two integration tests
One tests the UI.
One tests that the actual errors we are expecting come through in the form we expect to the Error Bottom Sheet.
They required different set up so I thought it was easier to do them separately. I wasn't sure if you would think we needed them both/ found them both useful. I do... but I am not super familiar and comfortable with integration testing.

image

image

In addition to running the jest tests, the easiest ways I found to test that the error sheet shows up and displays the real error are
1. Upload a config file that is not a config file
2. Go to Your Team → Invite Device. Turn off WiFi/Bluetooth mid-invite on the invited device after they accept the invite (have to be quick about it!). You will see the error.


src/frontend/sharedComponents/ErrorBottomSheet.integration.test.tsx
```diff
@@ -0,0 +1,89 @@
+import type {MapeoManager} from '@comapeo/core';
+import type {MapeoClientApi} from '@comapeo/ipc';
+
+import {createManager, setUpIPC} from '../../../tests/integration/helpers/core';
+
+/**
+ * Integration test that triggers real errors from CoMapeo Core
+ * to verify they have the expected structure (message, stack)
+ * TODO: Add check for Error codes once updated to comapeo-core@6.0.0
+ */
```
ErikSin: I'm not sure I fully understand why this matters. The format of the error doesnt change how the front end handles the error. Both the `message` and `stack` can be undefined, and the error bottom sheet can handle that.

What is your reasoning behind testing whether core returns a certain type of error?

src/frontend/sharedComponents/ErrorBottomSheet.test.tsx
ErikSin, line 1: Good job setting up these tests! This is a clever way of testing the ui.

src/frontend/sharedComponents/ErrorBottomSheet.tsx
```diff
@@ -1,44 +1,92 @@
...
+                style={styles.errorDetailsContainer}
+                contentContainerStyle={styles.errorDetailsContent}>
+                {errorCode && (
+                  <BodyText style={styles.errorText} variant="tinyMeta">
+                    {errorCode}
+                    {'\n\n'}
+                  </BodyText>
+                )}
+                <BodyText style={styles.errorText} variant="tinyMeta">
+                  {errorStack || errorMessage}
```
ErikSin, line 77: How did you decide to only show one of these? I think it would make sense to show both (and IMO i think we should show the error message first), but I am not sure if there is any context that I am missing

ErikSin (commented):
Good job (especially on the ui tests), clever way of creating tests.

Before I approve it i just want to get clarity on the other tests and your reasoning behind it. It feels like unnecessary tests and not something that should be in the code. But I know you also have spent alot more time thinking about errors, so I just want to understand your reasoning before we make the decision to leave it or take it out.

src/frontend/sharedComponents/ErrorBottomSheet.tsx
```diff
@@ -49,9 +97,56 @@ const styles = StyleSheet.create({
     height: '100%',
     alignItems: 'center',
     justifyContent: 'space-between',
+    padding: 30,
```
ErikSin: Also, this was a [mistake in the design](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1774465391139179) and that padding should be 20!

src/frontend/sharedComponents/ErrorBottomSheet.integration.test.tsx
cimigree: You're right that the component handles undefined values. 
But the integration test would catch if the backend changed their error structure in a breaking way.

For example, if CoMapeo Core changed errors to stop including .message and or stop including .stack
The test would fail, alerting us that we should (although we wouldn't _need to_) update the component's fallback logic. 
What if everything is now changed to **just** throw an Error.code (or something else)? We would display something, but we would not be taking advantage of what the backend is giving us and we would not necessarily know that. 
I think of it like a safety net for breaking changes in the backend error format.

The TODO comment is about adding checks for .code when backend adds error codes (like in Gregor's PR: https://github.com/digidem/comapeo-core-react/pull/163) - so we don't miss the opportunity to show better error messages. And then if the structure related to .code changes, the test will catch that.

Does that make sense as a regression test, or do you still think it's unnecessary?

src/frontend/sharedComponents/ErrorBottomSheet.tsx
cimigree: For standard JavaScript Error objects, .stack always includes the message at the top:

Error: <message>\n at <stack frames>
So showing .stack displays both the message and the trace

But I guess the backend could change and not use the standard Javascript Errors...

Would you rather it be?
{errorMessage && <BodyText>{errorMessage}</BodyText>}
{errorStack && <BodyText>{errorStack}</BodyText>}

The downside is the message would appear twice when a stack exists (once standalone, once at the start of the stack). But it's more defensive. Is this what you would prefer?

ErikSin: I did not know that. Using stack makes sense here!

src/frontend/sharedComponents/ErrorBottomSheet.integration.test.tsx
ErikSin: I think what I am concerned about is this is running in CI, but doesn't actually indicate whether the front app is broken or not.

If core/core-react change there error structure, then CI will fail. It just seems like we are testing something that is inconsequential to us but adding it to the CI tests-tests which are designed to be an indicator of whether our app is broken.

Also, changing the error structure would be a breaking change and would be included in the release notes. So we would know this happening when we update core and core-react.

Unless you feel strongly about keeping them, I think we should take these tests out

ErikSin (approved):

PR #1760, opened 2026-03-24
feat: map share fixes

### LINKED ISSUE #1756: `ReplaceBackgroundMap` actions not working

### LINKED ISSUE #1757: Map Reciever Cannot properly cancel a map being shared

### LINKED ISSUE #1759: Prevent screen from shutting off while sending bg map
When the background map is being sent we want to prevent the screen from falling asleep. 

## To Do:
- [ ] add [useKeepAwake](https://docs.expo.dev/versions/latest/sdk/keep-awake/#hooks) to [ReceivingBackgroundMap screen](https://github.com/digidem/comapeo-mobile/blob/0fbc12853872d6266ab3c627df7a69d0aef2d86c/src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx)
- [ ] add [useKeepAwake](https://docs.expo.dev/versions/latest/sdk/keep-awake/#hooks) to [SendingBackgroundMap screen](https://github.com/digidem/comapeo-mobile/blob/0fbc12853872d6266ab3c627df7a69d0aef2d86c/src/frontend/screens/BackgroundMaps/SendingBackgroundMap.tsx)

cimigree:
closes #1759 
closes #1757 
closes #1756 

I also included your fix just for simplicity.
This PR uses the keep awake so the Sending and Receiving background map screens don't black out even when map sharing takes a long time, which will cause the map sharing to cancel.
In order to prevent the other errors and failures where map cancelling throws errors, I had to look for, catch, and handle specific errors. Also looking for cancel or abort on the replace map page, although for me, the useEffect wasn't where that was caught... it was in the error catchers. But maybe it will be caught on a production version.
Doing all this error catching fixes the issues for the user even though I know it is not pretty.

I have an APK if it is helpful.
Here is a link to one: https://drive.google.com/file/d/1NbP5wwAywB0RasDDs3eHvBwc6rBWkfUv/view?usp=sharing

### DIFF
--- a/src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
+++ b/src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
@@ -3,6 +3,7 @@ import {StyleSheet, View, Pressable} from 'react-native';
 import {defineMessages, useIntl} from 'react-intl';
 import MaterialIcon from '@react-native-vector-icons/material-icons';
 import * as Sentry from '@sentry/react-native';
+import {useKeepAwake} from 'expo-keep-awake';
 
 import {
   useAbortReceivedMapShareDownload,
@@ -48,6 +49,7 @@ export function ReceivingBackgroundMap({
   const mapShare = useSingleReceivedMapShare({shareId});
 
   usePreventAndroidBackButton();
+  useKeepAwake();
 
   React.useEffect(() => {
     if (mapShare.status === 'canceled') {
@@ -71,7 +73,7 @@ export function ReceivingBackgroundMap({
         onError: (err: unknown) => {
           const error = toError(err, 'Failed to cancel map download');
           Sentry.captureException(error);
-          navigation.navigate('ErrorBottomSheet', {error});
+          navigation.replace('ErrorBottomSheet', {error});
         },
       },
     );
--- a/src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
+++ b/src/frontend/screens/BackgroundMaps/ReceivingBackgroundMap.tsx
@@ -52,7 +52,7 @@ export function ReceivingBackgroundMap({
   useKeepAwake();
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


ErikSin:
@gmaclennan can you take a look at this.

The crux of the problem is that the `mapShare.status` when cancelled by the Sender is not being received properly by the Receiver.

But when the Receiver tries to do another action, those action are throwing errors because they map share is cancelled.

Cindy's solution here is to check those errors, and if those errors indicate that the mapShare was cancelled, then do the appropriate action. I think it the best solution, but a clunky and fragile solution. The check works by looking for specific words in an error message. First the wording of that message could change in a future release and it would not be a breaking change or in the release notes. Second we could accidentally match error messages from situations that we didn't intend to handle since we are only checking a single word. 

I cannot think of any other way the front end can handle this.

I am wondering if this a case where we should fix this in the backend to avoid this fragile front end code?

PR #1771, opened 2026-03-24
chore: create plugin to remove permission

### LINKED ISSUE #1770: Remove FOREGROUND_SERVICE_MEDIA_PLAYBACK from android manifest
When expo-audio was added (removing the deprecated expo-av library). The library automatically changes the permissions and indicates to Google Play that we are seeking FOREGROUND_SERVICE_MEDIA_PLAYBACK. This feature is for playing audio while the app is in the background (which we do not do).

Google Play requires us to show a video showing why we need this. Since we are not using it we have to take it out to pass the google play requirements which is currently blocking the v10 release.

We need to take this out via an [expo config plugin](https://docs.expo.dev/config-plugins/introduction/).

It should be merged into develop and cherry picked into [hotfix v10.1](https://github.com/digidem/comapeo-mobile/pull/1769)

ErikSin:
closes #1770.

This plugin removes FOREGROUND_SERVICE_MEDIA_PLAYBACK from the android manifest.

This should be cherry picked into the [hotfix v10.1](https://github.com/digidem/comapeo-mobile/pull/1769) after it is merged with develop.

### DIFF
--- a/app.json
+++ b/app.json
@@ -81,6 +81,7 @@
       ["./expo-config-plugins/increaseJavaMemory.js"],
       ["./expo-config-plugins/flagSecure/index.js"],
       ["./expo-config-plugins/customPermissionText.js"],
+      ["./expo-config-plugins/removeMediaPlaybackPermission.js"],
       [
         "react-native-share",
         {
--- a/expo-config-plugins/removeMediaPlaybackPermission.js
+++ b/expo-config-plugins/removeMediaPlaybackPermission.js
@@ -1,0 +1,19 @@
+const {withAndroidManifest} = require('@expo/config-plugins');
+
+/**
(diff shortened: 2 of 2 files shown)


cimigree (approved):
I rebuilt and reinstalled permissions and all seems to work fine. It's super annoying because you can't even see a change in permissions in the settings. Glad you were able to fix.

achou11 (dismissed):
Drive-by observation, but you don't need to define your own plugin for this and can configure it using expo-audio's plugin instead  (see `Configuration in app config` section in https://docs.expo.dev/versions/latest/sdk/audio/).

You should be able to specify `"enableBackgroundPlayback": false` to achieve this.

In general, always check the docs to see if the module provides a way to configure it through their own plugin before building your own 😄

PR #1776, opened 2026-03-25
fix: app crashing on switching project in observation

### LINKED ISSUE #1705: Project invite while Editing Obvs / Acccept - app crash
**Describe the bug**
Being invited to a project while editing an observation causes the app to crash.
Notion: https://www.notion.so/digidem/Project-invite-while-Editing-Obvs-Acccept-app-crash-30b1b08162d58094bf71c3f7ebc3cd1a

**To Reproduce**
Device A editing an obvservation
Device B sends a project invite to Device A
Device A is done editing, sees project invite
Device A accepts the invite and  briefly sees success screen - app crash
Device B sees Device A has accepted invite
Device A can relaunch the app is on the invited project

ErikSin:
closes #1705

What was happening is that when accepting a new a project invite, we updated the project but didn't navigate away from the page they were on. So when a user was viewing an observation and the project was switched, we were attempting to fetch the observation (via the` observationId`) but in a new project where that observation did not exist.

I fixed this by resetting the navigation when the user accepts an invite. I reset the stack so only the Home Screen and the `InviteSuccessfullyAccepted` screen are the only 2 screens on the stack. And when the user closes the `InviteSuccessfullyAccepted` screen, that is when the projectId is switched

### DIFF
--- a/src/frontend/constants.ts
+++ b/src/frontend/constants.ts
@@ -48,6 +48,8 @@ export const EDITING_SCREEN_NAMES: (
   'Success',
   'AuthScreen',
   'ErrorBottomSheet',
+  'ConfirmDiscardObservationBottomSheet',
+  'ConfirmDiscardObservationEditBottomSheet',
 ];
 
 export const INVITE_SCREEN_NAME: (keyof AppStackParamsList)[] = [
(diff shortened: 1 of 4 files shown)


cimigree (changes requested):
I think there are still some issues when editing an observation.
Device Right editing an observation
Device Left sends invite to a project
Device Right goes to discard the observation
In the middle of discarding on Device Right, the invite pops up on Device Right
Device Right accepts the invite
Device Right gets navigated to the new project and then the Observation pops up.
Device Right tries to save the observation.
Gets error screen. Can discard the observation.
All seems normal on Device Left.

Is there anything that can be done about that weird flow? Doesn't seem ideal.

See video:

https://github.com/user-attachments/assets/dafee715-b059-46a7-8b18-61b9a65b43a1

ErikSin:
https://github.com/digidem/comapeo-mobile/pull/1776/commits/4934f38783f45ebfd3a718f06dc34621ab0a13b7 should fix that!

cimigree (approved):

Issue #1750, opened 2026-03-23
Add filtered data view in Early Access


ximenabb:
Hide https://github.com/digidem/comapeo-mobile/issues/1711 behind the early access flag

PR #1782, opened 2026-03-26
chore: add back early access

cimigree:
There wasn't an issue made for this... 
Toward #1750 

When we decided to hold off on implementing Early Access until we had a feature to put behind it, that was done by making an environment variable to hide it. Done in https://github.com/digidem/comapeo-mobile/pull/1551.

This PR removes that environment variable and re-exposes the ability for a user to opt in to early access and to see that as an option in the CoMapeo Settings menu. It also adds back an end to end test for it.

The look in the drawer menu isn't as nice as it used to be. Not sure if design might be able to give us a bit of a different look? See last two screenshots

Screenshots
---

image
image
image
image
image
image
image
image


ErikSin:
This looks all good, but I don't think we should merge this into develop (that is why there is not ticket for it). Lets merge it into #1795. Otherwise there is no way of testing whether this actually works or not. Currently the UI is all there, but I think the more crucial feature is being able to turn features on or off.

Once you switch the base I will approve.

cimigree:
> This looks all good, but I don't think we should merge this into develop (that is why there is not ticket for it). Lets merge it into #1795. Otherwise there is no way of testing whether this actually works or not. Currently the UI is all there, but I think the more crucial feature is being able to turn features on or off.
> 
> Once you switch the base I will approve.

Hmm. Not sure how to do that as #1795 is based on this PR?
When I try to change the base I get this:
Screenshot 2026-04-06 at 2 19 17 PM

ErikSin:
> Hmm. Not sure how to do that as #1795 is based on this PR? 

Oh right, Im confusing the order of operations. I guess lets just close this PR? Since everything is already in the other PR

PR #1795, opened 2026-03-30
feat: observation list toggle

### LINKED ISSUE #1711: Filtered data view
**Description**
Allow Coordinators and Participants to view only their own collected data or all collected data.

**Details**

- [ ] Coordinator and Participant can filter their data on the Observation List
- [ ] User (participant or coordinator) can “view all” by default
- [ ] User can "view all" or view "just you"
- [ ] At app restart, and after 1 hour,  CoMapeo displays “view all”
- [ ] user can navigate in the app, come back to the observation list to previously selected filter
- [ ] Filtering is project specific
- [ ] Data export is not affected by the filtering
- [ ] Joining project as a Participant or Coordinator - default “all”

**Design**

View All
Image

View Just You

Image

**Feature Spec**
https://www.notion.so/digidem/Filtered-data-view-MVP-17a1b08162d580f3837ee37b01dedc24?source=copy_link

cimigree:
added on top of #1782 
closes #1711 

## Summary

- When Early Access in enabled, Adds a filter toggle to the Observations List
- Users can switch between **All** (all observations) and **Just You** (only observations created on this device)
- Filter state persists across navigation and resets to "All" on app restart or after 1 hour of inactivity, and when switching projects

## Implementation details

- New `ObservationFilterContext` (Zustand + MMKV persistence) manages filter state
- Filtering uses the `createdBy` field on observation/track documents to match the current device ID
- Both observations and tracks are filtered

## Testing

- Adds E2E test (`tests/e2e/specs/observations/observation-filter-toggle.test.ts`) covering:
  - Toggle appears when Early Access is on, hidden when off
  - Default state is "All" active
  - Toggling between All and Just You updates active state correctly
  - Observations remain visible after toggling (single-device test)
  - I can't actually test whether or not the filter really works because I can't get the observations from another device, so I will add that to the manual testing document.
- Updated `docs/EndToEndTests/Observations.md`

## Screenshots

**No Toggle (no early access)**

image

**All**
image

**Just You**
image


src/frontend/contexts/ObservationFilterContext.tsx
```diff
@@ -0,0 +1,119 @@
+import {createContext, useContext} from 'react';
+import {createStore, useStore, type StoreApi} from 'zustand';
+import {
+  createJSONStorage,
+  persist as createPersistedState,
+} from 'zustand/middleware';
+import * as v from 'valibot';
```
ErikSin: I don't think we should use a persisted store here. The requirement is that when the app resets, the state reset (it goes back to showing all observations). In otherwords it does not need to know the state before it was reset since its alway shows all observations and therefore, the state does not need to be persisted.

src/frontend/contexts/AppProviders.tsx
```diff
@@ -108,7 +114,10 @@ export const AppProviders = ({
                                     }>
                                     <EarlyAccessStoreProvider
                                       value={earlyAccessStore}>
-                                      <AuthProvider>{children}</AuthProvider>
+                                      <ObservationFilterStoreProvider
```
ErikSin: I think we should scope this provider. The `ObservationFilterStoreProvider` is only being used by the observation list, so lets just wrap the observation list in this context so its not being consumed by the entire app

src/frontend/screens/ObservationsList/index.tsx
```diff
@@ -52,11 +60,49 @@ export const ObservationsList: React.FC<
...
+
+  React.useEffect(() => {
+    handleProjectChange(projectId);
+  }, [projectId, handleProjectChange]);
+
+  React.useEffect(() => {
+    if (isEarlyAccessEnabled) {
+      checkAndResetIfNeeded();
+    }
+  }, [isEarlyAccessEnabled, checkAndResetIfNeeded]);
```
ErikSin: this check should happen everytime the user is navigated back to this page. The requirement is that it should reset every hour. Right now it will only reset if early access is enabled or disabled.

There could be the case where the user is on this page for over an hour (and hasn't navigated away and back to this page). If thats the case, it wont reset, but I think that is ok. We would have to create a timer, which feel like overkill and I think it ok to just check on navigation events.

src/frontend/contexts/AppProviders.tsx
ErikSin: Upon further thought, I don't think we need a provider. The state can live all in the observation list

ErikSin (changes requested):
I think we can get ride of the provider completely and hold all the state in `observationsList/index.tsx`. Because it is a tab IT always remain mounted, so if the user sets state in that page, that state does not reset when the user navigates away (like it normally does in react). So therefore we dont actually need a context to hold the state.

Eg:
```tsx
export function ObservationList(){

 const isEarlyAccessEnabled = useEarlyAccessState(s => s.isEarlyAccessEnabled);
 const [filterMode, setFilterMode] = React.useState<'all' | 'mine'>('all');
 const lastInteractionRef = React.useRef<number | null>(null);

useFocusEffect(
    // check on every navigation event back to this page whether it has been an hour and the filter should reset
    React.useCallback(() => {
      if (
        lastInteractionRef.current !== null &&
        Date.now() - lastInteractionRef.current > 1_HOUR
      ) {
        setFilterMode('all');
        lastInteractionRef.current = null;
      }
    }, []),
  );

 function handleFilterModeChange(mode: 'all' | 'mine'){
   setFilterMode(mode);
   lastInteractionRef.current = Date.now();
 {

 const filteredData = //filtered data here
 
 return (
   {isEarlyAccessEnabled && (
        <ObservationFilterToggle
          filterMode={filterMode}
          onFilterModeChange={handleFilterModeChange}
        />
      )}

   < observations={filteredData}
     ...REST OF OBSERVATION LIST />
 )
}

```

What do you think? This feel a little more verbose, but simpler architecturally

cimigree:
@ErikSin,  I am afraid to say I didn't realize the state was not reset when the user navigated away! So I think your solution is a great one given that truth. 
I updated to include your changes. I also had to add a useEffect to reset when the early access mode is turned off...

ErikSin:
> @ErikSin, I am afraid to say I didn't realize the state was not reset when the user navigated away!

I only know this from spending DAYS trying to figure out why the location hook we were using in the map screen was not unmounting when the user navigated away.

src/frontend/screens/ObservationsList/index.tsx
```diff
@@ -52,11 +57,61 @@ export const ObservationsList: React.FC<
...
+  React.useEffect(() => {
+    if (!isEarlyAccessEnabled) {
+      setFilterMode('all');
+      lastInteractionRef.current = null;
+    }
+  }, [isEarlyAccessEnabled]);
+
+  function handleFilterModeChange(mode: 'all' | 'mine') {
+    setFilterMode(mode);
+    lastInteractionRef.current = Date.now();
```
ErikSin: There is a slight discrepancy in how we are setting the `lastInteractionRef`. In the useFocusEffect and the useEffect when `filterMode` is being set to `all` we are setting `lastInteractionRef` to null. In the `handleFilterModeChange` we are setting `lastInteractionRef` to a date regardless.

I think the best would be to just change it in `handleFilterModeChange`

eg:
```ts
function handleFilterModeChange(mode: 'all' | 'mine') {
  setFilterMode(mode);
  lastInteractionRef.current = mode === 'mine' ? Date.now() : null;
}
```

ErikSin (approved):
It looks good. Sorry you had to go through the work of creating and deleting the provider, but the subsequent changes are architecturally really simple which is nice.

1 minor problem, but not a blocker

src/frontend/screens/ObservationsList/index.tsx
```diff
@@ -52,11 +57,61 @@ export const ObservationsList: React.FC<
...
+      }
+    }, []),
+  );
+
+  React.useEffect(() => {
+    if (!isEarlyAccessEnabled) {
+      setFilterMode('all');
+      lastInteractionRef.current = null;
+    }
+  }, [isEarlyAccessEnabled]);
```
ErikSin: Sorry, one more thing, I think we can get rid of this useEffect entitely. I think the `filteredData` handles the early access state.

If the user had it `filterMode` set to 'mine' and then turned off the early access, the state would not be reset (but the user would still see ALL observations). And then if they turn early access on again (before 1 hour has passed), THEN the state would be set to 'mine' which i don't think is a problem

src/frontend/screens/ObservationsList/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1795/commits/4921d1ea5768f37c06c53cd23bc533b016fe5896

src/frontend/screens/ObservationsList/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1795/commits/4921d1ea5768f37c06c53cd23bc533b016fe5896

PR #1797, opened 2026-03-31
chore: upgrade comapeo dependencies

### LINKED ISSUE #1783: Upgrade Deps `comapeo/core` to v6 and `comapeo/core-react` v10
Upgrade core related deps. We are upgrading the core related deps in 2 parts. First we are going to update `core` to v6.0.2 and `core-react` to v10.1.0. When those are merged into `develop`, we are going to update the map share work from `develop`, and then update the map share branch to latest version of `core` and `core-react`.

## To Do:
- [ ] Upgrade `comapeo/core` to v6.0.2
- [ ] Upgrade `comapeo/core-react` to v10.1.0

cimigree:
closes #1783 
Upgrade @comapeo/core to v6.0.2 and @comapeo/core-react to v10.0.1 and @comapeo-ipc to 7.0.0

### Breaking change migrations:

- Replaces ClientApiProvider with the new ComapeoCoreProvider
- Removes useSetUpInvitesListeners() 
- Updates useManyMembers() in YourTeam to pass includeLeft: true so that we have access to past collaborators
- Fixes TypeScript error in useActiveArchiveServer — uses includeLeft: true so the returned type is MemberInfo[], which satisfies the isActiveArchiveServerMember type predicate
- useProjectOwnRoleChangeListener() — ProjectRemovalListener did not need to change because  it had projectApi.addListener separately (done earlier in anticipation by @achou11  in #1738 )
- Fixes RoleChangeEvent import to come from @comapeo/core root instead of the removed internal dist/mapeo-project subpath

### Map server wiring:
**Most of this code was written by Gregor in https://github.com/digidem/comapeo-mobile/pull/1693/ and ported over in here so that the map can load and new map files can be uploaded in the develop branch, and basically so it doesn't break!**
- Initializes @comapeo/map-server in the Node.js backend (app.js) and exposes it via createAppRpcServer
- Adds createAppRpc on the frontend to connect to the map server over IPC, providing getMapServerBaseUrl to ComapeoCoreProvider
- Replaces the old hand-rolled map hooks (useImportCustomMapFile, useRemoveCustomMapFile, useGetCustomMapInfo) with the equivalents now provided by @comapeo/core-react
- Updates BackgroundMaps screen to use expo-file-system's File.pickFileAsync() and the new useImportCustomMap / useGetCustomMapInfo hooks from @comapeo/core-react

### Infrastructure:

- Importing from @comapeo/map-server pulls in Node.js-only code (node:assert, node:http, etc.) which breaks Metro bundling for React Native — even though only DEFAULT_MAP_ID is used from it. I have worked around this with a Metro resolveRequest redirect (with help from AI), but the real fix should be upstream and I am hoping you will know what to do.
- Adds patches/expo+54.0.33.patch — extends Expo's fetch module to support uploading expo-file-system File objects directly (needed for map file upload) (Gregor wrote)
- Adds src/backend/patches/typebox+1.0.81.patch — replaces Unicode property escape regexes with ASCII equivalents for compatibility with older Android JS engines (Gregor wrote
- Updates jest.config.js to handle new ESM-only deps (custom-error-creator, secret-stream-http), unexported package.json subpath from @comapeo/core v6's stricter exports map, and redirects @comapeo/map-server to its constants file for the same reason as the Metro fix


achou11:
> useProjectOwnRoleChangeListener() — ProjectRemovalListener did not need to change because it had projectApi.addListener separately (I think this is correct. Andrew, can you verify?)

yep! done in #1738 in anticipation of this breaking change :)

src/backend/src/app.js
```diff
@@ -3,14 +3,18 @@ import { join } from 'path'
 import { mkdirSync } from 'fs'
 import { createRequire } from 'module'
 import { MapeoManager } from '@comapeo/core'
-import { createMapeoServer } from '@comapeo/ipc/server.js'
+import { createMapeoServer, createAppRpcServer } from '@comapeo/ipc/server.js'
+import { createServer as createMapServer } from '@comapeo/map-server'
+import { KeyManager } from '@mapeo/crypto'
```
achou11, line 8: should update the backend package.json to include `@mapeo/crypto@1.0.0-alpha.10` as a direct dep

src/backend/package.json
```diff
@@ -42,5 +44,8 @@
   "prettier": {
     "semi": false,
     "singleQuote": true
+  },
+  "overrides": {
+    "@hyperswarm/secret-stream": "6.7.1"
```
achou11, line 50: @gmaclennan - is this override to pin to a specific version or to "upgrade" the versions used by transitive deps? just wondering if it's actually needed as opposed to updating the lockfile

src/frontend/hooks/server/projects.ts
```diff
@@ -59,7 +59,7 @@ export function useActiveArchiveServer({
 }: {
   projectId: string;
 }): ArchiveServerMemberInfo | undefined {
-  const {data: members} = useManyMembers({projectId});
+  const {data: members} = useManyMembers({projectId, includeLeft: true});
```
achou11: although the `isActiveArchiveServerMember` used below is doing the appropriate role-based filtering, seems like ideally this would only get active members.
 
```suggestion
  const {data: members} = useManyMembers({projectId, includeLeft: false});
```

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -144,68 +145,59 @@ export function BackgroundMapsScreen() {
...
-          Sentry.captureException(err);
-          navigate('BackgroundMapErrorBottomSheet', {
-            title: t(m.importErrorTitle),
-            description: t(m.importErrorDesciption),
-          });
-        },
+      onError: err => {
+        if (
+          err instanceof Error &&
+          err.message.includes('cancelled by the user')
```
achou11, line 174: this is a super fragile check. took me a moment to realize that it's matching against an error thrown by expo-filesystem. not sure if I have a better suggestion given that.

a rather unfortunate API design decision...

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -238,39 +237,39 @@ function NoMapScreen({
...
-          </BodyText>
-          <View style={{alignItems: 'center', marginTop: 20}}>
-            <SecondaryDestructiveButton
-              fullSize
-              text={t(m.removeMapFile)}
-              onPress={onRemoveMapFile}
-            />
+      {error &&
+        /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+        (error as any).code !== 'MAP_NOT_FOUND' && (
```
achou11: Think you can use `isHttpError` from `@comapeo/core-react` to clean this up a bit:

```suggestion
        isHttpError(error) && error.code !== 'MAP_NOT_FOUND' && (
```

@gmaclennan - i thought we had some kind of error helper so that the `error.code` here would narrow to known maps-related error codes. Maybe I'm confusing things with core...

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
```diff
@@ -325,7 +324,7 @@ function MapInfoScreen({
           })}
         </BodyText>
 
-        <SecondaryButton
+        {/* <SecondaryButton
```
achou11, line 327: to gain some context: is this commented out because it's used in a follow-up?

src/frontend/sharedComponents/SelectDevice.tsx
achou11, line 1: I would suggest updating `getSelectableDevices()` to only accept active project members. Basically something like this:

```ts
type GetSelectableDevicesParams = {
  peers: PublicPeerInfo[];
  projectMembers: MemberApi.ActiveMemberInfo[];
  selectionMode: SelectionMode;
};

export function getSelectableDevices({
  peers,
  projectMembers,
  selectionMode,
}: GetSelectableDevicesParams): PublicPeerInfo[] {
    return peers.filter(device =>
      projectMembers.some(member => member.deviceId === device.deviceId),
    );
  }

  return peers.filter(device => {
    const existingMember = projectMembers.some(
      member =>
        member.deviceId === device.deviceId &&
        (member.role.roleId === COORDINATOR_ROLE_ID ||
          member.role.roleId === CREATOR_ROLE_ID ||
          member.role.roleId === MEMBER_ROLE_ID),
    );

    return !existingMember;
  });
}
```

However, this exposes a types issue in `useManyMembers({projectId})`: https://github.com/digidem/comapeo-core/issues/1256

Technically can update core-react to plaster over the issue but given that this PR wouldn't be able to bring in that fix anyways, I'm inclined to think that it makes more sense to fix the issue in core, which would subsequently fix the issue in core-react (maybe with some minor changes).

For the time being, you can explicitly specify `includeLeft: false` to work around the types issue.

metro.config.js
```diff
@@ -29,5 +29,17 @@ module.exports = {
     sourceExts: [...config.resolver.sourceExts, 'svg'],
     // Required for importing @comapeo/ipc and rpc-reflector without bundling the server/backend code also
     unstable_enablePackageExports: true,
+    // @comapeo/core-react imports DEFAULT_MAP_ID from the @comapeo/map-server root, which pulls in
```
achou11: This issue should be fixed in [`@comapeo/core-react@11.0.1`](https://github.com/digidem/comapeo-core-react/releases/tag/v11.0.1).

I think instead of updating the metro config here, it would make more sense to create a patch file for core-react to introduce the fix. That way when you eventually upgrade to that version of core-react, the only change needed is removing the patch.

jest.config.js
```diff
@@ -81,7 +83,13 @@ const config = {
   ],
   moduleNameMapper: {
     '\\.svg$': '<rootDir>/src/frontend/__mocks__/svg.tsx',
+    '^@comapeo/core/package\\.json$':
+      '<rootDir>/node_modules/@comapeo/core/package.json',
     '^@comapeo/ipc$': '<rootDir>/node_modules/@comapeo/ipc/dist/index.js',
+    '^custom-error-creator$':
+      '<rootDir>/node_modules/custom-error-creator/index.js',
+    '^@comapeo/map-server$':
```
achou11: assuming a patch file is introduced to fix the offending import, i think this could be updated to only affect the subpath import i.e.

```suggestion
    '^@comapeo/map-server/constants\.js$':
```

achou11 (changes requested):
More of an initial review as I haven't had the capacity to look through some of the more complex pieces or test any of the changes on a device. Overall the changes are going in the right direction and they surfaced a couple of issues in other comapeo modules.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: Yes! It is for map sharing.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: I can clean up all the errors with subsequent map work in any case.

src/backend/src/app.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

src/frontend/hooks/server/projects.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: Yeah sorry. Not sure what else to do.

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

metro.config.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

jest.config.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/f34ab6904fe60f664b2e0bcd87c2b3da59eb3f3f

cimigree:
Screenshot 2026-04-02 at 11 46 00 AM
 Going to address this in a separate PR to dev.

src/backend/patches/typebox+1.1.10.patch
achou11, line 1: minor thing but looks like typebox version used has been updated. should just be a matter of updating the file name if you want the patch-package warning to go away:

```console
Warning: patch-package detected a patch file version mismatch

  Don't worry! This is probably fine. The patch was still applied
  successfully. Here's the deets:

  Patch file created for

    typebox@1.0.81

  applied to

    typebox@1.1.10
  
  At path
  
    node_modules/typebox

  This warning is just to give you a heads-up. There is a small chance of
  breakage even though the patch was applied successfully. Make sure the package
  still behaves like you expect (you wrote tests, right?) and then run

    patch-package typebox

  to update the version in the patch file name and make this warning go away.

---
patch-package finished with 1 warning(s).
```

src/frontend/hooks/server/projects.ts
```diff
@@ -59,7 +59,10 @@ export function useActiveArchiveServer({
 }: {
   projectId: string;
 }): ArchiveServerMemberInfo | undefined {
-  const {data: members} = useManyMembers({projectId});
+  // Cast needed due to a types bug in @comapeo/core: https://github.com/digidem/comapeo-core/issues/1256
+  const {data: members} = useManyMembers({projectId, includeLeft: false}) as {
+    data: MemberApi.MemberInfo[];
+  };
```
achou11: the cast isn't needed here. the types issue upstream is for when `includeLeft` is either `undefined` or omitted.

The issue that's arising is related to the return type of this function being looser than the actually returned type. I would suggest:

1. Updating the return type of `isActiveArchiveServerMember()`:

```ts
export function isActiveArchiveServerMember(
  member: MemberApi.MemberInfo,
): member is ArchiveServerMemberInfo & MemberApi.ActiveMemberInfo {
  if (member.deviceType !== 'selfHostedServer') return false;
  if (!member.selfHostedServerDetails) return false;
  if (member.role.roleId !== MEMBER_ROLE_ID) return false;
  return true;
}
```

2. Updating or removing the return type of `useActiveArchiveServer()`. I would suggest removing it and letting the TS do the type inference for you.

```ts
export function useActiveArchiveServer({projectId}: {projectId: string}) {
  const {data: members} = useManyMembers({projectId, includeLeft: false});
  return members.find(isActiveArchiveServerMember);
}
```

src/frontend/sharedComponents/SelectDevice.test.ts
```diff
@@ -30,15 +25,14 @@ function mockMember(
...
+): MemberApi.ActiveMemberInfo {
   return {
     deviceId,
     name: `Device ${deviceId}`,
-    // @ts-expect-error Unsound but enough for testing purposes
     role: {roleId},
     deviceType,
     joinedAt: new Date().toISOString(),
-  };
+  } as MemberApi.ActiveMemberInfo;
```
achou11: instead of the type cast, i would suggest just adding an expect-error comment for the offending field i.e.

```ts
function mockMember(
  deviceId: string,
  roleId: MemberApi.RoleId,
  deviceType: 'mobile' | 'desktop' = 'mobile',
): MemberApi.ActiveMemberInfo {
  return {
    deviceId,
    name: `Device ${deviceId}`,
    role: {
      // @ts-expect-error Sufficient for testing purposes
      roleId,
    },
    deviceType,
    joinedAt: new Date().toISOString(),
  };
}
```

src/frontend/sharedComponents/SelectDevice.tsx
```diff
@@ -60,8 +58,15 @@ export const SelectDevice = ({
...
-  const projectMembersQuery = useManyMembers(projectId);
+  const {projectId} = useActiveProject();
+  // includeLeft: false is explicit here to get only active members.
+  // The cast is needed due to a types bug in @comapeo/core: https://github.com/digidem/comapeo-core/issues/1256
+  const projectMembersQuery = useManyMembers({
+    projectId,
+    includeLeft: false,
+  }) as {
+    data: MemberApi.ActiveMemberInfo[];
+  };
```
achou11: explained elsewhere but type cast isn't needed here

```suggestion
  });
```

src/frontend/screens/BackgroundMaps/BackgroundMaps.tsx
achou11: maybe at least add a comment about this coming from expo-filesystem? kind of puzzling at initial glance if you're unfamiliar

achou11 (changes requested):
Couple of small changes needed but not much needed after that.

src/frontend/hooks/server/projects.ts
achou11: not really in the scope of this PR, but would additionally argue that this hook is unnecessary and should instead just be some sort of helper function or done inline wherever it's used 😄

src/frontend/sharedComponents/SelectDevice.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/2784634ede97d66862006e0e6e3b16be0037415f

src/frontend/sharedComponents/SelectDevice.test.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/2784634ede97d66862006e0e6e3b16be0037415f

src/frontend/hooks/server/projects.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/2784634ede97d66862006e0e6e3b16be0037415f
(but left the hook)

src/backend/patches/typebox+1.1.10.patch
cimigree: https://github.com/digidem/comapeo-mobile/pull/1797/commits/2784634ede97d66862006e0e6e3b16be0037415f

cimigree:
@achou11 I think I updated everything as you suggested. I did some manual testing with my two devices -- inviting to as well as adding and removing from projects, adding a server. Exchanging. Adding and removing background maps.  As well as some basic stuff like creating a project, adding observations, recording audio, taking pictures. I was able to add a server and see it on the team screen but I am seemingly unable to remove it. I am just stuck in the loading space. I used this server:
'https://comapeo-mobile-e2e-01.fly.dev/'
I had to force close and then it was removed?

achou11 (commented):
Great job on this! Now on to the next upgrade 😂 

Regarding the remote archive removal issue, I think it's related to a bug in core-react (see https://github.com/digidem/comapeo-core-react/pull/171). Would you mind trying out the changes from the PR to see if it fixes the issue? Can do so by creating an updated patch file here or creating a tarball from that PR. If you go with the patch-file approach, make sure you update both the esm and cjs dist directories in core-react, just to make sure the patch actually makes it into the bundle.

I suspect that there may be a few issues that this bug causes related to doing write operations with members. If you can verify that the fix upstream works, you can include that as part of the patch file in this PR. Then when you upgrade to latest core-react the patch file can be removed.

Alternatively, if you don't anticipate too much time happening between now and the upgrade to latest comapeo deps, you can merge this PR and then introduce release with the fix(es) shortly after. What do you think?

achou11:
@cimigree I took it upon myself to add a commit that updates the patch file with the suspected fix. Mind trying the flow with the remote archive removal and seeing if it's fixed by this? It seemed to fix the issue for me when I tried it locally.

cimigree:
@achou11 Thanks for doing the patch but I still had the same issue. It loads endlessly and then when I force close and reopen, the remote archive is gone. Anything in particular I can log or do to troubleshoot?

achou11:
> @achou11 Thanks for doing the patch but I still had the same issue. It loads endlessly and then when I force close and reopen, the remote archive is gone. Anything in particular I can log or do to troubleshoot?

Hm not sure. That might be some unrelated issue. The issue I was seeing and addressing was the remote archive not disaplearing after removing succeeds. Thought the endless loading might be a related thing but maybe it's not?

achou11 (approved):
Approving this to unblock you on being able to do the next upgrade. This is under the assumptions that:

1. The CI failure is an unrelated flake and will eventually pass if tried enough.

2. The issue you mentioned around the archive server is either some kind of flaky behavior.

I could be wrong about both! You would know better.

Even if the mentioned issues are regressions, might be easier to address them as follow ups as opposed to endlessly updating patches in this PR. From what I can tell, the app code changes here seem solid so issues most likely require dep upgrades (as seen with the introduced patch file for core-react).

Issue #1764, opened 2026-03-24
Permission request/remove “record video” mention


ximenabb:
At app onboarding, it states “and record video” but we dont have that functionality

Update to:
Allow CoMapeo to take pictures?

https://www.notion.so/digidem/Permission-request-remove-record-video-mention-3261b08162d580c8a0b3fe7b00c5a3d7

ErikSin:
When i said this was possible originally, I actually misunderstood expo/RN. Its quite simple to [change the message in iOS](https://docs.expo.dev/guides/permissions/#ios) with the expo config. I originally thought you could do the same for Android I misundersood.

I thought we could change it directly in the Android code. In reading the android docs, the permissions are controlled by the Permission Controler, and the [customization of the Permission Controller](https://source.android.com/docs/core/ota/modular-system/permissioncontroller#customizing) can only be done by phone manufacturers (and the can only change colors and themes).

In short, this is not actually possible, so I am closing this issue,

PR #1811, opened 2026-04-01
fix: project removal listener wait for pending done

### LINKED ISSUE #1747: Removing device/ Inconsistent behaviour
**Describe the bug**
https://www.notion.so/digidem/Removing-device-Inconsistent-behaviour-3271b08162d580119d42d12af039727b?source=copy_link

cimigree:
closes #1747 
I was not able to reproduce the problem in a ticket but I did see another one.

What I saw a couple of times, was that when I was removed from a project (successfully) and then reinvited, on two occasions (out of 20) I saw the removed-from-project bottom sheet after successfully accepting the invite. I then close the bottom sheet and am removed from the project I was just invited to. No something went wrong or endless loop, but not ideal, and possibly related.
From logging I saw that a couple of times, when a device was re-invited to a project, the role query cache still held
the previous BLOCKED_ROLE_ID value while refetching the new role. This caused the removed-from-project sheet to
appear. So, I am now checking for isRoleRefetching and don't navigate if we are/ until that is done.

Also, of note, the changes in comapeo-core-react may address issues as it has not been upgraded in the front end yet.

### DIFF
--- a/src/frontend/sharedComponents/ProjectRemovalListener.tsx
+++ b/src/frontend/sharedComponents/ProjectRemovalListener.tsx
@@ -15,6 +15,7 @@ export const ProjectRemovalListener = () => {
   const {projectId} = useActiveProject();
   const {
     data: {roleId},
+    isRefetching: isRoleRefetching,
   } = useOwnRoleInProject({projectId});
 
   const navigation = useNavigationFromHomeTabs();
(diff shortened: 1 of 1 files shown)


ErikSin:
I don't think this definitively closes #1764.

Regardless we can merge the PR as I think it is still good for the app. But lets change the wording in this PR's description so the issue doesn't close.

ErikSin (approved):
Gets rid of some glitchyness!

cimigree:
> Regardless we can merge the PR as I think it is still good for the app. But lets change the wording in this PR's description so the issue doesn't close.

@ErikSin Ximena actually had found this bug and reported it, so I added in the PR Description that this PR closes that issue so everyone can see that the issue she reported and I found should have been resolved here.

PR #1812, opened 2026-04-02
fix: basic ftp override for security

cimigree:
## Summary

- Adds an `overrides` entry to force `basic-ftp` to `5.2.0`, patching a critical path traversal CVE (GHSA-5rq4-664w-9x2c) in its `downloadToDir()` method (as seen in Socket Security, see screenshot below)
- `basic-ftp` is a dev-only dependency pulled in via `@wdio/browserstack-service` → `proxy-agent` → `get-uri` → `basic-ftp` — it is never bundled or shipped in the app

## Notes

I think the vulnerability has no exploitable attack surface in this project since `downloadToDir()` is never called, but the override keeps Socket Security warning resolved.

### Screenshot from socket warning

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12003,9 +12003,9 @@
       "license": "MIT"
     },
     "node_modules/basic-ftp": {
-      "version": "5.0.5",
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks for catching this!

Issue #1813, opened 2026-04-07
Observation details (in observation view) do not appear on some android versions


gmaclennan:
**Describe the bug**
When viewing an observation with answers to the questions for details, the answers do not appear on certain versions on Android (15+) only.

**To Reproduce**
1. Create an observations
2. Add some answers to the "Details" section
3. Save the observation
4. View the observation
5. The answers will not be visible

**Expected behavior**
The answers should be visible

**Screenshots**
This observation has details which do not show. When exchanged with another device, the details are visible.

![Image](https://github.com/user-attachments/assets/feae84cc-b907-4c60-b070-92d529b0deff)

**Smartphone (please complete the following information):**
 - Device: Galaxy A35 5G
 - Model: SM-A356E/DS
 - OS: Android 16
 - Version 10.3

- Device: Tecno Pova 7
- Model: Tecno LJ6
- OS: Android 15 (HiOS 15.1.2)
- Version 10.3

**Additional context**

achou11:
for the sake of linking things, this is describing the same issue as https://github.com/digidem/comapeo-mobile/issues/1564

gmaclennan:
> for the sake of linking things, this is describing the same issue as [#1564](https://github.com/digidem/comapeo-mobile/issues/1564)

Describing the same observable bug, however could perhaps not be the same underlying cause. I will test tomorrow with longer text on the problem phones to see if it is a text cut-off issue vs. a text not appearing at all issue.

achou11:
I wonder if this is related to built-in styles from react native (or expo) that are dynamically set based on device theme. changing from dark mode (my usual) to light mode and then restarting the app fixes the issue for me.

it could be a combination of underlying changes from recent updates and conflicts with how the app defines some styles (which generally don't account for device theme).

<img width="300" src="https://github.com/user-attachments/assets/5ad61140-011b-4b00-a2e9-4dc0a3ae9c59" alt="Screenshot of CoMapeo Mobile displaying text showing up as expected in observation details" />

PR #1814, opened 2026-04-07
fix: fix missing observations details on some phones

gmaclennan:
Attempt at fixing https://github.com/digidem/comapeo-mobile/issues/1813. I think maybe setting `color: undefined` is being interpreted on some versions of Android as transparent text. We don't have this pattern anywhere else, so I think it's good to avoid this here even if it is not the cause of the bug.

### DIFF
--- a/src/frontend/screens/Observation/FieldDetails.tsx
+++ b/src/frontend/screens/Observation/FieldDetails.tsx
@@ -29,7 +29,7 @@ export const FieldDetails = ({
               <FormattedFieldProp field={field} propName="label" />
             </HeaderText>
             <BodyText
-              style={{color: value === undefined ? MEDIUM_GREY : undefined}}>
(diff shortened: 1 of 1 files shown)


cimigree:
@ErikSin I built an RC of this locally and tested on my Google Pixel 8a (where the text was not showing up for RC's or in Production) and with _this_ RC the text does show.

ErikSin:
@cimigree im just going to assign you as the reviewer

cimigree:
Just FYI, according to @achou11 it is related to Dark Mode. If someone who is having the problem changes to light mode, the text shows up. And I can confirm this worked for me also. He postulates that 

> react native (or expo?) changed the way their default platform styles work to deal with device theme which sometimes conflicts with how we define styles in-app (e.g. we have a expo config plugin to remove expo's text input style overrides)

cimigree (approved):

PR #1820, opened 2026-04-08
chore: monthly production upgrades

### LINKED ISSUE #1765: Upgrade Production Deps
Update all production deps to latest possible and exact versions.

## To Do
- [ ] run `npm upgrade-deps:production` and upgrade deps accordingly

cimigree:
closes #1765 
This PR updates production dependencies as part of our regular maintenance process using `npm run upgrade-deps:production`, `depcheck`, and `expo-doctor`.

---

### Updated Packages

| Package | From | To |
| ------- | ---- | -- |
| `@formatjs/intl-getcanonicallocales` | 3.2.1 | 3.2.2 |
| `@formatjs/intl-locale` | 5.2.1 | 5.3.1 |
| `@formatjs/intl-pluralrules` | 6.2.2 | 6.3.1 |
| `@formatjs/intl-relativetimeformat` | 12.2.2 | 12.3.1 |
| `@react-native-vector-icons/ant-design` | 12.4.0 | 13.0.0 |
| `@react-native-vector-icons/fontawesome` | 12.4.0 | 13.0.0 |
| `@react-native-vector-icons/fontisto` | 12.4.0 | 13.0.0 |
| `@react-native-vector-icons/ionicons` | 12.3.0 | 13.0.0 |
| `@react-native-vector-icons/material-design-icons` | 12.4.0 | 13.0.0 |
| `@react-native-vector-icons/material-icons` | 12.4.0 | 13.0.0 |
| `@react-native-vector-icons/octicons` | 20.4.0 | 21.0.0 |
| `@react-navigation/bottom-tabs` | 7.14.0 | 7.15.9 |
| `@react-navigation/native` | 7.1.28 | 7.2.2 |
| `@tanstack/react-query` | 5.90.21 | 5.96.2 |
| `fs-extra` | 11.3.3 | 11.3.4 |
| `posthog-react-native` | 4.34.0 | 4.41.0 |
| `react` | 19.1.0 | 19.1.4 |
| `react-hook-form` | 7.71.1 | 7.72.1 |
| `react-intl` | 8.1.3 | 10.1.1 |
| `react-native` | 0.81.5 | 0.81.6 |
| `react-native-confirmation-code-field` | 8.0.1 | 9.0.0 |
| `react-native-device-info` | 15.0.1 | 15.0.2 |
| `react-native-gesture-handler` | 2.30.0 | 2.31.0 |
| `react-native-mmkv` | 4.1.2 | 4.3.1 |
| `react-native-nitro-modules` | 0.32.1 | 0.35.4 |
| `react-native-safe-area-context` | 5.6.2 | 5.7.0 |
| `react-native-screens` | 4.23.0 | 4.24.0 |
| `react-native-svg` | 15.15.3 | 15.15.4 |
| `valibot` | 1.2.0 | 1.3.1 |
| `zustand` | 5.0.11 | 5.0.12 |

---

### Skipped Packages

| Package | Reason for Skipping |
| ------- | ------------------- |
| `@rnmapbox/maps` | Known Android crash issue when New Architecture is disabled; switching to MapLibre soon so not worth upgrading |
| `@sentry/react-native` | Holding until fully validated by Expo |
| `react-native-zeroconf` | Has custom patches that require Gregor's review before upgrading to v0.14 (16KB page size native changes) |
| `react-native-reanimated` | Requires New Architecture |
| `expo-*` | Managed by Expo SDK; pinned to SDK 54 compatibility — SDK 55 upgrade is a separate PR |
| `react-native-share` | Removed root-path from Android FileProvider config, silently breaking all file sharing. Staying on 12.1.0. |

---

### Security
Added an npm overrides entry to force axios to 1.15.0 across the dependency tree.

axios < 1.15.0 has a NO_PROXY hostname normalization bypass (CVE [GHSA-3p68-rc4w-qgx5](https://github.com/advisories/GHSA-3p68-rc4w-qgx5)) and that can allow Server-Side Request Forgery. We have it because of our E2E testing; axios is only a transitive dependency of @wdio/browserstack-service  — but the override should clean up the Socket alert.

### Removed Packages

These were identified as unused by `depcheck` and confirmed via manual review:

| Package | Reason |
| ------- | ------- |
| `@react-navigation/stack` | Codebase uses `@react-navigation/native-stack` instead; no imports found |
| `expo-dev-client` | Not configured as a plugin and not imported anywhere |
| `promise` | Native `Promise` is supported by Hermes; polyfill not needed |

---

### Patch Updates

| Patch | Change |
| ----- | ------ |
| `react-native-confirmation-code-field` | Updated patch from v8.0.1 to v9.0.0 — same one-line fix to `MaskSymbol` to prevent mask flickering when typing quickly |

---

### Compatibility Checks (`expo-doctor`)

Ran `npx expo-doctor` after upgrading. Getting some warnings about react native and others being ahead of suggested versions but that has been like that for many packages and the manual and automated testing all passes. 

I updated `react-native` just to the latest bug fix/ patch version, which required a slight `react` bump and `react-test-renderer` corresponding bump. Expo doctor also recommended we update `jest-expo` and `babel-preset-expo` so I did.

### Also
- While checking into usage, I discovered a file, CenteredText, that was not being used so I deleted it.

- When investigating`expo-system-ui` , which seemed unused, I noticed in the docs we were supposed to add the  plugin to enforce `userInterfaceStyle: light` on Android at the native level. My hunch is that this might help the invisible text we were seeing when users are in dark mode. 

- Found some invisible text, so I added an explicit color to it

- Sharing audio was being blocked by passcode, so I am now preventing passcode from being required when sharing Audio

- The title for exporting a track was wrong, so I updated that

- There was a button after being removed from a project that didn't fit well if the device name were at all long, so I fixed that

- The remote archive was sometimes showing the "remove from project" button and it shouldn't so I added an additional check for it. It seems that the hook `isArchiveServer` returns false if selfHostedServerDetails is not populated yet — which can happen before the device has fully synced with the server.


achou11:
> Also, when investigatingexpo-system-ui , which seemed unused, I noticed in the docs we were supposed to add the plugin to enforce userInterfaceStyle: light on Android at the native level. My hunch is that this might help the invisible text we were seeing when users are in dark mode.

Ah good catch! Didn't realize the plugin wasn't actually being used but yeah seems like that will most likely fix the recently emerging theming issues. Definitely worth doing a UI check throughout the app to make sure that using the plugin doesn't cause unexpected regressions.

cimigree:
> > Also, when investigatingexpo-system-ui , which seemed unused, I noticed in the docs we were supposed to add the plugin to enforce userInterfaceStyle: light on Android at the native level. My hunch is that this might help the invisible text we were seeing when users are in dark mode.
> 
> Ah good catch! Didn't realize the plugin wasn't actually being used but yeah seems like that will most likely fix the recently emerging theming issues. Definitely worth doing a UI check throughout the app to make sure that using the plugin doesn't cause unexpected regressions.

I am going to go through (just started) this whole document on the app which I think should be standard with upgrades like this. https://docs.google.com/document/d/1wU4oj5m7sHJvjLpAr-DgPIiaj_VBR8tkvy7IeecL00U/edit?tab=t.0#heading=h.fp6aj1o6j81t

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
```diff
@@ -89,13 +89,14 @@ export const CollaboratorInfo: NativeNavigationComponent<
     ownRole.roleId === CREATOR_ROLE_ID;
 
   const isDesktop = deviceType === 'desktop';
+  const isServer = isArchiveServer || deviceType === 'selfHostedServer';
```
ErikSin: You commented:

> The remote archive was sometimes showing the "remove from project" button and it shouldn't so I added an additional check for it. It seems that the hook isArchiveServer returns false if selfHostedServerDetails is not populated yet — which can happen before the device has fully synced with the server.

Can you elaborate more on why this is happening? I dont fully understand so perhaps I am missing some context, but I feel like we should adjust the hook instead of adding this line of code.

ErikSin (commented):
Looks good. I have 1 comment, that Im not sure is blocking, so i just want to understand it before i approve!

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
cimigree: Ah. Good question.
`isActiveArchiveServerMember` has three conditions:

`deviceType === 'selfHostedServer'`
selfHostedServerDetails is populated
`role.roleId === MEMBER_ROLE_ID`
The check for selfHostedServerDetails (condition 2) was designed to distinguish an active server from one that hasn't synced yet. 

But it means the function returns false for a server device that is legitimately present but hasn't fully synced — and then "Remove Device" shows on its card in the collaborator info screen

So maybe we need a new additional hook?
split it into two: `isArchiveServerDevice(member)` (just checks deviceType) and isActiveArchiveServerMember(member) (all three checks ) - just leave as is or use simple hook above and use it in `useActiveArchiveServer`? 
And we could use the simpler one where we just need to know "is this a server? which is just in the one spot in Collaborator info.

ErikSin: I understand. I suppose that is why Andrew named it `isACTIVEArchiveServerMember`. 

But can we just get rid of `isActiveArchiveServerMember` completely and do
`const isServer = deviceType === 'selfHostedServer`?

ErikSin (approved):
Did a pretty thorough manual review and everything seems to be working.

I kind of wanted to hold off to merge this to wait for the map server stuff... but that seems to be in forever limbo, so lets just merge this

I left a reply to that one comment, that i think should be adressed first before merging

src/frontend/screens/YourTeam/CollaboratorInfo.tsx
cimigree: I think so if you mean to just get rid of that in that file. But the hook itself I need to leave because it is used elsewhere.

PR #1833, opened 2026-04-14
feature: mark primary strings

### LINKED ISSUE #1774: Migrate Translated Strings to new Crowdin structure
We have to migrate the current translation into the new structure of `primary` and `secondary` strings. All translatable strings [should have been](https://github.com/digidem/comapeo-mobile/issues/1778) extracted from the app, and compiled into an "en-US" folder in the [messages folder](https://github.com/digidem/comapeo-mobile/blob/develop/messages).

## To Do
- [x] Use Claude to go through all the json files in the [messages folder](https://github.com/digidem/comapeo-mobile/tree/develop/messages), and take out any messages that have been translated in english for the non english languages. Due to an error in crowdin many strings have just had the english as the translation. We need to take these out of the translated json files. 
- [x] Go through Each of the json files and for each language code determine the [Region Subtag](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.4), and append the subtag to the language as the name of the json file. For example `fr.json` would be manually turned intp `fr-FR.json` (France french). For Portuguese use "pt-BR" and Spanish us "es-419". For the rest of the languages use the most comon region subtag.
- [x] Write a script to convert the current files into a new structure,
- [x] Manually upload each to crowdin
- [x] Turn on crowdin integration to download

### LINKED ISSUE #1778: Mark Primary Strings and Update Extraction Script
## To Note
- This step will cause all the primary strings to be untranslated as the Id is changing. This is ok, as we will be migrating at a later step
- This might break e2e tests as some things might not translate properly. If there are not many tests, try to comment them out in e2e so CI passes. If there are lots of tests broken, let them fail in CI and we will just force merge it (as eventually they will get fixed)
- This has to be done after #1777 

## To do
- [x] Go through source strings and mark the primary strings using this [PR](https://github.com/digidem/comapeo-mobile/pull/1722). ADD TO ID $1
- [ ] Update [extraction script](https://github.com/digidem/comapeo-mobile/blob/0b8b3ad603684b3f4776b5bd4f7f974f9e7fea23/package.json#L25) to compile the messages in a new folder called "en-US" in the [messages folder.](https://github.com/digidem/comapeo-mobile/tree/develop/messages) The extraction script should write all the strings marked as primary string into a `primary.json` file in the `en-US` folder. The extraction script should also write all the rest of the strings in a `secondary.json` file in the `en-US` folder.

### LINKED ISSUE #1779: Turn On Github Crowdin Integation (Upload Only)
With the new "en-US " structure turn on the Github Crowdin Integation to upload the source strings only. 

see: https://support.crowdin.com/github-integration/

To do: 
- Separate 2 folders, one for desktop, one for mobile
- In the mobile folder, there should be separation between primary and secondary

### LINKED ISSUE #1780: Update Compile Messages Script
Update the [build-translations script](https://github.com/digidem/comapeo-mobile/blob/develop/scripts/build-translations.mjs) to compile the translations, pulled from crowdin, into translations/message.json file* . This file is what the app uses as a source for all translations. 

Currently the build translations script reads all the `json` files in the [messages folder](https://github.com/digidem/comapeo-mobile/tree/develop/messages) which come from crowdin via the github-crowdin integration. With the work that addresses #1774 , the translations will be in a new format inside this messages folder. The work of this ticket is to update the script to pull from the new folder format.

In later work we can update/optimize how the app consumes the translations, which would require us to change how the translations are compiled. While this might be a good opportunity to address that now, so we can just update script once to compile to this optimized version, I think it would be a good idea to wait because we are making so many [updates to our translation process](https://github.com/digidem/comapeo-mobile/issues/1775) already. In the likely event that we introduce bugs it would be easier to pinpoint where the bug is if we update less code. 

*To note, the messages.json file is typically compiled just before building the app, and [NOT committed to github](https://github.com/digidem/comapeo-mobile/blob/bbbc9b09fb0a8ba533d92293326f5ff0d51a5f09/.gitignore#L56). See `Example message file` below

## To do
- [ ] Update the [build-translations script](https://github.com/digidem/comapeo-mobile/blob/develop/scripts/build-translations.mjs) to pull from the new format of the [messages folder](https://github.com/digidem/comapeo-mobile/tree/develop/messages)
- [ ] Ensure that the file it is compiled to the right format, see `Example message file` below

### Example message file
Here is an example [messages.json](https://github.com/user-attachments/files/26762568/messages.json) file that is built right before the app

cimigree:
closes #1778 
closes #1779 
closes #1774 
closes #1780

Merged in Erik's PR that has the comments that specify which strings should be primary
Added $1 to the beginning of all of the primary strings
Added an extraction script (in scripts and a command in package.json) to sort the primary strings into a primary file and the rest into a secondary file.
Ran the script and built those files.
Includes a temporary patch for the build-translations file to filter out directories so CI passes

**I left the comments in specifying which strings are primary but can easily delete them. What do you think???**


ErikSin:
>I left the comments in specifying which strings are primary but can easily delete them. What do you think???

I think we should delete them, its feels a little redundant at this point

messages/en.json
ErikSin, line 1: we should delete this file

ErikSin (approved):
Looks good. I think we should get rid of my "primary-strings" comments and also get rid of the en.json file

Good job with the node extraction script

ErikSin:
Also, it seems like it is just the language e2e tests that are failing. So i think the best strategy is to just comment them out so CI passes since it seems isolated to that

messages/en.json
cimigree: The build-translations.mjs script still reads from the messages/ folder and en.json is in there. So if we delete it now, the app's English strings would break until step 5. Are you sure we want to do that?

ErikSin (approved):

ErikSin:
THere is one failing test that is unrelated, so I am merging into develop!

ErikSin (approved):

PR #1839, opened 2026-04-16
chore: switch to map libre

### LINKED ISSUE #1736: Switch to Map Libre
Due to Mapbox becoming NOT open source, we are switching to map libre.

Andrew started the work in an experiment, so you can refer to [this PR](https://github.com/digidem/comapeo-mobile/tree/experiment/maplibre).

## To Do
- [ ] Switch to Map Libre

ErikSin:
closes #1736 

Switches from mapbox to maplibre.

The majority of the switch was a 1 to 1 mapping as map box and map libre have the same API.

In the [UserToolTipMarker](https://github.com/digidem/comapeo-mobile/pull/1839/changes#diff-54ec4ddbfaf3a74ca4f901d8952148cd0a22ac535a5e4a4e36c1fe92491af422) I changed the MarkerView to a Point Annotation. MarkerViews are meant to be interactive, and the tool tip is not something the user can interact with, so I made it a Point Annotation which is more optimized.


cimigree (approved):
Looks good and works great. Only two small things.
1. It looks like we will still need the MAPBOX_ACCESS_TOKEN when doing E2E tests.
Unless there is some kind of 
We need to document that somewhere somehow so it doesn't get removed accidentally.

Or, do you know if there is a MapLibre-hosted or open style URL that could replace the Mapbox one in the E2E fallback? [This possibly?](https://demotiles.maplibre.org/style.json)

2. I am seeing repeated logs in the console:
` LOG  MapLibre [info] Request failed due to a permanent error: Canceled  {"level": "warning", "message": "Request failed due to a permanent error: Canceled ", "tag": "Mbgl-HttpRequest"}`
especially when I move around on the map.

It doesn't appear to affect functionality but wanted to make sure you knew about it.

ErikSin:
> Or, do you know if there is a MapLibre-hosted or open style URL that could replace the Mapbox one in the E2E fallback? [This possibly?](https://demotiles.maplibre.org/style.json)

We still use the mapbox map as the main default map. We use maplibre to render the map, but the map tiles come from mapbox.

achou11:
> I am seeing repeated logs in the console:
LOG  MapLibre [info] Request failed due to a permanent error: Canceled  {"level": "warning", "message": "Request failed due to a permanent error: Canceled ", "tag": "Mbgl-HttpRequest"}
especially when I move around on the map.

already posted this elsewhere but for posterity...

tl;dr - basically, any mapbox assets (e.g. tiles, glyphs, etc) will fail to load on the map for devices running e2e tests on browserstack, due to using a hardcoded mapbox style url for the map style.

None of the asset urls that the hardcoded mapbox style url uses will include the access token with them, causing auth errors from mapbox and thus failing to load the assets. usually if you're using a mapbox sdk (like we were before), you set the access token to use client-side so that any asset requests the library makes will include it. however, since we're switching to maplibre, it doesn't do that and there's no way to update the urls requested to include it in the mobile sdks.

for context, our map server usually [transforms the urls](https://github.com/digidem/comapeo-map-server/blob/404a558c4fe8429c10c17721e2e4f5eada07f375/src/lib/style.ts#L20) to include the access token if available, so the auth errors aren't an issue when using the style url coming from the map server.

I don't think this is really a blocker for the PR in any sense since the e2e tests don't test anything related to the underlying map that's rendered. the ideal solution is to avoid the hardcoding of the style url used, but I'm assuming the browserstack issue related to the embedded http server still prevents that. otherwise, you could opt for a different hardcoded online style url like the maplibre demotiles one, but it won't really matter for the purposes of the e2e tests in the end.

PR #1844, opened 2026-04-16
chore: new folder and file structure for translations

### LINKED ISSUE #1774: Migrate Translated Strings to new Crowdin structure

cimigree:
closes #1774 

- Removes English from translations
- Creates folders with region subtags for each language
- Converts translated files into primary and secondary and converts the primary ids


ErikSin (commented):
Now would be the time to get rid of en.json, correct?

cimigree:
> Now would be the time to get rid of en.json, correct?

Yes! Want me to delete it here? It might make more sense for you to do it in your branch?

ErikSin:
> > Now would be the time to get rid of en.json, correct?
> 
> Yes! Want me to delete it here? It might make more sense for you to do it in your branch?

sure, i can delete it in my branch

ErikSin (approved):

PR #1848, opened 2026-04-16
fix: build international polyfills

cimigree:
Removes the region tag when building international polyfills.
Most of the changes are just linting based.
Adds Burmese (language of Myanmar) as a possibly language.

### DIFF
--- a/src/frontend/languages.json
+++ b/src/frontend/languages.json
@@ -339,9 +339,13 @@
     "nativeName": "Avañe'ẽ",
     "englishName": "Guarani"
   },
-  "gu-IN": {
+  "gu": {
     "nativeName": "ગુજરાતી",
     "englishName": "Gujarati"
+  },
+  "gu-IN": {
+    "nativeName": "ગુજરાતી",
+    "englishName": "Gujarati (India)"
   },
   "gx-GR": {
     "nativeName": "Ἑλληνική ἀρχαία",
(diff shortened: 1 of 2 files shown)


src/frontend/languages.json
```diff
@@ -583,6 +583,10 @@
     "nativeName": "ဗမာစကာ",
     "englishName": "Burmese"
   },
+  "my-MM": {
+    "nativeName": "ဗမာစကာ",
+    "englishName": "Burmese (Myanmar)"
+  },
```
ErikSin, line 597: we should also add the one without the regional variant. For each one that has a regional variant we need to make sure it the one without the regional variant

src/frontend/languages.json
ErikSin, line 1: Do we also need to add "gu" and "gu-IN"?

ErikSin (changes requested):
We need to add the language code without the regional variant as well so that they can fall back to each other.

src/frontend/languages.json
cimigree: It was already there:
```
  "my": {
    "nativeName": "ဗမာစကာ",
    "englishName": "Burmese"
  },
```

src/frontend/languages.json
cimigree: We had gu-IN but I added gu

cimigree: https://github.com/digidem/comapeo-mobile/pull/1848/commits/59a4284a54d2f700ce03c947320d34531d716687

cimigree:
Also added Mongolian without the region tag.
https://github.com/digidem/comapeo-mobile/pull/1848/commits/59a4284a54d2f700ce03c947320d34531d716687

ErikSin:
Sorry, this was not needed, my bad

PR #1855, opened 2026-04-20
fix: close keyboard on discard observation

### LINKED ISSUE #1853: Editing Observation or Track/ bottom sheet hidden by keyboard when canceling
https://www.notion.so/digidem/Editing-Observation-bottom-sheet-hidden-by-keyboard-when-canceling-3481b08162d58072bbb4d1ec9d19b976

Image

ErikSin:
closes #1853 

I also had `useKeyboardListener` use `useEffect` instead of the `useFocusEffect`. The `useFocusEffect` was used for turning on and off 2 subscriptions: 1 listening to the keyboard being visible, and the other listening to when the keyboard was being hidden. Based on whether the keyboard was visible or not, we change the ui in the <ActionsRow/> inside these listeners. 

This PR closes the keyboard when the user clicks on the `<headerLeft/>` while also opening the bottoms sheet. When the bottom sheet opens, the `useFocusEffect` return statement is being run, closing the subscriptions. But it seems to close the subscriptions before the action inside the subscription can run. In otherwords the actions row was still showing the UI as if the keyboard was open, even though the keyboard was closed.

By changing it to a useEffect, since the screen is still open in the background, the return statement is NOT run (aka the subscriptions are not closed) when the user navigates away, fixing that problem, as the actions inside the listeners are still running.

https://github.com/user-attachments/assets/ff42ab00-5149-4553-a49a-c4b5ca1246b7

### DIFF
--- a/src/frontend/hooks/useKeyboardListener.ts
+++ b/src/frontend/hooks/useKeyboardListener.ts
@@ -1,26 +1,33 @@
-import {useEffect, useState} from 'react';
+import {useFocusEffect} from '@react-navigation/native';
+import {useCallback, useState} from 'react';
 import {Keyboard} from 'react-native';
 
 export function useKeyboardListener() {
   const [keyboardVisible, setKeyboardVisible] = useState(false);
   const [keyboardHeight, setKeyboardHeight] = useState(0);
 
-  useEffect(() => {
-    const keyboardHideUnsub = Keyboard.addListener('keyboardDidHide', () => {
-      setKeyboardVisible(false);
-      setKeyboardHeight(0);
-    });
+  useFocusEffect(
+    useCallback(() => {
+      const keyboardHideUnsub = Keyboard.addListener('keyboardDidHide', () => {
+        setKeyboardVisible(false);
+        setKeyboardHeight(0);
+      });
 
-    const keyboardShowUnsub = Keyboard.addListener('keyboardDidShow', e => {
-      setKeyboardVisible(true);
-      setKeyboardHeight(e.endCoordinates.height);
-    });
+      const keyboardShowUnsub = Keyboard.addListener('keyboardDidShow', e => {
+        setKeyboardVisible(true);
+        setKeyboardHeight(e.endCoordinates.height);
+      });
 
-    return () => {
-      keyboardHideUnsub.remove();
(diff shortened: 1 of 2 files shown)


cimigree (approved):
This is a good change and fixes the problem in Observation Edit. However, the same problem happens when creating an Observation, or Creating or editing a track. You can use the down arrow to dismiss the keyboard but I feel like that is a little awkward.

Can you add `Keyboard.dismiss()` to the `handlePress` function in 
`src/frontend/screens/ObservationCreate/HeaderLeft.tsx`
`src/frontend/screens/SaveTrack/HeaderLeft.tsx`
`src/frontend/screens/TrackEdit/HeaderLeft.tsx`?

Or should I create a new issue?

ErikSin:
> This is a good change and fixes the problem in Observation Edit. However, the same problem happens when creating an Observation, or Creating or editing a track. You can use the down arrow to dismiss the keyboard but I feel like that is a little awkward.
> 
> Can you add `Keyboard.dismiss()` to the `handlePress` function in `src/frontend/screens/ObservationCreate/HeaderLeft.tsx` `src/frontend/screens/SaveTrack/HeaderLeft.tsx` `src/frontend/screens/TrackEdit/HeaderLeft.tsx`?
> 
> Or should I create a new issue?

Actually, I redid it so that the `useKeyboardListener` just closes the keyboard when the user navigates away. As a result, it actually fixes it in all the places you listed. Would you mind doing a quick re-review?

cimigree (approved):
Three things worked great.
ObservationCreate
ObservationEdit
TrackEdit.

Creating (Saving a new track) had the same problem as before. Looks like it doesn't use the keyboard listener.

See video.

https://github.com/user-attachments/assets/a78df6f6-22ff-4c23-a32b-9cf7d23580e3

I approved it because the problem that is still occurring was not part of the original issue.

ErikSin:
> Three things worked great. ObservationCreate ObservationEdit TrackEdit.
> 
> Creating (Saving a new track) had the same problem as before. Looks like it doesn't use the keyboard listener.
> 
> See video.
> screen-20260422-070257-1776877360131.mp4
> 
> I approved it because the problem that is still occurring was not part of the original issue.

ill create an issue for the tracks!

PR #1861, opened 2026-04-22
fix: something went wrong navigation bar padding

### LINKED ISSUE #1834: “Something went wrong” screen/ Restart app button
**Description**
Odd error state appears, inconsistently. And, the restart app button at the bottom was hidden

**Todo**
move button

https://www.notion.so/digidem/Something-went-wrong-screen-Restart-app-button-3271b08162d5807e88d8fc8b3fec2077?v=1051b08162d5807b950b000c3b2ceb7f&source=copy_link

cimigree:
closes #1834 
Adds padding on the bottom to make the button above the bottom navigation bar.

I would have preferred to use the the SafeAreaView from react-native-safe-area-context but for that we would need to move the SafeAreaProvider to the very top/ outside of the App and I wasn't sure of the consequences of that and if that would work. Seems risky... But let me know if you believe otherwise.

We can't use the react native SafeAreaView from React Native because it is deprecated and only works on iOS.

Google Pixel:
---
image

Samsung Galaxy A25
---
image

---
It is tricky to get to see this screen. What you can do temporarily...
What I did was load the app then in the return of App.tsx put
`<FatalErrorUntranslated />` instead of the usual content (I just commented it out).

**You have to load the app first** though or it just won't work. Or at least I couldn't get it to.

And even though it is the same styling you could also do the same with:
```
return (
    <LocaleStoreProvider value={persistedLocaleStore}>
      <IntlProvider>
        <FatalError />
      </IntlProvider>
    </LocaleStoreProvider>
  );
```

### DIFF
--- a/src/frontend/screens/FatalError.tsx
+++ b/src/frontend/screens/FatalError.tsx
@@ -49,6 +49,7 @@ const styles = StyleSheet.create({
     height: '100%',
     padding: 20,
     paddingTop: 80,
+    paddingBottom: 60,
     alignItems: 'center',
     justifyContent: 'space-between',
     backgroundColor: WHITE,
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
I didn't check since you have screen shots and your just adding padding!

PR #1862, opened 2026-04-22
chore: photo capture improvements

### LINKED ISSUE #1815: Photo Capture Improvements
**Description**
The feature @[Photo Capture Controls / Pinch to zoom](https://www.notion.so/digidem/Photo-Capture-Controls-Pinch-to-zoom-1791b08162d58054acbfc018b3174970?pvs=24) was added to the 2026 roadmap, and since it doesn't require UI change, we want to add some improvements based on logger user issues.

**Todo**

- [x] Add GPS accuracy info on Camera

- [x] Permissions Messaging

More details

https://www.notion.so/digidem/Photo-Capture-Improvements-31d1b08162d5809289dbf24dcb0db15a

cimigree:
closes #1815 
- Adds the GPS Pill to the Camera screen, which when pressed, navigates to the GPS metadata screen

   I removed the absolute position of the add button so that I could put the pill and the button in a flexed row with an absolute position

- Attempts to call request permission on camera screen. 

   This is tricky because that request won't run when permission is denied. 
   Since the request permission runs on app mount, permission is usually denied at that point. 
   The only time I could get that permission check to run on the camera screen was if I checked "Ask every time" in settings permissions.

- Adds different wording and Open Settings button on Camera when permission was denied

image

image


src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -39,10 +46,15 @@ export const CameraView = ({onAddPress}: Props) => {
   const accelerometerMeasurement =
     React.useRef<AccelerometerMeasurement | null>(null);
   const {hasPermission} = useCameraPermission();
+  useRequestCameraPermissionOnMount();
```
ErikSin: were calling useCameraPermission twice here now, which is not necessarily a problem, but can lead to subtle bugs and confusing behaviour down the line. To solve this we can just expose `hasPermission` in  `useRequestCameraPermissionOnMount().`

example:

```ts
export function useRequestCameraPermissionOnMount() {
  const {hasPermission, requestPermission} = useCameraPermission();
  const {mutateAsync} = useCameraPermissionMutation(requestPermission);

  React.useEffect(() => {
    if (!hasPermission) {
      mutateAsync();
    }
  }, [hasPermission, mutateAsync]);

  return {hasPermission}
}


// in CameraView.tsx
const {hasPermission} = useRequestCameraPermissionOnMount();

```

ErikSin: also perhaps we should add `!hasPermissions` to the `const disabled` ie:

```ts
 const disableButton = capturing || !cameraReady || !hasPermissions
```

src/frontend/sharedComponents/AddButton.tsx
ErikSin, line 1: I think this component was being used by the map before (based on the comments). But it is not anymore, so lets just get rid of this component and copy and paste all this code directly in the camera view (and get rid of any unnecessary props like isLoading)

src/frontend/hooks/useCameraPermissionTracker.ts
```diff
@@ -0,0 +1,29 @@
...
+  return useMutation({
+    mutationKey: CAMERA_PERMISSION_MUTATION_KEY,
+    mutationFn: fn,
+    networkMode: 'always',
+  });
+}
+
+export function useRequestCameraPermissionOnMount() {
+  const {hasPermission, requestPermission} = useCameraPermission();
+  const {mutateAsync} = useCameraPermissionMutation(requestPermission);
```
ErikSin, line 22: This isn't doing what it is supposed to be doing:

https://github.com/user-attachments/assets/885680a2-bc7c-478b-90ba-e43081899558

ErikSin (changes requested):
one blocking comment - the comment related to the passcode. But i think we should time box 1/4 of a day trying to figure it out (because im not sure if this is actually a solvable problem). So Look into it, and if we can solve it, great. if not we just need to document it in the PR, and let the rest of the team know that this is the behaviour.

src/frontend/hooks/useCameraPermissionTracker.ts
cimigree: Oh I am so sorry I didn't catch that! I think that was brought around with the change in this PR. https://github.com/digidem/comapeo-mobile/pull/1682
So I had to add a wrapper. See the comment.
I noticed the same thing was happening in Audio so I added it to that usage as well.
Tracks was fine.

src/frontend/sharedComponents/CameraView.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1862/commits/1b4402cb896a086f12eaf07469d9ba2c5e252398

src/frontend/sharedComponents/AddButton.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1862/commits/1b4402cb896a086f12eaf07469d9ba2c5e252398

src/frontend/sharedComponents/CameraView.tsx
```diff
@@ -1,30 +1,37 @@
...
     id: 'screens.CameraScreen.noCameraAccess',
-    defaultMessage: 'No access to camera. Please Allow access in setting',
+    defaultMessage: 'No access to camera. Please allow access in settings.',
   },
-  goToSettings: {
-    id: 'screens.CameraScreen.goToSettings',
-    defaultMessage: 'Go to Settings',
+  openSettings: {
+    id: 'screens.CameraScreen.openSettings',
+    defaultMessage: 'Open Settings',
```
ErikSin, line 34: I know that these were originally not marked as primary strings, but its more so that we didn't see these screens when reviewing the app (because i had all the permission). These are necessary for the basic functionality of the app and should both be marked as primary strings

ErikSin (approved):
looks good, and nice fix! Just one comment related to chaning the strings from secondary to primary. Just make sure the new en-us/secondary and en-us/primary get updated as well before you merge

PR #1867, opened 2026-04-27
feature: metric imperial choice

### LINKED ISSUE #1818: GPS / metric and imperial
**Description**
Ensure ensure that metric/imperial settings and GPS format settings are respected across the app (including sharing but not export)

Image

https://www.notion.so/digidem/GPS-metric-and-imperial-d7031e1de0da48dc85da30446bc1c1a0?v=2e61b08162d580acad71000c86505669&source=copy_link

cimigree:
closes #1818 
- Adds a property in settings for someone to select which unit they would like to use, metric or imperial.
- Adds the screen where they make that selection
- Adds the SVG for the icon for the line in settings
- Adds an E2E test for changing the value
- Adds this to context and providers and defaults to metric.
- Adds a utility function and a test to create that conversion by passing it what needs to be converted and the unit system and for one case, the decimals to use. For the one that uses decimals, there was only one place where we want to fixed(2) -- the observation tooltip, so I added that there. Is it definitely necessary?
- I used that converter in all of the places I could find that have distances.
- I updated some of the old Text components
- See attached screen shots.
Did I miss any? 

---

image
image
image
image
image
image
image
image


src/frontend/contexts/UnitSystemStoreContext.ts
```diff
@@ -0,0 +1,77 @@
...
+    actions,
+  };
+}
+
+export type UnitSystemStore = ReturnType<typeof createUnitSystemStore>;
+
+export const UnitSystemStoreContext = createContext<UnitSystemStore | null>(
+  null,
+);
+export const UnitSystemStoreProvider = UnitSystemStoreContext.Provider;
```
ErikSin: `Provider` will be [deprecated](https://react.dev/blog/2024/12/05/react-19#context-as-a-provider) in the next release, so lets start NOT using it.

src/frontend/lib/unitConversion.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function metersPerSecondOrConversion(
+  mps: number,
+  unitSystem: UnitSystem,
+): {value: string; unit: string} {
+  if (unitSystem === 'imperial') {
+    return {value: (mps * MS_TO_MPH).toFixed(2), unit: 'mph'};
+  }
+  return {value: mps.toFixed(2), unit: 'm/s'};
+}
```
ErikSin, line 52: Why are we converting meter per second into miles per hour? In terms of scale, meters and miles are quite different. Wouldn't ft/s be the more appropriate conversion here?

src/frontend/lib/unitConversion.ts
```diff
@@ -0,0 +1,37 @@
...
+// --- Conversion factors ---
+const M_TO_FT = 3.28084;
+const KM_TO_MI = 0.621371;
+const MS_TO_MPH = 2.23694;
+
+export function metersOrConversion(
+  meters: number,
+  unitSystem: UnitSystem,
+  decimals = 0,
+): {value: string; unit: string} {
```
ErikSin: We should strongly type the unit for all of these since string is too general. Currently there is no way of typescript to know whether it is feet, meters, mph, mps....etc since it can be ANY string. 

```ts
type Unit = "ft" | "m"
export function metersOrConversion(...params): {value: string; unit: Unit}(....rest)

```

src/frontend/screens/Settings/AppSettings/UnitSystemSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <ScrollView
+      testID="unitSystemScreen"
+      contentContainerStyle={{paddingTop: 25}}>
+      <SelectOne
+        value={unitSystem}
+        options={options}
+        onChange={selected => {
+          setUnitSystem(selected);
+        }}
+      />
```
ErikSin: Can you actually get rid of this `SelectOne` in this file?. This was migrated over from Mapeo, and its an unnecessary abstraction. I think we should just copy and paste the code from Select One twice - one for each unit.

src/frontend/lib/unitConversion.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (unitSystem === 'imperial') {
+    return {value: (meters * M_TO_FT).toFixed(decimals), unit: 'ft'};
+  }
+  return {value: meters.toFixed(decimals), unit: 'm'};
+}
+
+export function kmOrConversion(
+  km: number,
+  unitSystem: UnitSystem,
+): {value: string; unit: string} {
```
ErikSin: same comment about strongly typing unit here

src/frontend/lib/unitConversion.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (unitSystem === 'imperial') {
+    return {value: (km * KM_TO_MI).toFixed(2), unit: 'mi'};
+  }
+  return {value: km.toFixed(2), unit: 'km'};
+}
+
+export function metersPerSecondOrConversion(
+  mps: number,
+  unitSystem: UnitSystem,
+): {value: string; unit: string} {
```
ErikSin: same comment about strongly typing unit here

src/frontend/lib/unitConversion.ts
ErikSin, line 1: I think all these functions should NOT deal with the decimal places. the amount of decimal places can be dealt with directly with the consuming components. I know it makes the code a little more verbose, but simpler functions that do less things are better for testing purposes and avoids subtle bugs down the line

src/frontend/screens/ObservationMetadata.tsx
```diff
@@ -116,8 +121,16 @@ export const ObservationMetadata: NativeNavigationComponent<
     data: {name},
   } = useProjectSettings({projectId});
   const coordinateFormat = useCoordinateFormat();
+  const unitSystem = useUnitSystem();
   const openShare = useOpenShareDialog();
+
   const manualLocation = metadata?.manualLocation;
+  const lengthUnit = metersOrConversion(0, unitSystem).unit;
+  const speedUnit = metersPerSecondOrConversion(0, unitSystem).unit;
```
ErikSin: this is a bit akward, calling the same function with the dummy value 0.

Maybe just create 2 helper functions in unitConversion.ts

```ts
export function getMetersUnit(unitSystem: UnitSystem): Unit {
  return unitSystem === 'imperial' ? 'ft' : 'm';
}

export function getSpeedUnit(unitSystem: UnitSystem): Unit{
  return unitSystem === 'imperial' ? 'mph' : 'm/s';
}
```

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
```diff
@@ -222,9 +225,15 @@ export function MapReceivedBottomSheet({
...
-                        })
+                      ? (() => {
+                          const {value, unit} = kmOrConversion(
+                            warningInfo.distanceKm ?? 0,
+                            unitSystem,
+                          );
+                          return t(m.distanceAway, {
+                            distance: `${value} ${unit}`,
+                          });
+                        })()
```
ErikSin: This is a style thing, so I am not asking you to change it. But I personally think IIFE in react are really hard to read. Perhaps something we should discuss next week when we are together to determine if there are certain style things that should/can be avoided.

ErikSin (changes requested):
Overall looks good, there are just 3 small but blocking comments:
1. The functions should not be dealing with the decimal conversions
2. I don't think m/s should be converted to mph
3. using 0 in your functions to get the unit is awkward.

src/frontend/contexts/UnitSystemStoreContext.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/lib/unitConversion.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/lib/unitConversion.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/screens/Settings/AppSettings/UnitSystemSettings.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/lib/unitConversion.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/lib/unitConversion.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/lib/unitConversion.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/screens/ObservationMetadata.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/62701f6fc5fc5d7152d10f3a8a4894dc413c2c05

src/frontend/screens/BackgroundMaps/MapReceivedBottomSheet.tsx
cimigree: I agree.
Easier to read when not IIFE, for the most part. I left 1 in (in `LocationView`) because it was very simple and prevented extra function calls.
https://github.com/digidem/comapeo-mobile/pull/1867/commits/7e1c3cb3b4964260578af67b9d61cd1e7b8520e7

cimigree:
I had to make a guess about how the Unit should be formatted on the new Settings screen. Following the previous design choices made sense to me so that is what I did, showing either imperial or metric based on what they have. Should I confirm with design? It just seemed to follow logically...
image
Also, I followed what was being done with internationalization by only using it in the one place it was already being used, I think, in `AttachedPhotoPreviewModal`. In all the other places I had to add a unit conversion, Internationalization is not/ was not being used, so I did not add it because there are so many diffs in this file already and I wasn't sure it was desired behavior. However, I am wondering if we should have a ticket to internationalize all of the numbers?

src/frontend/contexts/UnitSystemStoreContext.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function UnitSystemStoreProvider({
+  value,
+  children,
+}: {
+  value: UnitSystemStore;
+  children: React.ReactNode;
+}) {
+  return React.createElement(UnitSystemStoreContext, {value}, children);
+}
```
ErikSin: We shouldn't be using createElement here. In general we should either use a jsx component instead of trying to force an react element inside of a non jsx component.

I think the better thing to do here is to NOT export a Provider and just use the Context directlty in AppProviders.tsx

eg:
```ts
//In AppProvider.tsx

return (
 <UnitSystemStoreContext value={unitSystemStore}>
  ...rest of the providers
 </UnitSystemStoreContext> )

```

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
```diff
@@ -33,6 +38,33 @@ import {useAppLanguageTag} from '../../hooks/useAppLanguageTag.ts';
 import {Accordian} from '../../sharedComponents/Accordian.tsx';
 import Octicons from '@react-native-vector-icons/octicons';
 
+function formatDistance(
+  meters: number,
+  unitSystem: UnitSystem,
+  formatNumber: (value: number, opts: object) => string,
```
ErikSin: This relates to your comment about intl. I think we either should use formatNumber for everything or not at all. Using it only for this component is a little confusing, expecially since we are dealing with the formatting manually in all the other components. I would argue to not use formatNumber here and just do it manually like the rest of the components

ErikSin (approved):
2 non blocking comments. Overall looking good, thanks for doing that

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/dbb984c42ad130d85c5c0fb78f23dd90c59d7034

Should we have a ticket to update all  numbers to use intl?

src/frontend/contexts/UnitSystemStoreContext.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1867/commits/dbb984c42ad130d85c5c0fb78f23dd90c59d7034

src/frontend/screens/PhotoPreviewModal/AttachedPhotoPreviewModal.tsx
ErikSin: i don't know. Do you think we should use intl? The manual formatting seems fine to me, but im not sure i fully understand the benefits of using intl

cimigree: I think it is some differences of commas and periods? Not a huge difference but somewhat of a difference. https://www.smartick.com/blog/other-contents/curiosities/decimal-separators/
German, French, Spanish: 1.234,56 (period as thousands separator, comma as decimal)
English: 1,234.56 (comma as thousands separator, period as decimal)
Maybe a low priority ticket? Find out what others think? I don't have strong opinions about it.

PR #1868, opened 2026-04-28
fix: invisible text privacy policy

### LINKED ISSUE #1819: Possible Invisible Text in Dark Mode
**Describe the bug**
We have discovered in a couple of places through use that when in Dark Mode on a recent version of Android, text can be white, which makes it invisible. See https://github.com/digidem/comapeo-mobile/issues/1709 and https://github.com/digidem/comapeo-mobile/issues/1813 and possibly https://github.com/digidem/comapeo-mobile/issues/1564

Default behavior probably changed (in React Native or Expo or Android) whereas undefined text used to be black (maybe?) now it is white if in Dark Mode.

I see this locally on my Samsung in dark mode:

Image

I think we should go through the whole app in Production mode (or an RC works too) and make sure this is not popping up in any more places. Better to get ahead of it...

cimigree:
closes #1819 
Updates the privacy policy text to use our shared text components and to make sure all of the text is visible, which it was not before in dark mode! You can see this screen during on boarding by pressing Learn More on the first screen after Get Started, but also from the Settings -> Data & Privacy -> Learn More once your app is onboarded

I also checked through the whole rest of the app for any RN Text components that do not have an explicit color and there are none others that need to be fixed.

DataPrivacyNotInvisible

### DIFF
--- a/src/frontend/screens/DataAndPrivacy/SettingsPrivacyPolicy.tsx
+++ b/src/frontend/screens/DataAndPrivacy/SettingsPrivacyPolicy.tsx
@@ -22,7 +22,6 @@ SettingsPrivacyPolicy.navTitle = m.navTitle;
 
 const styles = StyleSheet.create({
   scrollContent: {
-    backgroundColor: 'white',
     paddingVertical: 20,
     paddingHorizontal: 30,
   },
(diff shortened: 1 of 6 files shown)


ErikSin (approved):
Nice! Thanks for doing that, looks good. Also thanks for figuring out that it was due to dark mode!

PR #1876, opened 2026-04-30
chore: update settings menu

### LINKED ISSUE #1816: Settings Menu Improvements
**Description**
As we fine tune CoMapeo Mobile, a few screens need to be updated to align with the rest of the app and the direction taken on Desktop. Settings has not been updated since CoMapeo Mobile MVP was released.

Image

https://www.notion.so/digidem/Settings-Menu-Improvements-33b1b08162d58012ad43d9cc3e714f4a

cimigree:
closes #1816 

- Most of the changes in this PR are file moves and adjusting various imports
- Renames Settings to ComapeoSettings
- Makes sure all of the items that by UI are in ComapeoSettings, are in the ComapeoSettings Folder, including security, passcode, obscurepasscode etc, which had to be moved into that folder. 
- Took out the nested AppSettings folder since it is no longer relevant
- Took things out of that file folder that are not on the Comapeo Settings screen, such as Config
- Moves the DrawerMenu to a screen since it is a screen and not a shareable component.
- This file: src/frontend/screens/ComapeoSettings/index.tsx contains the UI changes for the Settings screen.
- Added an icon that I could not find in react-native-vector-icons
- Updated some E2E tests as needed.

Note: In the Figma and Notion, there is project data in the Sharing Permissions section of the Comapeo Setting menu. [According to a Slack message with Ximena](https://awanadigital.slack.com/archives/C6ZMBFATZ/p1777492846088849?thread_ts=1777492655.737909&cid=C6ZMBFATZ), though, the content of the menu should not change, just the UI, so I did not add that project statistics sharing to the Comapeo Settings menu as it lives in Coordinator Tools.

Note: the `MenuListItem` and `FullScreenMenu` are now only used in security since they are no longer used in the app settings/index. These probably don't need to be shareable components any more. However, there are so many diffs in this PR I thought it would make more sense to clean that up in a separate ticket/ pr.

Question: There are spanish translations in the Notion. Should I be adding those in Crowdin?

### Screenshots

**Regular Size**

image

**Large Size**
image
image


src/frontend/sharedComponents/DrawerMenu.tsx
ErikSin, line 1: I would argue that this is not a screen. The screens have there own page in the `Navigation/Stack/AppScreens` or in `Navigation/Tab`. The drawer is only used as a side component that is shared between tabs.

src/frontend/screens/Categories.tsx
ErikSin, line 1: Lets rename this file and export to "Categories" to match the new terminology.

src/frontend/screens/ComapeoSettings/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+AppSettings.navTitle = m.title;
+
+function SectionHeader({label}: {label: string}) {
+  return (
+    <BodyText variant="tinyMeta" style={styles.sectionHeader}>
+      {label}
+    </BodyText>
+  );
+}
```
ErikSin: This feels a little unnecessary as its just returning a `BodyText`. Lets get rid of the SectionHeader and just share the `styles.sectionHeader`

src/frontend/screens/ComapeoSettings/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  testID?: string;
+  materialIconName?: MaterialIconsIconName;
+  icon?: React.ReactNode;
+}) {
+  return (
+    <TouchableNativeFeedback
+      testID={testID}
+      onPress={onPress}
+      background={TouchableNativeFeedback.Ripple(VERY_LIGHT_BLUE, false)}>
+      <View style={styles.row}>
```
ErikSin: Lets use a TouchableOpacity Instead. TouchableNativeFeedback is a remnant from mapeo, and is not compatible with iOS

src/frontend/screens/ComapeoSettings/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  materialIconName,
+  icon,
+}: {
+  label: string;
+  onPress: () => void;
+  actionText?: string;
+  showArrow?: boolean;
+  testID?: string;
+  materialIconName?: MaterialIconsIconName;
+  icon?: React.ReactNode;
```
ErikSin: Instead of passing the MaterialIconName or Icon, lets just pass an `icon`. It will make the code a little more verbose as now we have to pass an entire react node, but it will get rid of the ambigous ternary code

src/frontend/screens/ComapeoSettings/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  actionText,
+  showArrow,
+  testID,
+  materialIconName,
+  icon,
+}: {
+  label: string;
+  onPress: () => void;
+  actionText?: string;
+  showArrow?: boolean;
```
ErikSin: Similar comment as the one related to the icon, this leads to some ambigous ternary code. IF there is both an `actionText` and a `showArrow` then only the action text will show up. Its leads to having unnecessary props that are not being used.

I think the better pattern here is to just have a prop that is a reactNode so that the developer can just directly pass in actionText or the Arrow icon

eg:
```ts
function SettingsRow{
  label,
  onPress,
  EndContent
  testID,
  Icon,
}: {
  label: string;
  onPress: () => void;
  EndContent:ReactNode
  testID: string;
  Icon: ReactNode
}){

return(
 <TouchableOpacity>
  {Icon}
  <HeaderText> {label}</HeaderText>
  {EndContent}
</TouchableOpacity>
)


}
```

ErikSin (changes requested):
Good job on this. A couple small but blocking comments related to the SettingsRow component. The way it is designed leads to ambiguous behaviour and some unnecessary props that can be solved by making the code slightly more verbose, while still keeping the re usability of the component

src/frontend/sharedComponents/DrawerMenu.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

src/frontend/screens/Categories.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

src/frontend/screens/ComapeoSettings/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

src/frontend/screens/ComapeoSettings/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

src/frontend/screens/ComapeoSettings/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

src/frontend/screens/ComapeoSettings/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1876/commits/7b14ce05cf5d7570ce9bb8d31ebcd2c29bd664f3

ErikSin (approved):
Looks good, Thank you

PR #1877, opened 2026-04-30
fix: color dot on inset map

### LINKED ISSUE #1744: Color dot on map not the same as in detail view
**Describe the bug**
The color dot of the observation on the map view is not the same as the color dot on the observation detail view.

**To Reproduce**
- Collect an observatiom
- - See the color of its observation on the map - example “green”
- Go to observation list and tap on the same observation
- See the dot on the map be orange and not green

**Expected behavior**
Color dot on map should be the same

https://www.notion.so/digidem/Color-dot-on-map-not-the-same-as-in-detail-view-3261b08162d580b6ad59d23f7cea17e9?source=copy_link

cimigree:
closes #1744 
- Passes the preset color to map inset 
- Makes a dot using css instead of using the orange dot svg.  I used the styling in the SVG to determine what css to use (size, border width, etc)
- Removes the orange dot svg since it is not used anywhere.

image


ErikSin (approved):
Thanks! Look good

PR #1878, opened 2026-05-05
feat: deep linking set up

### LINKED ISSUE #1766: Set Up Deep Linking
In preperation for invites over the internet, we need to set up [Deep Linking](https://reactnavigation.org/docs/deep-linking/)

# To Do
- [ ] Configure Deep Linking
- [ ] Test Deep Linking
- [ ] Determine and test how this works with passcode

cimigree:
closes #1766
###  What this adds
Sets up the deep linking infrastructure needed for invite-over-the-internet. When a user taps an invite link (comapeo://invite/<id> or https://app.comapeo.org/invite/<id>), the app opens and navigates to the InviteReceived screen (for now).

### Changes:
- Adds the configuration as [specified in the react navigation docs](https://reactnavigation.org/docs/deep-linking/) for deep linking.
- Also added logic in `RootStackNavigator` for two situations where the app isn't ready to navigate immediately: mid-onboarding and passcode locked. In both cases the invite ID is captured and held, then navigation happens once the app is ready. A requestAnimationFrame is used so that navigation only happens after the app navigator has finished mounting its screens.
- To support the above, a flag (`deepLinkReady`) in `deepLinkConfig.ts `tracks whether the app is ready. AppNavigator reads this flag to suppress React Navigation's own automatic Link handling while the app is not ready — otherwise React Navigation would try to navigate before the right screens exist.

### Questions:
- Obscured mode: Currently, if the app is in obscured mode (passcode screen showing), an invite link is saved and shown after unlock — same as the passcode scenario. Should invite links be suppressed entirely in obscured mode, or is showing them after unlock the right behavior?

- Has the domain for invite links been confirmed? The current implementation assumes app.comapeo.org based on the invite-over-internet in notion, but if that's not finalized the host will need to change in both `app.json` and `deepLinkConfig.ts`.

- assetlinks.json: Am I responsible for setting this up on app.comapeo.org? Without it, Android will fall back to the custom scheme only and won't intercept https:// links.

### To do next:
- Serve assetlinks.json from app.comapeo.org (or wherever) to enable Android App Links verification for https:// links
- Update `InviteReceived `to handle internet invites (the current screen is built for local wifi invites) or make new UI for it and it should handle expired and/ or invalid invite ids
- The hash fragment handling for sensitive invite data (noted in deepLinkConfig.ts) — `InviteReceived` (or whatever we create) will need to parse window.location.hash or the raw URL to retrieve the secret
- Web fallback page — for users without the app, https://app.comapeo.org/invite/<id> should serve a page explaining how to download CoMapeo, right?

### Testing:
Integration tests would not really work for this because it is all about timing and navigation which is something we would have to mock so much it wouldn't actually test what we want to test.
Instead, you can test manually.

Requires a dev build on a physical Android device.

Scenario 1 — Happy path:

`adb shell am start -W -a android.intent.action.VIEW -d "comapeo://invite/test-invite-id-123" com.comapeo.dev`
Expected: app opens but there will be an error message since that invite id is not handled by `InviteReceived`. I mocked out the response to it in InviteReceived just to check, which you can do if you wish. 

Scenario 2 — App backgrounded with passcode:

Set a passcode in Settings
Background the app and wait for it to lock
Run the adb command above
Enter the passcode Expected: same as above.


cimigree:
@ErikSin I know you suggested moving the DeepLinkListener (new name) into the AppScreens file but I could not figure out how to do it because  

`ERROR  [Error: A navigator can only contain 'Screen', 'Group' or 'React.Fragment' as its direct children (found 'DeepLinkListener'). To render this component in the navigator, pass it in the 'component' prop to 'Screen'.]`

so if you could provide a bit of guidance about what you were thinking I would appreciate it. And, should I try to add an integration test to this somehow once our testing is updated? I will need a few tips about the approach there too, I am afraid.

src/frontend/Navigation/Stack/DeepLinkListener.tsx
```diff
@@ -0,0 +1,31 @@
...
+import * as ExpoLinking from 'expo-linking';
+import {parseInviteUrl} from '../../lib/deepLinkConfig';
+import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
+
+export const DeepLinkListener = () => {
+  const navigation = useNavigationFromRoot();
+  const url = ExpoLinking.useURL();
+  const [pendingInviteId, setPendingInviteId] = React.useState<string | null>(
+    null,
+  );
```
ErikSin: I don't think I fully understand the architecture, but would the PendingInvitesListener also catch the invite? And if thats the case, would there be 2 navigation events?

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -56,6 +58,14 @@ export const RootStackNavigator = () => {
...
   const activeProjectId = useActiveProjectId();
   const {formatMessage} = useIntl();
+  const isNotReadyForInvite =
+    security.authState === 'unauthenticated' ||
+    !deviceInfo.name ||
+    !activeProjectId;
+
+  React.useEffect(() => {
+    setDeepLinkReady(!isNotReadyForInvite);
+  }, [isNotReadyForInvite]);
```
ErikSin: You can do this without a useEffect:

```ts
const isNotReadyForInvite =
    security.authState === 'unauthenticated' ||
    !deviceInfo.name ||
    !activeProjectId

//directly inline with no useEffecr
setDeepLinkReady(!isNotReadyForInvite);

```

src/frontend/Navigation/Stack/DeepLinkListener.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import * as ExpoLinking from 'expo-linking';
+import {parseInviteUrl} from '../../lib/deepLinkConfig';
+import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
+
+export const DeepLinkListener = () => {
+  const navigation = useNavigationFromRoot();
+  const url = ExpoLinking.useURL();
```
ErikSin: This is deprecated and you should use `useLinkingURL()`

src/frontend/lib/deepLinkConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const linking: LinkingOptions<AppStackParamsList> = {
+  prefixes: [ExpoLinking.createURL('/'), `https://${DEEP_LINK_HOST}`],
+  config: {
+    screens: {
+      InviteReceived: 'invite/:inviteId',
+    },
+  },
+};
```
ErikSin: the config property here is used for matching screens with the url. ie, allows you to have a 1 to 1 matching with the url, so as long as your url matches your screen pattern you can open that screen directly. But we are not doing that. We are always intercepting the url, and, grabbing the invite id, and then dispatching a navigation based on that invite id. This would cause 2 navigation events, so I think we need to get rid of it

src/frontend/Navigation/Stack/DeepLinkListener.tsx
```diff
@@ -0,0 +1,31 @@
...
+    null,
+  );
+
+  React.useEffect(() => {
+    if (!url) return;
+    const inviteId = parseInviteUrl(url);
+    if (inviteId) {
+      setPendingInviteId(inviteId);
+    }
+  }, [url]);
```
ErikSin: i think we can get rid of this useEffect entirely.

eg
```ts
const url = ExpoLinking.useURL();
const pendingInviteId = url ? parseInviteUrl(url) : undefined

React.useEffect(() => {
    if (!pendingInviteId) return;
    const waitUntilMount = requestAnimationFrame(() => {
      navigation.navigate('InviteReceived', {inviteId: pendingInviteId});
      setPendingInviteId(null);
    });
    return () => cancelAnimationFrame(waitUntilMount);
  }, [pendingInviteId, navigation]);

```

src/frontend/Navigation/Stack/DeepLinkListener.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  React.useEffect(() => {
+    if (!pendingInviteId) return;
+    const waitUntilMount = requestAnimationFrame(() => {
+      navigation.navigate('InviteReceived', {inviteId: pendingInviteId});
```
ErikSin: I think the better thing to do is follow the architecture of the other listeners, where they are aware of the nav route and uses `isInviteScreen` and `isEditingScreen`. This is also related to this comment:

>@ErikSin I know you suggested moving the DeepLinkListener (new name) into the AppScreens file but I could not figure out how to do it because
>ERROR  [Error: A navigator can only contain 'Screen', 'Group' or 'React.Fragment' as its direct children (found 'DeepLinkListener'). To render this component in the navigator, pass it in the 'component' prop to 'Screen'.]
>so if you could provide a bit of guidance about what you were thinking I would appreciate it. And, should I try to add an integration test to this somehow once our testing is updated? I will need a few tips about the approach there too, I am afraid.

What i was originally suggesting was just an in passing comment when we were looking at the conditional navigation, it wasnt a fully formed suggestion, just something to explore.

But i think also we can avoid using requestAnimationFrame if the listener is aware of the nav route. The nav route is set after the navigation has fully finished. So if we are passing the navigation route to this listener we know the navigation has fully finished before we try and dispatch the navigation action

ErikSin (changes requested):
Thanks for doing this work. I think your possibly dispatching the navigation event for 1 invite up to 3 times:
1. config property of the deeplinking
2. the DeepLinkListener
3. the PendingInviteListener (this one im not sure about)

Also, I think that we can simplify the listener a bit if it is aware of the nav route.

src/frontend/Navigation/Stack/DeepLinkListener.tsx
cimigree: No — PendingInvitesListener watches useManyInvites() which returns local peer-to-peer invites not the deep link invites, which is a completely different path — it never appears in that list. So there's no double navigation from that source.

src/frontend/Navigation/Stack/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1878/commits/c0770cacd287ac0b9e3b3048d1f2440d7eefe15a

src/frontend/Navigation/Stack/DeepLinkListener.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1878/commits/c0770cacd287ac0b9e3b3048d1f2440d7eefe15a

src/frontend/lib/deepLinkConfig.ts
cimigree: https://github.com/digidem/comapeo-mobile/pull/1878/commits/c0770cacd287ac0b9e3b3048d1f2440d7eefe15a

src/frontend/Navigation/Stack/DeepLinkListener.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1878/commits/c0770cacd287ac0b9e3b3048d1f2440d7eefe15a

src/frontend/Navigation/Stack/DeepLinkListener.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1878/commits/c0770cacd287ac0b9e3b3048d1f2440d7eefe15a

src/frontend/Navigation/Stack/index.tsx
```diff
@@ -56,6 +57,10 @@ export const RootStackNavigator = () => {
   const {data: deviceInfo} = useOwnDeviceInfo();
   const activeProjectId = useActiveProjectId();
   const {formatMessage} = useIntl();
+  const isNotReadyForInvite =
+    security.authState === 'unauthenticated' ||
```
ErikSin: > Obscured mode: Currently, if the app is in obscured mode (passcode screen showing), an invite link is saved and shown after unlock — same as the passcode scenario. Should invite links be suppressed entirely in obscured mode, or is showing them after unlock the right behavior?

I don't think we should show the invite link if the app is obscured. So this logic should be

```ts
cons isNotReadyForInvite = security.authstate !== authenticated ....rest

```

ErikSin (approved):
Sorry for the delay, i just wanted to have the meeting about the invites before i fully did a review.

to answer your question:
> assetlinks.json: Am I responsible for setting this up on app.comapeo.org? Without it, Android will fall back to the custom scheme only and won't intercept https:// links.

No, ill create an issue for this and we can do it later.

Just one non blocking comment that answers your question about obscure mode.

PR #1888, opened 2026-05-07
fix: nav back to camera after discarding observation

### LINKED ISSUE #290: Regression in navigation when returning to home tabs after discarding observation flow
Surfaced in https://github.com/digidem/comapeo-mobile/pull/257, but the regression can be reproduced with the following steps:

1. Navigate to camera tab
2. Press create observation button
3. Press button in top-left to discard observation (i.e. leave observation creation flow)
4. Returns to **_map screen_**

The result in step (4) is different than what happens in Mapeo Mobile. In Mapeo Mobile, the app returns to the **_camera tab_** in step (4). I don't expect the new behavior to be the intention, so I'm classifying this as a regression

ErikSin:
closes #290.

While the fix only adds a few lines of code, the logic is a little naive (and kind of convoluted to follow). We originally did not do this to avoid this added complication, but it seems like it was identified by enough people to be a problem.

### DIFF
--- a/src/frontend/screens/ObservationCreate/ConfirmDiscardObservationBottomSheet.tsx
+++ b/src/frontend/screens/ObservationCreate/ConfirmDiscardObservationBottomSheet.tsx
@@ -11,8 +11,8 @@ import {
   DestructiveButton,
   SecondaryButton,
 } from '../../sharedComponents/Buttons';
-import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
 import {useDraftObservationActions} from '../../contexts/DraftObservationContext';
+import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
 
 const m = defineMessages({
   discardTitle: {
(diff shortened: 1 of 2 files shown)


src/frontend/screens/ObservationCreate/ConfirmDiscardObservationBottomSheet.tsx
```diff
@@ -37,14 +37,27 @@ const m = defineMessages({
...
+    const homeTabRoute = navigation
+      .getState()
+      .routes.find(route => route.name === 'Home');
+    // the home screen is a nested navigator (the tabs)
+    // find the latest open tab in the nested tab navigator
+    const lastOpenedTab =
+      homeTabRoute?.state?.routes[homeTabRoute.state.index || 0]?.name;
+    // This assumes that the user can only navigate to this screen via the Camera or Map Screen
+    // If that changes, this will naively go back to the map screen.
+    navigation.popTo('Home', {
```
cimigree, line 58: A possible alternative just to be explicit that ObservationsList Tab is not an acceptable alternative at this time...
```
const targetTab =
  lastOpenedTab === 'Camera' || lastOpenedTab === 'Map'
    ? lastOpenedTab
    : 'Map';
```
`navigation.popTo('Home', {screen: targetTab});`

src/frontend/screens/ObservationCreate/ConfirmDiscardObservationBottomSheet.tsx
```diff
@@ -37,14 +37,27 @@ const m = defineMessages({
...
 
   function handleDiscard() {
     clearDraft();
-    navigation.popTo('Home', {screen: 'Map'});
+    // find the screen with the name "home"
+    const homeTabRoute = navigation
+      .getState()
+      .routes.find(route => route.name === 'Home');
+    // the home screen is a nested navigator (the tabs)
+    // find the latest open tab in the nested tab navigator
```
cimigree, line 53: Consider the possible alternative. The behavior and outcome is the same, but the ?? communicates the intent that when the state is undefined, we use 0, whereas your way, which still works, reads like a null guard when it is not because 0 will come through as falsey and 0 is what we will often have.

`const tabIndex = homeTabRoute?.state?.index ?? 0;`
`const lastOpenedTab = homeTabRoute?.state?.routes[tabIndex]?.name;`

cimigree (approved):
Ah. Kind of an unfortunate pain. Too bad there isn't a more elegant way to do it, but this  all works fine, but some possible alternative choices are noted below... just for clarity in what and why the code is checking for not in the actual working and outcome.

PR #1890, opened 2026-05-07
fix: observation title truncated

### LINKED ISSUE #1873: Large size font/ Observation title truncated
**Describe the bug**
https://www.notion.so/digidem/Large-size-font-Observation-title-truncated-34a1b08162d5809182fef615b18912b1?source=copy_link

Image

cimigree:
closes #1873
**Changes**
Removes fixed height from observation list so also have to remove getItemLayout optimization.

---
The reason we had fixed height was so that we could use this https://reactnative.dev/docs/flatlist#getitemlayout.
getItemLayout matters most when users scroll quickly through large lists — without it, FlatList has to measure every item lazily as it comes into view, which can cause blank frames and weirdness on fast scrolls.
So this is what it had looked like for some with large fonts:

---
image

---
However, even if we picked a super large magic number for height to allow for the most narrow of phones and the largest text (so we would have fixed height list items and could use getItemLayout) that doesn't seem like it would be optimal behavior. For most people the observation item would be too big. And even if we made it really big, for some it would still be too small.

So, instead, I opted to get rid of the getItemLayout feature and instead use a minHeight of 80 pixels. Now the height of each list item is variable depending on the content.

Please let me know if you would rather go a different way.

---
image

### DIFF
--- a/src/frontend/screens/ObservationsList/ObservationListItem.tsx
+++ b/src/frontend/screens/ObservationsList/ObservationListItem.tsx
@@ -43,7 +43,7 @@ function ObservationListItemNotMemoized({
     <TouchableOpacity
       onPress={() => onPress(observation.docId)}
       testID={testID}
-      style={{flex: 1, height: 80}}>
+      style={{flex: 1, minHeight: 80}}>
       <React.Suspense
         fallback={
           <View style={[styles.container, style]}>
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
looks good!

PR #1894, opened 2026-05-12
chore: update jest setup

ErikSin:
Updates Jest setup to automatically create a render function that mounts the entire navigation with a project already setup (as this will be needed for the majority of tests)

Writes a test of the setting screen to show how this helper function should be used.


cimigree (approved):
Exciting!

Issue #1883, opened 2026-05-06
Background tracking reliability improvements


gmaclennan:
**Describe the bug**
Tracks are consistently not being recorded when the app is in the background or the phone screen is off. This happened occasionally in the past - likely due to the app crashing or Android power management restricting the background usage - but in recent releases it happens consistently across all devices, including devices that worked fine in previous releases.

**To Reproduce**
1. Start a track and walk a short distance until you can confirm a track is being recorded.
2. Turn off the phone screen.
3. Walk a route that includes a bend, e.g. around two sides of a city block.
4. Turn the phone screen back on.

The result is a straight line between the location when the phone screen was turned off and when it was turned on.

**Expected behavior**
The track should continue to record while the app is in the background or the phone screen is off.

**Diagnosis**
"Background location" is really confusing on Android and with Expo. There are actually two mechanisms for getting location updates when the app is backgrounded:

1. Background location updates enabled by the `ACCESS_BACKGROUND_LOCATION` permission.
2. Location updates in a foreground service (FGS), which stays "in the foreground" while the app main activity is backgrounded, and does not require the `ACCESS_BACKGROUND_LOCATION` permission.

The `expo-location` module enables a foreground service (FGS) for location updates _if_ the `foregroundService` property of the [`LocationTaskOptions`](https://docs.expo.dev/versions/latest/sdk/location/#locationtaskoptions) is set, and `isAndroidForegroundServiceEnabled=true` in the [location plugin configurable properties](https://docs.expo.dev/versions/latest/sdk/location/#configurable-properties) (it defaults to true if `isAndroidBackgroundLocationEnabled=true`).

Without a FGS, background location updates are severely throttled — Android 8.0+ limits backgrounded apps to "only a few times each hour" regardless of the requested update interval ([docs](https://developer.android.com/about/versions/oreo/background-location-limits)). The FGS path bypasses this throttle because the OS treats a foreground-service-running app as foregrounded.

[...]

PR #1898, opened 2026-05-18
feat: enable foreground service for background tracks

### LINKED ISSUE #1884: Tracks not working when app is bg / Hotfix
**Describe the bug**
https://github.com/digidem/comapeo-mobile/issues/1883

add foreground service property / not internationalized

gmaclennan:
This is a quick partial fix for https://github.com/digidem/comapeo-mobile/issues/1883 (closes https://github.com/digidem/comapeo-mobile/issues/1884).

My understanding from reading the Expo docs and the `expo-location` source code is that a foreground service is only created if the `foregroundService.notificationTitle` and `foregroundService.notificationBody` properties are set when calling `Location.startLocationUpdatesAsync()`. A foreground service is the only way to reliably get location updates when the app is backgrounded or the screen is off.

Tomorrow I plan to test this around the neighbourhood to see if it fixes things.

This PR does not address internationalization of the notification about the foreground process working, nor does it request permissions for the notification, which means that on most devices, users will not see a notification anyway (however my understanding from the Android docs is that a Foreground Service will run regardless of whether the users gives permission for a notification, it's just that an app needs to "try").


cimigree:
I tried this out today from my house. No street corners where I live unfortunately, but the track did record all of what was happening even with the app in the background.
image

I don't have a screenshot, but when I did the same thing on a build with develop, the track worked for the out, but on the way back, it was a straight line from where I backgrounded the app to the when I reopened.

ErikSin (approved):

PR #1902, opened 2026-05-21
chore: simplify intl providers

### LINKED ISSUE #1881: Simplifiy Intl Providers
Currently we have 2 providers for Internationalization: 1. Locale Store and 2. the Intl Context. The Locale store is the persisted store that persists the language code, which can be null (indicating that the user is using the system preferences). The Intl Context, takes the language code and provides the translations based on the language code. Since the language code from the Locale Store can be null, we cannot take the language code directly from the Locale Store and we have to add an extra step of parsing the language code. 

This has caused a spaghetti code situation as the current locale resolution is split across three layers — LocaleStoreContext stores a nullable language tag, useAppLanguageTag resolves it by reading both the store and system locales, and IntlContext reads system locales again independently. This means useLocales() is called twice per render and the resolution logic is spread across multiple files.

We can simplify this by resolving the language tag once at store initialization (using the synchronous getLocales()), the persisted store is always in a valid, fully-resolved state from the moment the app starts. In otherwords, the persisted store always has a valid language tag and eliminates the intermediate useAppLanguageTag hook, reduces the number of moving parts, and makes the data flow easier to follow: user preference is stored, resolved at init, and consumed directly.

- [ ] Create function `getUsableLanguageTagFromSystemPreferences` that takes the `systemLanguageTags: Array<string>;` as a parameter and returns a language tag that is usable by the app. Should fallback to english if no language tag is useable
- [ ] Change LocaleState.languageTag from null | SupportedLanguageTag to SupportedLanguageTag — it should always hold a resolved value
- [ ] Update createInitialState in LocaleStoreContext to call getLocales() synchronously and resolve the language tag at startup (using `getUsableLanguageTagFromSystemPreferences`).
- [ ] Update the setLanguageTag action so that it cannot accept null, and can accept `SupportedLanguageTag` | `'SystemPrefrence'` and always resolves a language tag using getUsableLanguageTagFromSystemPreference if the user has set language to systemPreference
- [ ] Delete useAppLanguageTag hook and its test file — callers should use useLocaleState(s => s.languageTag) directly.
- [ ] Update IntlContext to read languageTag from useLocaleState instead of useAppLanguageTag — useLocales() is still called here, but only for building the regional variant message fallback chain
- [ ] Update LocaleStoreContext tests to reflect that the store is never in a null/unresolved state
 Check getAppLanguageTag in lib/intl.ts — its internal LocaleState param may need adjustment if the type no longer allows null
- [ ] Remove all uses of `useAppLanguageTag` and `getAppLanguageTag`
- [ ] Create a migration step for the LocaleStoreContext (since it no longer take null)

cimigree:
closes #1881 

The Ticket was very thorough so I just followed along!

1. Because we are changing the context/ provider — migration is included

`LocaleState` changed shape: `languageTag` is now always a `SupportedLanguageTag`, never null. I changed the persisted store version from 0 → 1 and added a `migrate()` function to handles existing versions. Users who had a chosen language keep it; users who had null (system preference) get it resolved on first launch. 

2. `setLanguageTag` Action changed — null replaced with `'SystemPreference'`

The old action was `setLanguageTag(null)` to reset to system preferences. The new action is `setLanguageTag('SystemPreference')`. 

3. `useLocales()` is now called once, not twice

Previously `IntlContext` called `useLocales()` and `useAppLanguageTag` also called `useLocales()` internally — two calls per render, as was noted in the ticket. Now only `IntlContext` calls it (for building the message fallback chain). The store resolves the language tag synchronously at init via `getLocales()` (non-hook, one-time).

4. `useSystemPreferences` is retained in the store even though `languageTag` is always resolved

You could ask: why keep `useSystemPreferences: boolean` at all if the tag is always resolved? It's kept because `LanguageSettings.tsx` needs to know which radio button to check — specifically whether the "Follow system preferences" row is selected, independent of what the resolved tag happens to be. Without it you couldn't distinguish "user chose English" from "system preference resolved to English".

5. `LocaleStoreProvider` is now the context object directly

Changed from a wrapper function using `React.createElement` to export const `LocaleStoreProvider = LocaleContext`, matching the pattern noted to be preferred to be used in `AppProviders` (e.g. `UnitSystemStoreContext`). Is that right?

6. We have good End to End tests for this, so I just modified and added to the unit tests. Would you prefer I also add Integration tests? I wasn't sure if they were necessary for this change.


src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -1,24 +1,52 @@
...
+  };
+  if (legacy.languageTag !== null) {
+    return {
+      languageTag: legacy.languageTag,
+      useSystemPreferences: legacy.useSystemPreferences,
+    };
+  }
+  // null means "use system preferences" — resolve now
+  return createInitialState();
+}
```
ErikSin: We should be leveraging valibot to these migrations, that is why we are using them in the first place. See comment above, we should be keeping the previous valibot schemas and renaming them, and then using the valibot functions to make the migrations

eg:

```ts
  if (version === 1) {
    // Validate instead of blindly casting — catches silent data corruption
    return v.parse(LocaleStateSchema, persistedState);
  }
  // version 0 → version 1
  try {
    const legacy = v.parse(LocaleStateV0Schema, persistedState);
    if (legacy.languageTag !== null) {
      return { languageTag: legacy.languageTag, useSystemPreferences: legacy.useSystemPreferences };
    }
  } catch { noop() }
  return createInitialState();
}

```

Alternatively you could also use safeParse if you don't want to use a try catch.

I know this seems trivial compared to what you did, they are almost identical, your just manually checking the type and manually coercing the new value. But what your doing requires implicit developer knowledge of the old store type. Secondly if there is a more complicated type, the manual checking will get more complicated. And lastly if we were to do more migrations this would get increasingly complicated. Valibot just allows us to get the type of the store without knowing which version of the store we have to migrate from.

src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -57,7 +104,7 @@ export type LocaleStore = ReturnType<typeof createLocaleStore>;
 
 const LocaleContext = createContext<LocaleStore | null>(null);
 
-export const LocaleStoreProvider = LocaleContext.Provider;
+export const LocaleStoreProvider = LocaleContext;
```
ErikSin: instead of exporting a provider, lets just use the context directly in `<AppProviders/>` similar to what we did for the metric/imperial context

ErikSin (approved):
One small comment about the exported provider, but otherwise looks good!

src/frontend/contexts/LocaleStoreContext.tsx
cimigree: Yes sorry. Still getting used to this pattern. https://github.com/digidem/comapeo-mobile/pull/1902/commits/4800fd401a4000b9e0231984f7f1ce346dfaa80e

PR #1903, opened 2026-05-21
chore: refactor observation fields

### LINKED ISSUE #1882: Refactor Observation Fields
Currently the logic for Observation Fields is quite nested an unclear. We can simplify the logic by removing the `<Questions>` layer and having the props be agnostic (currently they all just accept `field` as a prop). 

# To Do
- [ ] Update each component to take a `value` and a `changeValue` instead of just taking a whole `field` as a prop
- [ ] Remove `<Question>` and directly copy and past the logic into `index`

ErikSin:
closes: #1882 

I refactored the Observation Fields to follow our new [guidelines](https://github.com/digidem/comapeo-mobile/blob/develop/docs/EthosAndGuidelines.md).

Previously the ObservationField screen had several nested components. Those nested components had several unnecessary props.

This refactors introduces only one level of nesting, and the nested components have agnostic props instead of passing the entire `field` object.

I also created several tests. A general integration test that mounts the entire navigation, and tests the observation fields in a normal user interaction. As well as individual tests for each of the smaller components


ErikSin:
im not sure why, but the front end integration test is failing on CI, but is not faiing locally. Can you try running the tests locally @cimigree 

image

cimigree:
> im not sure why, but the front end integration test is failing on CI, but is not faiing locally. Can you try running the tests locally @cimigree
> 
> image

Yes they pass for me too. I can look into it as I review

ErikSin:
> > im not sure why, but the front end integration test is failing on CI, but is not faiing locally. Can you try running the tests locally @cimigree
> > image
> 
> Yes they pass for me too. I can look into it as I review

OK! I think the problem is because we mock/create individual instances of the entire backend in our tests. Since we are doing that, we never install the the backend packages when running the front end CI tests. I was importing the categories for the tests via the backend packages. That is also why it was working locally, because we both have those packages installed locally.

To fix this, i just added the default categories as a dev dep to the front end, and updated the import in the tests

src/frontend/screens/ObservationFields/index.tsx
```diff
@@ -39,118 +45,111 @@ export const ObservationFields = ({
...
     });
-  }, [navigation, current, onBackPress]);
+  }, [navigation, current, formatMessage, fieldIds.length, observationId]);
 
-  const fieldId = preset?.fieldRefs.map(({docId}) => docId)[current - 1];
-  const field = fields.find(val => val.docId === fieldId);
+  const field = fields.find(val => val.docId === fieldIds[current - 1]);
 
   if (!field) {
+    // should throw error here
```
cimigree: probably should remove comment and throw error

src/frontend/screens/ObservationFields/index.tsx
```diff
@@ -39,118 +45,111 @@ export const ObservationFields = ({
...
+  const tagValue = tags?.[tagKey];
+
   return (
     <ScrollView style={{flex: 1}} testID="OBS.add-details-scrn">
-      <Question field={field} />
+      <View style={styles.labelContainer}>
+        <HeaderText variant="header3">
+          <FormattedFieldProp field={field} propName="label" />
+        </HeaderText>
+        {<HeaderText variant="header5">{field.helperText}</HeaderText>}
```
cimigree: Why is there a wrapper around this? Also, field.helpText can be undefined. So then an empty header text renders leaving a big space. See photo
Maybe do this instead?
`{field.helperText && <HeaderText variant="header5">{field.helperText}</HeaderText>}`

Image

src/frontend/screens/ObservationFields/index.tsx
```diff
@@ -39,118 +45,111 @@ export const ObservationFields = ({
...
-  }, [current, navigation]);
+  const {updateTag} = useDraftObservationActions();
+  const tags = useDraftObservationState(state => state.value?.tags);
 
   React.useLayoutEffect(() => {
     navigation.setOptions({
+      headerTitle: formatMessage(m.title, {current, total: fieldIds.length}),
       headerLeft: props => (
-        <CustomHeaderLeft headerBackButtonProps={props} onPress={onBackPress} />
+        <CustomHeaderLeft
```
cimigree: So now the android back button goes away from the details and back to the Observation screen (as far as I can tell) when it used to do the same thing as the header back button, that is go back one question. Is this your intention?

src/frontend/screens/ObservationFields/Number.tsx
```diff
@@ -1,29 +1,26 @@
...
+        testID="OBS.number-inp"
+        value={typeof tagValue === 'number' ? String(tagValue) : ''}
         onChangeText={newVal =>
           updateTag(
-            field.tagKey,
-            newVal
-              .replace(/[^0-9.-]/g, '') // Allow digits, decimal, and negative sign
-              .replace(/(?!^)-/g, '') // Remove any minus sign that is not at the start
-              .replace(/(\..*?)\./g, '$1'), // Remove additional decimal points
+            parseFloat(
```
cimigree: parseFloat('-'), parseFloat('.'), and parseFloat('-0.') all return NaN It would be good to put something here so incomplete input doesn't save NaN into the observation tags

src/frontend/screens/ObservationFields/Number.tsx
```diff
@@ -1,29 +1,26 @@
...
+export const Number = ({updateTag, tagValue}: NumberProps) => {
   return (
-    <React.Fragment>
-      <QuestionLabel field={field} />
+    <>
       <TextInput
-        testID="OBS.details-inp"
-        value={typeof value === 'string' ? value : ''}
+        testID="OBS.number-inp"
+        value={typeof tagValue === 'number' ? String(tagValue) : ''}
```
cimigree: If someone is editing an existing observation that has a number field, it will show a blank field because this value used to be saved as a string.

src/frontend/screens/ObservationFields/Number.tsx
cimigree: This field used to be stored as a string but now it will be stored as a float. That does make sense, but it is a change. Is that for sure intentional?

src/frontend/screens/ObservationFields/index.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    const animalCategory = await screen.findByText('Animal');
+    expect(animalCategory).toBeVisible();
+
+    await user.press(animalCategory);
+
+    const details = await screen.findByText('Details');
+    await user.press(details);
+
+    // through manually clicking through the app, I know that this should be a text input
```
cimigree: This code comment should be a little less conversational for posterity...Maybe something like the category type Animal's first field is a text input (animal-notes).

src/frontend/screens/ObservationFields/Number.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+test('removes extra decimal points', () => {
+  const updateTag = jest.fn();
+
+  render(<Number updateTag={updateTag} tagValue={undefined} />);
+
+  fireEvent.changeText(screen.getByDisplayValue(''), '1.2.3');
+
+  expect(updateTag).toHaveBeenCalledWith(1.23);
+});
```
cimigree, line 102: If you had added fireEvent.changeText(screen.getByDisplayValue(''), '-');
  fireEvent.changeText(screen.getByDisplayValue(''), '.');
It would fail... Once you fix that, you can add that test!

cimigree (changes requested):
Nice refactor!
Applies the guidelines, as you said.
Sub-components now take agnostic props, logic lives at the top level, 
`<Question>` confusion is gone
Lots of great new integration tests.
**Cautions**
We will need to make sure that if the default-categories update, we update them in our frontend, i.e., keep them in sync manually.
The tests rely on real categories, which is probably fine, but if they change, the test will need to change.

**Questions/ Cautions**
The Android hardware back button goes back to the Observation screen instead of the previous question. Is that intentional?
Number now stores a float instead of a string. Are you sure that is merited?

**Must Change**
I am worried about someone entering incomplete info in the number field accidentally. Just a negative sign, or just a point (.) or -0. and it will be not a number. I can't test what actually happens because we don't have any number fields in the default categories, but will it crash the app? It definitely fails an integration test. It can be easily guarded against with `isFinite`(and check the input).
And then you can add a test to check those inputs.

And other little changes, see below.

src/frontend/screens/ObservationFields/Number.test.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1903/commits/33d8f7d92132158ce7a635f3b41ee1f955d36fcc

src/frontend/screens/ObservationFields/Number.tsx
ErikSin: handled now. if the value is a string (that can be parsed as a number) it will show the value. and then if they update the value, that string will be updated to a number. https://github.com/digidem/comapeo-mobile/pull/1903/commits/33d8f7d92132158ce7a635f3b41ee1f955d36fcc

src/frontend/screens/ObservationFields/Number.tsx
ErikSin: see comment below!

src/frontend/screens/ObservationFields/Number.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1903/commits/33d8f7d92132158ce7a635f3b41ee1f955d36fcc

src/frontend/screens/ObservationFields/index.tsx
ErikSin: You know what...i didn't realize that these were superflous wrapper. I think they were created in legacy mapeo with an attempt to guarantee a label, when labels were optional. but now labels are mandatory, so i think i can just get rid of the wrapper completely

src/frontend/screens/ObservationFields/Number.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1903/commits/e95471dc3e2ec98b7d295dc0cf94014536ad2f06 and https://github.com/digidem/comapeo-mobile/pull/1903/commits/33d8f7d92132158ce7a635f3b41ee1f955d36fcc

ErikSin: Wow number are annoying. I added so many tests to catch the edge cases! But also the tests made it really easy to find edge cases and determine if it breaks things!

src/frontend/screens/ObservationFields/index.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1903/commits/09e26cb0917c9db9b0ebb1c43fe182b0c8dc5eae

src/frontend/screens/ObservationFields/index.tsx
ErikSin: i removed all the wrappers entirely, which means i had to refactor a couple more pages. They were a remnant of the mapeo and they were really convoluted.

https://github.com/digidem/comapeo-mobile/pull/1903/commits/5fb9afb965dce900756f113f7a77801b7f9dd809

I also wrote several tests related to showing these details in an observation

cimigree (changes requested):
Wow this is quite a refactor!
With the new config I was able to test numbers!
One thing that worked was using the old version of a number (that was a string) and then switching branches to this new version of number (that is a number). I couldn't even tell it was processing. All worked as expected.

BUT, 
I am having 2 issues with numbers.
1. is that I cannot enter negative numbers using the keyboard given. Not sure if that is a problem. I can paste them in... Not sure if that is expected. So regex makes sense to keep but wondering about use case and needing negative numbers. It may be related to the numeric keyboard type. But I am not sure.
2. If I have a number in place and then I erase it so that there is no number (because I want to have "no answer") then it doesn't work. The first digit from the previous entry is kept and cannot be cleared. I can type 0 as a work around but I am not sure how intuitive that is and may be misleading or not what a user wants. See video.


https://github.com/user-attachments/assets/d7d2127d-98f0-4ead-96af-ff15d749989a



Also, my local jest tests are failing.
```
 FAIL  src/frontend/screens/ComapeoSettings/index.test.tsx (5.355 s)
  ● CoMapeo Settings Screen › opens drawer when header button is pressed

    TypeError: (0 , _reactNative.renderAsync) is not a function

      49 |     onTeardown.push(appProviders.teardown);
```
And 
 ```
FAIL  src/frontend/screens/ObservationFields/index.test.tsx
  ● Test suite failed to run

    Cannot find module '@comapeo/default-categories/package.json' from 'src/frontend/screens/ObservationFields/index.test.tsx'

      4 |
      5 | const DEFAULT_CONFIG_PATH = path.join(
    > 6 |   path.dirname(require.resolve('@comapeo/default-categories/package.json')),
        |                        ^
      7 |   'dist/comapeo-default-categories.comapeocat',
      8 | );
      9 |

      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.resolve (src/frontend/screens/ObservationFields/index.test.tsx:6:24)

```
And a bunch of tests on Browserstack are failing. Not sure if that is flakiness or the tests because it is Observations that are failing.

**Comment by ErikSin** [comment]
@cimigree this number issue seems device specific. I have updated the pr to be similar to the previous implementation. Can you try using numbers on `develop` branch, and then try using numbers on this branch. If they both have the same problem, then lets just merge this pr and open an issue

File: src/frontend/screens/ObservationFields/Number.tsx [hunk 8]
```diff
@@ -1,31 +1,40 @@
...
-          )
-        }
+        testID="OBS.number-inp"
+        value={text}
+        onChangeText={newVal => {
+          const sanitized = newVal
+            .replace(/[^0-9.-]/g, '') // Allow digits, decimal, and negative sign
+            .replace(/(?!^)-/g, '') // Remove any minus sign that is not at the start
+            .replace(/(\..*?)\./g, '$1') // Remove additional decimal points
+            // .replace(/^(-?)\./, '$10.') // Prefix bare decimal with 0
```
cimigree: I think this line that you commented out was the one that was breaking the functionality on the Samsung. Can you just delete it instead of leaving it commented out?

cimigree (approved):
It actually works fine on develop and on Samsung. Seems like all is fixed. I tested all manually though I see there are so many new and thorough tests. Way to go!

One non blocking comment!

PR #1907, opened 2026-05-26
fix: update scale bar with zoom

### LINKED ISSUE #1906: Scale Bar not updating
The scale bar on the map is not reacting to the zoom level, and therefore is never updating.

This happened in #1736. Previously the map component exposed the zoom level, and that exposed zoom level was passed to the scale bar. In 1736 I accidentally took out the logic that exposed that zoom level.

ErikSin:
closes #1906 

I don't think there is an easy way to test this, neither e2e or integration test. But luckily it is easy to test IRL:

https://github.com/user-attachments/assets/1fc38392-3598-4e9c-bac9-982e58774d4a

### DIFF
--- a/src/frontend/screens/MapScreen/index.tsx
+++ b/src/frontend/screens/MapScreen/index.tsx
@@ -154,6 +154,9 @@ export const MapScreen = ({
             if (event.properties.isUserInteraction && following) {
               setFollowing(false);
             }
+          }}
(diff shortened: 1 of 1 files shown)


cimigree (approved):
Works great. Very responsive. I checked on an offline background map that I had uploaded and all worked well.

PR #1909, opened 2026-05-27
chore: simplify intl.ts structure

ErikSin:
This is towards #1881.

#1902 started to address this, but it just revealed the tangle of code that is intl.

First, the intl.ts was exporting the type `SupportedLanguageTag` which was ALL possible language tags, including the languages that we did not support. This was being used as the language tag type in the intlProvider. 

We then were also exporting type `TranslatedLanguageTag` which are the only the language tag which our app supports.

These names are a little misleading as the TranslatedLanguageTag at the tags the app supports, while the `SupportedLanguageTag` is just ALL language tags.

As well, since the intlProvider had the type of SupportedLanguageTag, we had to check that that the language tag returned by the intlProvider was a language tag supported by the app everytime we used it. 

In order to simplify this, i got rid of SupportedLanguageTag entirely. Now the intl provider is of type `TranslatedLanguageTag`. So now we dont need to check everytime we use it because typescript guarantees it is of the right type.

I also updated the store migration to use valibot validation when migrating.

Lastly i added a onRehydrateCheck in zustand. The reason for this i that `TranslatedLanguageTag` can change. If for some reason the app no longer supportes a language the user was using, we just set the app language to system preference.

### DIFF
--- a/src/frontend/contexts/LocaleStoreContext.test.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.test.tsx
@@ -39,7 +39,7 @@ test('usage of state and actions hooks', () => {
 
   expect(stateHook.result.current).toStrictEqual({
     useSystemPreferences: false,
-    languageTag: 'pt-BR',
+    languageTag: 'pt',
   });
 
   act(() => {
--- a/src/frontend/contexts/LocaleStoreContext.test.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.test.tsx
@@ -34,7 +34,7 @@ test('usage of state and actions hooks', () => {
   });
 
   act(() => {
-    actionsHook.result.current.setLanguageTag('pt-BR');
+    actionsHook.result.current.setLanguageTag('pt');
   });
 
   expect(stateHook.result.current).toStrictEqual({
--- a/src/frontend/contexts/LocaleStoreContext.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.tsx
@@ -9,7 +9,7 @@ import {
 import {MMKVStoreInitializer} from '../hooks/persistedState/createPersistedState';
 import {
   getUsableLanguageTagFromSystemPreferences,
-  SupportedLanguageTag,
+  TranslatedLanguageTag,
 } from '../lib/intl';
 import {type LocaleState} from '../sharedTypes/locale';
 
@@ -34,6 +34,7 @@ function migrate(persistedState: unknown, version: number): LocaleState {
   }
   // version 0: languageTag was null | SupportedLanguageTag
   const legacy = persistedState as {
+    // @ts-expect-error will change
     languageTag: SupportedLanguageTag | null;
     useSystemPreferences: boolean;
   };
@@ -65,7 +66,7 @@ export function createLocaleStore({persist} = {persist: false}) {
 
   const actions = {
     setLanguageTag: (
-      languageTag: SupportedLanguageTag | 'SystemPreference',
+      languageTag: TranslatedLanguageTag | 'SystemPreference',
     ) => {
       if (languageTag === 'SystemPreference') {
         const systemLanguageTags = getLocales().map(l => l.languageTag);
--- a/src/frontend/contexts/LocaleStoreContext.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.tsx
@@ -61,6 +61,16 @@ export function createLocaleStore({persist} = {persist: false}) {
         storage: createJSONStorage(() => MMKVStoreInitializer),
         version: 1,
         migrate,
+        // usable language tags can change between app boot ups
+        // this just makes sure the language tag is still usable
+        // if not, then use system preferences
+        onRehydrateStorage: () => state => {
+          if (!state) return;
+          const usable = getUsableLanguageTag(state.languageTag);
+          if (!usable) {
+            store.setState(createInitialState());
+          }
+        },
       }),
     );
   } else {
--- a/src/frontend/contexts/LocaleStoreContext.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.tsx
@@ -11,7 +11,7 @@ import {MMKVStoreInitializer} from '../hooks/persistedState/createPersistedState
 import {
   getUsableLanguageTag,
   getUsableLanguageTagFromSystemPreferences,
-  TranslatedLanguageTag,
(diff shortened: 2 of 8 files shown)


src/frontend/contexts/LocaleStoreContext.test.tsx
```diff
@@ -34,7 +34,7 @@ test('usage of state and actions hooks', () => {
   });
 
   act(() => {
-    actionsHook.result.current.setLanguageTag('pt-BR');
+    actionsHook.result.current.setLanguageTag('pt');
```
ErikSin, line 37: As seem by this test, now it is just not possible to set a language tag that is not supported

src/frontend/sharedTypes/locale.ts
```diff
@@ -1,17 +1,44 @@
...
+import LANGUAGES from '../languages.json';
+import MESSAGES from '../../../translations/messages.json';
+
+type LegacySupportedLanguageTag = keyof typeof LANGUAGES;
+
+function isSupportedLanguageTag(
+  value: string,
+): value is LegacySupportedLanguageTag {
+  return value in LANGUAGES;
+}
```
ErikSin: this is purely to keep `LocaleStateSchemaV0` the same as before

src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -28,22 +34,20 @@ function createInitialState(): LocaleState {
  * Migrates persisted state from version 0 (where languageTag could be null)
  * to version 1 (where languageTag is always resolved).
  */
-function migrate(persistedState: unknown, version: number): LocaleState {
-  if (version === 1) {
-    return persistedState as LocaleState;
+function migrate(persistedState: unknown): LocaleState {
```
cimigree, line 37: Hm. You decided not to use the Zustand convention of version number in the migrate function, I see and are instead checking against versions of Schema that you create.
This works today, and is fine and is not something you need to change, but I am wondering what happens if another version gets added ? I think you'd need to add another schema rather than just adding a version number. This is a very defensive approach but I am wondering why you don't trust the version number? Is there a reason you don't trust the version number within the migrate function, given the rest of the codebase uses version checking (I think)?

src/frontend/lib/intl.ts
```diff
@@ -3,87 +3,38 @@ import MESSAGES from '../../../translations/messages.json';
...
- *
- * 1. They have at least one translated string
- * 2. They match a language that is found in the [supported languages file](../languages.json)
+ * Resolves a language tag from a list of system-preferred language tags.
+ * Falls back to `'en'` if no supported tag can be found.
  *
- * @param translatedLanguageTags List of language tags that may have translated messages
- * @returns List of languages that are usable within the app (see {@link UsableLanguage})
+ * @param systemLanguageTags Ordered list of language tags from system preferences (most preferred first)
+ * @returns A resolved SupportedLanguageTag
```
cimigree: SupportedLanguageTag no longer exists — should be AvailableLanguageTag.

src/frontend/sharedTypes/locale.ts
```diff
@@ -1,17 +1,41 @@
...
 
-import {isSupportedLanguageTag, SupportedLanguageTag} from '../lib/intl';
+import {AvailableLanguageTag} from '../lib/intl';
+
+import LANGUAGES from '../languages.json';
+import MESSAGES from '../../../translations/messages.json';
+
+type LegacySupportedLanguageTag = keyof typeof LANGUAGES;
+
+function isSupportedLanguageTag(
```
cimigree: You have the Github comment about isSupportedLanguageTag and (I assume -- LegacySupportedLanguageTag) —  only here to support LocaleStateSchemaV0. But that won't be in the codebase. So, probably worth a comment explaining that, since it's confusing to have a type checking against all the languages.json in a file that otherwise only cares about messages.json (the way fewer available ones).

src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -28,22 +34,20 @@ function createInitialState(): LocaleState {
  * Migrates persisted state from version 0 (where languageTag could be null)
```
cimigree, line 34: This comment is no longer relevant and should be rewritten since we are not using version numbers here any more.

cimigree (approved):
The code all looks good and makes sense but I just wanted to test thoroughly because I could not really understand why it had to be so complex before. I was wondering if I was missing something. 

So, I manually tested: fresh install, language selection, force close persistence, switching back to system preference, and then force close again. Migration with a chosen language, migration with a selected language, and the onRehydrateStorage path by temporarily removing a language from messages.json. And it all worked!

Mostly comments about comments and one question.

src/frontend/contexts/LocaleStoreContext.tsx
ErikSin: I think this comment is still relevant. We are still migrating from version 0 to version 1 even if i am not using the version number explicitly in the migration

src/frontend/contexts/LocaleStoreContext.tsx
ErikSin: The reason we started using valibot is to guarantee that the store is in the right shape, and by extension make sure that data isnt corrupted in migration.

I think it is a little hard to see in this case because the store is so simple, but valibot gets rid of implicit knowledge and easy developer mistakes. We could just check if the `languageTag` is null, and then manually coerce it to an actual value. But doing that requires implicit developer knowledge of the old store type. And if there is a more complicated type, the manual checking will get more complicated. Also the more versions we add, the more room for error gets added as we have to do more than one manual conversion. By keeping the old valibot schemas around we get rid of the implicit knowledge, as we are explictly using the same type. And by using valibot, there is much less room for developer error. 

As far as the actual version number, I didn't use it in the case because using valibot( for the reasons above) made the version number unnecessary, since there were only 2 types. I knew that if it was not the current schema, then it had to be the previous schema, so no point in using the version number. In other words, I wasn't explicitly not using the version number. If there were more than 2 version and several valibot schemas I definitely would use the version number to keep track of it all.

cimigree: Makes sense. Thanks for explaining.

src/frontend/contexts/LocaleStoreContext.tsx
cimigree: Fair point.

Issue #1872, opened 2026-04-30
Observation Photos in Portrait Mode landscaped - p2


ximenabb:
**Describe the bug**
Issue https://github.com/digidem/comapeo-mobile/issues/1675 not fixed

PR #1915, opened 2026-06-01
fix: photo orientation

cimigree:
towards #1872 
Turns out we cannot update to `react-native-vision-camera` version 5 without the new architecture. This is not well documented and took me a while to figure out. 

So this fix is attempting to override whatever EXIF orientation vision camera wrote (correct or not) with this command: `await Exify.write(raw.uri, {Orientation: 1}`;

Then after that we apply our home grown rotation using the accelerometer.

The `manipulateAsync` API got deprecated so now I made a wrapper/ helper for the suggested replacement (`ImageManipulator.manipulate`) since we use that behavior 3 times.

Also, in order to display the photo in the correct orientation, we have to calculate the layout using the accelerometer data 

At this point, this fix works for my three devices (two Samsungs and a Pixel running Android 15 & 16) **but not for a Motorola (Moto G4) testing device running Android 14**.

 It would be helpful to know why, so an idea is to see what the Exif value is for four different scenarios (portrait upright, landscape right, portrait upside down, landscape left). Then we can ascertain if possibly vision camera is writing wrong EXIF on that device. If not, we are stuck and have to wait for upgrade to vision camera 5.

I have added _temporary_ Sentry logging in case we want to debug this further so that we can see what the Exif Orientation is for a tester since different devices could possibly be using different values.

We could do an apk build and check Sentry for the Motorola to see what might be going on..

### DIFF
--- a/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
+++ b/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
@@ -161,7 +161,8 @@ export function createDraftObservationStore({persist}: {persist: boolean}) {
    * Only processes steps that are not already complete.
    */
   async function processPhoto(attachment: UnsavedPhotoAttachment) {
-    const {id, raw, original, thumbnail, preview, accelerometer} = attachment;
+    const {id, raw, original, thumbnail, preview, accelerometer, photoExif} =
+      attachment;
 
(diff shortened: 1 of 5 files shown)


ErikSin:
closing as we are descoping until we switch to the new architecture

PR #1927, opened 2026-06-04
chore: automates play store submission and upload to github releases

### LINKED ISSUE #1177: Update Github Actions/EAS actions to automate Publishing of App
Update the EAS flow and github actions to automatically publish and submit a release build. It should publish the APK to github and update the [link on our website](https://awana.digital/comapeo) with this new APK. As well it should also submit the app to the play store for review

To Do
- [x] Update eas commands to build both an AAB and APK
- [ ] [Automate Submission to the Play Store](https://docs.expo.dev/build/automate-submissions/)
- [ ] Automate updating the APK on github
- [x] Automate updating the APK on our website

cimigree:
closes #1177
### Summary

- Updates eas.json with submit config for Google Play
- Adds two new jobs to the Build Release workflow:
   1. GitHub Release — automatically creates a GitHub Release with the APK (downloaded from R2) and release notes attached if they exist, every time a release PR is merged
   2. Play Store submission — automatically submits the production AAB to the Play Store via eas submit after every release build

### Testing
I need to merge this to develop to test it. The workflow has a manual trigger (workflow_dispatch) that lets me run just the GitHub Release job with a specific build ID and version — without triggering a real release or touching the Play Store. That is a much more complicated piece of code and doesn't hurt much if we have to test it out a few times, unlike the play store submission. 

**My plan once merged**: run it manually with the v10.5 hotfix build. This gives us two things at once:

1. Confirms the workflow actually works
2. Gets v10.5 onto our GitHub releases page

@ErikSin  — Can you review this untested so I can merge this so I can run the test? The workflow_dispatch trigger only appears in the Actions tab once the workflow is on develop, unfortunately.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -29,10 +29,6 @@ on:
         description: 'EAS build ID (UUID from expo.dev build URL)'
         required: true
         type: string
-      version:
-        description: 'Release version short form (e.g. 10.5)'
-        required: true
-        type: string
 
 jobs:
   build:
@@ -106,8 +102,6 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request' && github.event.pull_request.merged == true
     needs: build
-    outputs:
-      release_version_short: ${{ steps.version.outputs.release_version_short }}
 
     steps:
       - name: Create GitHub App Token
@@ -149,7 +143,6 @@ jobs:
         id: inputs
         run: |
           echo "build_id=${{ inputs.build_id || needs.build.outputs.build_id }}" >> $GITHUB_OUTPUT
-          echo "version=${{ inputs.version || needs.tag.outputs.release_version_short }}" >> $GITHUB_OUTPUT
 
       - name: Create GitHub App Token
         uses: actions/create-github-app-token@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/build-release.yml
```diff
@@ -21,14 +23,25 @@ on:
...
+  workflow_dispatch:
+    inputs:
+      build_id:
+        description: 'EAS build ID (UUID from expo.dev build URL)'
+        required: true
+        type: string
+      version:
+        description: 'Release version short form (e.g. 10.5)'
+        required: true
+        type: string
```
ErikSin: lets get rid of this input and extract the version from EAS directly!

After the `Wait for EAS build to finish`

we can write this action:

```yml
-name: Extract version from EAS build
        id: eas_version
        env:
          BUILD_ID: ${{ steps.inputs.outputs.build_id }}
        run: |
          version=$(eas build:view "$BUILD_ID" --json | jq -r '.runtimeVersion')
          echo "version=$version" >> $GITHUB_OUTPUT

```

And then anywhere you have `steps.inputs.outputs.version` replace with `steps.eas_version.outputs.version`. and you can get rid of line 152.

Your original code allows the developer to type in the wrong version that may not match with the version of the apk. So these additions avoid that.

ErikSin (approved):
Since the workflow_dispatch flow is for testing, i am approving so we can try it out.

But there is one comment that we should try and address first

.github/workflows/build-release.yml
cimigree: https://github.com/digidem/comapeo-mobile/pull/1927/commits/45d2c7235c9f045c57d036ef57bc6f3e71dccf92

PR #1929, opened 2026-06-04
chore: precompile translations to AST

### LINKED ISSUE #539: Potentially address warning from  `@formatjs/intl` about not pre-compiling messages

### LINKED ISSUE #1879: Precompile translations into AST
https://formatjs.github.io/docs/guides/advanced-usage

ErikSin:
closes #1879 
closes #539

Also Lazy load translations so they are not all initially loaded in memory. Uses useQuery to deal with the async nature of dynamically loading the translations.

I also had to update the tsconfig to  `"module": "esnext"` as it was throwing an error with the dynamic imports


jest.setup.js
achou11, line 1: Mind explaining why the changes here are necessary? It's not immediately clear to me... (aside from tests failing without the change)

src/frontend/contexts/IntlContext.tsx
```diff
@@ -2,10 +2,13 @@ import * as React from 'react';
 import {IntlProvider as ReactIntlProvider, CustomFormats} from 'react-intl';
 import {StyleSheet, Text} from 'react-native';
 import {useLocales} from 'expo-localization';
+import {useQuery, keepPreviousData} from '@tanstack/react-query';
+import type {MessageFormatElement} from '@formatjs/icu-messageformat-parser';
```
achou11: Probably preferable to use the export from `react-intl` instead:

```suggestion
import type {MessageFormatElement} from 'react-intl';
```

scripts/build-translations.mjs
achou11, line 1: I would suggest updating this script to use the official formatjs tooling for compilation.

Implemented an example alternative script that maintains existing behavior but uses `@formatjs/cli-lib`: https://github.com/digidem/comapeo-mobile/commit/86219bccfc2e98e2d3c3fce6fc0dfdc117c74aaa

scripts/build-translations.mjs
```diff
@@ -3,17 +3,13 @@
 import path from 'node:path';
 import fs from 'node:fs';
 import {readFile, writeFile} from 'node:fs/promises';
+import {parse} from '@formatjs/icu-messageformat-parser';
```
achou11: see comment about updating this file as a whole. seeing usage of this module - which is pretty low-level and probably not meant for direct usage - raised some alarms for me.

src/frontend/contexts/IntlContext.tsx
```diff
@@ -29,20 +32,41 @@ export const IntlProvider = ({children}: {children: React.ReactNode}) => {
...
+      const merged: Record<string, MessageFormatElement[]> = {};
+      for (const msgs of results) {
+        Object.assign(merged, msgs);
+      }
+      return merged;
+    },
+    staleTime: Infinity,
+    // see: https://tanstack.com/query/latest/docs/react/guides/paginated-queries#better-paginated-queries-with-placeholderdata
+    placeholderData: keepPreviousData,
+  });
```
achou11: While this works, I think it'd be more idiomatic to use [`useQueries`](https://tanstack.com/query/latest/docs/framework/react/reference/useQueries) here. The main reasoning is to isolate each language with their own query to avoid unnecessary work when they change.

Implemented an example here: https://github.com/digidem/comapeo-mobile/commit/d7b11353deaa2642a17ea7717f82bc3cee304efe (note that it might need some adjustments...)

scripts/build-translations.mjs
```diff
@@ -22,13 +18,34 @@ async function run() {
...
+  );
+
+  const importLines = locales
+    .map(lang => `  ${lang}: () => import('./${lang}.json'),`)
+    .join('\n');
 
   await writeFile(
-    TRANSLATIONS_OUTPUT_PATH,
-    JSON.stringify(translations, null, 2),
+    path.join(TRANSLATIONS_DIR_PATH, 'index.ts'),
```
achou11: thoughts on this being located within `src/frontend/` and git checked? i've found it useful to have such files tracked by git to provide visibility when it changes (whether expected or not).

scripts/build-intl-polyfills.mjs
```diff
@@ -5,7 +5,17 @@ import {fileURLToPath} from 'node:url';
 import path from 'node:path';
 
 import languages from '../src/frontend/languages.json' with {type: 'json'};
-import messages from '../translations/messages.json' with {type: 'json'};
+
+const TRANSLATIONS_DIR = new URL('../translations/', import.meta.url).pathname;
```
achou11: Need to update this in order to be used on a non-posix OS.

```suggestion
const TRANSLATIONS_DIR = fileURLToPath(new URL('../translations/', import.meta.url));
```

achou11 (changes requested):
Functionality seems to work as intended, but think there's room to clean up the implementation.

I verified that dynamic imports are working by checking using expo atlas (see https://docs.expo.dev/guides/analyzing-bundles/).

I was trying to test changes to the system settings using an emulator but seems like `useLocales()` doesn't update when that occurs. Might be a dev-specific issue, as it doesn't work on dev branch either (but works with built APK). Would be good to build an apk and test that changing the language via phone settings has expected results in the app.

scripts/build-translations.mjs
achou11: on second thought, might not be ideal to have this git tracked. Updates to the messages directories are done via Crowdin PRs which would mean that you'd need a way to check if this file changes due to the updates.

scripts/build-intl-polyfills.mjs
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1929/commits/b3b31914ecf59d99de6fd5607b383fd9453cd2dc

scripts/build-translations.mjs
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1929/commits/3eac340c9037b020be22ab68f001220254756a6b

scripts/build-translations.mjs
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1929/commits/3eac340c9037b020be22ab68f001220254756a6b

src/frontend/contexts/IntlContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1929/commits/73838b5baa22a6a477837ed0699bbbee2d78513d

src/frontend/contexts/IntlContext.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/1929/commits/51c428eb8db9262298d3733911cc64b295b05763

jest.setup.js
ErikSin: jest doesn't support esm dynamic importing out of the box. So im just resolving the promise before hand in the jest environment (so it is no longer a dynamic import). We also dont need any other languages besides english in our jest tests, so i was only resolving english. But on reflection i'm just going to resolve all the languages in case we need them for future tests

scripts/build-translations.mjs
```diff
@@ -1,110 +1,112 @@
...
-    const lang = entry.name.split('-')[0];
-    if (result[lang]) {
-      throw new Error(
-        `Duplicate language code '${lang}': folders '${langToFolder[lang]}' and '${entry.name}' both map to the same code`,
-      );
+    if (
+      !Object.prototype.hasOwnProperty.call(
+        LANGUAGE_NAME_TRANSLATIONS,
+        languageCode,
+      )
```
achou11: Very subjective: this feels more readable at a glance and achieves the same thing:

```suggestion
    if (!(languageCode in LANGUAGE_NAME_TRANSLATIONS)) {
```

src/frontend/hooks/useLanguageQueries.ts
```diff
@@ -0,0 +1,36 @@
+import {useQueries, keepPreviousData} from '@tanstack/react-query';
+import {AvailableLanguageTag, localeImports} from '../../../translations/index';
+import {MessageFormatElement} from 'react-intl';
+
+/**
+ *
+ * @description The tanstack query option `placeholderData` is set to `keepPreviousData`. This means this query will only ever be `pending` on initial load
+ */
+export function useLanguageQueries(languageCodes: AvailableLanguageTag[]) {
```
achou11, line 9: defining a separate hook that just wraps a query hook seems unnecessary right now. especially since this only used in one place, I would recommend inlining it where it's used.

If you have a need for reusing it elsewhere, usually the recommended approach is to define a query options object and passing that into the relevant query hook in the calling component (see https://tkdodo.eu/blog/the-query-options-api).

achou11 (approved):
Couple of non-blocking suggestions but otherwise changes look good

PR #1930, opened 2026-06-08
chore: adds eslint rules for unused messages and duplicate keys

### LINKED ISSUE #1891: Create Eslint rule to catch unused message descriptors
Create a cutom eslint rule that throws an error when there are any unused message descriptors. 

Also if possible, try to create an eslint rule that check is any of the message descriptors ids are identical. This might not be fully possible, as eslint usually only compares the code in a singular file. And in order to do this it would have to check all message descriptor ids in all file - which might be a possibility. If it is possible, than do that. Otherwise just scope it to a single file and check that there are no identical ids in that one file.

cimigree:
closes #1891 
- Adds a folder for eslint rules. 
    - Adds a file in there to create rules for international (intl) messages. 
    - The files has two rules.
    - Within each individual file, it checks for unused messages and duplicate keys.
        -  If the file has a message defined by being looked up at runtime (see `src/frontend/screens/LocationInfoScreen.tsx`) there is no way for eslint to tell what has been used and unused, so then everything in that file passes.
        - If the file exports all of the messages, (see `src/frontend/screens/Onboarding/DataPrivacyMessages.ts`) there is no way with eslint to tell if the message is used, or unused, so that all passes.
    - Adds a test for that file that will run with the jest tests

- Adds a "throws" in the extract-messages script so that if there are duplicate keys with different values (the defaultMessage or description part), it will let the developer know! Right now, the first message with the duplicate key is just overwritten by the second.

- Creates a plugin for the custom EsLint rules
- Fixes the violations of the no unused messages.

- I got a lot of help from this post: https://neciudan.dev/master-eslint-rules and this site: https://astexplorer.net/


achou11:
Sorry for the drive-by comment, but worth noting that formatjs has an official eslint plugin: https://formatjs.github.io/docs/tooling/linter.

It probably doesn't cover everything you're trying to do but it's probably a good idea to integrate it regardless (if you think the rules it has are useful).

EDIT: Looking closer at the original issue, what you're trying to achieve is probably not covered by the official plugin. Just caught my eye because having to implement and maintain a custom eslint rule doesn't sound like a fun thing to maintain...

eslint-rules/intl.js
```diff
@@ -0,0 +1,155 @@
+// @ts-nocheck
```
ErikSin: by default `js` files have `// @ts-nocheck` so this is redundant.

However,  I think it would be good to use `// @ts-check` so that we can get typescript error checking in a js file.

You can use `/** @type {import('eslint').Rule.RuleModule} */` on your functions to type the `context` and `node` so you dont get ts errors

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1930/commits/7e95dcd262ca28deb5d596771fe2327c56b66f94

ErikSin (approved):
Thanks for doing that, excited to have this rule!

PR #1935, opened 2026-06-10
fix: guard audio recorder against use after teardown (COMAPEO-1Z7)

gmaclennan:
## Problem

Sentry [COMAPEO-1Z7](https://awana-digital.sentry.io/issues/7341419707/) — `Call to function 'AudioRecorder.record' has been rejected. → Caused by: Cannot use shared object that was already released` (127 events / 127 users).

The crash is at `recorder.record()` in `useAudioRecording.startRecording`:

```ts
  await recorder.prepareToRecordAsync();
  recorder.record();          // <-- crash
}, [recorder]);
```

`AudioRecorder` is a `SharedObject` created via expo-audio's `useReleasingSharedObject`, which calls `.release()` on the native object when the hook unmounts. `prepareToRecordAsync()` takes a moment on a real device, so if the recording screen is left during that await — i.e. **starting then leaving in quick succession** — the recorder is released before the awaited promise settles, and the subsequent `record()` runs against a released shared object and crashes.

This window is reachable in normal use: `usePreventBackButtonWhileRecording` only blocks back-navigation once recording has started (`shouldPrevent: !isLoading`), so the user can leave the screen during the prepare/loading phase.

## Fix

Track whether the hook is still mounted and skip `record()` if it was torn down during the `prepareToRecordAsync` await.

## Test

`useAudioRecording.test.tsx` reproduces the timing window at the hook level (an e2e test would be flaky for a race this tight): a deferred `prepareToRecordAsync` is started, the hook is unmounted, then the promise resolves — and we assert `record()` is **not** called on the released recorder. A second test asserts the happy path still records. Verified the regression test fails without the guard and passes with it.

### DIFF
--- a/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
+++ b/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
@@ -1,4 +1,4 @@
-import {useCallback} from 'react';
+import {useCallback, useEffect, useRef} from 'react';
 import {
   useAudioRecorder,
   useAudioRecorderState,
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -10,9 +10,21 @@ const RECORDING_OPTIONS = RecordingPresets.HIGH_QUALITY!;
 export function useAudioRecording() {
   const recorder = useAudioRecorder(RECORDING_OPTIONS);
   const status = useAudioRecorderState(recorder, 100);
+  const isMountedRef = useRef(true);
+
+  useEffect(() => {
+    isMountedRef.current = true;
```
cimigree, line 16: Not blocking, but this assignment is redundant as the ref is initialized to true above already, isn't it? And can be removed, right?

cimigree (approved):
One small non-blocking comment. 
I think this makes sense from a UI perspective because it is sort of unclear that the recording will start right away so I can imagine people backing out quickly. I can't do it on my phone, but I imagine it is possible on slower phones as we see a bunch of occurrences of the error.
I am glad to have the tests because as I said, I cannot reproduce. And good to have the happy path anyway (which I can reproduce. All seems to work fine still).

PR #1938, opened 2026-06-11
fix: reset nav stack after leaving a project so its screens unmount

### LINKED ISSUE #1937: Leave Project is causing failed refetches due to mounted screens in stack
### Description

 When leaving a project, `LeaveProject.tsx` navigates with `navigation.replace('LeftProjectConfirmation')`, which only swaps the top screen. The screens beneath it - `YourTeam` (with `useManyMembers`) and `CollaboratorInfo` (with `useSingleMember` for the user's own device) - stay mounted with active queries scoped to the project that was just left. Any refetch triggered by navigation focus or a reconnect event will fail (because the left project is now closed).

Additionally, the active project is only switched *after* navigating, and only if the default project was found in the (possibly not-yet-loaded) `useManyProjects` data — if not found, the left project remains active and Home will read from its closed datastores.

Related: A bug in `@comapeo/core-react` was refetching all queries for a left project on every leave, see digidem/comapeo-core-react#182 / digidem/comapeo-core-react#183.

 ### Sentry

  - [COMAPEO-1TK](https://awana-digital.sentry.io/issues/7072609149/) — `Error: Cannot await idle after closing` via `$member.getMany()` (186 events / 15 users)
  - [COMAPEO-223](https://awana-digital.sentry.io/issues/7434500086/), [COMAPEO-20E](https://awana-digital.sentry.io/issues/7383452114/) — same error via `$member.getById(ownDeviceId)`, surfaced as `MissingOwnDeviceInfoError`

Breadcrumbs show the identical sequence across users: YourTeam → CollaboratorInfo → LeaveProject → "Yes, Leave" → errors → LeftProjectConfirmation. I don't think these errors are visible to the user (the user has already navigated to the confirmation screen), but worth fixing to avoid the risk.

  ### Proposed fix

  In `LeaveProject.tsx`'s leave `onSuccess`:
  1. Set the new active project **before** navigating.
  2. Use `CommonActions.reset` to `[Home, LeftProjectConfirmation]` (the same pattern as `ProjectRemovalListener`) instead of `navigation.replace`, so no screen holding queries for the left project stays mounted.

gmaclennan:
Fixes #1937

## Problem

After a successful Leave Project, `navigation.replace('LeftProjectConfirmation')` only swapped the top screen. `YourTeam` and `CollaboratorInfo` stayed mounted beneath the confirmation screen, holding active queries scoped to the left project. Any refetch of those queries fails with an error (because the left project is closed) resulting in a burst of  `Cannot await idle after closing` / `MissingOwnDeviceInfoError` events in Sentry ([COMAPEO-1TK](https://awana-digital.sentry.io/issues/7072609149/), [COMAPEO-223](https://awana-digital.sentry.io/issues/7434500086/), [COMAPEO-20E](https://awana-digital.sentry.io/issues/7383452114/)).

The refetch trigger itself (broad query invalidation in `useLeaveProject`) is fixed in digidem/comapeo-core-react#183; this PR removes the app-side hazard so no observer of the left project survives the leave, protecting against any later refetch trigger (focus, reconnect, remount).

## Changes

In `LeaveProject.tsx`'s `onSuccess`:
- Switch the active project to the default project **before** navigating.
- Navigate with `CommonActions.reset({routes: [Home, LeftProjectConfirmation]})` (same pattern as `ProjectRemovalListener`) instead of `navigation.replace`, unmounting all screens scoped to the left project. `popToTop()` from the confirmation screen still lands on Home.

## Notes

- `RemovedFromProjectBottomSheet` (the other `useLeaveProject` consumer) already sets the active project and pops to top, including creating a default project when missing — unchanged.
- Known edge (pre-existing, unchanged): if the `useManyProjects` data hasn't loaded when leave succeeds, no default project is found and the active project remains the left one. Worth a follow-up.

## Testing

- `tsc --noEmit` clean (one pre-existing error on `develop`, unrelated)
- `npm run test:jest`: 56 suites / 658 tests pass
- Not yet manually verified on device — needs a QA pass of the leave flow: leave from YourTeam → confirmation screen → "Open {device}" lands on Home with the default project active, and no new Sentry events on COMAPEO-1TK.


ErikSin:
As a alternative I think we should consider doing this instead: https://github.com/digidem/comapeo-mobile/issues/2004

ErikSin (approved):

Issue #1940, opened 2026-06-11
Duplicate default (unnamed) projects can be created — unguarded createProject races


gmaclennan:
## Summary

We have seen duplicate default (unnamed) projects appear in production testing. There is no guard anywhere against creating more than one: `createProject()` happily creates another unnamed project every time, all "does a default exist?" checks read a render-time snapshot of the project list, and the creation calls are unawaited with no in-flight dedup. Project creation is also slow (keypair generation + project instance + default config import), so race windows are seconds wide on low-end devices.

## Duplicate-creation paths

1. **Double-tap on Close in `RemovedFromProjectBottomSheet`.** The button is only hidden by a re-render after the mutation status flips to `pending`. Two taps landing before that re-render run two `leaveProject.mutate` calls; both `onSuccess` callbacks see the same stale `defaultProject === undefined` closure and both fire `createProject` → two unnamed projects. The same latent double-tap issue exists on Join in `InviteReceived`.

2. **`ProjectRemovalListener` remount re-arms the button mid-flight.** The listener stays mounted in the tab navigator after it resets the stack to show the bottom sheet. Its `own-role-change` event path has no `currentRouteName !== 'RemovedFromProjectBottomSheet'` guard (unlike the initial-check effect above it), so a second role-change event re-dispatches the reset and remounts the sheet — discarding pending mutation state, showing the Close button again, and losing the first mutation's `onSuccess` (so `setActiveProjectId`/`popToTop` never run, making a second tap very likely while the first `createProject` is still in flight).

3. **Cross-flow races with fire-and-forget creates.** `InviteReceived` (accept) and `CreateOrNameSoloProject` (rename-solo flow) both fire `createProject` without awaiting it. Concrete production-shaped sequence: a user with no default project accepts an invite → unawaited create starts → the coordinator removes them shortly after → `RemovedFromProjectBottomSheet`'s project-list snapshot predates the create finishing → `defaultProject` is `undefined` → second create.

Also worth fixing while in there:

- `RemovedFromProjectBottomSheet`'s `leaveProject.mutate` has no `onError` — failures are silent (spinner clears, nothing happens).
- The sheet's `createProject` `onError` fallback sets `projects[0]` as active — from the stale pre-leave snapshot that can be the very project the user was just removed from, re-triggering the whole removal flow.

## Proposed fix

Centralize default-project resolution in one idempotent "ensure default project" helper that:

[...]

PR #1941, opened 2026-06-11
fix: ensure a default project is activated after leaving a project

### LINKED ISSUE #1939: Leaving a project can leave app stuck on the left project when no default project is found
## Summary

When leaving a project succeeds, the app tries to switch the active project to the default ("solo") project by searching the project list for a project with no name. If no unnamed project is found, `setActiveProjectId` is silently skipped and **the active project remains the project that was just left** — whose settings row has been deleted and whose data stores are closed. The app is then wedged: Home renders against a dead project, and relaunching doesn't help because `ActiveProjectIdStoreProvider` only falls back to another project when *no* project ID is persisted.

## Details

In `LeaveProject.tsx`, the leave `onSuccess` handler does:

```ts
const defaultProject = projects?.find(project => project.name === undefined);
if (defaultProject?.projectId) {
  setActiveProjectId(defaultProject.projectId);
}
// …nav reset to Home regardless
```

Problems:

1. **No fallback when no default is found.** Every path that creates the default project is best-effort/fire-and-forget (`InviteReceived`, `NameSoloProject` rename flow — both swallow creation errors), so users can legitimately have no unnamed project. For them, leaving a project silently leaves `activeProjectId` pointing at the left project.
2. **Render-time snapshot.** `projects` comes from the `useManyProjects` cache captured in a closure at button-press time. A default project created moments earlier (e.g. by the unawaited create in the invite-accept or rename flows) may not be in the snapshot, even though it exists.
3. **"Unnamed project" is a weak proxy for "default project".** It matches any unnamed project (including accidental duplicates — see the companion duplicate-defaults issue) and matches nothing when the default creation previously failed.
4. **Persistent wedge.** `leaveProject` in core sets `hasLeftProject: true` and deletes the project settings row, and `listProjects()` excludes left projects — so once wedged, the active project ID points at a project that no longer appears in any list, and the store's first-project fallback never runs because a project ID *is* persisted.

Related: core's `leaveProject` waits up to 45s for sync and can reject with "Sync timeout" *after* the project has already been marked left locally, in which case the error path also leaves `activeProjectId` pointing at the (actually left) project. A frontend fix should resolve the destination project from fresh data and never leave the active project pointing at a left project.

## Proposed fix

Resolve the destination project *inside* the leave success handler from fresh data (`clientApi.listProjects()`), with a fallback chain:

1. existing unnamed (default) project →
2. any other remaining joined project →
3. create a default project →
4. if creation fails, surface an error instead of silently staying on the left project.

See companion issue about duplicate default projects — the same shared, idempotent "ensure default project" helper should fix both.

gmaclennan:
## Description

Fixes #1939.

Leaving a project looked up the default (unnamed) project in a render-time snapshot of `useManyProjects` and silently skipped `setActiveProjectId` when none was found, leaving the app wedged on the left project — whose data stores are closed and which no longer appears in `listProjects()`. Relaunching didn't recover, because the active-project store's fallback only runs when no project ID is persisted.

### Changes

- New `useEnsureDefaultProject` hook: resolves the project that should become active from a **fresh** `listProjects()` call at decision time — existing unnamed project → create one → fall back to any other joined project if creation fails. Concurrent callers share a single in-flight create (one project max), which the follow-up PR for #1940 relies on.
- `LeaveProject` now uses it, so there is always a valid project to switch to (or a visible error), never a silent no-op.
- `LeaveProject` also recovers when core's `leaveProject` rejects with a sync timeout *after* the project has already been marked left locally (core persists `hasLeftProject` and deletes settings before waiting up to 45s for sync): on mutation error it re-checks `listProjects()` and runs the normal success path if the project is actually gone.

### Tests

- `useEnsureDefaultProject` hook tests (real `MapeoManager` over IPC): finds existing default, creates when missing, respects `excludeProjectId`, dedupes concurrent calls, falls back on creation failure, rejects when nothing is left to fall back to.
- `LeaveProject` component tests: leaving switches to the existing default; regression test that leaving with **no** default creates one and navigates instead of staying wedged on the left project.

> Stacked on #1938 (base branch `fix/leave-project-stale-queries`). Follow-up: duplicate-default-projects fix for #1940.

### DIFF
--- a/src/frontend/hooks/server/useEnsureDefaultProject.test.tsx
+++ b/src/frontend/hooks/server/useEnsureDefaultProject.test.tsx
@@ -1,0 +1,136 @@
+import {renderHook} from '@testing-library/react-native';
+import type {MapeoManager} from '@comapeo/core';
+import type {MapeoClientApi} from '@comapeo/ipc';
+import React, {type ReactNode} from 'react';
+
+import {useEnsureDefaultProject} from './useEnsureDefaultProject';
+import {
+  createManager,
+  setUpIPC,
+} from '../../../../tests/integration/helpers/core';
+import {MapeoApiWrapper} from '../../../../tests/integration/helpers/MapeoApiWrapper';
+
+function createWrapper(client: MapeoClientApi) {
+  return ({children}: {children: ReactNode}) => {
+    return <MapeoApiWrapper mapeoApi={client}>{children}</MapeoApiWrapper>;
+  };
+}
+
+function renderEnsureDefaultProject(client: MapeoClientApi) {
+  const {result} = renderHook(() => useEnsureDefaultProject(), {
+    wrapper: createWrapper(client),
+  });
+  return result.current;
+}
+
+function createFailingCreateClient(client: MapeoClientApi): MapeoClientApi {
+  return new Proxy(client, {
+    get(target, prop, receiver) {
+      if (prop === 'createProject') {
+        return () => Promise.reject(new Error('create failed'));
+      }
+      return Reflect.get(target, prop, receiver);
+    },
+  });
+}
+
+describe('useEnsureDefaultProject', () => {
+  let manager: MapeoManager;
+  let client: MapeoClientApi;
+  let onTeardown: Array<() => unknown> = [];
+
+  beforeEach(async () => {
+    onTeardown = [];
+
+    const managerSetup = await createManager({
+      name: 'test',
+      deviceType: 'mobile',
+    });
+    ({manager} = managerSetup);
+    const {fastifyController} = managerSetup;
+
+    const ipcSetup = setUpIPC({manager});
+    ({client} = ipcSetup);
+    const {stop} = ipcSetup;
+    onTeardown.push(stop);
+
+    await fastifyController.start();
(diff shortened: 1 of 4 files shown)


ErikSin (commented):
This feels overly complicated and the complication is coming from the `inFlightCreates`

In what scenario do we expect there to be more than one instance of creating a default project, considering that this is an imperative action, and would only be run if the user does some action?

If we get rid of inFlightCreates, we could simply check for the existence of the default project, and return it exists, or create one and return it if it exists.

ErikSin:
> This feels overly complicated and the complication is coming from the `inFlightCreates`
> 
> In what scenario do we expect there to be more than one instance of creating a default project, considering that this is an imperative action, and would only be run if the user does some action?
> 
> If we get rid of inFlightCreates, we could simply check for the existence of the default project, and return it exists, or create one and return it if it exists.

I just saw the [follow up issue](https://github.com/digidem/comapeo-mobile/issues/1940) and understand your reasoning a little better.

ErikSin:
see comment: https://github.com/digidem/comapeo-mobile/issues/1939#issuecomment-4784159715

ErikSin:
closing for a simpler solution

PR #1942, opened 2026-06-11
fix: prevent duplicate default projects from unguarded creation races

### LINKED ISSUE #1940: Duplicate default (unnamed) projects can be created — unguarded createProject races

gmaclennan:
## Description

Fixes #1940.

We have seen duplicate default (unnamed) projects in production testing. Every place that lazily created the default project checked a render-time snapshot of the project list and fired `createProject` without awaiting it, so concurrent flows could each decide "no default exists" and create one — e.g. the invite-accept flow's unawaited create racing the removed-from-project flow, or sequential invite sheets reading a stale cache while the first create was still in flight. Project creation is slow (keypair + project instance + default config import), so these windows are seconds wide on low-end devices.

### Changes

All lazy default-project creation now goes through `useEnsureDefaultProject` (from #1941), which checks a fresh `listProjects()` at call time and shares one in-flight create across all callers:

- `RemovedFromProjectBottomSheet`: replaces the inline find/create logic. Also adds the missing `onError` for the leave mutation (failures were silent), and the creation-failure fallback can no longer re-activate the project the user was just removed from (the helper excludes it; previously it fell back to a stale `projects[0]`).
- `InviteReceived`: replaces the render-time `hasDefaultProject` check + fire-and-forget create.
- `CreateOrNameSoloProject` (name-solo flow): replaces the fire-and-forget create after renaming the solo project.
- `ProjectRemovalListener`: the `own-role-change` event path now guards on the current route like the initial-check effect, so a repeat role-change event can't remount the bottom sheet and discard an in-progress leave (which left the user stuck with a re-enabled button).
- The sheet's Close button is guarded against double-taps landing before the disabling re-render.

### Tests

- Hook-level regression test: concurrent `ensureDefaultProject` calls from separately mounted components create exactly one project (the cross-flow race that produced the production duplicates).
- `RemovedFromProjectBottomSheet` component tests (real `MapeoManager` over IPC): closing switches to the existing default project and leaves the removed project; double-tapping Close creates at most one default project.

One investigation note: a same-component double-tap alone did *not* duplicate in the old code — React Query only fires mutate-level callbacks for the latest `mutate()` call on an observer — so the cross-component race above is the production-relevant path; the double-tap guard is belt-and-braces.

> Stacked on #1941 (base branch `fix/ensure-default-project-on-leave`), which is itself stacked on #1938.

### DIFF
--- a/src/frontend/hooks/server/useEnsureDefaultProject.test.tsx
+++ b/src/frontend/hooks/server/useEnsureDefaultProject.test.tsx
@@ -116,6 +116,22 @@ describe('useEnsureDefaultProject', () => {
     expect(projects).toHaveLength(1);
   });
 
+  // Regression test for #1940: e.g. the invite-accept flow and the
+  // removed-from-project flow each deciding to create a default project
+  test('concurrent calls from separately mounted components create exactly one project', async () => {
+    const ensureFromComponentA = renderEnsureDefaultProject(client);
+    const ensureFromComponentB = renderEnsureDefaultProject(client);
+
+    const [first, second] = await Promise.all([
+      ensureFromComponentA(),
+      ensureFromComponentB(),
+    ]);
+
+    expect(second).toEqual(first);
+    const projects = await client.listProjects();
+    expect(projects).toHaveLength(1);
+  });
+
   test('falls back to another joined project if creation fails', async () => {
     const namedProjectId = await client.createProject({name: 'named project'});
 
--- a/src/frontend/screens/Invites/InviteReceived.tsx
+++ b/src/frontend/screens/Invites/InviteReceived.tsx
@@ -7,9 +7,8 @@ import {
   useAcceptInvite,
   useRejectInvite,
   useSingleInvite,
-  useCreateProject,
-  useManyProjects,
 } from '@comapeo/core-react';
+import {useEnsureDefaultProject} from '../../hooks/server/useEnsureDefaultProject';
 import {HeaderText} from '../../sharedComponents/Text/HeaderText';
 import {BodyText} from '../../sharedComponents/Text/BodyText';
 import {UIActivityIndicator} from 'react-native-indicators';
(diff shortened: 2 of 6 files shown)


ErikSin:
see comment: https://github.com/digidem/comapeo-mobile/issues/1939#issuecomment-4784159715

ErikSin:
closing for a simpler solution

PR #1947, opened 2026-06-11
fix: duplicate option label bleeds into other selectOne fields

### LINKED ISSUE #1946: selectOne/selectMultiple option with duplicate label bleeds into subsequent fields
## Summary

A `selectOne` (or `selectMultiple`) field whose options include two entries with the **same `label`** (but different `value`) causes a rendering glitch: the duplicated option bleeds into *other* `selectOne` fields that are opened after it while paging through an observation's questions.

Originally reported against the config tooling in digidem/comapeocat#49, but the root cause is in this app.

## Root cause

`src/frontend/screens/ObservationFields/index.tsx` pages between questions with `navigation.popTo('ObservationFields', {question: n})` — the **same route**, so React Navigation keeps a single screen instance and re-renders `<Question>` with a different `field`.

Each field renders its options list keyed by `item.label`:
- `src/frontend/screens/ObservationFields/SelectOne.tsx` (`key={item.label}`)
- `src/frontend/screens/ObservationFields/SelectMultiple.tsx` (`key={item.label}`)

When a field has two options with the same label, the list has **duplicate React keys**. React can no longer cleanly reconcile the previous field's rendered children against the next field's, so a stale option element leaks into a field opened later — the glitch seen in comapeocat#49.

The config itself is valid (distinct `value`s); the bug is purely client-side rendering.

## Fix

Use a unique key (`` `${item.value}-${index}` ``) instead of `item.label` in both components. This is robust even if a malformed config duplicates both label and value, and `index` is a stable key here since `field.options` never reorders.

## Note

Tracking adding validation errors to config files to avoid duplicate labels in the first place in https://github.com/digidem/comapeocat/issues/49

gmaclennan:
## Problem

A `selectOne`/`selectMultiple` field with two options sharing the same `label` (different `value`) causes the duplicated option to bleed into *other* `selectOne` fields opened after it while paging through an observation's questions.

Reported against the config tooling in digidem/comapeocat#49, but the root cause is in this app.

## Root cause

`ObservationFields/index.tsx` pages between questions with `navigation.popTo('ObservationFields', {question: n})` — the same route, so one screen instance is reused and `<Question>` re-renders with a different `field`. Each field's options list was keyed by `item.label`, so two options with the same label produced **duplicate React keys**. React then can't reconcile the previous field's list against the next field's, leaking a stale option element into a later field.

The config is valid (distinct `value`s); the bug is purely client-side rendering.

## Fix

Key options by `` `${item.value}-${index}` `` instead of `item.label` in `SelectOne.tsx` and `SelectMultiple.tsx`. Unique even if a config duplicates both label and value, and `index` is stable here since `field.options` never reorders.

Fixes #1946
Refs digidem/comapeocat#49

### DIFF
--- a/src/frontend/screens/ObservationFields/SelectMultiple.tsx
+++ b/src/frontend/screens/ObservationFields/SelectMultiple.tsx
@@ -60,7 +60,7 @@ export const SelectMultiple = React.memo<Props>(({field}) => {
       <QuestionLabel field={field} />
       {field.options.map((item, index) => (
         <CheckItem
-          key={item.label}
+          key={`${item.value}-${index}`}
           onPress={() => handleChange(item.value)}
           checked={valueAsArray.includes(item.value)}
           label={item.label}
--- a/src/frontend/screens/ObservationFields/SelectOne.tsx
+++ b/src/frontend/screens/ObservationFields/SelectOne.tsx
@@ -49,7 +49,7 @@ export const SelectOne = React.memo<Props>(({field}) => {
       <QuestionLabel field={field} />
       {field.options.map((item, index) => (
         <RadioItem
-          key={item.label}
+          key={`${item.value}-${index}`}
           onPress={() => updateTag(field.tagKey, item.value)}
           checked={tags && item.value === tags[field.tagKey] ? true : false}
           label={item.label}


cimigree:
Might be best to close and reimplement since these files were refactored.

cimigree:
closing in favor of https://github.com/digidem/comapeo-mobile/pull/1980

Issue #1945, opened 2026-06-11
Photo storage-size label sometimes missing (expo-image disk-cache miss); make it always show


gmaclennan:
## Summary

The photo "storage size" label in the photo preview screen sometimes doesn't appear. It's derived from expo-image's on-device **disk cache**, which is timing/policy-dependent, so it misses when the image was served from memory, not yet written to disk, or evicted.

## Where

`src/frontend/lib/file-system.ts` `getExpoImageStorageSize` calls `ExpoImage.getCachePathAsync(url)`; for a blob-server URL that returns `null` on a disk-cache miss, so no size is shown. Used by `useImageLoadInfo` in `screens/PhotoPreviewModal/helpers.ts`.

Sentry: COMAPEO-20Z — https://awana-digital.sentry.io/issues/7400012137/

(A minimal fix landing now returns `null` on a cache miss so the label is simply omitted, rather than throwing/reporting to Sentry. This issue tracks making it **always** show.)

## What blocks "always shows"

The blob server (`@comapeo/core` `fastify-plugins/blobs.js`) streams blobs with only `Content-Type` — no `Content-Length` — so the client can't read the size from the URL without re-downloading the whole image. Tracked upstream in **digidem/comapeo-core#1281**.

## Proposed fix

Once the blob server sends `Content-Length` (comapeo-core#1281), have `getExpoImageStorageSize` read it for `http(s)` URLs (e.g. a `HEAD` request → `Content-Length`), falling back to the current `file://` `getInfoAsync` path. That makes the size available regardless of expo-image's cache state, so the label always shows.

PR #1951, opened 2026-06-11
fix: don't report expo-image disk-cache misses for photo storage size

### LINKED ISSUE #1950: Sentry error reported when fetching photo storage size on expo-image disk-cache miss (COMAPEO-20Z)
## Sentry

|---|---|---|---|
| COMAPEO-20Z | Error: Could not get size for image at … | 1 | https://awana-digital.sentry.io/issues/7400012137/ |

## Symptom

Sentry captured an `Error: Could not get size for image at …` exception. The image itself rendered fine for the user; only the optional storage-size label in the photo preview was affected. The error was caught by `useImageLoadInfo` and reported to Sentry rather than surfacing to the user.

## Root cause

`getExpoImageStorageSize` reads the photo size from expo-image's on-device **disk cache** (`getCachePathAsync`). For a blob-server URL, `getCachePathAsync` returns `null` on a disk-cache miss (image served from memory, not yet written, or evicted), so the function threw — and the only caller (`useImageLoadInfo`) caught it and reported to Sentry.

## Follow-up

Making the storage-size label **always** show (rather than being omitted on a cache miss) requires the blob server to send `Content-Length` (it currently sends only `Content-Type`), so the client can read the size directly. Tracked in:
- digidem/comapeo-core#1281 (blob server `Content-Length`)
- digidem/CoMapeo-mobile#1945 (read it so the label always shows)

gmaclennan:
## Sentry issue fixed

|---|---|---|---|
| COMAPEO-20Z | Error: Could not get size for image at … | 1 | https://awana-digital.sentry.io/issues/7400012137/ |

## Root cause

`getExpoImageStorageSize` reads the photo size from expo-image's on-device **disk cache** (`getCachePathAsync`). For a blob-server URL that returns `null` on a disk-cache miss (image served from memory, not yet written, or evicted), so the function threw — and the only caller (`useImageLoadInfo`) caught it and reported to Sentry. The image rendered fine; only the optional storage-size label was affected.

## Fix

`getExpoImageStorageSize` returns `Promise<number | null>` — `null` on a cache miss / missing file instead of throwing. `useImageLoadInfo` treats `null` as "no storage size" (`storageSize ?? undefined`). The UI already guards on `storageSize` being a number, so it degrades cleanly.

## Follow-up: make the label always show

Returning `null` omits the label on a cache miss rather than showing a wrong value. Making it **always** show requires the blob server to send `Content-Length` (it currently sends only `Content-Type`), so the client can read the size directly. Tracked in:
- digidem/comapeo-core#1281 (blob server `Content-Length`)
- digidem/CoMapeo-mobile#1945 (read it so the label always shows)

Closes #1950

### DIFF
--- a/src/frontend/lib/file-system.ts
+++ b/src/frontend/lib/file-system.ts
@@ -65,7 +65,7 @@ export async function selectFile({
  */
 export async function getExpoImageStorageSize(
   imageURL: string,
-): Promise<number> {
+): Promise<number | null> {
   let fileInfo: FileSystem.FileInfo;
 
   if (imageURL.startsWith('file://')) {
@@ -73,16 +73,14 @@ export async function getExpoImageStorageSize(
   } else {
     const cachePath = await ExpoImage.getCachePathAsync(imageURL);
 
-    if (!cachePath) {
-      throw new Error(`Could not get size for image at ${imageURL}`);
-    }
+    // null when the image isn't in expo-image's disk cache (served from memory,
+    // not yet written, or evicted) — a benign miss, not an error.
+    if (!cachePath) return null;
 
     fileInfo = await FileSystem.getInfoAsync(`file://${cachePath}`);
   }
 
-  if (!fileInfo.exists) {
-    throw new Error(`Could not get size for image at ${imageURL}`);
-  }
+  if (!fileInfo.exists) return null;
 
   return fileInfo.size;
 }
--- a/src/frontend/screens/PhotoPreviewModal/helpers.ts
+++ b/src/frontend/screens/PhotoPreviewModal/helpers.ts
@@ -194,7 +194,7 @@ export function useImageLoadInfo() {
       setImageLoadInfo({
         height: event.source.height,
         width: event.source.width,
-        storageSize,
+        storageSize: storageSize ?? undefined,
       });
     } catch (err) {
       captureException(err);


src/frontend/lib/file-system.ts
```diff
@@ -65,24 +65,22 @@ export async function selectFile({
...
+    // not yet written, or evicted) — a benign miss, not an error.
+    if (!cachePath) return null;
 
     fileInfo = await FileSystem.getInfoAsync(`file://${cachePath}`);
   }
 
-  if (!fileInfo.exists) {
-    throw new Error(`Could not get size for image at ${imageURL}`);
-  }
+  if (!fileInfo.exists) return null;
```
cimigree, line 83: Is it possible that a locally stored photo or file information that is genuinely missing on disc (not just the cache missing) might be swallowed by this?

cimigree (approved):
This makes sense. Seems like a cache miss is a normal thing to happen and no need to report the error to Sentry, especially as the image is rendering fine.

src/frontend/lib/file-system.ts
gmaclennan: yes, good point. It could be that this issue was mis-diagnosed - the real cause of the sentry issues could be a missing image, perhaps caused by the other issue we found with rare cases where images were never saved. I'm not feeling clear-headed enough to make a call on this right now, I can think about it tomorrow morning. I think your concern is valid.

Issue #1944, opened 2026-06-11
Disable Delete button while deletion is pending to avoid double-tap DocAlreadyDeletedError


gmaclennan:
## Summary

The Delete button on the confirm-delete sheets for observations and tracks can be tapped twice before the mutation flips to `pending`, firing two delete mutations. The second sees the doc already deleted and `@comapeo/core` throws `DocAlreadyDeletedError`.

## Where

- `src/frontend/screens/Observation/ConfirmDeleteObservationBottomSheet.tsx`
- `src/frontend/screens/Track/ConfirmDeleteTrackBottomSheet.tsx`

Sentry: COMAPEO-1YT — https://awana-digital.sentry.io/issues/7315369908/

(A fix landing now treats `DocAlreadyDeletedError` as success so it no longer surfaces as an error. This issue tracks the small UX hardening to prevent the double-tap in the first place.)

## Proposed fix

Disable the destructive Delete button while the mutation is pending (`disabled={isPending}`) — or guard the handler with an in-flight ref — so a rapid second tap can't start a second delete. The button is already hidden behind a spinner once `status === 'pending'`; the gap is the brief window before the first tap flips to pending.

PR #1960, opened 2026-06-11
fix: treat already-deleted doc as success on delete; fix stale-error capture

### LINKED ISSUE #1959: Deleting an already-deleted observation or track reports a "Doc already deleted" error (COMAPEO-1YT)
## Sentry

|---|---|---|---|
| COMAPEO-1YT | Error: Doc already deleted | 1 | https://awana-digital.sentry.io/issues/7315369908/ |

## Symptom

When a user deletes an observation or track that is already gone — for example by double-tapping the Delete button, or because the doc was already deleted on another device and synced — the confirm-delete flow surfaces an error. Sentry captures a `DocAlreadyDeletedError` ("Doc already deleted") even though, from the user's point of view, the document they wanted gone is gone.

## Root cause

Deleting an observation/track that is already deleted makes `@comapeo/core` `DataType.delete` throw `DocAlreadyDeletedError`. The throw itself is correct — the doc is already gone, which is the success case for the user — but both confirm-delete sheets (`ConfirmDeleteObservationBottomSheet.tsx` and `ConfirmDeleteTrackBottomSheet.tsx`) treated it as a failure, reporting it to Sentry and showing the error sheet.

The observation sheet additionally had a stale-variable bug: its `onError` handler captured the hook's previous-render `deleteObservationError` value instead of the fresh `err` argument passed to the handler.

The double-tap that produces this is tracked in digidem/CoMapeo-mobile#1944 (disable the Delete button while the deletion is pending).

gmaclennan:
## Sentry issues fixed

|---|---|---|---|
| COMAPEO-1YT | Error: Doc already deleted | 1 | https://awana-digital.sentry.io/issues/7315369908/ |

## Root cause

Deleting an observation/track that is already gone (a double-tap, or it was deleted on another device and synced) makes `@comapeo/core` `DataType.delete` throw `DocAlreadyDeletedError`. The throw is correct — the doc is already gone, which is the success case from the user's point of view — but both confirm-delete sheets reported it to Sentry. The observation sheet also had a **stale-variable bug**: `onError` captured the hook's previous-render `deleteObservationError` instead of the fresh `err` argument.

## Fix

`ConfirmDeleteObservationBottomSheet.tsx` and `ConfirmDeleteTrackBottomSheet.tsx`: in `onError`, when `err.code === DocAlreadyDeletedError.code` (imported from `@comapeo/core/errors.js`), finish quietly (`navigation.pop(2)`) without Sentry or the error sheet. The observation sheet now captures the fresh `err` (and drops the unused stale `error` from the hook destructure).

## Follow-up

The double-tap that produces this is tracked in digidem/CoMapeo-mobile#1944 (disable the Delete button while the deletion is pending).

Closes #1959

### DIFF
--- a/src/frontend/screens/Observation/ConfirmDeleteObservationBottomSheet.tsx
+++ b/src/frontend/screens/Observation/ConfirmDeleteObservationBottomSheet.tsx
@@ -12,6 +12,7 @@ import {
 } from '../../sharedComponents/Buttons';
 import {useNavigationFromRoot} from '../../hooks/useNavigationWithTypes';
 import {useDeleteDocument} from '@comapeo/core-react';
+import {DocAlreadyDeletedError, getErrorCode} from '@comapeo/core/errors.js';
 import * as Sentry from '@sentry/react-native';
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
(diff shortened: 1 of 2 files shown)


cimigree (approved):
It's great that we have the error codes now to be able to check and swallow this error.
I tested this manually with a normal delete and an attempt at a double-tap. Both worked cleanly and did not result in errors. I will be doing the follow up to disable the button during pending time in the upcoming sprint, FYI.

PR #1962, opened 2026-06-11
fix: navigate back to camera on completed observation

### LINKED ISSUE #1911: Navigate back to camera screen after observation
https://www.notion.so/digidem/Photo-Capture-Improvements-31d1b08162d5809289dbf24dcb0db15a?source=copy_link#3271b08162d58003a699f27c542052ca

ErikSin:
closes #1911 

https://github.com/user-attachments/assets/88b78178-a8a3-4930-8251-b20fdc755b45

### DIFF
--- a/src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
+++ b/src/frontend/screens/ObservationCreate/ObservationCreateSaveButton.tsx
@@ -117,7 +117,18 @@ export const ObservationCreateSaveButton = () => {
   const finalizeSave = () => {
     queryClient.invalidateQueries({queryKey: STORAGE_QUERY_KEY});
     clearDraft();
-    navigation.popTo('Home', {screen: 'Map'});
+    const homeTabRoute = navigation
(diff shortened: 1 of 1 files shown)


cimigree (approved):
This is good from the standpoint of saving and discarding a regular observation from the camera, even when there is no location.
I am not sure how important it is, but if you are in obscured mode, and take an observation from the camera, it goes back to the map. Comes from line 169 for me.
I don't _think_ that is blocking? But wanted to mention it as a discrepancy. Maybe you could lift lastOpenedTab up to component level?

ErikSin:
> This is good from the standpoint of saving and discarding a regular observation from the camera, even when there is no location. I am not sure how important it is, but if you are in obscured mode, and take an observation from the camera, it goes back to the map. Comes from line 169 for me. I don't _think_ that is blocking? But wanted to mention it as a discrepancy. Maybe you could lift lastOpenedTab up to component level?

lets not do it in obscure mode...one less thing to maintain....

PR #1971, opened 2026-06-17
fix: exchange index test

cimigree:
Adds a longer timeout for the sync with remote archive test. 
Attempt to close open network handles.
Trying to get that test to pass more reliably. Right now it is failing a LOT!
But, it passed on first attempt on this PR

### DIFF
--- a/tests/integration/helpers/core.ts
+++ b/tests/integration/helpers/core.ts
@@ -168,12 +168,11 @@ export async function inviteToProject(
 
 export const createTestServer = (): Promise<{
   serverBaseUrl: string;
-  close: () => void;
+  close: () => Promise<void>;
 }> =>
   new Promise((resolve, reject) => {
-    const startServerPath = require.resolve(
-      '../../../tests/integration/helpers/startTestCloudServer.mjs',
-    );
+    const startServerPath =
+      require.resolve('../../../tests/integration/helpers/startTestCloudServer.mjs');
     const childProcess = spawn('node', [startServerPath], {
       stdio: ['ignore', 'pipe', 'inherit'],
     });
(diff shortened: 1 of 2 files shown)


tests/integration/helpers/core.ts
```diff
@@ -186,9 +185,11 @@ export const createTestServer = (): Promise<{
...
         resolve({
           serverBaseUrl: url,
-          close: () => {
-            childProcess.kill('SIGTERM');
-          },
+          close: () =>
+            new Promise<void>(resolveClose => {
+              childProcess.once('close', () => resolveClose());
+              childProcess.kill('SIGTERM');
+            }),
```
ErikSin, line 192: What is this for? I don't think i fully understand the syntax and what it is trying to acheive

cimigree: I was getting a Jest "did not exit" warning running this test locally.

The test spins up a process --  (the cloud server) and afterEach needs to wait for it to fully stop before the next test starts. Without it, the server process lingers in the background with open network connections, which can interfere with the next test or cause that warning I saw.

It used to be 
```
close: () => {
  childProcess.kill('SIGTERM');
}
```
Sending SIGTERM just asks the process to stop — it doesn't wait for it to actually die and can exit really fast. So when afterEach (in the index test) did `await close()`, the await had nothing to wait on and returned immediately, while the child process was still shutting down in the background.

Now it is wrapping the whole thing in a Promise and uses the function resolveClose to fulfill the process and signal that the process has actually stopped.

The listener is set up first --  `childProcess.once('close', () => resolveClose())` so that it won't miss the close event. `once()` listens for the first firing of the `close` event, runs the callback (`resolveClose()`) and then removes itself.

Then the old way to kill the process. `childProcess.kill('SIGTERM');`

Then after the process succeeds in being "killed" then the `resolveClose` runs and the promise fulfills.

ErikSin (approved):
Hopefully this works!

PR #1980, opened 2026-06-18
fix: select duplicate label

cimigree:
Uses option value as key for select one and select multiple. 
Adds tests for making sure duplicate labels don't cause the wrong options to appear (which is why the key change was necessary)

This is a redo of https://github.com/digidem/comapeo-mobile/issues/1946 which was addressing this issue: https://github.com/digidem/comapeocat/issues/49

Summary
A selectOne (or selectMultiple) field whose options include two entries with the same label (but different value) causes a rendering glitch: the duplicated option bleeds into other selectOne fields that are opened after it while paging through an observation's questions.

Originally reported against the config tooling in https://github.com/digidem/comapeocat/issues/49, but the root cause is in this app.

It was redone for two reasons
1. The code for these files was totally refactored, so a merge with the develop branch would cause conflicts
 and 2. I am trying a slightly different approach.

This was Gregor's fix: 
Use a unique key (`${item.value}-${index}`) instead of item.label in both components. This is robust even if a malformed config duplicates both label and value, and index is a stable key here since field.options never reorders.

My take...
However, using the index is against the ESLint rule `no-array-index-key`. So instead, if we just convert the value to a string, that prevents the issue with duplicate labels. If the value is duplicated, there is a bigger problem with the options in the categories so it should probably show as wonky to the user as it is wonky!

However, if you think the array index should also be there in the key, I will add that.

### DIFF
--- a/src/frontend/screens/ObservationFields/SelectMultiple.test.tsx
+++ b/src/frontend/screens/ObservationFields/SelectMultiple.test.tsx
@@ -86,6 +86,38 @@ test('works when tagValue is a scalar string', () => {
   expect(updateTag).toHaveBeenCalledWith(['a', 'b']);
 });
 
+test('re-renders correctly when swapped to a field with duplicate labels', () => {
+  const DUPLICATE_LABEL_OPTIONS = [
+    {label: 'Same', value: 'x'},
+    {label: 'Same', value: 'y'},
+  ];
+  const NEXT_OPTIONS = [
+    {label: 'Red', value: 'red'},
+    {label: 'Blue', value: 'blue'},
+  ];
+  const updateTag = jest.fn();
+
+  const {rerender} = render(
+    <SelectMultiple
+      options={DUPLICATE_LABEL_OPTIONS}
+      updateTag={updateTag}
+      tagValue={[]}
+    />,
+  );
+
+  rerender(
+    <SelectMultiple
+      options={NEXT_OPTIONS}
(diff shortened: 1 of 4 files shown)


ErikSin:
> However, using the index is against the ESLint rule no-array-index-key. So instead, if we just convert the value to a string, that prevents the issue with duplicate labels. If the value is duplicated, there is a bigger problem with the options in the categories so it should probably show as wonky to the user as it is wonky!

if the value is duplicated, it doesnt really matter in terms of validation. Ideally there would not be 2 of the same value, but if there is, it wont corrupt the data or cause any validation issues. So i think having the UI break even if it wont break the backend is not a desirable outcome.

I think we should use the index key combined with the value, and just add a eslint ignore comment.

ErikSin (changes requested):
See comment above

cimigree:
> if the value is duplicated, it doesnt really matter in terms of validation. Ideally there would not be 2 of the same value, but if there is, it wont corrupt the data or cause any validation issues. So i think having the UI break even if it wont break the backend is not a desirable outcome.
> 
> I think we should use the index key combined with the value, and just add a eslint ignore comment.

https://github.com/digidem/comapeo-mobile/pull/1980/commits/0f94e84456aff70d8954761d15c8b0c823c1cf4a

ErikSin (approved):
Thanks for adding this!

PR #1981, opened 2026-06-18
fix: invite onboarding

### LINKED ISSUE #1910: Successful Invite not properly navigating in onboarding
https://www.notion.so/digidem/Duplicate-mapping-on-your-own-project-created-after-an-invite-error-message-despite-being-added-to--34a1b08162d5804ba201d325f725a401

ErikSin:
closes #1910 

The problem was that when a user joins a project, we reset the navigation before switching to the new project. The reason why we reset the navigation was because errors were being thrown if the user switched projects while viewing something from another project. For example if a user was viewing an observation from project A and gets invited to project B while viewing this observation, switching to project B without navigating away caused the app to crash, as that observation is not available to project B. To mitigate this, [we reset the navigation](https://github.com/digidem/comapeo-mobile/blob/a95583b0f9cc55c4470c123b6f8f4125e38e22fe/src/frontend/screens/Invites/InviteReceived.tsx#L95) before the user switches project, when they click the accept invitation button.

During onboarding, we attempt to reset the navigation to have `Home` as the first. But the `Home` screen does not exist yet, so the app just silently errors in the background. Since it a silent error, when the user presses the accept invite button, that first time, it looks like nothing happens. The user than presses the button the second time, where they try to join the project again. But they had joined the project already when they pressed the button the first time, so the app throws an error.

Unfortunately fixing this was not that trivial. Because of the nature of conditional rendering, [we dont want to navigate away when the projectId is set the first time](https://reactnavigation.org/docs/auth-flow#dont-manually-navigate-when-conditionally-rendering-screens). In the ideal world, the conditional rendering navigates the user to the home screen as the onboarding screens are no longer rendered. But the invitation screens are part of both the onboarding screen and the root stack. So when the rootstack was being rendered, it was also keeping the invitation screen open.

To fix this I had to [remove the shared screens using a nav key](https://reactnavigation.org/docs/auth-flow#removing-shared-screens-when-auth-state-changes). First i had to moved the shared screens into `Rootstack/index` (instead of having them duplicated like we did before). and then i added a `navigationkey` as the projectId. this means that when the projectId is changed, these screens get taken out of the nav stack.

https://github.com/user-attachments/assets/d752badd-308c-4299-9b89-c5f0ea34d161


src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -85,13 +85,31 @@ export const InviteReceived = ({
       {inviteId: inviteId},
       {
         onSuccess: projectId => {
+          // In versions before v6, the user did not have to have a default project
+          // Now we would like the user to always have a default project
+          // This guarantees that
```
cimigree, line 90: Is this comment incomplete?

src/frontend/screens/Invites/InviteReceived.tsx
```diff
@@ -85,13 +85,31 @@ export const InviteReceived = ({
...
           if (!hasDefaultProject) {
             createProject.mutate(undefined, {
               onError: err => {
                 Sentry.captureException(err);
               },
             });
           }
+
+          const isInOnboarding = navigation
```
cimigree, line 99: What if they are on a different onboarding screen? Will the invite wait? Is that deliberate to wait for that specific screen? I think this is the problem that I came across. Maybe best to check isOnboarding some other way.

tests/integration/helpers/setupIntegrationTest.tsx
```diff
@@ -78,3 +78,67 @@ export function setupIntegrationTest() {
...
+
+    await fastifyController.start();
+    onTeardown.push(() => fastifyController.stop());
+  });
+
+  afterEach(async () => {
+    for (const fn of onTeardown) await fn();
+  });
+
+  const renderNavigationAsync = async ({
```
cimigree, line 110: Why are you accepting the activeProjectId but not using it here? I think you should either remove it from the type or wire it through to createAppProvidersWrapper like the existing renderNavigation does

cimigree (changes requested):
I think this is the actual issue? https://github.com/digidem/comapeo-mobile/issues/1910

Problem. Because people don't always behave as we would like...
If Device A sends an invite during onboarding but Device B chooses Map On My Own at the last step (instead of the Join a Project Screen) I see the project invite bottom sheet on the Map On Your Own sheet and then I try to join the project (hit Join Project), I get this error.

```
 ERROR  The action 'RESET' with payload {"index":1,"routes":[{"name":"Home"},{"name":"InviteSuccessfullyAccepted","params":{"projectName":"Testing invite while editing","projectId":"8xxjdap6zyojxgotqb1k4e4oiykyo84fjei4rpzzz61fmzq9a9jo"}}]} was not handled by any navigator.

This is a development-only warning and won't be shown in production.
```
And I stay on the Map On Your Own screen with the bottom sheet open.
If I then press Join Project again, I get the Error bottom sheet with this error.

Image

I can then close that sheet and then I end up on my default project. I now have three projects. The project I was invited to and two default projects.

Seems like the same problem as before.

Also, 
I got this error when accepting an invite while within the app.
Not sure what it means or if it is a real error.
` ERROR  MapLibre error {com.comapeo.dev}[JNI]: Error setting property: circle-color [18]: Branch labels must be unique. {"level": "error", "message": "{com.comapeo.dev}[JNI]: Error setting property: circle-color [18]: Branch labels must be unique.", "tag": "Mbgl"}`

ErikSin:
> I think this is the actual issue? #1910
> 
> Problem. Because people don't always behave as we would like... If Device A sends an invite during onboarding but Device B chooses Map On My Own at the last step (instead of the Join a Project Screen) I see the project invite bottom sheet on the Map On Your Own sheet and then I try to join the project (hit Join Project), I get this error.
> 
> ```
>  ERROR  The action 'RESET' with payload {"index":1,"routes":[{"name":"Home"},{"name":"InviteSuccessfullyAccepted","params":{"projectName":"Testing invite while editing","projectId":"8xxjdap6zyojxgotqb1k4e4oiykyo84fjei4rpzzz61fmzq9a9jo"}}]} was not handled by any navigator.
> 
> This is a development-only warning and won't be shown in production.
> ```
> 
> And I stay on the Map On Your Own screen with the bottom sheet open. If I then press Join Project again, I get the Error bottom sheet with this error.
> Image
> 
> I can then close that sheet and then I end up on my default project. I now have three projects. The project I was invited to and two default projects.
> 
> Seems like the same problem as before.
> 
> Also, I got this error when accepting an invite while within the app. Not sure what it means or if it is a real error. ` ERROR MapLibre error {com.comapeo.dev}[JNI]: Error setting property: circle-color [18]: Branch labels must be unique. {"level": "error", "message": "{com.comapeo.dev}[JNI]: Error setting property: circle-color [18]: Branch labels must be unique.", "tag": "Mbgl"}`

fixed: https://github.com/digidem/comapeo-mobile/pull/1981/commits/366283ae76381be38f1150fe11f1a1a62457184f

tests/integration/helpers/setupIntegrationTest.tsx
ErikSin: if i wire it through to createAppProvidersWrapper than the test will have a project already set up. The point of this function is to have no project.

ErikSin: https://github.com/digidem/comapeo-mobile/pull/1981/commits/c2165fce12e8a63189950732205960c096693a12

src/frontend/screens/Invites/InviteReceived.tsx
ErikSin: The only should ONLY get an invite if they are on this screen. So i added the missing onboading screen to the exlusion list

src/frontend/screens/Invites/InviteReceived.tsx
ErikSin: No, IT is saying `Now we would like the user to always have a default project - this guarantees that`

cimigree (approved):
Works correctly now. I am sure this will be a helpful and important fix in the field
Can you check if this is closing the right issue (check the link/ number, please)?

PR #1984, opened 2026-06-23
fix: observation list faster rendering

### LINKED ISSUE #1967: Lazy load observation flat list
https://app.notion.com/p/digidem/Over-500-observations-created-1-2-mins-of-lag-time-on-list-3731b08162d5806888e2e11c38796a4a

cimigree:
closes #1967
### Summary
This PR is an attempt to make the loading and scrolling of the observation list faster and smoother, even if there are a lot of observations.

### Lifts fetching data up to the parent level 
Now the fetch only happens once (with two queries) instead of potentially 500+ times (with two queries each) on each list item. Especially as the useIsMyDocument is a backend IPC call, not just a cache lookup.
This also required fixing two other places that use these list item components — the observations inside a Track detail screen, and the tracks inside an Observation  screen. Since those are single items or small fixed lists (not 500+), they call the hooks locally rather than lifting them up.

### Adds and removes some FlatList props. 
Some have to do with the fact that I noticed some blank screen time as I was scrolling up only.

**windowSize={7}** -- We used to have 3, but as I said, I am seeing a blank screen on fast scroll. The default is 21. Having a greater number here keeps windows above and below in memory. So 7 is a compromise to save memory but to try to address the blank screen on fast upward scroll.

**updateCellsBatchingPeriod={20}** — How often (in milliseconds) FlatList checks whether it needs to render more items. Default is 50ms. Halving it to 20ms means blank cells get filled in faster after a fast scroll, at the cost of slightly more CPU work.

**Removed removeClippedSubviews** — The default for Android is true, so I just took it out for noise prevention

### Put the Suspense around the whole ObservationListItem
Because the hooks are in the parent, the boundary can be simplified to just wrap the item since they are effectively synchronous. This fallback now only appears briefly when the initial list loads rather then flickering per item when scrolling.

### Fixes for Creation of Test Data
While doing this work, I used the creation of test observations and noticed my input was invisible. So I fixed that in this PR. I also had to reload the app to get the observations to show up, so I tried to address that as well by refetching the observation list after all of the new observations are created.

### Before the changes were made rendering a large list is awkward and slow (sometimes)

https://github.com/user-attachments/assets/6b839cb7-f201-4538-9241-99756adaec93

### After the changes from this PR shows the loading of the Observation List is much quicker and smoother

https://github.com/user-attachments/assets/8ab64f6a-e452-4ecc-b0e8-1a1c7c80efaa

### DIFF
--- a/src/frontend/screens/ComapeoSettings/CreateTestData.tsx
+++ b/src/frontend/screens/ComapeoSettings/CreateTestData.tsx
@@ -1,5 +1,5 @@
 import {useOwnDeviceInfo, useCreateDocument} from '@comapeo/core-react';
-import {useMutation} from '@tanstack/react-query';
+import {useMutation, useQueryClient} from '@tanstack/react-query';
 import {lengthToDegrees} from '@turf/helpers';
 import {randomPosition} from '@turf/random';
 import {LocationObject} from 'expo-location';
@@ -10,7 +10,7 @@ import {StyleSheet, TextInput, ToastAndroid, View} from 'react-native';
 import {UIActivityIndicator} from 'react-native-indicators';
 
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
-import {LIGHT_GREY, RED, WHITE} from '../../lib/styles';
+import {BLACK, LIGHT_GREY, RED, WHITE} from '../../lib/styles';
 import {PrimaryButton} from '../../sharedComponents/Buttons';
 import {LocationView} from '../../sharedComponents/LocationView';
 import {ScreenContentWithDock} from '../../sharedComponents/ScreenContentWithDock';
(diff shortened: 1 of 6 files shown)


ErikSin:
Upon looking at your videos i don't think your recreating the original issue, and therefore I don't know if this is fixing it. Can you create an APK of this, and well send it to Ximena to test on her device?

cimigree:
> Upon looking at your videos i don't think your recreating the original issue, and therefore I don't know if this is fixing it. Can you create an APK of this, and well send it to Ximena to test on her device?

Ok I added an APK to this folder. https://drive.google.com/drive/folders/1sk4ab-1xhuGXlTzqeXYcgkTtJ8Uq_swv
Called observation-list-test-6-26-26.apk

ErikSin:
Closing this as it has been open for a long time. When someone with a device that has this problem is more available we can re-open this PR for testing.

PR #1988, opened 2026-06-24
fix: read storage when it is updated

### LINKED ISSUE #1706: Updating storage on phone not resetting low storage indicator
**Describe the bug**
When someone sees that they have low storage, they leave the app and remove some files. When they go back to the app, if they haven't force closed, then the low storage indicator is still there.

**To Reproduce**
1. Make sure your phone is full
2. Open comapeo
3. See that there is a low storage indicator
4. Remove files from phone to free up storage
5. Go back to the app (without having force closed)
6. The low storage indicator will still be there

**Expected behavior**
When someone frees up storage, they should see that they now have enough storage

**Screenshots**
N/A

**Additional context**
Gregor came across this when developing on an emulator

cimigree:
closes #1706 
### Summary

With useQuery, refetchOnWindowFocus defaults to true [("If set to true, the query will refetch on window focus if the data is stale."](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)
but it doesn't actually do anything because it only works without wiring on browsers. For react native, you need to make sure the focus manager is set in order for react query to know when the app is active.

> Instead of event listeners on window, React Native provides focus information through the [AppState module](https://reactnative.dev/docs/appstate#app-states). You can use the AppState "change" event to trigger an update when the app state changes to "active":

Based on the docs: https://tanstack.com/query/latest/docs/framework/react/react-native#refetch-on-app-focus

So adds a listener in App State so that react query knows when the app is active, and thus when to refetch queries.
That way, when someone clears up some of their storage and comes back to the app, the low storage warnings update accordingly.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -19,10 +19,10 @@ Logger.setLogCallback(log => {
 // https://github.com/maplibre/maplibre-react-native/blob/6f99de530eec2e06de485ef86f4be61f941e0e09/docs/content/modules/mlrn-module.md#setconnectedconnected
 setConnected(true);
 
-import {QueryClient} from '@tanstack/react-query';
+import {QueryClient, focusManager} from '@tanstack/react-query';
 import {AppNavigator} from './AppNavigator';
 import {initializeNodejs} from './initializeNodejs';
-import {PermissionsAndroid} from 'react-native';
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks!

PR #1991, opened 2026-06-24
fix: loader in delete track and observation 

### LINKED ISSUE #1944: Disable Delete button while deletion is pending to avoid double-tap DocAlreadyDeletedError

### LINKED ISSUE #1977: Block Observation and Track Deletion UI
Add Loader to deletion buttons

See this issue for same but more detailed instructions https://github.com/digidem/comapeo-mobile/issues/1944

ErikSin:
closes #1977.
closes #1944.

I have no screen shots as the deletion happens quite quickly and so i dont have enough time to get a screen shot.

### DIFF
--- a/src/frontend/screens/Observation/ConfirmDeleteObservationBottomSheet.tsx
+++ b/src/frontend/screens/Observation/ConfirmDeleteObservationBottomSheet.tsx
@@ -16,6 +16,7 @@ import {DocAlreadyDeletedError, getErrorCode} from '@comapeo/core/errors.js';
 import * as Sentry from '@sentry/react-native';
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 import {NativeRootNavigationProps} from '../../sharedTypes/navigation';
+import {UIActivityIndicator} from 'react-native-indicators';
 
 const m = defineMessages({
   deleteTitle: {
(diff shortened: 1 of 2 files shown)


cimigree (approved):
I can't really trigger the double tap or see the spinner either, but this seems like programmatically the correct approach.

PR #1993, opened 2026-06-29
enhancement: set sentry qa name

### LINKED ISSUE #1968: Set QA device names for Sentry
We can set tags in Sentry for easier searching of issues. Lets create a screen that prompts the user to set a QA device name that we add as a tag in Sentry, allowing us to easier search for error in sentry.

We need to create a custom screen that asks the QA users to type in a QA specific name. We should also include copy that explain what this name is being used for, and asking the QA user to write down that name to share to developers. This screen should only show if the user is using a Release Candidate or preRelease.

Store the name in a new mmkv/zustand persisted store.

To Do
- [ ] Create mmkv/zustand persisted store that stores the name (the name should be optional as it is not for all builds
- [ ] Create new screen prompting the user to type in a QA name. Can look similar to the [CreateTestData Screen](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/ComapeoSettings/CreateTestData.tsx).
- [ ] Conditionally render the screen in the [RootStackNavigator](https://github.com/digidem/comapeo-mobile/blob/62daaf19e41282f66db569317ea1f37b764d479d/src/frontend/Navigation/Stack/index.tsx#L119). If they are using a ReleaseCandidate/prerelease and do not have a name set, this should be conditionally rendered first.
- [ ] when the name is set, [set the Sentry Tag](https://docs.sentry.io/platforms/react-native/enriching-events/tags/)with the key`QA_Device_Name`
- [ ] Create basic test where if there is no persisted `QAName` then this screen will show
- [ ] Build APK and manually test that these screen work.

cimigree:
closes #1968 

- Creates a component to show a QA Name Setting screen if one hasn't been set yet.
- The component is split into two pieces. A nav screen to use in settings and a pure component (with no navigation) to use outside of the Navigation Container.
- The screen shows first thing, whether or not the RC has completed onboarding
- I also chose to add it to the settings screen, with the test data creation, in case the QA completer decides that a new name is needed or a mistake was made on initial setting (and allows someone to check what name was chosen). It is gated there.
- Shows no heading if the screen appears first thing; if the name hasn't been set yet, because the back button is not useful (and not available without navigation). On the initial setting component, some styling (and safe area context) had to be added so the screen showed properly.

### Manual Testing
I tested it with a passcode and without.

**Suggestions for testing:**
- First build a release-candidate apk
- You can start with a current RC in which you have already been onboarded. Install the APK and you will see the Screen to input the name show first thing. Once you hit save, you see a toast (similar to how it is done for creating test data) and then you will land on the map screen.
- Then delete the RC of the app entirely or clear all of the data. Install this release candidate version and you should see the screen to input the qa name and when you hit save (and see the toast) you will proceed into onboarding.
OR
- temporarily change the App variant to development. You can clear your data and try it before onboarding. Or if you don't want to clear data, it will show up first thing when you open the app.

### Some screen recordings and screen shots.
When I did the screen recordings, I hadn't added the margin to the Create QA Device Name screen yet.
image

---

What the QA device name creation save looks like when someone has already been onboarded.
https://github.com/user-attachments/assets/0ae5bfaf-773f-4324-b72a-2d76cc67b7fe

---
And when someone is installing the APP for the first time (no onboarding completed)

https://github.com/user-attachments/assets/390c3122-b6a6-493a-8cde-eb98b0c1d64c

Here is the QA_Device_name showing up in Sentry!
Screenshot 2026-06-29 at 5 56 17 PM


src/frontend/App.tsx
```diff
@@ -96,6 +97,18 @@ if (appMetricsOptIn) {
...
+const initialQADeviceName = qaDeviceNameStore.instance.getState().qaDeviceName;
+if (initialQADeviceName) {
+  Sentry.setTag('QA_Device_Name', initialQADeviceName);
+}
+qaDeviceNameStore.instance.subscribe((current, previous) => {
+  if (current.qaDeviceName !== previous.qaDeviceName && current.qaDeviceName) {
+    Sentry.setTag('QA_Device_Name', current.qaDeviceName);
+  }
+});
+
```
ErikSin, line 118: Is it possible to do this directly in the store? If so, then I think we should, so were not adding an extra listener here

src/frontend/screens/SetQADeviceName.tsx
```diff
@@ -0,0 +1,137 @@
...
+      contentContainerStyle={styles.content}
+      dockContent={
+        <PrimaryButton fullSize text="Save Name" onPress={handleSave} />
+      }>
+      <View style={styles.section}>
+        <HeaderText variant="header2">Set QA Device Name</HeaderText>
+        <BodyText style={styles.description}>
+          This name is used for the developers to be able to search through
+          events in Sentry and identify the relevant device. Spaces are allowed
+          and there is a maximum of 200 characters and no new lines.
```
ErikSin, line 51: Can you also add text asking them to write down the name they chose OR take a screenshot of the name so that it can be shared with developers (obviously you're a better writer than i, so dont write what i just wrote 😬)

src/frontend/AppNavigator.tsx
```diff
@@ -11,6 +11,12 @@ import type Sentry from '@sentry/react-native';
 import {PostHogProvider} from 'posthog-react-native';
 import {postHog} from './lib/posthog';
 import {linking} from './lib/deepLinkConfig';
+import {useQADeviceName} from './contexts/QADeviceNameStoreContext';
+import {SetQADeviceNameScreen} from './screens/SetQADeviceName';
+import {APP_VARIANT} from './lib/appVariant';
+
+const isQABuild =
+  APP_VARIANT === 'releaseCandidate' || APP_VARIANT === 'preRelease';
```
ErikSin: this check is done twice, so lets just add `isQABuild` as an exported variable in lib/appVariant!

tests/integration/helpers/react.tsx
```diff
@@ -171,6 +174,9 @@ export function createAppProvidersWrapper({
 
   const unitSystemStore = createUnitSystemStore({persist: false});
 
+  const qaDeviceNameStore =
+    qaDeviceNameStoreOverride ?? createQADeviceNameStore({persist: false});
```
ErikSin: instead of adding an override store, I think we can just add a `QaDeviceName` prop. And then if there is a QaDeviceName, we set it in the store when we instantiate of the store.

I think it would be cleaner than having 2 possible instances of the store, while still having the same control of whether or not the store has a name, which will allow you to keep your tests logic as is.

Also it would follow the other architecture of the projectId, which does a simlar thing in this file!

src/frontend/AppNavigator.tsx
ErikSin, line 1: More of a meta question for you, do you think we should rename this? This doesn't always return the App Navigator anymore, but also does that matter? What do you think?

ErikSin (changes requested):
1 blocking comment related to how you create the override store. I just dont want to add more complication to the testing environment, and having 2 possible stores just feels a little fragile.

src/frontend/AppNavigator.tsx
cimigree: I think it is fine as it is. It has a small "gate" but it still handles the navigation of the user through the app. It just now starts the navigation for QA with the question about a Sentry name and only if they don't have one already. If you want to rename that is fine with me, but I don't think it is needed for clarity or readability.

src/frontend/App.tsx
cimigree: I made a callback and wired it up in App.tsx rather than inside the store module itself, so the store doesn't have to import Sentry and more importantly and the reason for that is then, tests can create the store without triggering any Sentry calls (more testable). I think it would set a tag every time the store is initialized, including in the test helper. It felt cleaner (more readable) to set the tag in App.tsx where all of the other Sentry setup type calls are. But I can move it if I am missing something or you think it makes more sense.

cimigree: https://github.com/digidem/comapeo-mobile/pull/1993/commits/1bb0b5d8fd1e992ba5211f629f778785a1467294

src/frontend/screens/SetQADeviceName.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1993/commits/1bb0b5d8fd1e992ba5211f629f778785a1467294

src/frontend/AppNavigator.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1993/commits/1bb0b5d8fd1e992ba5211f629f778785a1467294

cimigree: ANd https://github.com/digidem/comapeo-mobile/pull/1993/commits/e4ef67c3280861d3c0605528180712320e351b0d (missed some earlier)

tests/integration/helpers/react.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/1993/commits/1bb0b5d8fd1e992ba5211f629f778785a1467294

cimigree:
I had to add a keyboard avoiding view for the edit screen.
Here is the new text.
image

src/frontend/App.tsx
ErikSin: I think we can get rid of all this code, now that its in the setter, there is no point on checking on every open of the app.

Unless Sentry requires us to set a tag for every session?

ErikSin: Also Sentry is never initialized in our testing environment so all sentry calls are just no-op function in jest anyway.

Sentry is also a global export, so i think just importing it directly in the Store Context is alot simpler/cleaner than passing it as a function. I am fairly certain you can just go

```ts
//In QADeviceNameStoreContext
import * as Sentry from @senty/react-native

...restOfContext
const actions: QADeviceNameActions = {
    setQADeviceName: (name: string) => {
      store.setState({qaDeviceName: name});
      Sentry.setTag('QA_Device_Name', name)
    },

```

ErikSin (approved):
i think you can simplify the tag initialization, but not a blocking comment

src/frontend/App.tsx
cimigree: > Unless Sentry requires us to set a tag for every session?

Yes, it does.

PR #1995, opened 2026-06-29
chore: bump default categories to latest

ErikSin:
Uses latest default categories, in preparation for adding Saamaka translations.


cimigree (approved):
Looks good. I loaded and tested basic observation flow and all was good.

PR #2000, opened 2026-06-30
fix: switch to RN native ActivityIndicator

gmaclennan:
Replaces the unmaintained `react-native-indicators` spinner with React Native's built-in `ActivityIndicator`.

`react-native-indicators` is unmaintained and incompatible with the New Architecture — it was causing rendering issues and pinning CPU to 100%. A new `LoadingIndicator` shared component wraps `ActivityIndicator` and all call sites switch to it.

Note: this is a visual change. The spinner is now the platform-native Material Design spinner on Android.

Split out of #1994 to reduce that PR's review surface; #1994 is stacked on top of this branch.


cimigree:
Do you think design should weigh on in the new look?
The color? The size? The new behavior? 


https://github.com/user-attachments/assets/358943aa-eed5-4ee3-b77c-3abe9f6ebff4

gmaclennan:
> Do you think design should weigh on in the new look? The color? The size? The new behavior?
> 
Yes, good idea, can you consult? Is the behavior different? I think color and size we can tweak immediately. The work of fixing the old spinner style, or making our own spinner to match the style of the old one, is too much work short-term unless we feel really strongly about having our own spinner.

cimigree:
> > Do you think design should weigh on in the new look? The color? The size? The new behavior?
> 
> Yes, good idea, can you consult? Is the behavior different? I think color and size we can tweak immediately. The work of fixing the old spinner style, or making our own spinner to match the style of the old one, is too much work short-term unless we feel really strongly about having our own spinner.

I asked and am waiting on a response as of July 21.

cimigree (approved):
@ErikSin or @gmaclennan  do you want to look at the changes I made? I set the default color to comapeo-blue as Aldo suggested, fixed some missing usages of the new Loading Indicator and merged with develop. Earlier I had gone through and looked at all of the places it is used and they all seemed to do the trick as far as I could tell. I just feel a little weird approving code that I wrote? But it is only a bit at the end.

ErikSin:
@cimigree Do you think it is a good idea to maintain a number size for android, and then a different prop for iOS? If so, then i think we need to create the component that automatically maintains that, so we don't have to do that for each instance of it.

IMO i think we should try to use `small` and `large` for android as well, so we don't have to maintain it. But I think you have more context than i do.

cimigree:
> @cimigree Do you think it is a good idea to maintain a number size for android, and then a different prop for iOS? If so, then i think we need to create the component that automatically maintains that, so we don't have to do that for each instance of it.
> 
> IMO i think we should try to use `small` and `large` for android as well, so we don't have to maintain it. But I think you have more context than i do.

I agree. I think small and large look fine to me and if and until design has a strong opinion otherwise, than I would rather just have one look/ one set of code for ios and android. There will be enough places where we _need_ to have different code, so here this seems to work to share it.

ErikSin:
> I agree. I think small and large look fine to me and if and until design has a strong opinion otherwise, than I would rather just have one look/ one set of code for ios and android. There will be enough places where we _need_ to have different code, so here this seems to work to share it.

Oh I see that is already happening. Were passing a size, but then forcing that into a "small" and "large". I didn't fully understand the code.

I think we should just get rid of that and just have the prop "Large" or "small", and make that part of this refactor. It will just make it more consistent, and less confusing instead of using a number that is then turned into an enum

cimigree:
> Oh I see that is already happening. Were passing a size, but then forcing that into a "small" and "large". I didn't fully understand the code.
> 
> I think we should just get rid of that and just have the prop "Large" or "small", and make that part of this refactor. It will just make it more consistent, and less confusing instead of using a number that is then turned into an enum

https://github.com/digidem/comapeo-mobile/pull/2000/commits/fae79fb754d8939714653d9d46ccf72b90f7da87

src/frontend/sharedComponents/LoadingIndicator.tsx
```diff
@@ -0,0 +1,14 @@
+import {ActivityIndicator, type ActivityIndicatorProps} from 'react-native';
+import {COMAPEO_BLUE} from '../lib/styles';
+
+type Props = ActivityIndicatorProps;
+
+// React Native's ActivityIndicator renders the platform's native indeterminate
+// spinner, animated by the OS view itself — no JS, Animated, or reanimated work
+// per frame. It replaces react-native-indicators, which is unmaintained and
+// whose radial-transform layout collapses to a line under the New Architecture.
+export const LoadingIndicator = ({
```
ErikSin, line 18: Can you add a jsdoc comment explicitly saying that defaults to large and defaults to `COMAPEO_BLUE`?

src/frontend/sharedComponents/FullScreenCenteredLoader.tsx
ErikSin, line 1: We should rename this to something that more obviously separates this from the `LoadingIndicator`.

Perhaps `FullScreenCenteredLoader` or something along those lines

src/frontend/sharedComponents/Loading.tsx
```diff
@@ -1,25 +1,19 @@
...
+import {LoadingIndicator} from './LoadingIndicator';
 
 export const Loading = ({
   color,
   size,
   style,
 }: {
   color?: string;
-  size?: number;
+  size?: 'small' | 'large';
```
ErikSin: i dont think we should allow a size prop here. the one place where it is `small` is the `BackgroundMapsScreen` and it is almost hard to see. I think this should always be Large, and we should not even allow any sizing prop

src/frontend/sharedComponents/LoadingIndicator.tsx
```diff
@@ -0,0 +1,14 @@
...
+  size = 'large',
+  color = COMAPEO_BLUE,
+  ...props
+}: Props) => <ActivityIndicator size={size} color={color} {...props} />;
```
ErikSin: Some of the components using this are turning off the animation for the e2e tests. Should we just not turn of ALL the ActivityIndicators in e2e, instead of doing it for each disparate component? In otherwords, just control that prop directly in this. That would simplify the code and make the e2e faster all around, no?

ErikSin (changes requested):
I have one blocking comment about the e2e test animations. It would simplify the code in many place, and make e2e overall faster if we just turned off the animations at the component level instead of the consuming level

src/frontend/sharedComponents/LoadingIndicator.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2000/commits/295dfeac4b2aae52a2f154c20a48a818e8204ff9

src/frontend/sharedComponents/Loading.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2000/commits/295dfeac4b2aae52a2f154c20a48a818e8204ff9

src/frontend/sharedComponents/FullScreenCenteredLoader.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2000/commits/295dfeac4b2aae52a2f154c20a48a818e8204ff9

ErikSin (approved):
Looks good! thanks

PR #2003, opened 2026-06-30
enhancement: adds sentry replay

### LINKED ISSUE #1970: Turn on Sentry Session Replay for Sentry in QA BUild
[Set up session replays for ReleaseCandidate and preRelease builds](https://docs.sentry.io/platforms/react-native/session-replay/)

To Do
- [ ] Enable Session Replays for QA builds
- [ ] [Replay Captures on Error](https://docs.sentry.io/platforms/react-native/session-replay/#replay-captures-on-errors-only) only?

cimigree:
closes #1970 

- Enables Sentry replay on QA builds on error only. This is for release-candidates not pre-release as currently configured.
- I am passing the option to not mask anything since it is just for QA.
- The Map will probably capture as black screen until we can implement captureSurfaceViews as an option, but that is not available until [version 8.12](https://github.com/getsentry/sentry-react-native/releases/tag/8.12.0) and we are on version 8.0

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -81,7 +81,6 @@ Sentry.init({
   debug: false, // this added alot of unneccesary noise to the console.
   initialScope: {user: {id: sentryUserId}},
   enableMetrics: false,
-  replaysSessionSampleRate: 0,
   replaysOnErrorSampleRate: sentryEnvironment === 'qa' ? 1.0 : 0,
   integrations:
     sentryEnvironment === 'qa'
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -73,8 +73,8 @@ const sentryEnvironment: SentryEnvironment =
 
 const appMetricsOptIn = sentryEnvironment !== 'production';
 let navigationIntegration:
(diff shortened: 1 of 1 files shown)


src/frontend/App.tsx
```diff
@@ -85,6 +85,17 @@ Sentry.init({
   debug: false, // this added alot of unneccesary noise to the console.
   initialScope: {user: {id: sentryUserId}},
   enableMetrics: false,
+  replaysOnErrorSampleRate: sentryEnvironment === 'qa' ? 1.0 : 0,
```
ErikSin, line 89: we have up to 100K replays before they charge us, so I think we can also set the `replaysSessionSampleRate` to `1.0` when the `sentryEnvironment === 'qa'`

ErikSin (changes requested):
Just one thing to add!

src/frontend/App.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2003/commits/c7a586353efa15fd27086c90b75882b71680ffbf

ErikSin (approved):

PR #2006, opened 2026-07-02
feat: date input

### LINKED ISSUE #1817: Input Formats (date)
**Description**
Users want to add a “details” question format that allows them to input a date (

Image

Image

https://www.notion.so/digidem/Input-Formats-date-2e71b08162d580748b86e00cba090910

ErikSin:
closes #1817 

Much of the work for this exists outside of this repo. comapeocat had to be updated to include date as a new field. And then a new category set had to be created to use this field.

The other major complication was internationalization. react-intl has the months and days of most languages already translated. But it does not support some of the languages that we support, so I had to add translation strings in order for us to allow us to translate it in crowdin. But because most of the languages are already supported, i wanted to show the translated strings from react-intl if they were not available via crowdin. The added complication is that we have fallback languages, so i needed to check the fallback language translation as well, but i wanted to make sure that we check the react-intl translations of the first languages before trying to check crowdin of the fallback language. For example, if the user sets the language to Gujarati and has a fallback language of Bahasa, the order of operation is
1. check crowdin for Gujarati
2. (if no translation available from the step above) check react-intl for Gujarti
3. (if no translation available from the step above) check crowdin for Bahasa
4. (if no translation available from the step above) check react-intl for Bahasa.
5. ...continue steps 3 and 4 for each fall back langauge
6. if still no translations, resort to using the english translations

This order of operation matters because we want the crowdin translations to take priorty over the react-intl translations (if a group has put the effort into translating a local dialect for a training, we want that to be used). BUT we don't want the crowdin translations of a fallback language to take precedent over the react-intl translations of a priority language. If we had Bahasa translations in crowdin but no Gujarati translations in crowdin, the user still should see the Gujarati translations if it is avaialble in react-intl.

The other complication is that we need to check the crowdin translation for each language to see if they are available. Because of the nature of React-Intl we are merging all the translations of the chosen language, the fall back languages, and english into one object. This object guarantees that ALL strings are available, even if that string is just in english. So for example, if I had Gujarti and Bahasa as my chosen languages, all the available Gujarati translation would be in this object, then all the available Bahasa strings (that are not available in Gujarti) would be in this object, and the rest of the string would be in english. But this object is not aware of what string is in what language. So if i was to check this object for the existence of the month or date, it would always return something which is not helpful for determining if the translation is available.

In order to check, i had to break up the language useQueries and no longer combine the results right away. Based on the separated query results i can check each language to see if the translation is available in that language. And then i I can combine them like it was combined before to be provided by the app. Previously the useQueries combined them automatically 

### Testing assets
Category file with the date picker can be found in this folder, called `comapeo-cat-with-date.comapeocat`.
https://drive.google.com/drive/u/0/folders/1lqd-v6yhxQBq6LL5DBLM9nyYLTPAM4f3

The date picker is the last question in the observation details of the Category `Tree`


gmaclennan:
great work on this, internationalization can be really hard. One note: it's Hermes (not react-intl) that's providing the `Intl` API, which is what gives the translations of day and month names (via the `Intl.Datetimeformat`), and Hermes have a limited subset of languages - it's not just indigenous languages it's missing. Formatjs have [polyfills](https://formatjs.github.io/docs/guides/react-native-hermes/) for many more locales than Hermes supports, so we could include the locale polyfills for `@formatjs/intl-datetimeformat` like we do already for `@formatjs/intl-relativetimeformat` in  `./scripts/build-translations.mjs`. That would ensure coverage (without needing to wait for translators) for more languages, and crowdin can be the fallback for languages not covered by the polyfills.

src/frontend/screens/ObservationFields/Date.tsx
```diff
@@ -0,0 +1,139 @@
...
+              name={direction === 'left' ? 'chevron-back' : 'chevron-forward'}
+              size={20}
+              color={NEW_DARK_GREY}
+            />
+          )}
+          onDayPress={handleDayPress}
+          markedDates={
+            valueAsDate
+              ? {
+                  [valueAsDate.slice(0, 10)]: {
```
cimigree: I think this might be problematic for people significantly ahead of UTC, like in Indonesia (UTC+7 to UTC+9). Here's how I think it will work:

new Date("2025-07-02T00:00:00").toISOString()
→ local midnight July 2nd in UTC+7 = July 1st 17:00 UTC
Stored value: "2025-07-01T17:00:00.000Z"
valueAsDate.slice(0, 10) = "2025-07-01" → July 1st is highlighted, not July 2nd

src/frontend/screens/ObservationFields/Date.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  return (
+    <>
+      <View style={styles.container}>
+        {valueAsDate && (
+          <View style={styles.labelContainer}>
+            <Text style={styles.labelText}>{formatMessage(m.date)}</Text>
+          </View>
+        )}
+        <TouchableOpacity
```
cimigree, line 75: I thought we liked to use Pressable with tappable text labels?

```
<Pressable
  onPress={() => setCalendarVisible(val => !val)}
  testID="OBS.DateInput"
  hitSlop={{top: 12, bottom: 12, left: 12, right: 12}}>
```

src/frontend/screens/ObservationFields/Date.tsx
```diff
@@ -0,0 +1,139 @@
...
+            {valueAsDate
+              ? formatDate(valueAsDate, {
+                  dateStyle: 'medium',
+                })
+              : formatMessage(m.date)}
+          </Text>
+        </TouchableOpacity>
+      </View>
+      {calendarVisible && (
+        <Calendar
```
cimigree, line 89: I am able to select a future date here, which is supposed to be greyed our and blocked, right?
I think there is a maxDate prop...

Image

src/frontend/screens/ObservationFields/Date.tsx
```diff
@@ -0,0 +1,139 @@
+import {useState} from 'react';
```
cimigree: Was it too difficult to do the dismiss on outside tap as the Notion requested? "Tapping anywhere outside of the date picker closes it"

src/frontend/lib/intl.ts
```diff
@@ -101,3 +110,205 @@ function extractLanguageCode(languageTag: string): string {
...
+ */
+
+function getMonth({
+  month,
+  translatedMessages,
+}: {
+  month: Month;
+  translatedMessages: TranslatedMessages;
+}): string {
+  const cache = createIntlCache();
```
cimigree: Each of the calls (getMonth, getShortDay, getLongDay) creates a new cache that is discarded after the function returns, so I am not sure what the benefit is.
Is it possible to do a module-level cache shared across all three functions (and across the 12 month + 14 day calls inside configureCalendarLocale) instead?

src/frontend/sharedComponents/FormattedData.tsx
```diff
@@ -115,3 +124,10 @@ export function getFieldAnswerText({
 
   return undefined;
 }
+
+export function isISODateString(val: unknown): val is string {
```
cimigree, line 128: If you add an additional check such as `&& !isNaN(new Date(val).getTime())` it would guarantee against weirdness that fits the shape of a date but isn't a date such as "2024-99-99T00:00:00Z"

src/frontend/screens/ObservationFields/Date.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ *
+ * @description the `tagValue`can be any type (not a date), but this component will only update it to date
+ *
+ */
+export const DatePicker = ({tagValue, updateTag}: DatePickerProps) => {
+  const [calendarVisible, setCalendarVisible] = useState(false);
+  const {formatMessage, formatDate} = useIntl();
+  const valueAsDate = isISODateString(tagValue) ? tagValue : null;
```
cimigree, line 40: The Notion says "Today's Date is automatically filled by default" but valueAsDate starts as null when tagValue is undefined, so the field shows the "Date" placeholder instead of today's date. Is there a reason for that?

cimigree (changes requested):
Woohoo way to go! The date storage, display, and deselect logic is clean and the tests are very helpful and thorough! 
I left a few comments: 
the most important blocking ones are 
1. that future dates aren't blocked, 
2. the calendar highlight can show the wrong day in UTC+ timezones like Indonesia, and 3. the default value doesn't match what it says in Notion (today's date should be pre-filled) - but not sure if this one is a blocker. 
I also added some comments and questions that you can take or leave, like with TouchableOpacity vs Pressable and the missing ability to dismiss if you tap outside the calendar behavior.  I thought the language logic seemed well done, but I will defer to Gregor on that!

ErikSin:
> great work on this, internationalization can be really hard. One note: it's Hermes (not react-intl) that's providing the `Intl` API, which is what gives the translations of day and month names (via the `Intl.Datetimeformat`), and Hermes have a limited subset of languages - it's not just indigenous languages it's missing. Formatjs have [polyfills](https://formatjs.github.io/docs/guides/react-native-hermes/) for many more locales than Hermes supports, so we could include the locale polyfills for `@formatjs/intl-datetimeformat` like we do already for `@formatjs/intl-relativetimeformat` in `./scripts/build-translations.mjs`. That would ensure coverage (without needing to wait for translators) for more languages, and crowdin can be the fallback for languages not covered by the polyfills.

Cool, good to know! Ill add those polyfills now!

src/frontend/screens/ObservationFields/Date.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2006/commits/b6c16a865e0672d3eee007996a413bcb27ba31d5

src/frontend/screens/ObservationFields/Date.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2006/commits/b6c16a865e0672d3eee007996a413bcb27ba31d5

src/frontend/screens/ObservationFields/Date.tsx
ErikSin: its not trivial unfortantly. But I added it to the bottom the of the screen: https://github.com/digidem/comapeo-mobile/pull/2006/commits/0912c2011fcca7eac54f3a7838d7e65441e69745

IMO it is not worth the refactor required of ALL of the fields, in order for this to work. If QA brings it up, lets deal with it then, otherwise i think just having the bottom of the screen dismiss the calendar is good enough

src/frontend/lib/intl.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2006/commits/01dec74e0c62fc212e245d299bc8f81a403eae90

src/frontend/screens/ObservationFields/Date.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2006/commits/cf810f6de81a49b62ac89bad7da6d780395f02bf

cimigree (approved):
This all looks good.
I tested manually and looked at the code.
This will be a good option to have.
Thanks for doing this; Dates and times are such a pain!

PR #2015, opened 2026-07-02
chore:update core

ErikSin:
For showing type errors in core/core react


ErikSin:
@RangerMauve It is [this hook in core-react](https://github.com/digidem/comapeo-core-react/blob/fa90b71b512023731be261beebc26aa5f302545e/src/hooks/documents.ts#L48) that is not returning null (it core 7.2.0 it was not returning null)

PR #2022, opened 2026-07-03
chore: remove usage of deprecated `useDocumentCreatedBy` hook

### LINKED ISSUE #2001: Remove use of `useDocumentCreatedBy` from `@comapeo/core-react`
The hook has been marked as deprecated in `@comapeo/core-react` since [6.3.1](https://github.com/digidem/comapeo-core-react/releases/tag/v6.3.1) and the intention is to remove it entirely at some point (see https://github.com/digidem/comapeo-core-react/pull/144).

This hook is used internally in a couple of custom hooks:

1. [`useGetCreatedBy`](https://github.com/digidem/comapeo-mobile/blob/abf262bb2d35146fb1efcc8f65dad951d1735655/src/frontend/hooks/server/useGetCreatedBy.ts)

2. [`useIsMyDocument`](https://github.com/digidem/comapeo-mobile/blob/abf262bb2d35146fb1efcc8f65dad951d1735655/src/frontend/hooks/server/useIsMyDocument.ts)

(1) is straightforward to get rid of entirely. (2) requires more work.

Noting that this is relatively non-urgent but would allow core and core-react to reduce surface area a little bit.

achou11:
Closes #2001 . Enables removal of `useGetCreatedBy` and `useIsMyDocument` custom hooks. Tried to keep changes minimal even though there were opportunities to remove some additional custom hooks. I did not test these changes on a device, although I don't expect there to be notable visible changes as the changes don't add more suspense boundaries than before (if anything, some boundaries have been removed).


src/frontend/screens/ObservationsList/TrackListItem.tsx
```diff
@@ -48,7 +48,12 @@ const TrackObservationItemNotMemoized = ({
       onPress={onPress}
       testID={testID}
       style={styles.touchable}>
-      <View style={[styles.container, style, !isMine && sharedStyles.synced]}>
+      <View
+        style={[
+          styles.container,
+          style,
+          showSyncedIndicator && sharedStyles.synced,
```
achou11, line 55: felt more appropriate to treat this as more presentational and define an additional prop that describes how it's used i.e. to display the sync indicator. consumers of the component then hold the logic for determining the prop it i.e. if the document is yours or not.

open to doing whatever is preferred.

src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -59,17 +64,23 @@ function ObservationListItemNotMemoized({
...
   const photos = observation.attachments.filter(isSavedPhoto);
-  const isMine = useIsMyDocument(observation.originalVersionId);
 
   return (
-    <View style={[styles.container, style, !isMine && sharedStyles.synced]}>
+    <View
+      style={[
+        styles.container,
+        style,
+        showSyncedIndicator && sharedStyles.synced,
```
achou11, line 82: similar comment about specifying a component prop for this as opposed to handling it in this component.

src/frontend/screens/ObservationsList/TrackListItem.tsx
cimigree: Agree. I think this behavior is more in line with our frontend guidelines/ ethos

cimigree (approved):
All looks good. 
I created data and manually tested and all worked as expected.
Thanks for doing this.

src/frontend/screens/ObservationsList/ObservationListItem.tsx
cimigree: See above

Issue #1966, opened 2026-06-17
Upgrade Developement Deps


ErikSin:
6 month development dependency upgrage. 

## To Do
- [ ] run `npm run upgrade-deps:dev`
- [ ] upgrade dev deps in src/backend

PR #2023, opened 2026-07-06
chore: upgrade eslint

cimigree:
towards #1966 

Our eslint react native config package did not support eslint 10 so we could not upgrade. The suggestion was to swap it for @react-native/eslint-config, but that package still pulls in eslint-plugin-react-native as a dependency, so it wouldn't have unblocked us. eslint-plugin-react-native is also low-maintenance (last release Dec 2024) and hasn't declared support for ESLint 10.

So instead of patching it to add ESLint 10 as a peer dependency, I removed it entirely. We were only actively using two of its rules — no-unused-styles and no-single-element-style-arrays — and had the other five turned off. 

Rather than stay stuck on ESLint 9, I:

Dropped eslint-plugin-react-native
Re-implemented no-single-element-style-arrays as a custom ESLint rule in eslint-rules/react-native.js (following the same pattern as our existing eslint-rules/intl.js), since it's small and self-contained.
Accepted losing no-unused-styles — I couldn't figure out how to make a patch file for it. It is quite complex.

After that, this PR updates all of the Eslint packages to be compatible with v10.
New rules we've now turned on:

`react-hooks/set-state-in-effect` — warns when you update state directly inside a useEffect body. We fixed two of these and suppressed one in AudioRecording where the fix would require a larger refactor that I am not sure how to do.
`react-hooks/incompatible-library `— warns when a library's API isn't compatible with the React Compiler. We suppressed this inline in EditProjectDetails where react-hook-form's watch() is used. I wanted to have visibility when we violate that rule so left it on in general.
`@eslint-react/use-state `— setter must follow the setX naming convention matching its state variable. We fixed the one violation (setShareButtonLoading → setIsShareButtonLoading) and left the rule on.

Some other rules are not relevant for us and are turned off.
Rules we're turning off (noise for React Native):

`@eslint-react/purity `— flags new Date() during render. Not relevant for React Native — this rule exists to prevent hydration mismatches in server-rendered apps, which don't apply to us.
`@eslint-react/no-unnecessary-use-prefix` — flags mock functions named use* that don't call hooks. Our mocks intentionally mirror the names of the real hooks they replace.
`@eslint-react/naming-convention-ref-name `— requires all refs to be named *Ref. Not a convention we follow, and descriptive names like camera or timer are more readable.
`@eslint-react/set-state-in-effect` — duplicate of react-hooks/set-state-in-effect (same issue, now flagged by two plugins). Turned off to avoid double-reporting.
`@eslint-react/web-api-no-leaked-timeout` — requires setTimeout to be assigned to a variable for cleanup. Not relevant for React Native (no DOM).

This rule
`preserve-caught-error` — Is a new eslint v10 built-in rule that requires attaching the original caught error as { cause: err } when re-throwing it. We turned it off because enabling it requires adding es2022.error to our TypeScript library configuration, which would mean duplicating and overriding the entire lib array from our React Native base tsconfig. It is only being thrown by one file.

**Other dependency changes required by the upgrade**
zod-validation-error added (4.0.2) — eslint-plugin-react-hooks@7.1.1 imports it via a ./v4 subpath that only exists in v4+; without pinning it we got a module resolution error.
@eslint/config-helpers added (0.6.0) — includeIgnoreFile moved here; the old import from @eslint/compat is now deprecated.
require-addon pinned to 1.1.0 in overrides — a transitive native-addon loader. Version 1.2.0 switched to conditional exports where Jest doesn't honor the "node" condition, which broke sodium-native loading and caused widespread test failures. Pinning to 1.1.0 restores the working behavior.

**Manual Testing**

- I manually tested the audio component (stops recording and saves after 5 minutes)
- the Team page (for initially connected peers)
- The sharing of an observation (for the variable name change)
- And that the input passcode wrong passcode clears and shows the error


ErikSin (approved):
Looks good! thanks for creating the custom rule!

PR #2028, opened 2026-07-08
chore: upgrade react test library

### LINKED ISSUE #1966: Upgrade Developement Deps

cimigree:
closes #1966 
based off #2023 

**Package changes...**
@testing-library/react-native 13.3.3 → 14.0.1
Added test-renderer@1.2.0 (v14's replacement for the deprecated react-test-renderer)
Removed react-test-renderer@19.1.4 

Had to update about 26? I lost count... test files

**v14 made everything async**

render / rerender / unmount → await render(...), etc. Any test callback that got a new await was made async.

fireEvent.changeText / fireEvent.press, etc. → awaited. (Note: this is different from the @testing-library/react, where fireEvent is still sync and this caused a need to adjust the eslint config (see below))

renderHook → also need to be awaited

renderAsync/unmountAsync → folded into async render/unmount in the integration helper

In v14, renderHook's result.current is not available synchronously. So any act(() => { someHook.result.current.doThing() }) threw Cannot read properties of null. The fix was converting those to await act(async () => { ... }). This affected ~53 act blocks across the tests.

v14 types fireEvent.changeText's text parameter as string. Two tests deliberately pass non-string values (null, {foo: 'bar'}) to verify the field handles bad input — I cast those with as unknown as string and left a comment explaining the intent, rather than change what the tests check. 

**ESLint rule adjustments**
eslint-plugin-testing-library@7.16.2 (the latest release) hasn't caught up to RN v14's async fireEvent — its react preset still assumes fireEvent is synchronous. Two rules had to be adjusted so linting didn't fail in the tests.

**Tests**
All the tests pass locally and the only one that is not passing here is the usual exchange one.


ErikSin (approved):
Thanks for doing this, this looked insanely tedious to do. But also, the async rendering makes so much more sense for react testing library! we had to circumvent so much of suspense rendering before it was async, so im glad they made it async by default.

PR #2032, opened 2026-07-15
feat: hypercore frontend

### LINKED ISSUE #1904: Hypercore update
**Description**
Prompt the user to update hypercore  (database storage) to the latest version - requires copying user data into the new storage format.

See details : https://www.notion.so/digidem/Update-hypercore-2e71b08162d580c39ff8ff497bc7032f?source=copy_link

### Update CoMapeo — Not Enough Space (when applicable)

Image

### Updating CoMapeo (when enough space available)

Image

### Out of storage space (Error)

Image

### Update Complete!

Image

cimigree:
closes #1904 

image

image

image

image


src/backend/src/app.js
```diff
@@ -40,6 +40,54 @@ const log = debug('mapeo:app')
...
+  if (reason instanceof Error) return reason
+  return new Error(typeof reason === 'string' ? reason : 'unknown rejection')
+}
+
+/**
+ * @param {unknown} message
+ * @returns {{ forceSkipMigrate: boolean, availableDiskSpace?: number }}
+ */
+function parseServerRestartMessage(message) {
+  if (typeof message !== 'object' || message === null) {
```
RangerMauve: Is this ever actually going to happen? Can we just assume the message type instead?

src/frontend/lib/retryServerStart.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Handled by the 'server:restart' listener in src/backend/src/app.js — the
+ * Node runtime can only start once per app process, so retrying re-runs the
+ * backend's init in-process.
+ */
+export function retryServerStart({
+  forceSkipMigrate,
+}: {
+  forceSkipMigrate: boolean;
```
RangerMauve: Is this ever going to be used with `forceSkipMigrate: false`? If not I think we should just omit it.

src/backend/src/app.js
```diff
@@ -74,15 +117,16 @@ process.on('exit', (code) => {
  * @param {string} options.defaultConfigPath
  * @param {boolean} options.forceSkipMigrate
  */
-export async function init({
```
RangerMauve, line 68: Was this change necessary? Maybe revert it if not.

src/backend/src/app.js
```diff
@@ -95,46 +139,55 @@ export async function init({
...
+        }
+        serverStatus.setState('STARTING')
+      } catch (reason) {
+        serverStatus.setState('MIGRATION_ERROR', { error: asError(reason) })
+        return
+      }
+    } else {
+      if (reason === MIGRATION_REASON_NO_SPACE) {
+        serverStatus.setState('LOW_SPACE')
+        initOptionsToRetryAfterLowSpace = options
```
RangerMauve: We should pass all these params into the RPC method args instead. We should avoid global state like this in the backend.

RangerMauve (changes requested):
I'll add some commits in here for the needed space flag on low space

src/frontend/lib/retryServerStart.ts
cimigree: Yes, when we restart after coming back from freeing up space in storage.

src/backend/src/app.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/2032/commits/25e3dac55b51dc849cc488171b63102482602cbf

src/backend/src/app.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/2032/commits/25e3dac55b51dc849cc488171b63102482602cbf

src/backend/src/app.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/2032/commits/25e3dac55b51dc849cc488171b63102482602cbf

src/backend/src/app.js
```diff
@@ -173,8 +171,7 @@ export async function init(options) {
       }
     } else {
       if (reason === MIGRATION_REASON_NO_SPACE) {
-        serverStatus.setState('LOW_SPACE')
-        initOptionsToRetryAfterLowSpace = options
+        serverStatus.setState('LOW_SPACE', { context: String(spaceNeeded) })
```
RangerMauve: Do we use constructors like this elsewhere? I think it'd be best to use `toString` or `spaceNeeded?.toString()`

src/backend/src/app.js
```diff
@@ -189,18 +186,27 @@ export async function init(options) {
...
+      fastify,
+      defaultConfigPath,
+      defaultIsArchiveDevice: true,
+      defaultOnlineStyleUrl: DEFAULT_ONLINE_MAP_STYLE_URL,
+      customMapPath: join(customMapsDir, DEFAULT_CUSTOM_MAP_FILE_NAME),
+    })
+  } catch (reason) {
+    // This can still fail on its own (it runs its own DB migrations), even
+    // when the storage migration above went fine. Same error screen either way.
+    serverStatus.setState('MIGRATION_ERROR', { error: asError(reason) })
```
RangerMauve: Are you sure we're getting migration errors during the constructor? If there are then that's a major bug in the backend and we need to fix it. Is there a minimal reproducable case and a stack trace I could work with?

RangerMauve (changes requested):
Looks good other than some nitpics

src/backend/src/app.js
cimigree: Yes it happened. No it wasn't a core bug (it was my own path mistake, now fixed), but I believe it is worth keeping because before this fix, any of those would leave the app stuck rendering nothing (STARTING forever) instead of showing the error screen. It's making sure whatever fails there gets surfaced to the user and to Sentry instead of hanging silently. I think it is possible that any of these things might hit that error:
A bad or corrupted rootKey passed into the manager
A missing/malformed bundled file (this is what I caused)
A corrupted or partially-written database file on the device (crashed mid-write from a previous session)
Disk-full or permission errors when the manager tries to open/create its sqlite files
A future regression in comapeo-core-old's own drizzle migration logic, or a schema mismatch if the device's DB was touched by an incompatible version at some point.
So what do you think? Ok to keep?

src/backend/src/app.js
cimigree: https://github.com/digidem/comapeo-mobile/pull/2032/commits/b3f6f8f7ead5bb0ffb03b629c68b800e53085b68

src/backend/src/app.js
RangerMauve: That's fair. Does it need to be labelled as a migration error? I assumed we were catching general errors during init.

cimigree: https://github.com/digidem/comapeo-mobile/pull/2032/commits/a44bde5891b0968646200ff9bed6b708de514853

RangerMauve (dismissed):
Backend bits look good. This will end up changing once I update core-react-native with hypercore11 support.

ErikSin (changes requested):
Just want to make sure this doesn't accidently get merged while we are waiting on the core-react-native piece, so i am "requesting a change".

PR #2033, opened 2026-07-16
fix: use updated settings language

### LINKED ISSUE #1989: System-preferred language not used after switching app language then returning to using system preference
**Describe the bug**

When a user interacts with the app language settings such that they go from `system-preferred -> specific language -> system-preferred`, the app no longer accounts for changes to the system preferred languages on subsequent reopenings of the app.

**To Reproduce**

1. Open app, go through onboarding
2. Navigate to app language settings screen
3. Select any language option
4. Select `Follow System Preferences Option`
5. Close the app
6. Open system language settings and change preferred language to some other language
7. Open the app
8. Navigate to language settings screen

You'll notice that despite the system language being whatever's selected in step (6), the app uses and displays a _previous_ system-preferred language. It seems to use the system-preferred language that was set when performing step (4).

**Expected behavior**

Using the `Follow System Preferences` option in the app should align with the system-preferred language (assuming it's a language the app supports).

**Screenshots**

This was done in an emulator. However, since it's in the dev environment, I was able to verify by logging values in the app code.

In this recording, I do the following:

1. Starting with a system-preferred language of English, I update the app language settings to Spanish, and then restore the app language setting to use the system preferences. 

2. Close the app.

3. I change the system-preferred language to Spanish.

4. I reopen the app. The app language settings screen shows that the system-preferred language is English, even though it should be Spanish.

5. I update the app language settings to French, and then restore the app language setting to use the system preferences. **Notice how the detected language gets corrected to Spanish when doing this**.

6. Close the app.

7. I change the system-preferred language to Portuguese.

8. I reopen the app. The app language settings screen shows that the system-preferred language is Spanish, even though it should be Portuguese.

https://github.com/user-attachments/assets/88dbda1d-f498-4f40-9643-60eb118edcc0

 - Device: Android Emulator
 - OS: Android
 - Version 16

**Additional context**

This issue is NOT present in a released version of the app. This issue would be part of an upcoming v13 release.

The repro steps + video are done using the [latest version of the develop branch](https://github.com/digidem/comapeo-mobile/tree/c250f3cead7e9512ebff30bfde8dd4a695ff61a0) at the time of writing. An early guess is that it's a regression caused by the changes in https://github.com/digidem/comapeo-mobile/pull/1902, but I have not done a git bisect to verify this.

**Based on testing this, the issue can be worked around by changing the app language settings back and forth. However, you'll run into the same issue each time you restart the app.**

ErikSin:
closes #1989 

When we simplified the intl provider in #1902, we updated the architecture to persist the language code, and only update the language in the app when the language is changed by user. This means that when the language is changed in the android settings, the provider doesn't get updated, causing the problem in 1902.

Initially i tried to change this in [onRehydrateStorage](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#onrehydratestorage). I tried to hydrate the store with the new language code provided by the android settings. But this was not actually updating the store properly. It seems like it is not properly documented, but due to the synchronous nature of the store, onRehydrateStorage tries to update the store BEFORE the store is initialized (As found by Claude). If we catch the error in the `onRehydrateStorage` you see that it is trying to update a store that is undefined and throwing an error. In the documentation it looks like `onRehydrateStorage` should be used to only update states outside the store.

I then used the [merge](https://zustand.docs.pmnd.rs/reference/integrations/persisting-store-data#merge) function which directly updates the store as seen in the docs. I also moved all the functions originally in the onRehydrateStorage into the `merge` function

Essentially we are checking for the language code in the android settings, and merging that language code into the store state. This means that if the android setting language has changed, that will be reflected in the provider.

One caveat is that this will only update if the app is completely closed and restarted. Trying to update the app language when the app has not restarted would require us to listen for when the app has been backgrounded, and then update the provider in a `useEffect`. While this is something that we could do, I think the complication of that code is not worth doing for a very unlikely edge case.

### DIFF
--- a/src/frontend/contexts/LocaleStoreContext.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.tsx
@@ -61,15 +61,30 @@ export function createLocaleStore({persist} = {persist: false}) {
         storage: createJSONStorage(() => MMKVStoreInitializer),
         version: 1,
         migrate,
-        // usable language tags can change between app boot ups
-        // this just makes sure the language tag is still usable
-        // if not, then use system preferences
-        onRehydrateStorage: () => state => {
-          if (!state) return;
-          const usable = getUsableLanguageTag(state.languageTag);
-          if (!usable) {
-            store.setState(createInitialState());
+        // Runs on every hydration (unlike `migrate`, which only runs on a
+        // version bump). Its return value becomes the applied state
+        // directly, so there's no need to reach for the store instance,
+        // which doesn't exist in the onRehydrateStorage.
+        merge: (persistedState, currentState) => {
+          const state = persistedState as LocaleState;
+
+          if (state.useSystemPreferences) {
+            // usable language tags can change between app boot ups;
+            // re-resolve from system preferences in case they changed
+            const systemLanguageTags = getLocales().map(l => l.languageTag);
+            return {
+              ...currentState,
+              languageTag:
+                getUsableLanguageTagFromSystemPreferences(systemLanguageTags),
+            };
           }
+
+          // checks that the language tag is still an available language
+          if (!getUsableLanguageTag(state.languageTag)) {
+            return {...currentState, ...createInitialState()};
+          }
+
+          return {...currentState, ...state};
         },
       }),
     );
--- a/src/frontend/contexts/LocaleStoreContext.tsx
+++ b/src/frontend/contexts/LocaleStoreContext.tsx
@@ -66,25 +66,32 @@ export function createLocaleStore({persist} = {persist: false}) {
         // directly, so there's no need to reach for the store instance,
         // which doesn't exist in the onRehydrateStorage.
         merge: (persistedState, currentState) => {
-          const state = persistedState as LocaleState;
+          const parseResults = v.safeParse(LocaleStateSchema, persistedState);
 
-          if (state.useSystemPreferences) {
+          if (!parseResults.success) {
+            return currentState;
+          }
+
+          const validatedPersistedState = parseResults.output;
+
+          if (validatedPersistedState.useSystemPreferences) {
             // usable language tags can change between app boot ups;
             // re-resolve from system preferences in case they changed
             const systemLanguageTags = getLocales().map(l => l.languageTag);
             return {
               ...currentState,
+              ...validatedPersistedState,
               languageTag:
                 getUsableLanguageTagFromSystemPreferences(systemLanguageTags),
             };
           }
 
           // checks that the language tag is still an available language
-          if (!getUsableLanguageTag(state.languageTag)) {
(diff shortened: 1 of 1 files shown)


src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -61,15 +61,30 @@ export function createLocaleStore({persist} = {persist: false}) {
...
-        onRehydrateStorage: () => state => {
-          if (!state) return;
-          const usable = getUsableLanguageTag(state.languageTag);
-          if (!usable) {
-            store.setState(createInitialState());
+        // Runs on every hydration (unlike `migrate`, which only runs on a
+        // version bump). Its return value becomes the applied state
+        // directly, so there's no need to reach for the store instance,
+        // which doesn't exist in the onRehydrateStorage.
+        merge: (persistedState, currentState) => {
```
cimigree, line 68: One thing I noticed. On a fresh install (uninstall + reinstall), zustand calls merge(undefined, currentState). state.useSystemPreferences then throws on undefined. The old onRehydrateStorage had `if (!state) return`; up top for exactly this situation. I think it just didn't make the jump over to merge? Or was that a deliberate omission?

It's silent (I think it is swallowed in the catch) and nothing the user can see now. But it leaves persist.hasHydrated() permanently false for this store, which possibly could cause a problem later on.

I think you could just add` if (!persistedState) return currentState;` at the top

src/frontend/contexts/LocaleStoreContext.tsx
```diff
@@ -61,15 +61,30 @@ export function createLocaleStore({persist} = {persist: false}) {
...
+          const state = persistedState as LocaleState;
+
+          if (state.useSystemPreferences) {
+            // usable language tags can change between app boot ups;
+            // re-resolve from system preferences in case they changed
+            const systemLanguageTags = getLocales().map(l => l.languageTag);
+            return {
```
cimigree, line 81: Small nit, not blocking: `useSystemPreferences` here comes from `...currentState` rather than being set explicitly. Works fine today since currentState is always createInitialState() at this point, but if we ever add a manual rehydrate() call somewhere down the line this could quietly pick up the wrong value. Might be worth just spelling out useSystemPreferences: true here for clarity, but not a big deal either way.

cimigree (approved):
Nice fix, this resolves the original bug for me on repeated system-language changes!
But there is one  issue on fresh installs. It is not blocking exactly but it is sort of blocking. I will leave that to you to decide.

ErikSin:
Your comment made me realize that i was quite naively validating, and we use valibot in order to have stronger validations. So i refactored it using valibot instead

cimigree (approved):
Nice. Sorry I forgot about that too! This is resolved much better than I suggested.

PR #2035, opened 2026-07-21
fix: track bottom sheet (android)

ErikSin:
closes #2019 

After spending a long time playing around with the app, the entering [layout animation](https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/layout-transitions) was causing problems. It seems like it was causing the bottom sheet to 1. not render properly and 2. the buttons inside the bottoms sheet were un-clickable. In order to fix this, I changed the animation to use [`useSharedValue`](https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue) instead.

Unfortunately the animation logic is slightly more complicated now, but it is working consistently.

I also [updated the babel config plugin](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/getting-started/#react-native-community-cli), as our was outdated with the v2 plugin.

PR #2037, opened 2026-07-21
fix: add permission so tracks does not crash

### LINKED ISSUE #2036: Tracks crashing on initial startup
In the [new architecture branch](https://github.com/digidem/comapeo-mobile/pull/1994), the entire app crashes and shuts down when tracks is started.

No error is caught on the JS side, which implies that it is most likely happening on the native side

ErikSin:
closes #2036.

Starting a track caused the entire app to hard-crash (FATAL EXCEPTION in TaskBroadcastReceiver, SIG 9) as soon as the first background location update was delivered.

Expo-location's background location task consumer schedules a persisted JobScheduler job on each location report. Android requires the RECEIVE_BOOT_COMPLETED permission for any persisted job request, or it throws IllegalArgumentException — and since this happens inside a native BroadcastReceiver, it's unrecoverable from JS 

Adding `android.permission.RECEIVE_BOOT_COMPLETED` to the Android permissions list in app.json fixes the issue.

This is stacked ontop of #2035 because the tracks sheet is inacessible without it.

### DIFF
--- a/app.json
+++ b/app.json
@@ -116,7 +116,8 @@
         "android.permission.FOREGROUND_SERVICE_LOCATION",
         "android.permission.CAMERA",
         "android.permission.RECORD_AUDIO",
-        "android.permission.CHANGE_WIFI_MULTICAST_STATE"
+        "android.permission.CHANGE_WIFI_MULTICAST_STATE",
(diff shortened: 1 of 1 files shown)


cimigree (approved):
Looks good. Thanks.

PR #2040, opened 2026-07-22
chore: always run expo start in dev client mode

achou11:
`expo-dev-client` was removed in https://github.com/digidem/comapeo-mobile/pull/1820 under the assumption that it isn't directly used anywhere. however - from what I can tell - running `expo start` will have different default behavior based on whether this module is installed (even if it's not configured in the `app.json`). without the module installed, the development server runs in "Expo Go" mode:

```console
› Using Expo Go
› Press s │ switch to development build
```

This mode generally does not work for our application because of the use of native modules that aren't registered with Expo Go. It's also generally recommended by Expo to run the server in development client mode for non-trivial apps.

If using `expo-dev-client` is not desirable, you can force Expo to use the development build by passing `--dev-client` to `expo start`, which is what this PR introduces. The console output will show this instead:

```console
› Using development build
› Press s │ switch to Expo Go
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "ios": "npm run build:backend && npm run ios-no-backend-rebuild",
     "ios-no-backend-rebuild": "expo run:ios",
     "prestart": "npm run build:translations && npm run build:intl-polyfills",
-    "start": "expo start",
+    "start": "expo start --dev-client",
     "lint:prettier": "prettier \"src/**/*.{js,ts,jsx,tsx}\" --check",
     "lint:eslint": "eslint --cache .",
     "lint:types": "tsc --noEmit",


ErikSin:
Thank jesus. I keep forgetting to switch modes and expo go gets un-consensually downloaded on my phone at least once a week.

ErikSin (approved):
THANKS!

achou11:
> Thank jesus. I keep forgetting to switch modes and expo go gets un-consensually downloaded on my phone at least once a week.

haha yeah I didn't realize it actually does that but makes sense.

I guess the question is whether `expo-dev-client` should be reintroduced. Maybe not a priority for now though.

ErikSin:
> > Thank jesus. I keep forgetting to switch modes and expo go gets un-consensually downloaded on my phone at least once a week.
> 
> haha yeah I didn't realize it actually does that but makes sense.
> 
> I guess the question is whether `expo-dev-client` should be reintroduced. Maybe not a priority for now though.

i did some research of expo-dev-client. I personally don't use any of the features it offers. That is not to say we shouldn't use it, because i imagine there are probably some useful features that we could use. so yeah, i agree not a priority for right now, but not out of the picture.

PR #2045, opened 2026-07-22
fix: jest-test-fixes

### LINKED ISSUE #2017: Fix Jest Tests in core-react-native-branch
The jest tests are breaking due to the new architecture. I explored a bit into why it is breaking and it so far has to do with the dependencies and imports of the new libraries. I suspect we need to mock some new things/ update certain libraries to fix jest

The Pr should be based of [this PR](https://github.com/digidem/comapeo-mobile/pull/1994) and merged into that PR (not into `develop`).

# To Do
- [ ] Investigate why jest tests are breaking
- [ ] Fix jest tests

cimigree:
closes #2017 
At first, every test file failed with "Cannot find module 'expo-modules-core'." 
This PR fixes that and all the issues that surfaced after that was resolved; now all tests (68 suites / 760 tests) pass again.

1. `expo-modules-core` was only a transitive dependency of expo and wasn't hoisted to the top level of node modules, so jest-expo's setup couldn't require() it. I pinned it as a direct dependency, with an overrides entry so its optional peer on react-native-worklets resolves to the version this app already pins (because otherwise the npm install failed).
2. `react-native-worklets` 0.10 crashes under Jest. So, I wired up the jest resolver react-native-worklets ships for this problem exactly, so tests load its JS mock instead.
3. @sentry/core and other @sentry/* sub-packages ship ESM-only and weren't in the Babel transform allowlist.
4. jest-expo now substitutes the global fetch with a stub that does nothing — this broke two tests that need to spin up a real local HTTP server and expect fetch to reach it (`sendMetricsData.test.ts` and `Exchange/index.test.tsx`, the latter via @comapeo/core's member-api adding a server peer). So I added a shared helper that swaps in [undici](https://github.com/nodejs/undici)'s real implementation for those cases. Because of using undici, jest.setup.js now needs a patch for performance.markResourceTiming, which undici reads at import time, so it has to happen before any test file's own imports run.
5. `packageJson.test.ts` compared frontend and backend package.json dependency versions; that comparison is irrelevant now that src/backend/ is gone, so I removed that test/ check.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,7 +6,7 @@ const NODE_MODULE_PATTERNS_TO_TRANSFORM = [
   '@react-native',
   '@?expo',
   '@?react-navigation',
-  '@sentry/react-native',
+  '@sentry/',
   'native-base',
   // Awana modules distributed as ESM
   '@comapeo/',
@@ -61,6 +61,7 @@ const NODE_MODULE_PATTERNS_TO_TRANSFORM = [
 const config = {
   preset: 'jest-expo',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  resolver: './node_modules/react-native-worklets/jest/resolver.js',
   // https://react-native-documents.github.io/docs/sponsor-only/jest-mocks
   setupFiles: [
     './node_modules/@react-native-documents/picker/jest/build/jest/setup.js',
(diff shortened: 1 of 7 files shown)


ErikSin (approved):
Im glad its was just a matter a configuration and that we did not need to rewrite the tests!

PR #2049, opened 2026-07-23
fix: do not specify coordinates when initializing a draft observation

### LINKED ISSUE #2048: Observation can end up at Null Island when created before a location is registered
**Describe the bug**

When creating an observation, if a location is not registered before saving the observation, the observation ends up at Null Island (0, 0).

[Relevant field support thread in Slack](https://awanadigital.slack.com/archives/CK58XTJD7/p1784722010992659)

**To Reproduce**

1. Create observation very quickly OR disable location services on the device and then create observation.
2. View observation details
3. Inset map shows observation at Null Island. Observation metadata shows lat of 0 and lon of 0

**Expected behavior**

I would expect the observation to not have any coordinates recorded for it, such that there's no inset map view displayed when viewing its details and the coordinates are omitted in the observation metadata

**Screenshots**

https://github.com/user-attachments/assets/25323dfd-ec42-4d58-bfdf-b240282ee009

Image

 - Device: Pixel 6
 - OS: Android 17

**Additional context**

The following lines are most likely the cause:

https://github.com/digidem/comapeo-mobile/blob/282162a1b132eb594fcd980a43489fa9250b8b82/src/frontend/contexts/PersistedStores/DraftObservationStore.ts#L512-L513

When creating an observation, it's initialized with lat 0 and lon 0. If there aren't any updates from the location between init and saving, then the observation ends up using those coordinates.

achou11:
Fixes #2048 

Removes the (0,0) coordinate initialization for an "empty" observation. This prevents the issue of an observation defaulting to Null Island when a location isn't registered by the time it's created. Instead, no coordinates are specified for the saved observation, which I _think_ makes more sense in this case.

---

Example observation when doing the repro steps in original issue with this change applied:

image

image

### DIFF
--- a/src/frontend/contexts/DraftObservationContext.test.tsx
+++ b/src/frontend/contexts/DraftObservationContext.test.tsx
@@ -179,8 +179,6 @@ describe('useDraftObservationActions()', () => {
       expect(stateHook.result.current.unsavedAttachments).toStrictEqual([]);
       expect(stateHook.result.current.value).toStrictEqual({
         schemaName: 'observation',
-        lat: 0,
-        lon: 0,
         metadata: {manualLocation: false},
         tags: {notes: ''},
         attachments: [],
--- a/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
+++ b/src/frontend/contexts/PersistedStores/DraftObservationStore.ts
@@ -509,8 +509,6 @@ function createEmptyStoreState(): DraftStateEmpty {
 function createEmptyObservationValue(): ObservationValueWithPreset {
   return {
     schemaName: 'observation',
-    lat: 0,
-    lon: 0,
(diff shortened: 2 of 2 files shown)


ErikSin:
@gmaclennan I copied this code from the original context you made, was there a reason you had set it to [0,0] initially?

gmaclennan:
No reason, that would be a bug in the original code.

gmaclennan:
FYI the change was from https://github.com/digidem/comapeo-mobile/pull/630 - I think I just missed this when prototyping the store. The behaviour previously in Mapeo legacy had been to have no coordinates when there is no device location, which is what this PR restores. I think that's the correct behaviour.

ErikSin (approved):
Thanks Andrew!

PR #2050, opened 2026-07-24
refactor: crnm e2e tests

cimigree:
Stacked on https://github.com/digidem/comapeo-mobile/pull/1994
[These tests now pass on Browserstack.](https://github.com/digidem/comapeo-mobile/actions/runs/30556520215/job/90924392870)

1. The bottom sheet needed additional correction on the bottom for the nav bar

2. The early access line in settings was causing a bit of trouble in testing only because it is just on the edge of the screen, which means it is in view, so no further scrolling is done. The scroll into view is still useful when I test locally and should be there in the code on the observation filter toggle test in case it goes off screen if we change which device we use on Browserstack also.

3. Early access in the settings test is a little tricky. On Browserstack, the very top edge of the Early Access "box" is visible (see screenshot from the test). But that doesn't mean the contents of it are visible on the page. In fact they are not! However, scrolling to the element using the existing test ID  just doesn't work, because the element IS in fact "visible" on the screen. The very top edge is there. So the test thinks it has been scrolled to.

To fix that in the settings test, which needs the whole early access element to be visible, I had to add a test id for the text within the box. That way, the whole accessible element is in the frame and accessible to the test rather than just the first appearance of the box which goes around that particular setting. I tried so many other ways to scroll, scroll to end, etc. and none of them worked consistently. So I know it is another parameter, but it definitely fixes the test flakiness and the test consistently not working.

Screenshot 2026-07-29 at 6 34 02 PM

4. I also had to adjust the back button clicks/ repetition of the same element. Appium seems not to be able to reassign the same variable to the same element on a different page. I have had to make that fix in other tests before so am just proactively making that fix everywhere, which may eliminate some flakiness.

5. The Exchange settings (everything/ previews) was being flaky. So I added a test id to the button which was sometimes not getting clicked which seems to make it much more reliable.

6. The main map screen is _sometimes_ taking longer to resolve now after onboarding so I had to extend the timeout there or else all of the tests end up failing.


ErikSin:
> 3. To fix that in the settings test, which needs the whole early access element to be visible, I had to scroll to the element below early settings so the whole accessible element is fetched/ resolved rather than just the first appearance of the bounding box, which is apparently what Appium does. I tried so many other ways to scroll and none of them worked consistently. Any other ideas?

Can you explain this one a little more? I don't think i fully understand what exactly the problem is

src/frontend/sharedComponents/BottomSheetWrapper.tsx
```diff
@@ -71,8 +73,9 @@ const AnimateBottomSheetContainer = ({
         <Animated.View
           style={{
             backgroundColor: WHITE,
-            padding: 20,
+            paddingHorizontal: 20,
             paddingTop: 40,
+            paddingBottom: 20 + insets.bottom,
```
ErikSin, line 78: I think this is actually an animation issue. I noticed when i was doing the thorough QA that sometimes the bottom sheet would fully show up, and sometimes the bottom would get cut off (but it wasn't consistent). I think it doesn't hurt to leave this in for now as it consistently fixes the problem and allows e2e to pass. But just saying it out loud in case you notice it as well. If so, we should fix the animation instead so there isn't inconsistent padding

cimigree:
> > 3. To fix that in the settings test, which needs the whole early access element to be visible, I had to scroll to the element below early settings so the whole accessible element is fetched/ resolved rather than just the first appearance of the bounding box, which is apparently what Appium does. I tried so many other ways to scroll and none of them worked consistently. Any other ideas?
> 
> Can you explain this one a little more? I don't think i fully understand what exactly the problem is

I updated above to this:
Early access in the settings test is a little tricky. On Browserstack, the very top edge of the Early Access "box" is visible (see screenshot from the test). But that doesn't mean the contents of it are visible on the page. In fact they are not! However, scrolling to the element using the existing test ID just doesn't work, because the element IS in fact "visible" on the screen. The very top edge is there. So the test thinks it has been scrolled to.

To fix that in the settings test, which needs the whole early access element to be visible, I had to add a test id for the text within the box. That way, the whole accessible element is in the frame and accessible to the test rather than just the first appearance of the box which goes around that particular setting. I tried so many other ways to scroll, scroll to end, etc. and none of them worked consistently. So I know it is another parameter, but it definitely fixes the test flakiness and the test consistently not working.

It looks like this before any scrolling/ swiping happens.
Screenshot 2026-07-29 at 6 34 02 PM

ErikSin (approved):
I hate that we have to rewrite the testIds and text, but i suppose it actually does make sense. Referencing them as variable by definition should only reference the one thing and was kind of surprised that js let just us do that in the first place!

Issue #2057, opened 2026-08-03
No suspense boundary in drawer menu


gmaclennan:
The [Drawer menu](../blob/develop/src/frontend/Navigation/Tab/index.tsx) contains several suspense hooks (`useManyProjects`, `useProjectSettings`, `useOwnDeviceInfo`, `useOwnRoleInProject`, and the storage query) but no suspense boundary (or error boundary). This means that if any of these queries suspends (e.g. by switching project) then the entire homescreen is replaced. This could be related to #1613, by causing a rapid unmount and re-mount of the drawer.

The fix is to put a suspense boundary in the drawer menu with a spinner. Not sure if we want an error boundary in here too?

cimigree:
On a **project switch**, the boundary that actually fires isn't inside the drawer. It's the one in `Navigation/Stack/index.tsx`, catching `useSingleProject` in `ActiveProjectProvider`, which sits *above* `HomeTabs`. 

To work out which boundary was firing, I temporarily rendered a tiny component inside each Suspense `fallback` — the one in
`Navigation/Stack/index.tsx` and the `fallback={null}` one in `AppNavigator.tsx` — that did nothing but log on mount and unmount, each with its own label. It only mounted when that boundary actually shows its fallback, so it tells you *which* boundary fired.

Result: switching to a project that isn't cached yet fired the `Navigation/Stack` boundary log every time. The `AppNavigator` one only ever fired at app start.

The thing suspending beneath that boundary is `ActiveProjectProvider` —`useSingleProject` gets a new query key on a project switch. It sits above `HomeTabs`, and a parent renders before its children, so when it suspends,`DrawerMenu` never renders at all. A boundary only catches suspends from below it, so the drawer's boundary is never in the path for this case.

Where it *would* help is when the drawer's own queries are cold or invalidated while the parent's are "warm" — the drawer suspends on its own, and today that escalates to the same top-level boundary and replaces the whole homescreen. So the boundary is still worth adding, it just won't cover the project-switch issues noted in #1613.

On #1613 — that turned out to be a separate problem and is fixed in #2061. The suspense hide/reveal was the **trigger**, but the actual failure was the drawer container being laid out with a width of 0. So these two issues/ PRs can be decoupled; this one can stand on its own.

[...]

PR #2058, opened 2026-08-03
fix: fix blank drawer menu (experimental)

### LINKED ISSUE #1613: Blank Drawer Menu
The drawer-menu is sometimes showing up blank SOMETIMES after a user gets removed from a project Or gets invited to a new project. I have determined one way to consistently recreate the issue (See below). 

BUT It is also happening sometime after a user either gets removed from a project OR they get invited to a new project - but not consistently. I have tried and recorded several different combinations of user interactions and have not been able to determine a way to consistently recreate the issue outside of the one flow below. It is happening not that often.

If the user closes the app and opens it again the menu will show up again. 

### How to consistently recreate the issues
1. Device A is on v6.2 or less and renames their solo/default project. Now device A has no default project
2. Device B invites Device A to a project
3. Device A switches to another project
4. Device B removes device A from the project
5. Device A does NOT switch back to the project they have been removed from, closes the app, and upgrades to v8.2 or higher
6. Device A open the app (now on v8.2) and switches to the project they have been removed from. They will get notified that they have been removed. When they open the menu it is blank

** Steps 3 and 6 are critical in order to recreate this issue

**Video Recreating Step above**

https://github.com/user-attachments/assets/cf9e5ad6-50ed-4237-a1b2-14134da6ae1d

## Screen Shot:

Image

gmaclennan:
Fixes #1613

Ran a Claude Fable pass on this, and nothing from our code was showing any suspects for causing this bug. The one issue identified was the `removeClippedSubviews` setting on Android, which could cause the drawer contents not to be re-attached in certain circumstances when there is a race on re-rendering, which could happen on a project switch, related to #2057 which causes the drawer to be removed and re-rendered during the suspense phase from when a project is switched.

### DIFF
--- a/patches/README.md
+++ b/patches/README.md
@@ -80,3 +80,9 @@ See: [Reviewer context](https://github.com/digidem/comapeo-mobile/pull/1225).
 
 Modifies Expo's native fetch implementation to stream files directly to the fetch request rather than loading them into memory. This patch enables streaming uploads by modifying the native Android code to read and transmit files in chunks. This prevents out-of-memory (OOM) errors when uploading large files (e.g., over 200MB).
 Refer to this [issue](https://github.com/expo/expo/issues) and this [PR](https://github.com/expo/expo/pull) which have been created to address this limitation in Expo's fetch implementation.
+
+## `react-native-drawer-layout`
+
+### [Experimental fix for empty drawer on Android](./react-native-drawer-layout+4.2.4.patch)
+
+Hard-to-reproduce issue where the drawer is empty on Android. It's possible that this is caused by `removeClippedSubviews` being set to `true` on Android, which could result in the drawer not being re-attached in some circumstances. The cost of setting `removeClippedSubviews` to `false` is that the drawer will always be rendered, even when it's not visible, which should have minimal impact on performance given that the drawer is small and there is only one of them.
--- a/patches/react-native-drawer-layout+4.2.4.patch
+++ b/patches/react-native-drawer-layout+4.2.4.patch
@@ -1,0 +1,26 @@
+diff --git a/node_modules/react-native-drawer-layout/lib/module/views/Drawer.native.js b/node_modules/react-native-drawer-layout/lib/module/views/Drawer.native.js
+index 5fb761c..4a7face 100644
+--- a/node_modules/react-native-drawer-layout/lib/module/views/Drawer.native.js
++++ b/node_modules/react-native-drawer-layout/lib/module/views/Drawer.native.js
+@@ -296,7 +296,7 @@ export function Drawer({
+                 accessibilityLabel: overlayAccessibilityLabel
+               }) : null]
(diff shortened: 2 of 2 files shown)


cimigree:
Thanks for attempting to solve this annoying and non deterministic problem, and for writing up this up — the suspense angle is what pointed me in the right direction for the fix I applied here: #2061 

I added a ton of logs for the drawer and caught the blank drawer in the act: on a project switch (which I captured during invitation to a new project) the suspense boundary in `Navigation/Stack/index.tsx` hides and re-shows the tree, and the drawer is torn down and re-mounted exactly as you described in #2057 .

When I captured the native view tree with `adb shell dumpsys activity top` during a live failure, the drawer container was still there — a `ReactViewGroup`, still marked visible, laid out at `0,0-0,2256` — with **zero child views**. Not children rendered at zero width; no children in the tree at all. Everything else on screen was laid out normally at 1080. Which is what you were seeing.

With `removeClippedSubviews` on, Android keeps a rectangle describing the visible area of a container and detaches any child falling outside it — an optimization for long lists. A container that's 0 wide has a rectangle with no area, so every child falls outside it and gets detached. That's why the drawer's contents were missing outright rather than just invisible.

**But** the ordering is what makes it downstream. Clipping only decides which children are attached to a native view — it can't change the parent's own measured size. The container measured 0 wide on its own, and the detachment followed from that. So turning clipping off re-attaches the children into a container that's still 0 wide, and the drawer will stay blank.  A rectangle with no area is a consequence of the container being 0 wide, not a cause of it.

Forcing the container's width to 0 reproduces the blank drawer on its own, with clipping left exactly as it is today.

So I think the fix ends up being one prop on `<Drawer>` in `Navigation/Tab`, no patch needed (see PR mentioned above).

So, I think this PR can be closed, but I'd rather leave that call to you or @ErikSin in case you want to keep the patch for other reasons.

ErikSin:
> So, I think this PR can be closed, but I'd rather leave that call to you or @ErikSin in case you want to keep the patch for other reasons.

@cimigree is there a specific other reason that you are referring too here?

cimigree:
> > So, I think this PR can be closed, but I'd rather leave that call to you or @ErikSin in case you want to keep the patch for other reasons.
> 
> @cimigree is there a specific other reason that you are referring too here?

No I meant some reason I don't know about or can't see.

ErikSin:
I have no context outside of this PR and what you have written up @cimigree. So i have nothing to add as why we should keep this PR. If you think we should close this PR, then lets close it!

gmaclennan:
Agreed, Cindy's reasoning seems sound and this was only a shot-in-the-dark experiment to try to fix it. Thanks for doing the hard work with the detailed debugging Cindy! We can close this.

gmaclennan:
Agreed, Cindy's reasoning seems sound and this was only a shot-in-the-dark experiment to try to fix it. Thanks for doing the hard work with the detailed debugging Cindy! We can close this.

PR #2061, opened 2026-08-04
fix: blank drawer

### LINKED ISSUE #1613: Blank Drawer Menu

cimigree:
closes #1613 
## Summary

Fixes when the drawer menu sometimes opens blank.

It happens after switching projects or accepting an invite. Once it happens, the drawer stays blank until the app is restarted.

**Cause:** the sliding container that holds the drawer menu gets laid out as **0 wide**. The menu is still there, it just has no designated space to show itself.

Why it goes to zero: `react-native-drawer-layout` caps the panel with `maxWidth: '100%'`. That is a percentage, measured against the parent. During a project switch, a Suspense boundary above the drawer briefly hides and re-shows the whole screen. If the panel is measured in that moment, the parent is zero wide — so 100% of it is zero.

It never recovers because nothing tells React to re-measure. 

## The Fix

One prop on `<Drawer>`:

```tsx
drawerStyle={{width: drawerWidth, maxWidth: drawerWidth}}
```

A fixed number cannot be a percentage of zero.

`width` is passed too because the library uses that same number to decide how far the map slides when the drawer opens. Passing it keeps the two the same.

### How I know this is the right fix

Through the magic of AI and luck (I was able to reproduce the blank drawer a few times) I captured the native view tree on a real device while the drawer was blank. 

It showed the panel was present but had a **width of 0 with zero children**. No crash, no error.

I then forced a re-measure with `adb shell wm size` (no restart). The drawer's width jumped from **0 back to its normal 304** and the whole menu reappeared instantly 

I modified `node_modules` and forced the failure by setting the library's max width to 0%. Then it was a blank drawer every time, with the same numbers as the real bug (the drawer container's measured size - specifically 0 wide × 752 tall (in dp)).

Then I re-ran that last bit (max width set to 0%) with this fix in this PR applied — and the drawer rendered normally. Same setup as was causing the blank drawer; only the fix changed.

This modification of the `node_modules` makes the bug (blank drawer) happen on demand, so the fix is proven rather than guessed.

### Suggestions for testing / verifying
The drawer should look exactly the same as before. Any visible difference is a bug.
I checked on a Pixel fold because I was worried about differences on a large screen but in this branch and develop the drawer looked exactly the same.

### DIFF
--- a/src/frontend/Navigation/Tab/index.tsx
+++ b/src/frontend/Navigation/Tab/index.tsx
@@ -23,15 +23,16 @@ import {useAuthContext} from '../../contexts/AuthContext';
 
 const Tab = createBottomTabNavigator<HomeTabsParamsList>();
 
-const APPROX_APP_BAR_HEIGHT = 56;
+// These mirror react-native-drawer-layout's own defaults, so the drawer keeps
+// exactly the size it had before. The gap leaves a strip of screen uncovered
+// beside the open drawer so it can be tapped to dismiss — the library calls it
+// APPROX_APP_BAR_HEIGHT. 360 is the Material Design 3 standard drawer width:
+// https://m3.material.io/components/navigation-drawer/specs
+const DRAWER_EDGE_GAP = 56;
 const MAX_DRAWER_WIDTH = 360;
 
-// Mirrors react-native-drawer-layout's default so the drawer keeps its usual size.
 function getDrawerWidth(windowWidth: number) {
-  return Math.max(
-    Math.min(windowWidth - APPROX_APP_BAR_HEIGHT, MAX_DRAWER_WIDTH),
-    0,
-  );
+  return Math.max(Math.min(windowWidth - DRAWER_EDGE_GAP, MAX_DRAWER_WIDTH), 0);
 }
 
 export const HomeTabs = ({navigation}: NativeRootNavigationProps<'Home'>) => {
--- a/src/frontend/Navigation/Tab/index.tsx
+++ b/src/frontend/Navigation/Tab/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {TouchableOpacity} from 'react-native';
+import {TouchableOpacity, useWindowDimensions} from 'react-native';
 import {createBottomTabNavigator} from '@react-navigation/bottom-tabs';
 import {CameraScreen} from '../../screens/CameraScreen';
 import {MapScreen} from '../../screens/MapScreen';
@@ -23,8 +23,20 @@ import {useAuthContext} from '../../contexts/AuthContext';
 
 const Tab = createBottomTabNavigator<HomeTabsParamsList>();
 
+const APPROX_APP_BAR_HEIGHT = 56;
+const MAX_DRAWER_WIDTH = 360;
+
+// Mirrors react-native-drawer-layout's default so the drawer keeps its usual size.
+function getDrawerWidth(windowWidth: number) {
+  return Math.max(
+    Math.min(windowWidth - APPROX_APP_BAR_HEIGHT, MAX_DRAWER_WIDTH),
+    0,
+  );
+}
+
 export const HomeTabs = ({navigation}: NativeRootNavigationProps<'Home'>) => {
   const [drawerOpen, setDrawerOpen] = useOpenDrawer();
+  const drawerWidth = getDrawerWidth(useWindowDimensions().width);
 
   return (
     <>
@@ -39,6 +51,11 @@ export const HomeTabs = ({navigation}: NativeRootNavigationProps<'Home'>) => {
         }}
         drawerType="slide"
         swipeEnabled={false}
+        // The library's `maxWidth: '100%'` resolves to 0 if the panel lays out
+        // while an ancestor is momentarily zero-width — which a suspense
+        // hide/reveal on project switch does, blanking the drawer until restart
+        // (#1613). `width` matches so the library's drawerWidth stays in sync.
+        drawerStyle={{width: drawerWidth, maxWidth: drawerWidth}}
         renderDrawerContent={() => (
           <DrawerMenu
             closeMenu={() => {
--- a/src/frontend/Navigation/Tab/index.tsx
+++ b/src/frontend/Navigation/Tab/index.tsx
@@ -26,9 +26,10 @@ const Tab = createBottomTabNavigator<HomeTabsParamsList>();
 // These mirror react-native-drawer-layout's own defaults, so the drawer keeps
 // exactly the size it had before. The gap leaves a strip of screen uncovered
 // beside the open drawer so it can be tapped to dismiss — the library calls it
-// APPROX_APP_BAR_HEIGHT. 360 is the Material Design 3 standard drawer width:
-// https://m3.material.io/components/navigation-drawer/specs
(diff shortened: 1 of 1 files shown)


src/frontend/Navigation/Tab/index.tsx
```diff
@@ -23,8 +23,20 @@ import {useAuthContext} from '../../contexts/AuthContext';
 
 const Tab = createBottomTabNavigator<HomeTabsParamsList>();
 
+const APPROX_APP_BAR_HEIGHT = 56;
```
ErikSin: What is `App Bar` you are referring to here? And why does it's height matter for the width of the drawer?

src/frontend/Navigation/Tab/index.tsx
```diff
@@ -23,8 +23,20 @@ import {useAuthContext} from '../../contexts/AuthContext';
...
+const MAX_DRAWER_WIDTH = 360;
```
ErikSin, line 33: how did you determine this `MAX_DRAWER_WIDTH`? Is it just an arbitrary number?

src/frontend/Navigation/Tab/index.tsx
cimigree: Good question! The name is kind of misleading/ confusing, and it isn't my choice exactly. Both constants are copied straight from `react-native-drawer-layout` (`src/utils/getDrawerWidth.tsx` as you can see in node_modules), which is why I kept the names: so it's obvious which library values we're mirroring.

The library uses 56dp as a **horizontal gap** — the strip of screen left uncovered on the far edge when the drawer is open, so you can always see and tap the screen behind it to dismiss the drawer.

Why 56? It's the standard Material app bar height, reused as that margin by Material convention. So on a 360dp phone: 360 − 56 = a 304dp drawer, leaving a 56dp strip of map showing.

I'd be happy to rename it with a note that it mirrors the library; the name is pretty unclear! The only cost is that it's less searchable against the library source. So I will wait to hear what you think and can keep, keep & comment, or rename and comment.

src/frontend/Navigation/Tab/index.tsx
cimigree: No it's not arbitrary. It's the Material Design 3 standard navigation drawer width, and it's the library's own `DEFAULT_DRAWER_WIDTH` (see above). The library links the dimension directly:

https://m3.material.io/components/navigation-drawer/specs

Full formula, straight from the library:
`drawer width = min(screen width − 56, 360)`

So the 360 cap only kicks in above 416dp wide, like on tablets and unfolded foldables (which is where I tested), where you don't want the drawer stretching across the whole screen.

The reason I just copied the formula rather than inventing one is that the drawer has to be **exactly** the width the library would have computed, so that this PR doesn't change how the drawer looks on any device. It only stops the width being overridden by a percentage that can collapse to zero.

src/frontend/Navigation/Tab/index.tsx
ErikSin: > The only cost is that it's less searchable against the library source.

Is the concern that if someone was to do a text search of this variable, if the variable name is not the same, they will not be able to find the reference in the node_module? I personally don't search node_modules ever, not to say my work flow is the only work flow, but i just think i don't fully understand the concern.

src/frontend/Navigation/Tab/index.tsx
ErikSin: That makes sense. lets add a comment explaining why you chose this number so we have a reference to it in the future!

src/frontend/Navigation/Tab/index.tsx
cimigree: I guess yes that plus attributing where it came from - it is not really my invented variable/ constant. I am just reusing the one from the library so I chose to use the same name. But just let me know if you want me to change it to make it more clear, but I think I would a comment that the name is changed along with where it came from. What do you think?

ErikSin: If I was to do it, I would rename it to something more semantically correct, and then comment that it comes from `react-native-drawer-layout`. But if your workflow means having the same name is important, Im open to keeping the name, but I am just trying to understand your workflow.

Is it important to your workflow to match the variable names so your text searches find the references in the node_modules? And why does finding the reference in the node_modules matter? Or are you just concerned about giving credit where credit is due?

If it is the latter, would a comment referencing where it came from not suffice?

Im open to keeping the name if it is important to you. But I am just trying to understand why it is important.

cimigree: No workflow reason. I only occasionally search node_modules. It was just attribution (credit where credit is due), and you're right that a comment covers that fine.

src/frontend/Navigation/Tab/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2061/commits/8e4960b05ec213c3922a395529dec18681aaec33

src/frontend/Navigation/Tab/index.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2061/commits/8e4960b05ec213c3922a395529dec18681aaec33

src/frontend/Navigation/Tab/index.tsx
```diff
@@ -23,8 +23,21 @@ import {useAuthContext} from '../../contexts/AuthContext';
 
 const Tab = createBottomTabNavigator<HomeTabsParamsList>();
 
+// These mirror react-native-drawer-layout's own defaults, so the drawer keeps
+// exactly the size it had before. The gap leaves a strip of screen uncovered
+// beside the open drawer so it can be tapped to dismiss — the library calls it
+// APPROX_APP_BAR_HEIGHT. 360 is the Material Design 3 standard drawer width:
+// https://m3.material.io/components/navigation-drawer/specs
```
ErikSin: small thing, can you add the `360 is the Material Design 3 standard drawer width` AFTER line 31, right above `MAX_DRAWER_WIDTH `

ErikSin (approved):
Thanks!

Yeah, its just seems a lot simpler to me to have the gap named something more self descriptive rather than use the confusing name that the library uses - which IMO is very counter intuitive

PR #2062, opened 2026-08-04
fix: observation details translation

### LINKED ISSUE #2052: Details when viewing a recording observation show questions untranslated
**Describe the bug**
When viewing the details on a recorded observation, the questions show untranslated and the answer for multi or single select options are also untranslated

https://app.notion.com/p/digidem/Details-when-viewing-a-recording-observation-show-questions-untranslated-3aa1b08162d5807cae8cf3324f06f724?v=c816a539054149c7b7606183bf5d7bf1&source=copy_link

cimigree:
closes #2052 

This PR makes sure that Observation details when viewing an observation (both questions and answers) are translated.

I made a reusable hook called useFieldsQuery to tie together the two field lookups to make sure they behave the same.
Previously, the ObservationFields screen (used when creating/ editing an observation) had the query done correctly (passing the language tag). 

BUT, the observation fields displayed when viewing an observation did not have the lookup correct (no language passed). 

So I made one hook so that they are both correct and linked together. A single call site so they wouldn't drift apart again.

If you would rather just have the code repeated, I can undo that slight refactor, but just thought it made sense to link them so they are sure to behave the same. However, I don't feel strongly about it.

A bonus! This PR also fixes the share text, which had English labels and option values for the same reason — Buttons.tsx builds its label: value pairs from the same fields array that had been looked up without the language tag.

image

### DIFF
--- a/src/frontend/hooks/server/fields.ts
+++ b/src/frontend/hooks/server/fields.ts
@@ -1,0 +1,13 @@
+import {useManyDocs} from '@comapeo/core-react';
+import {useActiveProject} from '../../contexts/ActiveProjectContext';
+import {useLocaleState} from '../../contexts/LocaleStoreContext';
+
+export function useFieldsQuery() {
+  const {projectId} = useActiveProject();
+  const languageTag = useLocaleState(s => s.languageTag);
+  return useManyDocs({
+    projectId,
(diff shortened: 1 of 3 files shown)


ErikSin (approved):
Thanks!

PR #2063, opened 2026-08-06
fix: naming casting unmounting

cimigree:
1. This PR updates to the latest version of comapeo/core-react
It also moves comapeo/ipc to a dev dependency per instruction

Also, when IPC changed previously, all(?) the exported symbol names were changed.
2. This PR reflects those name changes.

3. Also `rpc-reflector` (a dependency of `@comapeo/ipc`) bumped from 3.0.1 to 4.3.1, changing `MessagePortLike`'s expected interface, which no longer structurally matches Node's MessagePort typing. That was fixed with a type assertion derived from the functions' own parameter types (no new import needed).

4. SetQADeviceName.test.tsx and ActiveProjectIdStoreContext.test.tsx were failing because the new @comapeo/ipc throws ClientClosed/RpcChannelClosed when a React component's useEffect cleanup (removeListener) runs after the RPC client is closed. 

IN this PR, this failure was fixed in those tests by explicitly unmounting rendered trees/hooks before the test's afterEach closes the client, rather than relying on unclear ordering between React Testing Library's automatic cleanup and the file's own teardown.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,10 +9,9 @@
       "version": "1.14.0-pre",
       "hasInstallScript": true,
       "dependencies": {
-        "@comapeo/core-react": "12.0.2",
+        "@comapeo/core-react": "12.0.3",
         "@comapeo/core-react-native": "1.0.0-pre.10",
         "@comapeo/default-categories": "1.1.2",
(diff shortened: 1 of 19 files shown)


ErikSin (approved):
Thanks!

PR #2064, opened 2026-08-06
fix: reinstates audio blue screen animation

### LINKED ISSUE #2046: Audio Blue screen animation is not working
While recording audio, the screen should be blue in proportion the the amount of time left in the audio.

Currently there seems to be no animation:

Image

cimigree:
closes #2046 
Changes needed to get the animated blue background on audio to show.
There are two changes in this PR.

One is the change to `useWindowDimensions() `which was made while investigating the bug. It turned out not to be the cause but I think it is worth keeping because the old way, `Dimensions.get()`, reads the screen size once, when it is mounted, and never updates.
`useWindowDimensions()` is React's recommended reactive equivalent. It re-renders the component if the dimensions change (rotation, split-screen, or a late-arriving correct value).
This app already fixed the same `Dimensions.get() `pattern for the same reason in `TrackBottomSheet`, so this keeps `AnimatedBackground` consistent.
However, if you can think of a reason to revert to Dimensions.get() I would be happy to change it back.

One is the needed fix. 
It was a stacking order issue. Fabric doesn't reliably honor negative z index. In the new architecture, a later rendered sibling renders on top of the earlier sibling. So I just changed the order so the background renders first (and got rid of the unneeded negative zIndex). Now you can see the background again!

Screen shot:
image

Screen recording:

https://github.com/user-attachments/assets/2fee065d-5d7a-47d8-9c97-0ecbffdd4f41

### DIFF
--- a/src/frontend/screens/Audio/AudioRecording/AnimatedBackground.tsx
+++ b/src/frontend/screens/Audio/AudioRecording/AnimatedBackground.tsx
@@ -1,4 +1,4 @@
-import {Dimensions, StyleSheet} from 'react-native';
+import {StyleSheet, useWindowDimensions} from 'react-native';
 import Animated, {
   useAnimatedStyle,
   useDerivedValue,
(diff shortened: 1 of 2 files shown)


ErikSin (approved):
Thanks and good to know about the z-index!

PR #2065, opened 2026-08-06
refactor: wire sentry backend

### LINKED ISSUE #2056: Wire Sentry to Backend
In #1994, Sentry needs to be connected to the new back end.

All the changes for Sentry should be in the [pre-9 release](https://github.com/digidem/comapeo-core-react-native/releases/tag/v1.0.0-pre.9).

Docs [here](https://github.com/digidem/comapeo-core-react-native/blob/main/docs/sentry-integration.md)

## To Do
- [ ] Integrate Sentry

cimigree:
closes #2056 
Based on https://github.com/digidem/comapeo-core-react-native/blob/main/docs/sentry-integration.md sections 4.1 and 9.
Completes the App's Sentry Consent Setup (connects frontend to backend wiring through consent)

1. Sets the two defaults related to Sentry reporting explicitly (for consistency and visibility) in the app config. 
   a. Device diagnostics is set to true by default (which was always that way) but obviously is overridden by user choice. 
   b. App Usage Stats is set to be on for non production builds so we can have info in Sentry for QA, RCs, Dev builds, etc. But **I am not sure if this is the desired behavior. Could you adjust if needed?** 

2. Uses the existing Data and Privacy checkboxes to also opt in to the two Sentry integration types in addition to what they already do. 
    a. The more basic one, the one that is opted in by default (and already was that way with the other metrics reporting we do) is the Device Diagnostics Metrics. As per the docs, it is a restart to activate. Toggling the checkbox takes effect for Sentry starting the next app launch not immediately. 
   b. The one that is off by default is the App Usage Diagnostics. That checkbox is currently wired to PostHog but now it is also wired to Sentry. Same restart to activate behavior (based on comapeo/core-react-native docs).
   
I did not add any new UI or stores/ contexts. This PR reuses the existing, already-persisted MetricsDiagnosticsStoreContext and AppUsageStatsContext so this adds to what those checkboxes already do and is not a behavior change for existing users.

Just FYI, I haven't confirmed on my device that toggling actually reaches Sentry.


src/frontend/App.tsx
gmaclennan, line 78: This should be coming from the actual user choice of app usage metrics opt-in - it affects navigation integration and the tag we use to track when someone has opted in.

gmaclennan (changes requested):
Generally looks ok, but there is a `appMetricsOptIn` var that should be using our actual app usage persisted state - I think rename to make it consistent?

Also, because core-react-native now manages the state of appMetricsOptIn and appUsageDataOptIn, and posthog also manages its optIn state, I think we should drop our stores - keeping two persisted values in-sync is a recipe for a bug down the line. I'm ok with that being a follow-up.

ErikSin:
> Also, because core-react-native now manages the state of appMetricsOptIn and appUsageDataOptIn, and posthog also manages its optIn state, I think we should drop our stores - keeping two persisted values in-sync is a recipe for a bug down the line. I'm ok with that being a follow-up.

The backend manages the persisted state? If that's the case then we should get rid of those stores now

ErikSin:
@gmaclennan since we are using non reactive states, and we need the app to be restarted when we update the value, do you think we should force restart the app when the user turns on/off app usage or metrics?

ErikSin:
Also is the [AppDiagnosticMetrics](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/metrics/AppDiagnosticMetrics.ts) and [DeviceDiagnosticMetrics](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/metrics/DeviceDiagnosticMetrics.ts) part of app metrics or usage stats?

Previously we were considering them AppMetrics, but i think that might be wrong?

src/frontend/App.tsx
```diff
@@ -63,47 +64,32 @@ import {createQADeviceNameStore} from './contexts/QADeviceNameStoreContext.tsx';
...
-// which owns the Sentry.init call across the RN, Node, and Android-FGS hubs —
-// we pass only the allowlisted extensions. Tracing stays env-gated via the
-// plugin's tracesSampleRate; the full runtime consent model (and a stable user
-// id) migrates later with an updated core-react-native.
-const appMetricsOptIn = APP_VARIANT !== 'production';
 let navigationIntegration:
   ReturnType<(typeof Sentry)['reactNavigationIntegration']> | undefined =
   undefined;
 
+const backendAppDiagnosticsEnabled = getApplicationUsageData();
```
cimigree: I think this name might be unclear about what the variable is referencing? I am a bit confused.
Doesn't this hold usage data? Why does the name reference AppDiagnostics (backendAppDiagnostics)?

src/frontend/App.tsx
```diff
@@ -187,16 +155,45 @@ TaskManager.defineTask(
...
     postHog.optOut();
+    appDiagnosticMetrics.setEnabled(false);
+    setApplicationUsageData(false).catch(err => {
+      Sentry.captureException(err);
+    });
   },
 });
 
+const deviceDiagnosticMetrics = new DeviceDiagnosticMetrics();
+const backendDeviceDiagnosticsEnabled = getDiagnosticsEnabled();
```
cimigree: I am wondering if this name choice is a bit confusing. I am thinking the name should not include the word device? I thought this (getDiagnosticsEnabled();) was more app Diagnostics?

src/frontend/sharedComponents/MetricsDiagnosticsPermissionToggle.tsx
```diff
@@ -18,19 +17,23 @@ const m = defineMessages({
...
       <Text style={styles.permissionText}>
         {formatMessage(m.shareDiagnostics)}
       </Text>
       <Checkbox
-        value={isEnabled}
+        value={reactiveDiagnosticsEnabled}
         error={false}
         onPress={() => {
-          setIsEnabled(!isEnabled);
+          const newDiagnosticEnabledValue = !reactiveDiagnosticsEnabled;
```
cimigree, line 34: Should this also call deviceDiagnosticMetrics.setEnabled(newDiagnosticEnabledValue)?
 Similar to how in App.tsx line 177/185, where the App Usage checkbox does this for appDiagnosticMetrics?
I feel like that might have gotten lost when the store was deleted but I am super confused about how all of this is working frankly.

src/frontend/screens/Onboarding/DeviceNaming.tsx
```diff
@@ -79,7 +76,13 @@ export const DeviceNaming = ({
       },
       {
         onSuccess: () => {
-          navigation.replace('Success');
+          // Always restart the app here
```
cimigree, line 79: Just clarifying...
This restart is running always, not just when the diagnostics toggle actually changed. I assume it's because there is no store to be able to check and I assume the user never notices anything and that there is a restart. Just confirming that's the intended reasoning.

cimigree (commented):
I admit I got a little lost with the different metrics and how they are or are not updated and when. So I have some comments and they may have surfaced a bug or I may just be confused. So I am not requesting changes, but I do have comments.

ErikSin:
> So I have some comments and they may have surfaced a bug

What bug are you referring to?

src/frontend/App.tsx
ErikSin: im also confused tbh....ill try clearing up the names

src/frontend/sharedComponents/MetricsDiagnosticsPermissionToggle.tsx
cimigree: @ErikSin This is the one ^^ I think might be a bug.

ErikSin: I was a little confused about this as well, but `deviceDiagnosticMetrics` is not available here. `deviceDiagnosticMetrics` is instantiated outside of the react tree in app.tsx. I think the only way to make it available here is if we passed it via a context, but it seemed like overkill to have a context for one value used in one place.

Since the entire app has to be reset in order for the new value to take place, I just ran `deviceDiagnosticMetrics.setEnabled` in app.tsx and synced that value with the new value that is set in `@comapeo/core-react-native/sentry`.

For `appDiagnosticMetrics.setEnabled` we did pass it to a context. But that context made sense because it keeps track of several values - the values related to whether we prompt the user to enable app diagnostics. But to make sure the values were in sync, i also did the same thing and ran appDiagnosticMetrics.setEnabled in app.tsx and synced it the value in `@comapeo/core-react-native/sentry`

src/frontend/screens/Onboarding/DeviceNaming.tsx
ErikSin: yeah, i figured that if we just always reset we will always be in sync with the new or old value. In otherwords if they change the value, then we are in sync, and if they don't change the value, it is inconsequential because the user experience doesn't change

src/frontend/App.tsx
ErikSin: I cleaned up all the names to match the backend names

src/frontend/App.tsx
ErikSin: see comment above

src/frontend/App.tsx
cimigree: This is great with the added bonus of fixing the class names. Thanks!

src/frontend/sharedComponents/MetricsDiagnosticsPermissionToggle.tsx
cimigree: If they uncheck being willing to share, and we are still sending data, that doesn't seem quite right, but it is only until an app restarts and we are not sending PII though so it is not ideal, but it is fine in the end.

PR #2070, opened 2026-08-12
fix: share observation crashing when no `original` photo

### LINKED ISSUE #2043: Share Observation sends a Json
Sharing an observation sends a json file instead of creating a text/whats app message/email.

This is not happening with all observations but it is deterministically happening when it does happen. determining what those conditions actually are is a little confusing...

ErikSin:
closes #2043 

When an original photo was not available while sharing an observation, the sharing module throws an error and returns a json with the error message. 

This fix catches the error, and then attempts to find the `preview` photo. 

I also made it that if there are any errors at all (eg In both attempts to fetch `original` and `preview`) it doesn't throw an error at all. I am still sending the error to sentry, but I figure it is better to still share the other information then throwing an error completely.

### DIFF
--- a/src/frontend/screens/Observation/Buttons.tsx
+++ b/src/frontend/screens/Observation/Buttons.tsx
@@ -111,9 +111,8 @@ export const ButtonFields = ({
   }
 
   async function fetchAttachmentBase64(attachment: SavedPhoto) {
-    const originalUrl = await getAttachmentUrl(attachment, 'original');
-
     try {
+      const originalUrl = await getAttachmentUrl(attachment, 'original');
       return await convertUrlToBase64(originalUrl);
     } catch {
       // The original may fail to load depending on media sync settings.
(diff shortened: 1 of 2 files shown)


src/frontend/screens/Observation/Buttons.tsx
```diff
@@ -97,7 +98,32 @@ export const ButtonFields = ({
...
+    return projectApi.$blobs.getUrl({
+      driveId: attachment.driveDiscoveryId,
+      name: attachment.name,
+      type: 'photo',
+      variant,
+    });
+  }
+
+  async function fetchAttachmentBase64(attachment: SavedPhoto) {
+    const originalUrl = await getAttachmentUrl(attachment, 'original');
```
achou11: mostly cautionary, but might be worth adding including this within the try part here too. unlikely for it to reject but worth handling if it does.

achou11 (approved):
One non-blocking comment but otherwise change lgtm.

Out of scope, but there's definitely an opportunity to update `handlePressShare` to be mutation-based, which could make the surrounding code a little bit easier to follow.

PR #2087, opened 2026-08-24
chore: add a disabled button

### LINKED ISSUE #2078: Disabled Button State
Introduce an optional param to button that disables the button. 

## To Do
- [ ] Introduce 'disabled' param as an optional param (that defaults to `false`)
- [ ] Grey out button when disabled
- [ ] Change text and icon to white when disabled (test this out on a few svg icons, as svgs sometimes don't properly take on the color)

### Screenshots

Image

### Figma
https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=6086-3910&t=aGoSqDFBqSxYYGxr-4

cimigree:
closes #2078 
Adds disabled as a param and a type for buttons.

The param goes to the base button. Then, within each button type, the disabled button can be exported if the param is received.

image

I checked through all of the SVG icons used with buttons and they all took a color or were already white. Is that ok? It just seemed like overkill to change them to take currentColor and pass white when they were already white.

### DIFF
--- a/src/frontend/sharedComponents/Buttons.tsx
+++ b/src/frontend/sharedComponents/Buttons.tsx
@@ -10,6 +10,7 @@ type ButtonProps = {
   fullSize: boolean;
   text: string;
   onPress: () => void;
+  disabled?: boolean;
   testID?: string;
   style?: ViewStyleProp;
 };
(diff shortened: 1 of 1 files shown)


ErikSin (approved):
Thanks!

PR #2096, opened 2026-08-27
feat: removes asking user for background permission

### LINKED ISSUE #2012: Remove request for background location updates
We currently use the permission `ACCESS_BACKGROUND_LOCATION` and request background location access when the user first starting recording a track (this is a separate permission to the foreground location which we have generally already asked for).

We don't actually need background location permission, because following #1898 we are using a foreground service (FGS) which Android considers as "foreground location access" (see [Android location permissions docs](https://developer.android.com/develop/sensors-and-location/location/permissions)):

> "The system considers your app to be using foreground location if ... Your app is running a foreground service ... Your app retains access when it's placed in the background, such as when the user presses the Home button on their device or turns their device's display off"

Background location permission is actually for apps which are permanently tracking user location in the background without user interaction or notification. Tracks fits the foreground location definition of "receives location information ... for a defined amount of time".

Removing this permission removes a barrier to Play store publishing approval (in the past Google have changed the restrictions on background location and blocked approval). It also simplifies the UX by removing the additional location permissions request.

This requires some UX work to redefine the states of the tracks bottom sheet to remove the states for "no background location permission given".

The frontend task is to remove the permissions from the app.json config and remove the runtime request.

ErikSin:
closes #2012 

The user facing feature change is that the user no longer sees the bottom sheet which prompts them to update their location permission to `Allow all the time`. Previously we thought this screen was needed as we thought this location permission was needed in order for `tracks` to continue getting location updates even when the app is backgrounded on the phone. It turn out this permission is not needed (see #2012 for more info).

Firstly the logic for prompting this bottom sheet was updated. Previously when the user opened tracks, [we checked if the background permission is enabled (aka the location permission is set to `Allow all the time`), and if that was not enabled we rendered the bottom sheet which prompts them to update their location permission.](https://github.com/digidem/comapeo-mobile/blob/develop/src/frontend/screens/MapScreen/TrackBottomSheet/index.tsx#L90). Now we no longer check for the background permission and removed that screen completely.

Since the background permission is not necessary I was also able to update the expo permission which deal with location permissions at the native level.


app.json
```diff
@@ -41,7 +41,6 @@
         "expo-location",
         {
           "isIosBackgroundLocationEnabled": true,
```
cimigree: Should probably remove that too.
And instead you _could_ add 
`isAndroidForegroundServiceEnabled: true`
We hard code that below so it is not necessary but might make the behavior clearer? Not sure.

cimigree (changes requested):
1. The changes in the two messages files were not commited here and should be.
2. I assume you will handle any changes needed in the Play console?
3. There is a now unused SVG that can be deleted, src/frontend/images/LocationNoFollow.svg
4. `isIosBackgroundLocationEnabled: true` is left behind in `app.json:43.` and can be removed

Great! This will remove friction for users and for Play store approval! Yippee!

app.json
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2096/commits/30e2991a01032f1064cf3a743d90bca4436e8ad7

cimigree (approved):

PR #2098, opened 2026-08-31
fix: remote detection alert test data and rendering

### LINKED ISSUE #2054: Custom background map causes remote detection alerts to be (partially?) hidden
**Describe the bug**
When a project has remote detection alerts to display, the existence of a custom background map causes (parts of) it to no longer appear on the map.

**To Reproduce**
Reproduction steps:

1. Remove custom background map if one is being used
2. View a project with remote detection alerts. The alerts should be visible on the map
3. Add a custom background map
4. The alerts (or some part of them) disappear.

https://app.notion.com/p/digidem/Custom-background-map-causes-remote-detection-alerts-to-be-partially-hidden-3ab1b08162d58076bb23c297396f4c0e?v=c816a539054149c7b7606183bf5d7bf1&source=copy_link

cimigree:
closes #2054 
Includes a couple of bumps in comapeo/core-react-native that has the real fixes so that remote alerts and their labels will show on all of the different maps. Aside from that,

**What this PR does...**
Creates a way to make test alerts. I put it alongside/ after the creating observations test data that can be reached through settings within the menu. I extracted the logic to a different file.

image

This PR also removes a hardcoded font line for labels.
With the line gone, MapLibre falls back to its own default (`Open Sans Regular`, `Arial Unicode MS Regular`). 
For our custom maps the result is identical because both names miss and get substituted. But the extra font in the default gives better coverage for non-Latin characters on maps that do serve fonts directly, like the default Mapbox one.

To test:
1. Go to the Create Test Data screen. Scroll down and press "Create Alerts"
2. Check default map. Should see several alerts and labels (see screenshot).
3. Turn off wifi or turn on airplane mode. The backup, basic map should show up and the labels and alerts should still be there.
4. Turn wifi back on. Upload a custom map. Same thing. Should see alerts and labels.

Some screenshots.
**Default Map:**
image

**Custom Map**
image

**Fall back Map for no wifi/ internet**
image


ErikSin (changes requested):
It's currently not running on iOS I don't think that's your fault. I think it was just the order which PR reviews have taken so can you try updating it from develop and running on iOS and I can re-review it once that's done.

cimigree:
I was able to get it to work on iOS with alerts and labels.

Custom Map
---
IMG_2586


Default Map
---
IMG_2587

Fallback Map
---
IMG_2589

ErikSin (approved):
Thanks!

PR #2108, opened 2026-09-02
fix: stops tracks from crashing due to too many re-renders

### LINKED ISSUE #2107: Tracks Causing App to crash in iOS

ErikSin:
closes #2107 

On my physical iPhone and simulator turning on tracks was causing the entire app to crash. The `<MarkerView / >` had a key that was reset every second. The constant re-rendering seems to have caused the app to crash.

Removes the key entirely


cimigree (approved):
Tracks work great on Android.
On iOS the marker view flashes intermittently. It is not related to this fix though... but I want to add info about it here for posterity.
See screen recording below

https://github.com/user-attachments/assets/f5439769-bfe1-4277-b76c-522a7b138283

PR #2110, opened 2026-09-02
fix: adjust setting so we can record audio

### LINKED ISSUE #2105: Audio not recording

ErikSin:
closes #2105 

In iOS, `allowAudioRecording` is set to `false` [by default](https://docs.expo.dev/versions/latest/sdk/audio/#audiomode). In order to record we need to explicitly set it to true.

I also explicity set the [RecordingDirectory](https://docs.expo.dev/versions/latest/sdk/audio/#recordingdirectory) to `Document`. The cache was being purged when the app was closed. So if the user had not saved an observation with an audio file, and then closed the app, we would lose that audio file before it was saved to the db.

### DIFF
--- a/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
+++ b/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
@@ -28,8 +28,7 @@ export function useAudioRecording() {
   const startRecording = useCallback(async () => {
     // iOS gates recording on allowsRecording, which defaults to false; without
     // this the native recorder throws as soon as record() is called.
-    // This is not documented in expo, but can be seen in the error that is thrown
-    await setAudioModeAsync({allowsRecording: true, playsInSilentMode: true});
+    await setAudioModeAsync({allowsRecording: true});
     await recorder.prepareToRecordAsync();
     // expo-audio releases the recorder's native object when this hook unmounts.
     // If the screen unmounts while prepareToRecordAsync is still awaiting,
--- a/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
+++ b/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
@@ -2,10 +2,16 @@ import {useCallback, useEffect, useRef} from 'react';
 import {
   useAudioRecorder,
   useAudioRecorderState,
+  setAudioModeAsync,
   RecordingPresets,
 } from 'expo-audio';
 
-const RECORDING_OPTIONS = RecordingPresets.HIGH_QUALITY!;
+// `directory: 'document'` keeps recordings out of the cache dir, which iOS can
+// purge under storage pressure before the draft observation gets saved.
+const RECORDING_OPTIONS = {
+  ...RecordingPresets.HIGH_QUALITY!,
+  directory: 'document' as const,
+};
 
 export function useAudioRecording() {
   const recorder = useAudioRecorder(RECORDING_OPTIONS);
@@ -20,6 +26,10 @@ export function useAudioRecording() {
   }, []);
 
   const startRecording = useCallback(async () => {
+    // iOS gates recording on allowsRecording, which defaults to false; without
+    // this the native recorder throws as soon as record() is called.
+    // This is not documented in expo, but can be seen in the error that is thrown
+    await setAudioModeAsync({allowsRecording: true, playsInSilentMode: true});
     await recorder.prepareToRecordAsync();
     // expo-audio releases the recorder's native object when this hook unmounts.
     // If the screen unmounts while prepareToRecordAsync is still awaiting,
--- a/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
+++ b/src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
@@ -28,7 +28,7 @@ export function useAudioRecording() {
   const startRecording = useCallback(async () => {
     // iOS gates recording on allowsRecording, which defaults to false; without
     // this the native recorder throws as soon as record() is called.
-    await setAudioModeAsync({allowsRecording: true});
+    await setAudioModeAsync({allowsRecording: true, playsInSilentMode: true});
     await recorder.prepareToRecordAsync();
     // expo-audio releases the recorder's native object when this hook unmounts.
     // If the screen unmounts while prepareToRecordAsync is still awaiting,
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -86,6 +86,7 @@ jest.mock('expo-audio', () => {
       HIGH_QUALITY: {ios: {}, android: {}, web: {}},
       LOW_QUALITY: {ios: {}, android: {}, web: {}},
     },
+    setAudioModeAsync: jest.fn(() => Promise.resolve()),
   };
 });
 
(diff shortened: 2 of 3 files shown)


cimigree (changes requested):
Sorry but I was not able to record.
I think you mean closes #2109 
I got this error (see screenshot attached)
From the error it looks like it needs to add playsInSilentMode: true.

Image

When I added that, I still could only play the audio before I saved the observation, not after.
After I saved the observation and I went to play the the audio, It showed up as 0:00 out of 0:00 and pressing the play button does nothing.

Image

Before I delve further, as far as the document part, I think these things need to be addressed.

ErikSin:
> After I saved the observation and I went to play the the audio, It showed up as 0:00 out of 0:00 and pressing the play button does nothing.

I actually broke up the original ticket. so this fixes audio not recording, and there is another PR that i have that will fix audio not playing

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -2,10 +2,16 @@ import {useCallback, useEffect, useRef} from 'react';
...
   useAudioRecorder,
   useAudioRecorderState,
+  setAudioModeAsync,
   RecordingPresets,
 } from 'expo-audio';
 
-const RECORDING_OPTIONS = RecordingPresets.HIGH_QUALITY!;
+// `directory: 'document'` keeps recordings out of the cache dir, which iOS can
+// purge under storage pressure before the draft observation gets saved.
+const RECORDING_OPTIONS = {
```
cimigree, line 11: It's fine to save to documents, but we need to clear this afterwards once it has been copied into core and if it's deleted or if the draft is abandoned (like backing out of saving the observation)  because every recording is double the storage with no way for a user to open up that space. Right now it is never deleted and it should be. It will even be included in iCloud backups! https://developer.apple.com/documentation/foundation/optimizing-your-app-s-data-for-icloud-backup
Should the deletion be a separate PR? I kind of feel like it belongs in here since you are introducing the document saving in here. I will let you decide, but we really should delete the document in the situations I mentioned.

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -20,6 +26,9 @@ export function useAudioRecording() {
   }, []);
 
   const startRecording = useCallback(async () => {
+    // iOS gates recording on allowsRecording, which defaults to false; without
+    // this the native recorder throws as soon as record() is called.
+    await setAudioModeAsync({allowsRecording: true, playsInSilentMode: true});
```
cimigree, line 31: This sets playsInSilentMode globally and nothing resets it (like for when someone goes to play the recording, it will play in silent mode). I think that is probably what we want, but just checking.

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin: [I actually reverted it back to "cache"](https://github.com/digidem/comapeo-mobile/pull/2110/commits/2c19bcfa5df31007a2e9c2a377225f838fadb180). I realized saving it as a `document` was not actually solving the original problem like i had thought. I will create a separate issue for the problem of audio not being available when the user closes the app

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin: yes, i think this is the desired behaviour.

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -2,10 +2,17 @@ import {useCallback, useEffect, useRef} from 'react';
 import {
   useAudioRecorder,
   useAudioRecorderState,
+  setAudioModeAsync,
   RecordingPresets,
+  RecordingOptions,
 } from 'expo-audio';
 
-const RECORDING_OPTIONS = RecordingPresets.HIGH_QUALITY!;
+// `directory: 'document'` keeps recordings out of the cache dir, which iOS can
```
cimigree: Can you update the comment to match the code please?

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -2,10 +2,17 @@ import {useCallback, useEffect, useRef} from 'react';
...
+// purge under storage pressure before the draft observation gets saved.
+const RECORDING_OPTIONS: RecordingOptions = {
```
cimigree, line 10: The director being cache is the default, so you don't need this explicit line here.
Can just go back to `const RECORDING_OPTIONS = RecordingPresets.HIGH_QUALITY!;`

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
```diff
@@ -20,6 +27,9 @@ export function useAudioRecording() {
   }, []);
 
   const startRecording = useCallback(async () => {
+    // iOS gates recording on allowsRecording, which defaults to false; without
+    // this the native recorder throws as soon as record() is called.
+    await setAudioModeAsync({allowsRecording: true, playsInSilentMode: true});
```
cimigree, line 29: I ran into another issue when testing this out on Android. It seems to work fine on iOS actually.
If it is the first time recording Audio and I need to grant microphone permission, after I grant it, I get [an error.](https://awana-digital.sentry.io/issues/7616044038/?environment=development&project=4507170965618688&query=&referrer=issue-stream&sort=recommended)

I can close it and then record, but that does not seem ideal for someone the first time using Audio on android.

Basically it is the same bug we fixed with the guard on line 37. So we probably need to repeat that guard above the await above. Like line 32.

Or perhaps it can be fixed by setting the hasNavigatedRef in the askPermissionMutation (similar in how it is done when navigating from Settings) because navigating from settings works fine.

cimigree (changes requested):
Tested on iOS and Android and it recorded.
There are two non-blocking comments and one blocking comment that I surfaced while testing on Android.

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2110/commits/2d9f0595c3697c56d7dd0e1717a059b19c84ac12

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2110/commits/6d5ab5ce403d86dbbe5cda20d02237709f8055c9

src/frontend/screens/Audio/AudioRecording/useAudioRecording.ts
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2110/commits/3fd20d65589ad810b52651ba13a262d4e4087e6d

cimigree (approved):
I tested on iOS and Android and both seem to work fine now.
There is a little issue with coming back from settings on iOS but I will fix that in my permissions PR. I am pretty sure it is not introduced here.

PR #2113, opened 2026-09-03
feat: new onboarding flow

### LINKED ISSUE #2102: Onboarding improvements
## Description
While improving the Tracks flow and updating the permissions request, we are taking the opportunity to update the Onboarding process.

What’s new
- Remove Location & Camera Permissions prior to Welcome screen
- Add Notifications Permission Request interstitial after Name Device screen
- Update Device Ready screen (see below)
- Integrate Location and Camera Permission Requests on their respective Main Tabs on First-Time Use see https://github.com/digidem/comapeo-mobile/issues/2101

Image
Launch > Welcome > Data & Privacy > Name Device > Notification Request > Device Ready

## Launch Screen

Image

- [ ] Update screen

## Welcome screen

Image

- [ ] Update screen

## Data & Privacy

Image

- [ ] Update “Learn More” button to tertiary link style (blue text link).

## Name Device

Image

- [ ] Update the header copy from “Name Your Device” to “Name This Device”

## ~~New: Notifications Permission Request (During Onboarding)~~
**This screen is being done in #2111 instead**

Image

- [ ] Integrate screen as designed
- [ ]  Tertiary button style (blue text link) for “Not Now”
- [ ] - **Openmojis**
    - [Mobile Message](https://openmoji.org/library/emoji-E096/)
    - [GPS](https://openmoji.org/library/emoji-E1CD/)
    - [Counterclockwise Arrows Button](https://openmoji.org/library/emoji-1F504/)

##  Device Ready 

Image

- [ ] Update secondary header to “Coordinate with team to start or map on your own.”
- [ ] Remove “Join a Project” button
- [ ] Swap “Mapping On Your Own” white button for the new “Start Mapping” blue button.
- [ ] User can be invited to a project in this screen

## Figma

https://www.figma.com/design/iUeC0Qzhb4H0unuPfxQAsM/-Mobile--CoMapeo?node-id=12318-6757
## Primary strings

Always Stay Notified
Start Mapping

ErikSin:
closes #2102 

## Things completed in this PR:
1. I got rid of the manual closing of the splash screen. Previously we waited for the user to choose their permission before closing the splash screen. Since we are moving away from permission being asked on first open, we no longer need to wait, and the splash screen now just automatically closes
2. Changed the PNG Asset for the splash screen
3. No longer asking for permissions in first app startup. I moved the location permission to the map screen. The map screen will eventually be updated to show a different screen based on the permissions available, but that is out of scope of this PR
4. Since we are getting rid of prompting the user to be invited to a project during onboarding I remove the ability for the user to get invited to a project during onboarding. 
5. I also update the e2e tests and jest tests to account for this new onboarding flow


src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -3,119 +3,135 @@ import {StyleSheet, View} from 'react-native';
...
-          </View>
-          <BodyText style={styles.bodyText}>{t(m.chooseProject)}</BodyText>
+    <ScreenContentWithDock
+      contentContainerStyle={{paddingTop: 85, alignItems: 'center', gap: 20}}
+      dockContent={
+        status === 'pending' ? (
+          <LoadingIndicator />
+        ) : (
+          <PrimaryButton
+            testID="ONBOARDING.map-on-your-own-btn"
```
cimigree, line 78: Should the id change to Start Mapping?

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
```diff
@@ -1,96 +1,41 @@
...
-            {formatMessage(m.designedFor)}
-          </BodyText>
-        </View>
-      </ScrollView>
-      <View style={styles.buttonContainer}>
+    <ScreenContentWithDock
+      contentContainerStyle={{flex: 1, padding: 40}}
+      dockContent={
         <PrimaryButton
           testID="ONBOARDING.get-started-btn"
```
cimigree, line 32: Should this button now be named Start btn?

src/frontend/screens/Onboarding/Success.tsx
```diff
@@ -3,119 +3,135 @@ import {StyleSheet, View} from 'react-native';
...
 import {BodyText} from '../../sharedComponents/Text/BodyText';
-import {PrimaryButton, SecondaryButton} from '../../sharedComponents/Buttons';
-import {WHITE, DARK_GREEN, COMAPEO_BLUE} from '../../lib/styles';
-import {useOwnDeviceInfo} from '@comapeo/core-react';
-
+import {PrimaryButton} from '../../sharedComponents/Buttons';
+import {WHITE, DARK_GREEN, NEW_DARK_GREY} from '../../lib/styles';
+import {useCreateProject, useOwnDeviceInfo} from '@comapeo/core-react';
+import {ScreenContentWithDock} from '../../sharedComponents/ScreenContentWithDock';
+import MatericalIcon from '@react-native-vector-icons/material-icons';
```
cimigree, line 15: MaterialIcon (spelling error)

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
```diff
@@ -1,96 +1,41 @@
...
-        </View>
-        <View style={styles.designedForContainer}>
-          <BodyText variant="smallMeta" style={styles.designedForText}>
-            {formatMessage(m.designedFor)}
-          </BodyText>
-        </View>
-      </ScrollView>
-      <View style={styles.buttonContainer}>
+    <ScreenContentWithDock
+      contentContainerStyle={{flex: 1, padding: 40}}
```
cimigree, line 29: Not sure what the solution is here but the screen looks a lot better on a small phone than a big phone and better matches the Figma on a small phone.

You can also see the white status bar (mentioned in summary) and the lack of blue background.

Image

cimigree (commented):
Reviewed on both Android (on a Pixel) and iOS (on the SE simulator and my iPhone).
Most of what's below is some things #2102 asked for that you didn't do
plus a few other things.


1. Issue #2102 says the user can be invited to a project on the Device Ready screen but this PR makes that impossible. If that should not happen at that stage, can you let me know? Or comment in the issue? I know in your PR you say you removed it but it reads as a direct contradiction of the ticket and I thought I remembered that Aldo and Ximena said we wanted to be able to receive invites on that screen (and it wasn't changed in the ticket).

2. Three of the tickets items don't seem to have been done.

- Data & Privacy "Learn More" was supposed to change to a  tertiary blue link. 
- "Name Your Device" was supposed to change to "Name This Device" but it didn't. Plus it doesn't look great on iOS.
(screenshot attached)


Image

- The welcome screen background and status bar. It is showing as white instead of the pale blue in the Figma. And because it used to be a dark screen, we set the status bar to light. Now it is white on a white background so it is invisible.

3. Splash sequence is not looking great on Android. On a fresh Android install the cold
start is now: splash → black → splash again → white flash → Welcome. iOS is fast
enough that I can't see it, but Android is struggling.

Not blocking but definitely a thing.
Did you know that iOS asks for Local Network permission on first launch, before the user has a project or any peers, because discovery is set up in App.tsx? I am not sure what we can do about this, but I just want to make sure everyone is aware there is a permission still being asked on iOS from the getgo.

Image

See some teeny tiny nitpick comments about code in the code.

src/frontend/screens/Onboarding/Success.tsx
ErikSin: do you think the `testId` need to match the text? if thats the case, why are we not just using the text instead of the testId in our tests?

cimigree: No. That was a bad suggestion, I guess. But it would be helpful if the test ID focused on what the button did and I feel like map on your own is a little misleading because that is an old screen and now that button is creating a project and then starting to map. I also think it is helpful when you are looking at the failing E2E test (especially in the Github Actions or Browserstack) to be able to identify quickly what went wrong by seeing the testID that is being used or missed and if that testID is descriptive, that makes it even more helpful.

I think using text matches is a little slower in Browserstack and can be ambiguous (not in this case, but sometimes). It would make sense to me to test actions and behavior with testIds and check that the user is seeing the right words by checking text matches.

But it is a nit and not a big deal at all.

ErikSin: Passive aggressive comment aside, I agree with most of what you're saying. I think that if that's the case we should make the test ID is more generic and tied to the function of the button rather than the text of the button. So I think the test ID should generically describe the function of the button and not be tied to the text at all. For example, perhaps this test ID should be `ONBOARDING.navigate-to-map-screen`?

cimigree: Ok sounds good to make the test ID generally describe the function. Many of our buttons have btn as a suffix, which makes it easier to scan in test results/ failure reporting. So maybe `ONBOARDING.navigate-to-map-btn `? Not blocking either way.

ErikSin:
> 3\. Splash sequence is not looking great on Android. On a fresh Android install the cold
>         start is now: splash → black → splash again → white flash → Welcome. iOS is fast
>         enough that I can't see it, but Android is struggling.

This is actually due to testing on a development build. If you look at the warning [in this documentation](https://docs.expo.dev/develop/user-interface/splash-screen-and-app-icon/#splash-screen), it says not to test splash screen on a development build and to build an APK. So I tried it on two phones using an APK and it seems to be working fine, while in a development it is doing what you're saying it is doing.

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
ErikSin: Let's leave it for now and we'll see what design says when they do a full review.

src/frontend/screens/Onboarding/Success.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2113/commits/becad1764f5d5cd9cac3091abcb4df5b2e86a9ff

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2113/commits/861c0933913b7fa17427045e6300b8396641eb31

src/frontend/screens/Onboarding/Success.tsx
ErikSin: https://github.com/digidem/comapeo-mobile/pull/2113/commits/861c0933913b7fa17427045e6300b8396641eb31

ErikSin:
Everything is fixed. I left the connectivity permission prompts and design can determine whether we need to put it somewhere else when they do the review. 

Currently, when the user accepts an invite in the on boarding after they get navigated to the map screen, there is a navigation error. that error already existed and only shows up in dev mode. There's a bigger fix that I think is at a scope of this PR only because it touches the bottom sheet model and that bottom sheet model is used in a lot of places so I don't think it's wise to implement that fix now considering that that problem already existed before.

src/frontend/sharedComponents/BottomSheetWrapper.tsx
```diff
@@ -12,41 +11,13 @@ import {usePreventAndroidBackButton} from '../hooks/usePreventAndroidBackButton'
...
-  usePreventAndroidBackButton();
-  return <AnimateBottomSheetContainer>{children}</AnimateBottomSheetContainer>;
-};
-
-const AnimateBottomSheetContainer = ({
-  children,
-}: {
-  children: React.ReactNode;
-}) => {
+export const BottomSheetWrapper = ({children}: {children: React.ReactNode}) => {
```
cimigree: Could you please explain these changes? I am wondering if they are left over from trying to fix the invite received during onboarding and are accidentally included or are deliberate and if deliberate, what do they stem from? Is it so that the Android back button is no longer supposed to work on All Projects? That was the only change I noticed and the bottom sheets all looked fine, so not blocking. Just curious.

src/frontend/screens/Onboarding/IntroToCoMapeo.tsx
```diff
@@ -1,96 +1,47 @@
...
-        </View>
-        <View style={styles.designedForContainer}>
-          <BodyText variant="smallMeta" style={styles.designedForText}>
-            {formatMessage(m.designedFor)}
-          </BodyText>
-        </View>
-      </ScrollView>
-      <View style={styles.buttonContainer}>
+    <ScreenContentWithDock
+      contentContainerStyle={{
```
cimigree: On iPhone, one can scroll the screen and if so, there is this white stripe. See screenshot mid scroll.
Not sure how to fix and it is not blocking but it's not ideal.

Image

src/frontend/screens/Onboarding/DataPrivacy.tsx
```diff
@@ -28,6 +27,14 @@ export const DataPrivacy = ({
     <ScreenContentWithDock
       dockContent={
         <View style={styles.buttonsContainer}>
+          <TouchableOpacity
```
cimigree: Non blocking. The tap area is just where the text is. Could you use Pressable and add a hitSlop?

cimigree (approved):
I tested on RC builds on iPhone and two Androids. The onboarding fixes all look good and inviting during onboarding worked fine (I had no idea there was an error).

I have some non blocking comments for you to decide if you want to address.
Thanks. It will be great to have this.

src/frontend/sharedComponents/BottomSheetWrapper.tsx
ErikSin: Yes, it was when I was trying to fix the bottom sheet. I have taken out all the changes.

PR #2115, opened 2026-09-07
fix: status bar error

### LINKED ISSUE #2073: Status Bar Error
When opening the camera, the `<StatusBar/>` is throwing the error: `RCTStatusBarManager Module requires that the UIViewControllBasedStatusBarAppearance key in the Info.plist is set to no`

Image

cimigree:
closes #2073 

We cannot use the <StatusBar /> component in the camera on iOS with this key in the app.json `UIViewControllerBasedStatusBarAppearance": true` set to false. However, if we set it to false, we can't use the status bar styling we have set up in the whole rest of the app. So this PR only uses the <StatusBar /> component in Android. See below for an image of the dark status bar on iOS. It is only an issue when the phone camera is not pointed at anything (like the phone is sitting on a desk, for example). It is pretty tricky to change that styling to light content on iOS.

### DIFF
--- a/app.json
+++ b/app.json
@@ -125,7 +125,7 @@
         "ITSAppUsesNonExemptEncryption": false,
         "NSLocalNetworkUsageDescription": "CoMapeo uses the local network to discover and sync with nearby devices in your project.",
         "NSBonjourServices": ["_comapeo._tcp"],
-        "UIViewControllerBasedStatusBarAppearance": true
+        "UIViewControllerBasedStatusBarAppearance": false
       }
     },
     "extra": {
--- a/app.json
+++ b/app.json
@@ -125,7 +125,7 @@
         "ITSAppUsesNonExemptEncryption": false,
         "NSLocalNetworkUsageDescription": "CoMapeo uses the local network to discover and sync with nearby devices in your project.",
         "NSBonjourServices": ["_comapeo._tcp"],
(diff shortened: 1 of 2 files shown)


ErikSin (changes requested):
I seem to be getting this error as soon as I open the app now: 

Image

cimigree:
> I seem to be getting this error as soon as I open the app now:
> 
Did you rebuild (npx expo prebuild --clean -p android)? Because of the change in the app.js you have to do that.

ErikSin:
> > I seem to be getting this error as soon as I open the app now:
> 
> Did you rebuild (npx expo prebuild --clean -p android)? Because of the change in the app.js you have to do that.

yes, but i did `npx expo prebuild --clean -p ios` (`ios` not `android`) and ran it on an apple device. I assume that is what you meant? Or does the prebuild for android fix this on ios?

cimigree:
Ok I can see where that error is coming from now. 
No you did not need to build on Android.
Not sure why I wasn't getting it earlier. Should be fixed but of course now we have a boolean for iOS and Android in the camera whether or not to use <StatusBar />.
Because of that the status bar info on iOS is dark. But it is still viewable. Just not perfect for accessibility I think. It is pretty tricky to change it though, so I left it alone for now.
IMG_2530

ErikSin (approved):
Looks good, thanks!

PR #2116, opened 2026-09-07
fix: awkward grey bottom bar ios

### LINKED ISSUE #2077: Akward Bottom Bar
In iOS there is a persistent grey bottom bar on all screens

Image

## To Do
- [ ] Remove bottom bar

cimigree:
closes #2077 
The safe area on iOS is technically the home activity indicator area. Since there is no home button, this is the part of the screen a user presses for that function. On Android (on my Samsung for example) it is the bottom navigation bar. So we had set the color to be grey to set that Android navigation bar apart from our app. On iOS the different color looks weird. So this PR separates the behavior between iOS and Android. It is probably the first of many visual changes we will need to make...

So on iOS we make the color white. On Android, we keep it medium grey, which is what it was before.

IMG_2517

### DIFF
--- a/src/frontend/Navigation/Stack/index.tsx
+++ b/src/frontend/Navigation/Stack/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import {Platform} from 'react-native';
 import {NativeStackNavigationOptions} from '@react-navigation/native-stack';
 import {SafeAreaView} from 'react-native-safe-area-context';
 import {WHITE, MEDIUM_GREY} from '../../lib/styles';
@@ -28,6 +29,11 @@ export type NavigatorLayout = NonNullable<
 export type NavigatorScreenLayout = NonNullable<
   React.ComponentProps<typeof RootStack.Navigator>['screenLayout']
 >;
+
+// Android's bottom inset sits behind the system navigation bar, where grey is a
+// deliberate visual choice (#1671). iOS's is the home activity indicator, where anything but
+// color of the tabs doesn't fit visually (#2077).
+const BOTTOM_INSET_COLOR = Platform.OS === 'android' ? MEDIUM_GREY : WHITE;
 
 const NavigatorScreenOptions: NativeStackNavigationOptions = {
   presentation: 'card',
@@ -69,7 +75,7 @@ export const RootStackNavigator = () => {
   const layout: NavigatorLayout = ({children, state, navigation}) => (
     <SafeAreaView
       edges={['bottom']}
-      style={{flex: 1, backgroundColor: MEDIUM_GREY}}>
+      style={{flex: 1, backgroundColor: BOTTOM_INSET_COLOR}}>
       <React.Suspense fallback={<FullScreenCenteredLoader />}>
         <PendingInvitesListener
           currentRouteName={state.routes[state.index]?.name}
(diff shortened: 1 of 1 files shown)


ErikSin:
Im not sure if this is difficult or even possible, but what if we just got rid of the bottom bar/safe area entirely for ios? I think it just looks a little akward with that bottom space turned white

cimigree:
It is possible to remove that space although not recommended.
See the transcript of this [design talk when they first introduced the home indicator. ](https://developer.apple.com/videos/play/tech-talks/801/?time=603). Go to transcript. See 9:29.
This PR makes everything in our app shift down to the gesture area on iOS. There are interactive apple controls there so it is not ideal to have our buttons and tabs and things overlap apples gesture area. But it is what you asked for.
Here is what it would look like. It's hard to see in the screenshots so I suggest you run this on a device or simulator.
IMG_2531
IMG_2532
IMG_2533

Let me know what you think.

ErikSin:
> It is possible to remove that space although not recommended. See the transcript of this [design talk when they first introduced the home indicator. ](https://developer.apple.com/videos/play/tech-talks/801/?time=603). Go to transcript. See 9:29. This PR makes everything in our app shift down to the gesture area on iOS. There are interactive apple controls there so it is not ideal to have our buttons and tabs and things overlap apples gesture area. But it is what you asked for. Here is what it would look like. It's hard to see in the screenshots so I suggest you run this on a device or simulator.

I don't think I fully understand what you are saying here. You say that it is not recommended to get rid of the bottom bar, but then proceeded to get rid of the bottom bar.

I was stating a visual preference, but I don't think that my personal visual preference should take precedence over a technical reason. Are you implying it should, or is there another reason?

cimigree:
> 
> I don't think I fully understand what you are saying here. You say that it is not recommended to get rid of the bottom bar, but then proceeded to get rid of the bottom bar.
> 
> I was stating a visual preference, but I don't think that my personal visual preference should take precedence over a technical reason. Are you implying it should, or is there another reason?

I just wanted you to be aware of all of the pros and cons. Wanted you to see what it would look like but also to know why it is not recommended. That way as time goes by and we keep seeing it, we remember and know about all the possible alternatives. 

I can put it back to keeping that extra white space because of needing to have the room for a touch bar. But my question now is, knowing that it is supposed to be space for a touch bar, does it make sense to have it white? Or should it be gray (to designate its separateness). I would prefer it to be white because iOS users know that is the space to do the gestures and it doesn't matter what color it is. How about you?

ErikSin:
> I would prefer it to be white because iOS users know that is the space to do the gestures and it doesn't matter what color it is. How about you?

As a daily ios user, do you know if the bottom bar is normally a different color? We should probably follow the ios convention. If there is no real ios convention, then yes, I agree, the white bottom bar looks better!

cimigree:
> > I would prefer it to be white because iOS users know that is the space to do the gestures and it doesn't matter what color it is. How about you?
> 
> As a daily ios user, do you know if the bottom bar is normally a different color? We should probably follow the ios convention. If there is no real ios convention, then yes, I agree, the white bottom bar looks better!

Yes, convention is for it to be the same color. So for us would be white. I looked through a whole bunch of my apps to confirm!

cimigree:
OK, @ErikSin I put the white bottom bar back, but I did notice that it needs to be a different color on a couple of screens.
So there is a little change with an object and the screens with different colors to pass in.
I also noticed the bottom sheet has the background poking through but I could not for the life of me figure out how to get rid of that transparent background without getting rid of that bottom strip. So I left it. It doesn't look too bad.
IMG_2564

Here are the other screens that needed a special color:
IMG_2561
IMG_2562

ErikSin (approved):

PR #2117, opened 2026-09-07
fix: back button screen name on iOS

### LINKED ISSUE #2074: Unwanted iOS Back Button UI
The back button in iOS includes the screen name of the previous screen. This crowds the header, and is different UI from Android

## To Do
- [ ] Remove name of previous screen

cimigree:
closes #2074 
Changes the display mode for the back buttons to minimal for iOS so that the screen name won't show. 

Not sure why but `HeaderBackButton` from @react-navigation/elements defaults `displayMode` to 'default' on iOS and 'minimal' on Android. 

'default' renders the previous screen's title next to the icon and 'minimal' does not render the screen title. 

So this PR keeps the exact same behavior for Android and makes the iOS match it.

IMG_2519

### DIFF
--- a/src/frontend/sharedComponents/CustomHeaderLeft.tsx
+++ b/src/frontend/sharedComponents/CustomHeaderLeft.tsx
@@ -5,7 +5,7 @@ import {HeaderBackButtonProps} from '@react-navigation/elements';
 import {BackIcon} from './icons';
 import {BLACK} from '../lib/styles';
 import {useNavigationFromRoot} from '../hooks/useNavigationWithTypes';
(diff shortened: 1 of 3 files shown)


ErikSin:
I think we should fix the akward centering/spacing of the arrow. Do you see how it is not centered in the bubble Are you opposed to that for any reason?
image

Maybe just getting rid of that extra space inside the bubble area.

cimigree:
Ok done.
IMG_2534
IMG_2535

ErikSin (approved):
Thanks!

PR #2118, opened 2026-09-07
fix: showing no wifi screen correctly

### LINKED ISSUE #2076: Exchange showing no WiFi when connected to wifi
No Wifi screen is showing, but phone is connected to wifi.

I was able to successfully invite another device onto a project directly before this, implying there is a wifi connection

Image

## To Do
- [ ] Diagnose and fix issue

cimigree:
closes #2076 
This PR addresses several issues related to wifi and connection that are different in iOS than in Android and fixes some little glitches in Android too.

#### Changes
1. Adds functions at the end of `src/frontend/contexts/LocalDiscoveryContext.tsx` to deal with the differences between Android and iOS related to wifi connection and network availability. `isWifiEnabled` is Android only. Also adds a check of the correct wifi IP so discovery only starts when wifi is joined to a network (not just when it is "on"). 
2. In order to show the wifi network (SSID) on iOS we need to add an entitlement and a configuration in App.tsx. More on that at the end.
3. Relating to the remote archive and the metrics sharing, iOS does not have a native internet reachability check, so we were polling every 2 seconds (a value set in [`LocalDiscoverContext`](https://github.com/digidem/comapeo-mobile/blob/bded730502111f27f91be00dc021f359292ae849/src/frontend/contexts/LocalDiscoveryContext.tsx#L37). This PR fixes that in the configuration in App.tsx and also uses a utility function to correctly get the internet status with precise types, instead of a boolean, to prevent flickering for the user.
4. Adds a test for the new utility function

IMG_2521

### About Entitlement
Apple also only allows the app to see the SSID if the user granted precise location. We already ask for location but it should be noted that anyone who picks "Approximate" will still see -- at the top of the screen (not the name of the wifi).

The [next eas build should enable Access Wi-Fi Information and regenerate the profile](https://docs.expo.dev/build-reference/ios-capabilities/). It'll ask for an Apple login when we do any kind of `eas build ...`. Each variant is its own App ID (com.comapeo.dev, .rc, .pre, com.comapeo), so each picks it up on its first build. It needs to happen before the next signed device build. If the profile doesn't carry the entitlement, signing will fail.

### DIFF
--- a/src/frontend/contexts/LocalDiscoveryContext.tsx
+++ b/src/frontend/contexts/LocalDiscoveryContext.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {AppState, AppStateStatus} from 'react-native';
+import {AppState, AppStateStatus, Platform} from 'react-native';
 import NetInfo, {
   type NetInfoWifiState,
   type NetInfoState,
@@ -235,8 +235,8 @@ export function createLocalDiscoveryController(mapeoApi: ComapeoCoreClientApi) {
     // Don't do anything if the app is not active (starting and stopping
     // discovery based on appState is handled below in onAppState)
     if (appState !== 'active') return;
-    const nextIpAddress = nextNetInfo.details.ipAddress;
-    const prevIpAddress = netInfo?.details ? netInfo.details.ipAddress : null;
+    const nextIpAddress = getWifiIpAddress(nextNetInfo);
+    const prevIpAddress = getWifiIpAddress(netInfo);
     // netInfo.isConnected is true (I think) only if the wifi network has
     // internet access. We use the presence of an IP address to detect if wifi
     // is connected to a local network.
@@ -274,7 +274,7 @@ export function createLocalDiscoveryController(mapeoApi: ComapeoCoreClientApi) {
     }
     netInfo = nextNetInfo;
     updateState({
-      wifiStatus: nextNetInfo.isWifiEnabled ? 'on' : 'off',
+      wifiStatus: getWifiStatus(nextNetInfo, isLocalWifiConnected),
       wifiConnection: isLocalWifiConnected ? 'connected' : 'disconnected',
       wifiLinkSpeed: nextNetInfo.details.linkSpeed,
       ssid: nextNetInfo.details.ssid,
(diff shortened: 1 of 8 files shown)


src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -274,7 +274,7 @@ export function createLocalDiscoveryController(mapeoApi: ComapeoCoreClientApi) {
     }
     netInfo = nextNetInfo;
     updateState({
-      wifiStatus: nextNetInfo.isWifiEnabled ? 'on' : 'off',
+      wifiStatus: getWifiStatus(nextNetInfo, isLocalWifiConnected),
       wifiConnection: isLocalWifiConnected ? 'connected' : 'disconnected',
```
ErikSin: `wifiConnection` is never used. I think we can get rid of it. If I remember correctly, we had a screen that needed `wifiConnection`, but we got rid of that screen and we should've got rid of this variable at the time that we got rid of that screen.

src/frontend/contexts/LocalDiscoveryContext.tsx
```diff
@@ -274,7 +274,7 @@ export function createLocalDiscoveryController(mapeoApi: ComapeoCoreClientApi) {
     }
     netInfo = nextNetInfo;
     updateState({
-      wifiStatus: nextNetInfo.isWifiEnabled ? 'on' : 'off',
+      wifiStatus: getWifiStatus(nextNetInfo, isLocalWifiConnected),
```
ErikSin: I think we can simplify this as well. I think we can do the same check that you do for iOS but also do it for android so in other words, we can get rid of get `getWifiStatus` and just do the ternary check `isLocalWifiConnected ? 'on' : 'off'`.

ErikSin (changes requested):
One blocking comment just to simplify the code. There are some complications in the code that you didn't introduce. They were introduced a while ago when we had different features for exchange that we are no longer using so I think we should take this opportunity to clean that up.

src/frontend/contexts/LocalDiscoveryContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2118/commits/43ccbf8b79e07d17d0aaf4073cda0b526d48940a

src/frontend/contexts/LocalDiscoveryContext.tsx
cimigree: https://github.com/digidem/comapeo-mobile/pull/2118/commits/43ccbf8b79e07d17d0aaf4073cda0b526d48940a

cimigree:
Just as you thought, nothing used `wifiConnection` anywhere. 
While I was poking around in there, nothing reads `wifiLinkSpeed` either. It's only mentioned in a JSDoc example. Want me to take that out too, or do you know if it is destined to be used for something coming up?
For `getWifiStatus` you're right about that too. Fetch('wifi') always comes back as type wifi.  So, I took that out, but that is a change on Android.  
The change is, if wifi is on but not joined to a wifi network, it now shows the No Wi-Fi screen. Before, Android with wifi on but not joined showed the exchange screen while discovery was stopped, so you'd just wait on that screen and never see another device (peer). I think that's more correct but please say if you disagree.

ErikSin:
> The change is, if wifi is on but not joined to a wifi network, it now shows the No Wi-Fi screen. Before, Android with wifi on but not joined showed the exchange screen while discovery was stopped, so you'd just wait on that screen and never see another device (peer). I think that's more correct but please say if you disagree.

That was unintentional on my part but I agree that it is better this way so let's leave it and if design and or someone doesn't like it then we can revert it but for now I think it's good as is

ErikSin:
> While I was poking around in there, nothing reads wifiLinkSpeed either.
I think, but I could be wrong., we wanted to use this for metrics eventually, so let's leave it in for now.

ErikSin (approved):
Thanks

PR #2119, opened 2026-09-08
fix: no error when cancelling download ios

### LINKED ISSUE #2075: Cancelling observation GeoJson Share throws error
When the user is downloading observation and decides to cancel the download, the error bottom sheet shows saying `OPERATION_CANCELLED`. 

This should not throw an error, and the app should silently swallow the operation cancelled error

## To Do
- [ ] Swallow cancel error and have user just be navigated back to the previous screen

Image

cimigree:
closes #2075 
[React native documents (picker) ships with error codes.](https://react-native-documents.github.io/docs/sponsor-only/errors) The docs say that will be changed in a future PR, but it is very helpful for now as it matches on both iOS and Android. 
So, this PR checks for the error code instead of using a string match as the string is different on iOS and Android.
I tested export on both Android and iOS both with and without photo files and all went smoothly.

### DIFF
--- a/src/frontend/hooks/server/projects.ts
+++ b/src/frontend/hooks/server/projects.ts
@@ -10,7 +10,11 @@ import {useMutation, useQuery} from '@tanstack/react-query';
 
 import {useActiveProject} from '../../contexts/ActiveProjectContext';
 import {MEMBER_ROLE_ID} from '../../sharedTypes';
-import {saveDocuments} from '@react-native-documents/picker';
+import {
+  errorCodes,
+  isErrorWithCode,
+  saveDocuments,
+} from '@react-native-documents/picker';
 import {Exports} from '../../sharedTypes/navigation';
 import * as FileSystem from 'expo-file-system/legacy';
 import {useLocaleState} from '../../contexts/LocaleStoreContext';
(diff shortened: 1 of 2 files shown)


src/frontend/hooks/server/projects.ts
```diff
@@ -10,20 +10,19 @@ import {useMutation, useQuery} from '@tanstack/react-query';
...
 import noop from '../../lib/noop';
 
 export function isUserCancelled(err: unknown): boolean {
-  return (
-    typeof err === 'object' &&
-    err !== null &&
-    'message' in err &&
-    (err as {message?: string}).message === 'user canceled the document picker'
-  );
+  return isErrorWithCode(err) && err.code === errorCodes.OPERATION_CANCELED;
```
ErikSin: Since this is one line logic, I think we should get rid of this function completely and just do the error logic in ExportObservation.tsx. 1. for simplicity. 2. (this was a problem before that neither of us noticed) this function ONLY works for `@react-native-documents/picker`, but having it as a reusable function implies it would work for other modules

cimigree: https://github.com/digidem/comapeo-mobile/pull/2119/commits/bb7571ab645698cbee0a46625b4d3c44d38ef2f8

ErikSin (approved):
Thank you!

PR #2120, opened 2026-09-08
fix: modal behavior ios

### LINKED ISSUE #2106: React Navigation Modals always remain on top in iOS

cimigree:
closes #2106 
## The problem

On iOS a modal stays on top of anything opened after it. So navigating from a bottom sheet to another screen left the modal covering the newly navigated to screen,  and Close went back to the wrong screen. 

Android is fine because it orders screens by stack index.

This PR applies the same fix for both platforms:  that is, I use `replace` instead of `navigate`.

## Changes

- Adds a new method `openSheetAction` to the root stack in the navigator. The invite, map share and deep link listeners now replace an open modal instead of stacking on it. It reads `presentation` off the navigator's own descriptors, so that it can be applied to all of the modals programmatically. I did not test the deep link listener, so we should make sure to do that when we work on Invite over the Internet.
- All other modals that had navigate no use "replace" so that they can replace the previous screen or modal.
- Moves Confirm Delete Photo and Remove Archive into the modal group as a clean up.
- While I was testing, I could not actually start a track. The app crashed whenever I did. So, I added a fix here. I know it is unrelated but it was necessary to test out the behavior of the tracks modals.
Restores `isIosBackgroundLocationEnabled` in app.json. It was dropped in #2096  and without it `startLocationUpdatesAsync` throws, so tracks cannot start on iOS at all. The flag only adds `location` to `UIBackgroundModes`. It does not ask the user for background permission, so the rest of #2096 still works.

## Needs a prebuild

`app.json` changed. iOS needs `npx expo prebuild` and a rebuild. Also need to do `npm run ios `again.

## Behaviour change

Closing a screen you reached from a sheet no longer returns you to the sheet, because the sheet is gone. Back from New Collaboration goes to the drawer. Closing an error sheet goes to whatever was under the confirmation. Same on
Android.

## Testing

- Switch Project, then New Collaboration
- Start a track, Switch Project, tap another project, Stop Tracks
- Receive an invite while recording a track
- Receive a map while the Exchange settings sheet is open
- Receive an invite while a sheet is open
- Remove a CoMapeo server from a project and receive an invite or a map
- Same list on Android

---
I ran all the E2E tests locally on my Samsung device and they all passed.


ErikSin (approved):
I don't love this solution, but I also realize that it's the only thing that we can really do right now.

It just add the level of complication for a very minor UI component. There is a lot of overhead in always having to consider whether we need to use navigate or replace and whether the user is already on a modal but I also don't know of a better solution.

Issue #1402, opened 2025-09-17
Remove Onboarding


ErikSin:
Remove onboarding screens introduced in #1326. A new onboarding flow is being created, and we would like to create an RC beforehand for release. This will avoid unnecessary QA'ing of the onboarding flow.

PR #2126, opened 2026-09-14
fix: map libre zoom on android

### LINKED ISSUE #2123: Map Zoom Broken
The Map is zooming out without user intervention.

## To Do
- [ ] Determine what is causing the problem and fix it!

cimigree:
closes #2123 
Adds a patch to Map Libre for broken map zoom with new architecture on Android. I had to use AI to make the patch, but I tested it out myself on my 3 Android devices with these steps.
1. Open the map. It should start at the right zoom level and centered on the user.
2. Zoom in and walk/ move around. The zoom should hold and not snap back out.
3. Pinch/ zoom out. There should be no jump to another location.
4. Tracking. Should be able to zoom in and out several times as much as you want and keep it zoomed while tracking.
5. The Locate button. It should recenter at the same zoom as it started.

Problem:
After the switch to the new architecture, MapLibre v10's Camera keeps re-applying its last camera stop on Android. Every replay resets the zoom and interrupts pinches. In non technical terms... On Android, the map keeps snapping back to the last zoom and location the app gave it, so it undoes whatever zooming you have just done and and when you try to pinch to zoom, the map moves in the middle of the pinch and the zoom stops. Very frustrating!

This is a known issue in MapLibre. It's reported as [#1402](https://github.com/maplibre/maplibre-react-native/issues/1402) and [#1464](https://github.com/maplibre/maplibre-react-native/issues/1464), both closed with "upgrade to v11". rnmapbox hit the same bug and fixed it by not using setNativeProps for the camera ([rnmapbox/maps#3376](https://github.com/rnmapbox/maps/pull/3376)). I think these issues might be related also, but I am not sure. https://github.com/maplibre/maplibre-react-native/issues/950, https://github.com/maplibre/maplibre-react-native/issues/781, https://github.com/maplibre/maplibre-react-native/issues/741

So this patch is needed until we upgrade to v11.
Map Libre v11 rebuilt the Camera for the New Architecture, so we won't need the patch once we upgrade although it would be wise to run through the 5 steps listed above when we upgrade.


ErikSin (approved):
Thanks for the fix!

